Issue #2, opened 2018-04-03
Clang: Flexible array member with non-trivial destruction


mratsim:
Currently tests are failing for me on MacOS with Clang:

```
Error: execution of an external compiler program 'clang++ -c  -w -I/Users/<User>/.nimble/pkgs/ttmath-0.5.0  -I'/Users/<User>/.choosenim/toolchains/nim-#devel/lib' -o /Users/<User>/Programming/Status/nimbus/tests/nimcache/nimbus_stack_test.o /Users/<User>/Programming/Status/nimbus/tests/nimcache/nimbus_stack_test.cpp' failed with exit code: 1

/Users/<User>/Programming/Status/nimbus/tests/nimcache/nimbus_stack_test.cpp:242:29: error: flexible array member 'data' of type 'TY_fSgjjjotfhO8wSAxXj6bfw []' with non-trivial destruction
  TY_fSgjjjotfhO8wSAxXj6bfw data[SEQ_DECL_SIZE];
```

The offending line is:

```C++
struct tySequence_sp47cI5UpQ9bwXDfsT164pg : TGenericSeq {
  TY_fSgjjjotfhO8wSAxXj6bfw data[SEQ_DECL_SIZE];
};
```

It is used at several places in the generated C++ code, full output available [here](https://gist.github.com/mratsim/d3ebfef95a134f66f91d3d6535d32189).

Edit: GCC-7 does not have the same issue

arnetheduck:
The issue here is `SEQ_DECL_SIZE` which is declared empty (in `nimbase.h`) - this leads to an empty ("flexible") array - that's not part of the `C++` standard (it's a `C` only feature) so llvm is right to reject it, even if it happens to work for gcc (I wouldn't be surprised if gcc rejected it in pedantic mode or with extensions turned off)

mratsim:
Closed by https://github.com/status-im/nimbus/commit/1d81a7c7ba2ce6f6ea24ffa0341f67e8bd9d72bb and #5

PR #3, opened 2018-04-03
Refactor tests: remove bash/unix shell dependency for tests

mratsim:
This remove the dependency on bash/scripting shell and change the following:

- Test files start with test_ (Note that without any config `nimble test` tests file starting with a t in the tests directory by default)
- Due to #2 and https://github.com/status-im/nim-ttmath/issues/10 a macOS specific nim.cfg is added to compile Nimbus with GCC-7 (homebrew gcc to get it)

This does not address the following:
- Appveyor windows config
- Some tests are there but were unused:
    - test_opcode
    - test_vm
    - test_vm_json
- Switching of nim-rlp repo from `alehander42/nim-rlp#fix-ordinal` to `status-im/nim-rlp#master`


.travis.yml
```diff
@@ -5,4 +5,4 @@ before_install:
   - docker pull yglukhov/nim-base
 script:
   - docker run yglukhov/nim-base nim --version
-  - docker run -v "$(pwd):/project" -w /project yglukhov/nim-base /bin/bash -c "nimble install -y && ./tests/build_test.sh && ./tests/test.sh"
+  - docker run -v "$(pwd):/project" -w /project yglukhov/nim-base /bin/bash -c "nimble install -y && nimble test"
```
yglukhov: Let's use `statusteam/nim-base` instead.

.travis.yml
```diff
@@ -4,5 +4,6 @@ services:
 before_install:
   - docker pull yglukhov/nim-base
 script:
-  - docker run yglukhov/nim-base nim --version
-  - docker run -v "$(pwd):/project" -w /project yglukhov/nim-base /bin/bash -c "nimble install -y && ./tests/build_test.sh && ./tests/test.sh"
+  - docker run -v "$(pwd):/project" -w /project yglukhov/nim-base nim --version
+  - docker run -v "$(pwd):/project" -w /project yglukhov/nim-base nimble install -y
```
yglukhov: This will not work, `nimble install -y` and `nimble test` have to be in the same `docker run`. Also let's use `statusteam/nim-base` instead of `yglukhov/nim-base` please ;)

.travis.yml
```diff
@@ -4,5 +4,6 @@ services:
 before_install:
   - docker pull yglukhov/nim-base
 script:
-  - docker run yglukhov/nim-base nim --version
-  - docker run -v "$(pwd):/project" -w /project yglukhov/nim-base /bin/bash -c "nimble install -y && ./tests/build_test.sh && ./tests/test.sh"
+  - docker run -v "$(pwd):/project" -w /project yglukhov/nim-base nim --version
+  - docker run -v "$(pwd):/project" -w /project yglukhov/nim-base nimble refresh
```
yglukhov: Nope, still won't work. Use this please:
```
- docker run -v "$(pwd):/project" -w /project statusteam/nim-base sh -c "nimble install -dy && nimble test"
```

.travis.yml
mratsim: ah, I pasted from secp256k1 and didn't see that the docker was wrong here too

PR #12, opened 2018-04-12
Fixed TTInt rlp encoding

yglukhov:
The code looked wrong to me. The TTInt was encoded as a list of ints. While we probably want to encode it as a blob. This also fixes clang compilation.


mratsim:
Excellent, can you remove MacOS from the allowed failures as well now that it works? Thanks

PR #22, opened 2018-04-28
ENode object and utility procedures.

cheatfate:
ENode handling procedures and tests.

### DIFF
--- a/nimbus.nimble
+++ b/nimbus.nimble
@@ -8,10 +8,10 @@ license       = "Apache License 2.0"
 skipDirs      = @["tests"]
 
 requires "nim >= 0.18.1",
-         "https://github.com/cheatfate/nimcrypto#master",
+         "https://github.com/cheatfate/nimcrypto",
          "https://github.com/status-im/nim-rlp",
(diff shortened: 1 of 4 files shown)


cheatfate:
@yglukhov actually everything from nim-eth-p2p will go to nimbus.

yglukhov:
Why not just move eth-p2p to nimbus altogether now?

zah:
@cheatfate, we haven't decided to move everything from `p2p` to Nimbus yet. The last plan was that we'll try to define some of the required dependencies as concepts (e.g. things like the state database) and we'll then supply concrete instances from the Nimbus code. This may turn out to be more trouble than it's worth it, but we should give it a shot.

Otherwise, the code looks good.

zah:
The possibility to create projects such as https://github.com/holisticode/pss-chat-demo looks interesting and this would be the primary motivation about keeping such separation.

On the other hand, someone can just create a dependency on Nimbus and then import only specific modules.

cheatfate:
Not right place for this PR

PR #24, opened 2018-05-06
Migrate from ttmath to stint

mratsim:


src/utils_numeric.nim
```diff
@@ -5,33 +5,36 @@
...
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-import ttmath, constants, strformat, strutils, sequtils, endians, macros, utils / padding, rlp
+import stint, constants, strformat, strutils, sequtils, endians, macros, utils / padding, rlp
 
 # some methods based on py-evm utils/numeric
 
 # TODO improve
 
 proc intToBigEndian*(value: UInt256): Bytes =
```
yglukhov: These procs seem to know too much about stint internal structure. Can we just move them to stint lib?

src/vm/value.nim
```diff
@@ -21,7 +21,7 @@ type
 
 # TODO: The Int256 value is stored as array[32, byte], and we bitcast it
 # back and forth. This is a hacky workaround for the problem that clang
-# doesn't let you store ttmath types inside nim variant types (unions). Things
+# doesn't let you store stint types inside nim variant types (unions). Things
```
yglukhov: I suppose this is no longer the case, so we can just revert this hack completely.

src/utils_numeric.nim
mratsim: Yes I will move them this week, for now I just focused on porting 1-to-1 and just exposing Stint internal but tested functionality (like `countLeadingZeroBits`) and avoid introducing new procs/tests/reviews in Stint.

src/vm/value.nim
mratsim: Good catch, it seems like only vint is used externally here: https://github.com/status-im/nimbus/blob/master/src/runner.nim#L86

Will remove that hack as well.

Issue #27, opened 2018-05-14
Multiple RLP/Stint (?) linked issues in test_vm_json.nim


mratsim:
To be investigated, many tests but not all fails at rlp toBytes proc even simple byte operations like equality, xor, and, compare:

```
  [FAILED] tests/fixtures/VMTests/vmBitwiseLogicOperation/or2.json
vmBitwiseLogicOperationsgt3.json
  [OK] tests/fixtures/VMTests/vmBitwiseLogicOperation/sgt3.json
vmBitwiseLogicOperationxor2.json
test_helpers.nim(48)     test_vm_json
computation.nim(278)     testFixture
computation.nim(272)     :anonymous
storage.nim(40)          sstore
rlp.nim(360)             setStorage
rlp.nim(306)             read
rlp.nim(209)             toBytes

    Unhandled exception:
  [FAILED] tests/fixtures/VMTests/vmBitwiseLogicOperation/xor2.json
vmBitwiseLogicOperationbyte4.json
test_helpers.nim(48)     test_vm_json
computation.nim(278)     testFixture
computation.nim(272)     :anonymous
storage.nim(40)          sstore
rlp.nim(360)             setStorage
rlp.nim(306)             read
rlp.nim(209)             toBytes

    Unhandled exception:
  [FAILED] tests/fixtures/VMTests/vmBitwiseLogicOperation/byte4.json
vmBitwiseLogicOperationnot5.json
test_helpers.nim(48)     test_vm_json
computation.nim(278)     testFixture
computation.nim(272)     :anonymous
storage.nim(40)          sstore
rlp.nim(360)             setStorage
rlp.nim(306)             read
rlp.nim(209)             toBytes

    Unhandled exception:
  [FAILED] tests/fixtures/VMTests/vmBitwiseLogicOperation/not5.json
vmBitwiseLogicOperationeq2.json
test_helpers.nim(48)     test_vm_json
computation.nim(278)     testFixture
computation.nim(272)     :anonymous
storage.nim(40)          sstore
rlp.nim(360)             setStorage
rlp.nim(306)             read
rlp.nim(209)             toBytes

    Unhandled exception:
  [FAILED] tests/fixtures/VMTests/vmBitwiseLogicOperation/eq2.json
vmBitwiseLogicOperationbyte8.json
  [OK] tests/fixtures/VMTests/vmBitwiseLogicOperation/byte8.json
vmBitwiseLogicOperationand0.json
test_helpers.nim(48)     test_vm_json
[...]

PR #28, opened 2018-05-14
Re-enable exp and byteOp - pass 5 more tests

mratsim:
Exp and byteOp tests are re-enabled following pow implementation in Stint (https://github.com/status-im/nim-stint/pull/46 and https://github.com/status-im/nim-stint/commit/1283d582a10a4c9a2879e29e25855fb8bef04f47).

Multiple tests that used to work with ttmath are still failing apparently due to RLP or Memranges compatibility with Stint, see https://github.com/status-im/nimbus/issues/27.


mratsim:
#27 is fixed, waiting for the Appveyor tests before merging. Between #27 and this PR Nimbus passes 140 new tests (to be fair most of them were due to regressions)

Issue #36, opened 2018-05-23
Gas: Disentangle gas computation from operator logic


mratsim:
Followup on #34 

Currently operator logic, Ethereum forks specificities and gas costs are interleaved making maintenance, updates and having a global view hard.

Examples

- call instruction:
https://github.com/status-im/nimbus/blob/762b0b24d2eedba865ed885c036328a8bf1edddf/src/logic/call.nim#L15-L47

- create instruction:
https://github.com/status-im/nimbus/blob/762b0b24d2eedba865ed885c036328a8bf1edddf/src/logic/system_ops.nim#L20-L25

- selfdestruct instruction (notice how the EIP150 and EIP160 cost gas while the "default has no gas cost):
https://github.com/status-im/nimbus/blob/762b0b24d2eedba865ed885c036328a8bf1edddf/src/logic/system_ops.nim#L93-L140

mratsim:
Note that both Parity and Geth are separating logic from gas:

  - [Geth](https://github.com/ethereum/go-ethereum/blob/13af27641829f61d1e6b383e37aab6caae22f2c1/core/vm/gas_table.go)
  - [Parity](https://github.com/paritytech/parity/blob/master/ethcore/evm/src/interpreter/gasometer.rs)

mratsim:
#49

Issue #35, opened 2018-05-22
Gas computation: use uint64 instead of uint256


mratsim:
Using uint64 instead of uint256 for gas computation like in Go-Ethereum would be much better for devices from a memory and performance standpoint.

See Go-Ethereum PR - https://github.com/ethereum/go-ethereum/pull/3674.

Furthermore, gas prices can then be stored in a const array and replaced at instantiation site at compile-time instead of being loaded into memory.

This can be tested with
```Nim
const a = [1111, 2222, 3333, 4444]
echo a[2]
```
and checking the generated C code.

mratsim:
Note that block gas limit and so transaction gas limits hovers arount 8M currently so a int64 can be better to avoid uint surprises when substracting. [(Yellow paper, 4.3.4)](https://ethereum.github.io/yellowpaper/paper.pdf)

There is still a question about dealing with int exceptions (on debug only) but if there are gas computation issues it might be safer to crash anyway.

Issue #34, opened 2018-05-18
Support gas prices pre-EIP 150


mratsim:
Some tests are failing because the environment is [pre-EIP 150](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-150.md) which increased a lot of gas prices.

For example SLOAD opcode changed from 50 to 200 and impacts this test:
- [expXY_success.json](https://github.com/status-im/nimbus/blob/master/tests/fixtures/VMTests/vmArithmeticTest/expXY_success.json)

Decompiling the code yields
```Nim
import ../src/vm/code_stream, ../src/opcode_values, strformat

var c = newCodeStreamFromUnescaped("0x6000356000556020356001556001546000540a600255")

let opcodes = c.decompile()
for op in opcodes:
  echo &"[{op[0]}]\t{op[1]}\t{op[2]}"

# [1]     PUSH1   0x00
# [2]     CALLDATALOAD
# [4]     PUSH1   0x00
# [5]     SSTORE
# [7]     PUSH1   0x20
# [8]     CALLDATALOAD
# [10]    PUSH1   0x01
# [11]    SSTORE
# [13]    PUSH1   0x01
# [14]    SLOAD
# [16]    PUSH1   0x00
# [17]    SLOAD
# [18]    EXP
# [20]    PUSH1   0x02
# [21]    SSTORE
# [-1]    STOP
```
Expected gas balance is 39853, current resulting gas balance was 39553

Other almost certainly impacted (gas not matching in logs) includes:
- expXY
- blockhash257Block
- blockhashInRange
- coinbase
- blockhashOutOfRange
- blockhashMyBlock
- blockhashNotExistingBlock
- blockhash258Block
- smod5 (out of gas error)
- smod7 (out of gas error)
- signextend_Overflow_dj42

mratsim:
Note that the current [constants.nim](https://github.com/status-im/nimbus/blob/386d8e8327f48f8f3c6c054f114c7afc1e1801dd/src/constants.nim) was probably translated from py-evm while py-evm was also undergoing gas refactoring or many constants were introduced temporarily pending further refactoring:

The [current py-evm](https://github.com/ethereum/py-evm/blob/7cc98c1a5754e952d45b84064954eb21fe7dc42a/evm/constants.py) does not have the following:

- `GAS_SLOAD` (200 **post**-EIP150) vs `GAS_SLOAD_COST` (20 **pre**-EIP150)
[...]

PR #37, opened 2018-05-23
[Ready for merge] 1st gas refactoring - handle different gas costs in Ethereum forks

mratsim:
After diving into Py-EVM and the translated Nimbus codebase and stubs, I believe several refactorings for gas are needed.

At least:

  1. #34 - Supporting different gas cost depending on block number/VM forks
  2. #35 - use uint64 instead of uint256
  3. #36 - Isolate gas computation to reduce maintenance burden, currently it's interleaved with operator logic like for `CALL`, `CREATE` and `SELFDESTRUCT` instructions, hence the large number of files updated in this PR.

This PR only addresses the 1st point.

Status - Work in progress:
2018-05-23:
  - Introduce a GasCostKind enum and an array that maps GasCostKind -> GasCost. Different VMs will have their own mapping.
  - Rename VMs by date for chronological ordering
  - Stub Tangerine Whistle VM (hardfork EIP150) that introduced the new VM prices with the same methods as the FrontierVM
  - Change Frontier opcode gas cost to their actual value
  - **Not implemented yet**: using the proper VM depending on block number processed.


mratsim:
1st part of refactoring done, Nimbus can now be easily extended to support different gas costs depending on block number.

The main example of #34 `expXY_success` is addressed.
Overall only one new test is passed, but there are probably more dependencies.

Given than 31 files are already changed by this PR, I prefer to merge it as it is know.

src/types.nim
```diff
@@ -45,5 +46,29 @@ type
     ##   it uses the peek methods of the stack and calculates the cost
     ##   then it actually pops/pushes stuff in exec
     ## I followed the py-evm approach which does that in opcode logic
-    gasCostConstant*: UInt256
+    gasCostKind*: GasCostKind
     runLogic*:  proc(computation: var BaseComputation)
+
+  GasCostKind* = enum
+    GasZero
```
zah: I guess these are the same constants from `opcodes_table.nim`?

I think we should avoid the use of style-insensitive identifiers in our own code. They are reasonable only when dealing with third-party libraries. (to clarify, I see that we have `GasBase` here and `GAS_BASE` in opcodes_table).

src/types.nim
```diff
@@ -45,5 +46,29 @@ type
...
+    GasSelfDestruct
+    GasInHandler
+    GasRefundSclear
+
+    GasBalance
+    GasCall
+    GasExp
+    GasSHA3
+
+  GasCosts* = array[GasCostKind, UInt256]
```
zah: Does this need to be an array? Perhaps it can work as an object instead, because `initOpcodes` can use the given identifiers as field names. In practice, this won't make much of a difference, but `nimsuggest` would work a bit better for example.

src/types.nim
mratsim: Yes they are, you're right, I'll unify the identifier style.

src/vm/forks/f20161018_tangerine_whistle/tangerine_block.nim
```diff
@@ -0,0 +1,21 @@
+# Nimbus
+# Copyright (c) 2018 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or http://www.apache.org/licenses/LICENSE-2.0)
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT) or http://opensource.org/licenses/MIT)
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+import
+  ../../../logging, ../../../constants, ../../../errors, ../../../transaction,
```
zah:
```nim
../../../[logging, constants, errors, transaction]
```
You know you can do this, right?

src/types.nim
mratsim: I don't think I can use an object here: with arrays I can call `gasCosts[gasCostKind]`.

Instead, assuming I have the following object:
```Nim
type GasCosts = object
  gasVeryLow, gasLow, gasMid, gasHigh: Natural

const a = GasCosts(
  gasVeryLow: 1111,
  gasLow: 2222,
  gasMid: 3333,
  gasHigh: 4444
)
```
I would need a case statement before choosing the proper field incurring extra branching overhead for every VM opcode:
```Nim
proc consumeGas(gck: GasCostKind, costs: GasCosts): Natural =
  case gck
  of GasVeryLow: costs.gasVeryLow
  of GasLow: costs.gasLow
  of GasMid: costs.gasMid
  of GasHigh: costs.gasHigh
```

mratsim: Note: since those are known at compile-time for each fork, it can be inlined at compile-time after #35 (can't use uint256 at compile-time unfortunately) so there wouldn't be any runtime difference. But compilation time would be shorter with arrays as it avoids case statement checking.

zah: But where is this code that needs to use `GasCostKind` in dynamic fashion (storing in a field or a table and then using it for lookups)? Can you point me to it?

mratsim: Here: https://github.com/status-im/nimbus/blob/7a9a862dbfdf1eb49240507cc711b03a3bad42e5/src/opcode.nim#L16

Also, as I want to decouple gas consumption from opcode execution like in Parity and as mentioned by Alexander, it is easier to have it in an array/table.

Issue #39, opened 2018-05-24
`^`(int, int): Uint256 overloading is surprising and conflicts with stdlib


mratsim:
https://github.com/status-im/nimbus/blob/502941a388b6110adaab9697f9a375e0dc598eda/src/constants.nim#L49-L55

This is in conflict with Nim math standard lib and only saves a few characters in constants.nim. (i.e. `2 ^ 256` instead of `2.u256 ^ 256`)

PR #40, opened 2018-05-24
[WIP] Gas & Memory ops refactoring - uses native integer instead of uint256

mratsim:
This refactoring is according to #35, see full rationale there.

Other implementations:
  - Go-Ethereum uses [uint64 with explicit overflow checks](https://github.com/ethereum/go-ethereum/blob/8771c3061f340451d0966adcc547338a25f2231f/common/math/integer.go#L81-L99)
  - Py-EVM uses Python arbitrary size integer but has an [explicit gas limit](https://github.com/ethereum/py-evm/blob/7f33392e92d1bd0a56b24fdec210d077449d3533/evm/constants.py#L104) of 2^63 - 1
  - Parity uses [uint256 but tries to drop down to uint64](https://github.com/paritytech/parity/blob/e346f3058ef806ae397571beef5901f8723116d9/ethcore/evm/src/evm.rs#L58-L76) as much as possible.

Details:
  - An alias `GasInt` has been introduced to easily switch from int64 to what we want in the future, for example `range[0'i64 .. high(int64)]`.
  - Int64 will throw exceptions on gas overflow if compiled with --overflowChecks. I'm pretty sure the EVM does it as well, but to be confirmed.
  - I solved #39 as well, i.e. overloading `^`(int, int): Uint256
  - I moved `Message` and `GasMeter` types in `types.nim` to avoid recursive imports
  - EVM memory operations (pointer + length) were changed to use `Natural` instead of `uint256` as well

Pending:
  - I didn't try to compile vm_tests_json today, so don't merge yet.

What did I learn:
  - `call.nim` code is 80% about dealing with the various gas costs depending on the block number and vm fork and not about `CALL` opcode logic. This will be addressed by #36 .

### DIFF
--- a/src/computation.nim
+++ b/src/computation.nim
@@ -6,7 +6,7 @@
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
 import
-  strformat, strutils, sequtils, tables, macros, stint, terminal,
+  strformat, strutils, sequtils, tables, macros, stint, terminal, math,
   constants, errors, utils/hexadecimal, utils_numeric, validation, vm_state, logging, opcode_values, types,
   vm / [code_stream, gas_meter, memory, message, stack],
 
@@ -15,7 +15,7 @@ import
   vm/forks/f20150730_frontier/frontier_vm_state,
   vm/forks/f20161018_tangerine_whistle/tangerine_vm_state
 
-proc memoryGasCost*(sizeInBytes: UInt256): UInt256 =
+proc memoryGasCost*(sizeInBytes: Natural): GasInt =
   let
     sizeInWords = ceil32(sizeInBytes) div 32
     linearCost = sizeInWords * GAS_MEMORY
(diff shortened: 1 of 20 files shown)


mratsim:
Note: the linux build succeded but the mac build failed for unrelated reason:
```
$ nimble install -dy && nimble test
   Warning: Package 'nimbus' has an incorrect structure. It should contain a single directory hierarchy for source files, named 'nimbus', but file 'decompile_smart_contract.nim' is in a directory named 'examples' instead. This will be an error in the future.
      Hint: If 'examples' contains source files for building 'nimbus', rename it to 'nimbus'. Otherwise, prevent its installation by adding `skipDirs = @["examples"]` to the .nimble file.
  Verifying dependencies for nimbus@0.1.0
    Prompt: No local packages.json found, download it from internet? -> [forced yes]
Downloading Official package list
       Tip: 8 messages have been suppressed, use --verbose to show them.
     Error: Refresh failed
        ... Could not download: 
        ... Additional info: nodename nor servname provided, or not known
```

PR #45, opened 2018-05-30
Hash256/EthAddrres refactoring

yglukhov:
- Use `EthAddress` instead of `string`
- Use `Hash256` instead of `string`
- Stack simplified, but still `seq[byte]` is supported, I want to eventually get rid of that.
- Use `GasInt`
- Use `BlockNumber`
- Uncomment (and fix) some bits of `DB`


nimbus/db/backends/memory_backend.nim
```diff
@@ -5,26 +5,54 @@
...
+    genericHash
+    blockNumberToHash
+    blockHashToScore
+
+  DbKey* = object
+    case kind: DBKeyKind
+    of genericHash, blockHashToScore:
+      h: Hash256
+    of blockNumberToHash:
+      u: BlockNumber
```
coffeepots: Is it a good idea to have single letter fields here?

tests/test_stack.nim
```diff
@@ -54,11 +54,6 @@ suite "stack":
       stack.push(element)
     check(stack.popInt == 3.u256)
 
-    stack = newStack()
-    stack.push("1".toBytes)
-    check(stack.popBinary == "1".toBytes.pad32)
-
-
```
coffeepots: Why was the single byte test removed?

nimbus/vm/forks/f20150730_frontier/frontier_validation.nim
```diff
@@ -10,7 +10,7 @@ import
   ../../../constants, ../../../errors, ../../../vm_state, ../../../transaction, ../../../utils/header
 
 proc validateFrontierTransaction*(vmState: BaseVmState, transaction: BaseTransaction) =
-  let gasCost = transaction.gas * transaction.gasPrice
+  let gasCost = u256(transaction.gas * transaction.gasPrice)
```
mratsim: Is it normal that this is back to uint256 again?

nimbus/chain.nim
```diff
@@ -23,21 +23,21 @@ type
...
-    blockNumber*: UInt256
+    blockNumber*: BlockNumber
     difficulty*: UInt256
-    gasLimit*: UInt256
-    parentHash*: string
-    coinbase*: string
+    gasLimit*: GasInt
+    parentHash*: Hash256
+    coinbase*: EthAddress
     nonce*: string
```
mratsim: I'm pretty sure the nonce is either an `uint64` or an hex string representing an uint64. But that can be done later.

See in nim-ethash: https://github.com/status-im/nim-ethash/blob/8486a31f4f8897421fac07920b51ae16003e94bc/src/proof_of_work.nim#L146-L154

nimbus/vm/forks/f20161018_tangerine_whistle/tangerine_validation.nim
```diff
@@ -10,7 +10,7 @@ import
   ../../../constants, ../../../errors, ../../../vm_state, ../../../transaction, ../../../utils/header
 
 proc validateTangerineTransaction*(vmState: BaseVmState, transaction: BaseTransaction) =
-  let gasCost = transaction.gas * transaction.gasPrice
+  let gasCost = u256(transaction.gas * transaction.gasPrice)
   var senderBalance: UInt256
```
mratsim: Actually I think senderBalance should be Gasint

nimbus/db/backends/memory_backend.nim
yglukhov: Well yes, it's pretty private and local so I didn't bother with better naming. I can change it that bothers you.

nimbus/chain.nim
yglukhov: You're right. This pr is intentionally focused on Hash/EthAddress to not introduce even more changes. Unfortunately it also includes some chain db changes with which I started originally :)

tests/test_stack.nim
yglukhov: It is because `Stack.popBinary` is removed. I think `popBinary` should not be there because it's inefficient (allocates a seq) and from what i can see currently not that necessary.

nimbus/db/backends/memory_backend.nim
coffeepots: Nah, since it's not visible there's no harm in shorthand.

PR #56, opened 2018-06-20
Persist* functions

yglukhov:
This ports more functions to chain db, in particular the persist funcitons for state syncing. Some bits  (e.g. transactions) are waiting for a proper hexarytrie, and thus commented out.

### DIFF
--- a/nimbus/db/db_chain.nim
+++ b/nimbus/db/db_chain.nim
@@ -5,8 +5,9 @@
 #  * MIT license ([LICENSE-MIT](LICENSE-MIT) or http://opensource.org/licenses/MIT)
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-import stint, tables, rlp, ranges, state_db, backends / memory_backend,
-  ../errors, ../utils/header, ../constants, eth_common, byteutils
+import stint, tables, sequtils, algorithm, rlp, ranges, state_db, nimcrypto,
+  backends / memory_backend,
+  ../errors, ../block_types, ../utils/header, ../constants, eth_common, byteutils
 
 type
   BaseChainDB* = ref object
@@ -16,6 +17,10 @@ type
   KeyType = enum
     blockNumberToHash
     blockHashToScore
+
+  TransactionKey = tuple
+    blockNumber: BlockNumber
+    index: int
 
 proc newBaseChainDB*(db: MemoryDB): BaseChainDB =
   new(result)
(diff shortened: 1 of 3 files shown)


nimbus/db/db_chain.nim
```diff
@@ -60,35 +68,86 @@ proc getCanonicalBlockHeaderByNumber*(self: BaseChainDB; n: BlockNumber): BlockH
...
+  self.db.delete(transactionHashToBlockKey(transactionHash))
+
+proc setAsCanonicalChainHead(self: BaseChainDB; headerHash: Hash256): seq[BlockHeader] =
+  ##         Sets the header as the canonical chain HEAD.
+
+  let header = self.getBlockHeaderByHash(headerHash)
+
+  var newCanonicalHeaders = sequtils.toSeq(findNewAncestors(self, header))
+  reverse(newCanonicalHeaders)
+  for h in newCanonicalHeaders:
```
mratsim: We are iterating 4 times over the same data :/

1. toSeq(findNewAncestors
2. reverse
3. this
4. line 127

nimbus/db/db_chain.nim
```diff
@@ -99,45 +158,55 @@ proc headerExists*(self: BaseChainDB; blockHash: Hash256): bool =
...
+
+proc persistHeaderToDb*(self: BaseChainDB; header: BlockHeader): seq[BlockHeader] =
+  let isGenesis = header.parentHash == GENESIS_PARENT_HASH
+  if not isGenesis and not self.headerExists(header.parentHash):
+    raise newException(ParentNotFound, "Cannot persist block header " &
+        $header.hash & " with unknown parent " & $header.parentHash)
+  self.db.set(genericHashKey(header.hash), rlp.encode(header).toSeq())
+  let score = if isGenesis: header.difficulty
+              else: self.getScore(header.parentHash).u256 + header.difficulty
+  self.db.set(blockHashToScoreKey(header.hash), rlp.encode(score).toSeq())
```
mratsim: Is it be possible to have an `toOpenArray` for Rlp @zah?

Otherwise maybe db.set should be a template that takes an iterator to avoid extra seq allocation.

yglukhov: I was thinking about `db` to have both `openarray` and `ByteRange` interfaces.

zah: The database will prefer the `BytesRange` type, so these `toSeq` calls should be eliminated once it's integrated.

PR #57, opened 2018-06-25
Few steps towards having 64-bit block numbers

zah:
Some of the other clients don't bother to store block numbers as 256-bit integers (64 bits or even 32 bit are enough).

The changes here are based on helpers introduced in eth_common that try to abstract away all the conversion points in the code between VM words and block numbers stored in variables in fields.

Unfortunately, this turn out to be more work than anticipated and I've postponed finishing the refactoring for now (Block numbers are still 256 bit).


mratsim:
Is that supposed to be WIP, or will you do multiple PRs?

##### Note:

Current block number is 5_856_855 (can be seen https://etherscan.io/blocks), we reached this stage in about 3 years.

`high(int32) = 2^31+1 = 2_147_483_649 ~= 367 *  5_856_855`

so we can use native ints even on 32-bit platforms, for about 1000 years provided block generation speed doesn't change.

zah:
I prefer merging this as it is right now. We'll find the time to work on it again in the future in separate PRs.

Issue #14, opened 2018-04-12
Remove unnecessary ref object


mratsim:
Py2nim created every types as a `ref object` because in Python everything is a ref.

As we don't need inheritance or reference semantics (a shared context for example) for several nimbus types, for those types heap allocations are an unnecessary cost.

arnetheduck:
Obsolete

PR #65, opened 2018-07-03
Refactor interpreter dispatch

### LINKED ISSUE #62: EVM is using exceptions for non-exceptional situations
The current EVM port uses exception to terminate interpretation of normal opcodes:
  - Return
  - Revert
  - SelfDestruct

https://github.com/status-im/nimbus/blob/6f28d1186675f648e0587f44e730b9dc23291bab/nimbus/vm/interpreter/opcodes_impl/system_ops.nim#L145-L176

Using exception for control flow is contrary to our exception policy to keep exceptions for unexpected cases, i.e. overflows, invariants being breached, non-recoverable errors ...

### LINKED ISSUE #63: Boolean opcodes do not compile on latest devel
The current boolean implementation relies on `0 == false` in a corner macro case that was fixed in devel.

Consequently, Nimbus does not compile on latest devel

```
.../nimbus/nimbus/vm/interpreter/opcodes_impl/comparison.nim(11, 13) template/generic instantiation from here
.../nimbus/nimbus/vm/interpreter/opcodes_impl/helpers.nim(33, 12) template/generic instantiation from here
lib/system.nim(414, 10) Error: type mismatch: got <bool, int literal(0)>
but expected one of:
proc `==`[Enum: enum](x, y: Enum): bool
  first type mismatch at position: 1
  required type: Enum: enum
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`(x, y: pointer): bool
  first type mismatch at position: 1
  required type: pointer
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`(x, y: string): bool
  first type mismatch at position: 1
  required type: string
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`(x, y: char): bool
  first type mismatch at position: 1
  required type: char
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`(x, y: bool): bool
  first type mismatch at position: 2
  required type: bool
  but expression '0' is of type: int literal(0)
proc `==`[T](x, y: set[T]): bool
  first type mismatch at position: 1
  required type: set[T]
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`[T](x, y: ref T): bool
  first type mismatch at position: 1
  required type: ref T
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`[T](x, y: ptr T): bool
  first type mismatch at position: 1
  required type: ptr T
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`[T: proc](x, y: T): bool
  first type mismatch at position: 1
  required type: T: proc
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`(x, y: int16): bool
  first type mismatch at position: 1
  required type: int16
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`(x, y: int64): bool
  first type mismatch at position: 1
  required type: int64
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`[bits: static[int]](a343728: StUint[bits]; b343730: int{lit}): bool
  first type mismatch at position: 1
  required type: StUint[==.bits]
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`[T: SomeUnsignedInt](x, y: T): bool
  first type mismatch at position: 1
  required type: T: SomeUnsignedInt
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`(x, y: int8): bool
  first type mismatch at position: 1
  required type: int8
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`(x, y: float32): bool
  first type mismatch at position: 1
  required type: float32
  but expression '<(leftSigned, rightSigned)' is of type: bool
func `==`(x329224, y329226: StUint): bool
  first type mismatch at position: 1
  required type: StUint
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`(x, y: float): bool
  first type mismatch at position: 1
  required type: float
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`(x, y: int32): bool
  first type mismatch at position: 1
  required type: int32
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`(x, y: cstring): bool
  first type mismatch at position: 1
  required type: cstring
  but expression '<(leftSigned, rightSigned)' is of type: bool
func `==`(x, y: IntImpl): bool
  first type mismatch at position: 1
  required type: IntImpl
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`[bits: static[int]](a2532798: SomeInteger;
                           b2532800: StInt[bits]): bool
  first type mismatch at position: 1
  required type: SomeInteger
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`[bits: static[int]](a2532794: StInt[bits]; b2532796: SomeInteger): bool
  first type mismatch at position: 1
  required type: StInt[==.bits]
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`[T](x, y: seq[T]): bool
  first type mismatch at position: 1
  required type: seq[T]
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`[bits: static[int]](a343736: StInt[bits]; b343738: int{lit}): bool
  first type mismatch at position: 1
  required type: StInt[==.bits]
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`(x, y: int): bool
  first type mismatch at position: 1
  required type: int
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`[T: tuple |
    object](x, y: T): bool
  first type mismatch at position: 1
  required type: T: tuple or object
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`[I, T](x, y: array[I, T]): bool
  first type mismatch at position: 1
  required type: array[I, T]
  but expression '<(leftSigned, rightSigned)' is of type: bool
func `==`(x339224, y339226: StInt): bool
  first type mismatch at position: 1
  required type: StInt
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`[bits: static[int]](a2532786: StUint[bits]; b2532788: SomeInteger): bool
  first type mismatch at position: 1
  required type: StUint[==.bits]
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`(a, b: NimIdent): bool
  first type mismatch at position: 1
  required type: NimIdent
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`(a, b: NimSym): bool
  first type mismatch at position: 1
  required type: NimSym
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`[bits: static[int]](a343740: int{lit};
                           b343742: StInt[bits]): bool
  first type mismatch at position: 1
  required type: int
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`[bits: static[int]](a2532790: SomeInteger;
                           b2532792: StUint[bits]): bool
  first type mismatch at position: 1
  required type: SomeInteger
  but expression '<(leftSigned, rightSigned)' is of type: bool
func `==`(x, y: UintImpl): bool
  first type mismatch at position: 1
  required type: UintImpl
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`[bits: static[int]](a343732: int{lit};
                           b343734: StUint[bits]): bool
  first type mismatch at position: 1
  required type: int
  but expression '<(leftSigned, rightSigned)' is of type: bool
proc `==`(a, b: NimNode): bool
  first type mismatch at position: 1
  required type: NimNode
  but expression '<(leftSigned, rightSigned)' is of type: bool

expression: <(leftSigned, rightSigned) == 0
```

https://github.com/status-im/nimbus/blob/6f28d1186675f648e0587f44e730b9dc23291bab/nimbus/vm/interpreter/opcodes_impl/helpers.nim#L13-L39

mratsim:
This completely removes the need of an OO approach in the EVM as in Py-EVM to use a more classical "switch" as in Geth, Parity and Cpp-Ethereum (and NimVM).

It is a follow-up to #49 and #52.

Summary:
  - 1300 lines added, 2200 deleted
  - 29 files deleted
  - Arguably much easier to maintain
  - Very probably much better performance
  - Needed step if we want a Nimbus EVM stand-alone (like EVMJIT, Hera)

## Fixes:

  - Fix #63 boolean opcodes do not compile on latest devel. The new implementation compiles.
  - Fix #62 current EVM is using exceptions for control flow. The new implementation just breaks out of the opcode execution loop.
  - Help on #14, several ref objects allocated in tight loops, especially "Opcode", were removed.

## Expected benefits:

  - Maintenance is significantly improved for complex opcodes:
     - CALL needed to be special-cased depending on forks https://github.com/status-im/nimbus/blob/6f28d1186675f648e0587f44e730b9dc23291bab/nimbus/vm/interpreter/opcodes_impl/call.nim#L19-L41

      - CREATE and SELFDESTRUCT and more generally all procs with a different logic depending on hardforks beyond "simple" gas changes introduced dispatch issues https://github.com/status-im/nimbus/blob/6f28d1186675f648e0587f44e730b9dc23291bab/nimbus/vm/interpreter/opcodes_impl/system_ops.nim#L19-L24

      - Previous dispatching logic used exceptions for control flow (#62) in Return, Revert and SelfDestruct: https://github.com/status-im/nimbus/blob/6f28d1186675f648e0587f44e730b9dc23291bab/nimbus/vm/computation.nim#L247-L291

       - Reimplementation of opcodes in single file of 750 lines instead of 18 files totaling 1220 lines

       - No inheritance instances to track.

  - Performance

    - CALL, CREATE, SELFDESTRUCT special-casing require slower runtime dispatch
    - The heap allocated opcode type has been killed, which significantly reduce GC pressure as it avoids allocation at each opcode read https://github.com/status-im/nimbus/blob/6f28d1186675f648e0587f44e730b9dc23291bab/nimbus/vm_types.nim#L40-L45. VM Forks that would have been created on the heap for each smart contract CALL will also be avoided
    - Computed Gotos as researched here are the fastest portable interpreter dispatch technique with speed benefits of about 20%. See research notes in the wiki: https://github.com/status-im/nimbus/wiki/Interpreter-optimization-resources

  - Nimbus EVM as standalone

    - Some interest has been expressed to use Nimbus EVM as a standalone library with EVM-C interface, that can be called from any language. To avoid having to load the Nim runtime, inheritance and all GC-ed types should be avoided in that case

## Note on regressions

There are regressions on `arith`, `boolean` and `mktx` tests that test that the CALL opcode which is stubbed.

Similarly the `expXY_success` test passes even though it uses SLOAD. It passes because SLOAD is stubbed with a non-zero value.

## Note on ~new jump tests passed~ specific jump tests

~I'm not too sure why I pass new jump/control flow tests, the new tests were enabled by this commit https://github.com/status-im/nimbus/pull/65/commits/63b6fc251f77146b60b85bf032429cd5489f9e33, where I changed log proc back to templates as in master (I had some import/symbol visibility issue with templates during branch implementation).~

The following tests are passing with  -d:release and failing in debug mode:
  - BlockNumberDynamicJumpiOutsideBoundary
  - DynamicJumpiOutsideBoundary
  - jumpiOutsideBoundary
  - kv1

## CI failure:

~Pending - https://github.com/status-im/nim-rocksdb/issues/3~ Solved


nimbus/vm/forks/f20150730_frontier/frontier_vm_state.nim
```diff
@@ -1,170 +0,0 @@
...
-#     prevHeaders*: seq[BlockHeader]
-#     receipts*: void
-#     computationClass*: Any
-#     _chaindb*: BaseChainDB
-#     accessLogs*: AccessLogs
-#     blockHeader*: BlockHeader
-
-# proc _executeFrontierTransaction*(vmState: FrontierVMState;
-#                                  transaction: FrontierTransaction): FrontierComputation =
-#   transaction.validate()
```
zah: The deleted code here seems quite complicated. Was it translated in any way in the new implementation?

nimbus/vm/interpreter/opcodes_impl.nim
```diff
@@ -0,0 +1,759 @@
...
+  ../../vm_state, ../../errors, ../../constants, ../../vm_types, ../../logging,
+  ../../db/[db_chain, state_db],
+  ../../utils/[bytes, padding, address] # TODO remove those dependencies
+
+# ##################################
+# 0s: Stop and Arithmetic Operations
+
+op add, FkFrontier, inline = true, lhs, rhs:
+  ## 0x01, Addition
+  push: lhs + rhs
```
zah: The `push` construct seems a bit unnatural, because it suggests that an arbitrary block may be given, but is this actually true? You could have defined it as a simple dirty template accepting a single expression and making use of the `computation` variable.

nimbus/vm/interpreter/opcodes_impl.nim
```diff
@@ -0,0 +1,759 @@
+# Nimbus
```
zah: The fact that all opcode implementations were moved will make merging this with the other state-ops branch very hard. Any suggestions about how should we solve this problem?

nimbus/vm/interpreter_dispatch.nim
```diff
@@ -0,0 +1,231 @@
...
+  ../vm_types, ../errors,
+  ../logging, ./stack, ./computation, terminal # Those are only needed for logging
+
+func invalidInstruction*(computation: var BaseComputation) {.inline.} =
+  raise newException(ValueError, "Invalid instruction, received an opcode not implemented in the current fork.")
+
+let FrontierOpDispatch {.compileTime.}: array[Op, NimNode] = block:
+  fill_enum_table_holes(Op, newIdentNode"invalidInstruction"):
+    [
+      Stop: newIdentNode "toBeReplacedByBreak",
```
zah: We should try to avoid programming constructs that break grepping, "go to definition" and ntags. The previous opcode table was nicer in this regard, because it used the actual proc names and I believe we can keep this property.

Can you explain in more details how the different forks will feature different opcode implementations? My previous understanding was that this is a rare occurrence and we'll handle it by defining separate op code implementation only where differences exist (and these separate implementations will have different proc names). Any shared code may be defined as helper templates. 

EDIT: I noticed how `genCall` is defined, but I still think the whole suffix-appending approach is probably unnecessary (the `op` macro doesn't need the `ForkName` parameter).

nimbus/vm/interpreter/opcodes_impl/system_ops.nim
```diff
@@ -1,177 +0,0 @@
...
-
-method runLogic*(create: CreateByzantium, computation) =
-  if computation.msg.isStatic:
-    raise newException(WriteProtection, "Cannot modify state while inside of a STATICCALL context")
-  procCall runLogic(create, computation)
-
-proc selfdestructEIP150(computation) =
-  let beneficiary = stack.popAddress()
-  # TODO: with
-  # with computation.vm_state.state_db(read_only=True) as state_db:
```
zah: I find these TODO items and commented code quite useful. We should try to keep them.

nimbus/vm/interpreter/opcodes_impl.nim
mratsim: Yeah, I did too much macros recently and completely forgot about dirty templates

nimbus/vm/interpreter_dispatch.nim
mratsim: I agree, I will suffix the changed ops with their EIP (like Py-EVM) or fork (might be more dev friendly) but the default one will have an easy name.

I also agree with the grepping issue, I was often wondering where `ceil32` was defined and it was defined by `ceilXX(32)`

nimbus/vm/interpreter/opcodes_impl.nim
mratsim: Fixed in https://github.com/status-im/nimbus/pull/65/commits/3151bba0d13f9f92c4fb874a4ab50b26269c293e

nimbus/vm/interpreter_dispatch.nim
mratsim: Fixed in https://github.com/status-im/nimbus/pull/65/commits/00baaf384e996657c08221f5efb94e209b1f61b6

nimbus/vm/interpreter/opcodes_impl.nim
mratsim: I feel like most of the changes in https://github.com/status-im/nimbus/tree/state-ops-implementation are pretty self-contained.

State-ops can be merged first, and I rebase and port changes over.
I also have to carry over the changes from https://github.com/status-im/nimbus/tree/blocknumber-size-refactoring

nimbus/vm/forks/f20150730_frontier/frontier_vm_state.nim
mratsim: No, I feel that the code is much too complex and it has been refactored/deleted in Py-EVM since our port from January so it's better if we rewrite it from scratch.

Notice how there is no execute_frontier_transaction in the following screenshot and how all methods are about 10 lines long:

nimbus/vm/interpreter/opcodes_impl/system_ops.nim
mratsim: Yes I find them useful as well, I try to keep them:

- Create: https://github.com/status-im/nimbus/blob/00baaf384e996657c08221f5efb94e209b1f61b6/nimbus/vm/interpreter/opcodes_impl.nim#L492

- Call: https://github.com/status-im/nimbus/blob/00baaf384e996657c08221f5efb94e209b1f61b6/nimbus/vm/interpreter/opcodes_impl.nim#L664

I preferred to delete it to reduce cognitive load given that:
- the comment didn't match with the current Nim syntax,
- the op is relatively straightforward,
- I can always use git history to get it back.

and I intend to finish this as soon as this PR and state ops are merged.

mratsim:
I've rebased on the new master. I still have to port to opcode_impl.nim some of the new opcode implementations that include the #59 additions.

@zah I'm getting Illegal storage accesses with the new hexaryTree:

```Nim
[Suite] parse bytecode
  [OK] accepts bytes
  [OK] next returns the correct opcode
  [OK] peek returns next opcode without changing location
  [OK] stop opcode is returned when end reached
  [OK] [] returns opcode
  [OK] isValidOpcode invalidates after PUSHXX
  [OK] isValidOpcode 0
  [OK] isValidOpcode 1
  [OK] right number of bytes invalidates

[Suite] gasMeter
  [OK] consume spends
  [OK] consume errors
  [OK] return refund works correctly

[Suite] memory
  [OK] write
  [OK] write rejects valyes beyond memory size
  [OK] extends appropriately extends memory
  [OK] read returns correct bytes

[Suite] stack
    test_stack.nim(30, 16): Expect Failed, unexpected exception was thrown.
  [FAILED] push only valid
  [OK] push does not allow stack to exceed 1024
  [OK] dup does not allow stack to exceed 1024
  [OK] pop returns latest stack item
  [OK] swap correct
  [OK] dup correct
  [OK] pop raises InsufficientStack appropriately
  [OK] swap raises InsufficientStack appropriately
  [OK] dup raises InsufficientStack appropriately
  [OK] binary operations raises InsufficientStack appropriately

[Suite] opcodes
  [OK] add
Traceback (most recent call last)
test_opcode.nim(79)      test_opcode
test_opcode.nim(45)      testCode
interpreter_dispatch.nim(227) executeOpcodes
interpreter_dispatch.nim(194) frontierVM
opcodes_impl.nim(195)    balance
vm_state.nim(122)        readOnlyStateDB
db_chain.nim(231)        getStateDb
state_db.nim(26)         newAccountStateDB
hexary.nim(55)           initHexaryTrie
assign.nim(124)          genericAssign
assign.nim(100)          genericAssignAux
assign.nim(24)           genericAssignAux
assign.nim(21)           genericAssignAux
assign.nim(106)          genericAssignAux
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
```

zah:
I'm a bit worried about the separation of VM/VmState that is no longer present, but we can always restore it in the future if necessary. I'm still looking forward to see how the fork-specific opcode implementations will work out (we've discussed on Slack that these may be mapped to an additional static parameter for the forked instructions).

All in all, let's merge this, so the rest of the team doesn't accumulate too many conflicts.

PR #68, opened 2018-07-06
Delete utilities: address, padding, bytes

mratsim:
This PR deletes 3 utilities that are not needed in Nim:

## Address

Address was moved out of Py-EVM and put into https://github.com/ethereum/eth-utils/blob/master/eth_utils/address.py.

I looked into usage for some procs like:
  - is_normalized_address
  - is_canonical_address
  - is_binary_address

and didn't find any in Py-EVM. In any case if it's ever needed this can be handled by the type system in Nim.

## Padding

PR https://github.com/status-im/nim-stint/pull/57 brought padding support in Stint, for padding in string, strformat is enough though due to https://github.com/nim-lang/Nim/issues/8220 you need some workaround for strformat in a quote do section: https://github.com/nim-lang/Nim/issues/8220

## Bytes

By using `seq[byte]` instead of `string` for the default representation in BaseComputation you avoid a lot of useless conversion within the EVM.

## Future TODOs:

Message is still using strings https://github.com/status-im/nimbus/blob/d280ea0d88281f29c97a06229e399a2672894818/nimbus/vm_types.nim#L54-L82.

I had some issue with "0x" prefixes in test so I prefer keep that for later.


nimbus/vm/computation.nim
```diff
@@ -45,13 +45,17 @@ proc shouldBurnGas*(c: BaseComputation): bool =
 proc shouldEraseReturnData*(c: BaseComputation): bool =
   c.isError and c.error.erasesReturnData
 
+func bytesToHex(x: openarray[byte]): string {.inline.} =
```
zah: There is a function in Nimcrypto for this: `nimcrypto/utils.toHex`

tests/test_stack.nim
```diff
@@ -21,6 +21,12 @@ template testFailPush(value: untyped): untyped =
   expect(ValidationError):
     stack.push(value)
 
+func toBytes(s: string): seq[byte] =
```
zah: hrm, I'm not sure this is safe. Have you used it for a long time? @yglukhov?

yglukhov: On the other hand, this trick was demonstrated by Araq some time ago, so maybe it's going to break anytime soon. I'd say let it be for now.

zah: @data-man had a pull request about turning a `string` into a `openarray[byte]`. I think we should convince Araq to merge it.

Otherwise, this PR is good for merging for me.

mratsim: It's just for testing in any case

nimbus/vm/computation.nim
mratsim: Both the `nimcrypto` and `bytesutils` openarray[byte] to hex string proc do not prepend the `0x` which for some reason is needed.

When message.code is changed to `seq[byte]` instead of string this will be removed.

PR #71, opened 2018-07-17
Fixed sstore gas computation, empty sha3, smod, sdiv

yglukhov:
Fixed some tests, `smod5` and `smod7` are broken now.

EDIT: Fixed smod and sdiv

### DIFF
--- a/nimbus/vm/interpreter/opcodes_impl.nim
+++ b/nimbus/vm/interpreter/opcodes_impl.nim
@@ -46,7 +46,7 @@ op sdiv, inline = true, lhs, rhs:
   ## 0x05, Signed division
   var r: UInt256
   if rhs != 0:
-    let min = (one(UInt256) shl 256) + one(UInt256)
+    const min = (1.u256 shl 255) - 1.u256
     var a = lhs
     var b = rhs
     var signA, signB: bool
--- a/nimbus/vm/interpreter/opcodes_impl.nim
+++ b/nimbus/vm/interpreter/opcodes_impl.nim
@@ -182,8 +182,11 @@ op sha3, inline = true, startPos, length:
 
   computation.memory.extend(pos, len)
   let endRange = min(pos + len, computation.memory.len) - 1
-  push:
-    keccak256.digest computation.memory.bytes.toOpenArray(pos, endRange)
+  if endRange == -1:
+    push(EMPTY_SHA3)
+  else:
+    push:
+      keccak256.digest computation.memory.bytes.toOpenArray(pos, endRange)
 
 # ##########################################
 # 30s: Environmental Information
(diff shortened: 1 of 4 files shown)


nimbus/vm/interpreter/opcodes_impl.nim
```diff
@@ -44,12 +44,20 @@ op divide, inline = true, lhs, rhs:
...
+  var r: UInt256
+  if rhs != 0:
+    let min = (one(UInt256) shl 256) + one(UInt256)
+    var a = lhs
+    var b = rhs
+    var signA, signB: bool
+    extractSign(a, signA)
+    extractSign(b, signB)
+    if a == min and b == not zero(UInt256):
+      r = min
```
mratsim: Special check for `low(Int256)` is always false here because you flipped the sign already.

This should be done right at the start and return early.

nimbus/vm/interpreter/opcodes_impl.nim
```diff
@@ -44,12 +44,20 @@ op divide, inline = true, lhs, rhs:
 
 op sdiv, inline = true, lhs, rhs:
   ## 0x05, Signed division
-  push:
-    if rhs == 0: zero(Uint256)
+  var r: UInt256
+  if rhs != 0:
+    let min = (one(UInt256) shl 256) + one(UInt256)
```
mratsim: * This can be `const`.

* It should be `shl 255` to set the most significant bit. For some reason `shl x` with x > 255 also work but it may be a bug in stint.

* here min (= low signed int256) representation is: 1000...0001, the int256 minimum is 1000...0000 (but you cannot negate it) so the `+ one(Uint256)` is not needed.  
  See [stint high/low signed int implementation](https://github.com/status-im/nim-stint/blob/406f1aa317b0dac6ce1c8fe1f3aa1cb6e3603f10/stint/private/int_highlow.nim#L19-L23)
  ```Nim
  func high*[T](_: typedesc[IntImpl[T]]): IntImpl[T] {.inline.}=
    # The highest signed int has representation
    # 0b0111_1111_1111_1111 ....
    # so we only have to unset the most significant bit.
    result = not result
    most_significant_word(result) = most_significant_word(result) shr 1

  func low*[T](_: typedesc[IntImpl[T]]): IntImpl[T] {.inline.}=
    # The lowest signed int has representation
    # 0b1000_0000_0000_0000 ....
    # so we only have to set the most significant bit.
    not high(IntImpl[T])
  ```

Ideally I'd like an test in test_opcodes to catch tricky low(Int256) special casing.

nimbus/vm/interpreter/utils/utils_numeric.nim
```diff
@@ -38,3 +28,15 @@ func wordCount*(length: Natural): Natural {.inline.}=
   # Returns the number of EVM words corresponding to a specific size.
   # EVM words is rounded up
   length.ceil32 shr 5 # equivalent to `div 32` (32 = 2^5)
+
+proc flipSign(value: var UInt256) =
+  value = not value
+  value += 1.u256
```
mratsim: For an i4, the lowest value is
- 16 = 0b1000

Flipping bits: 0b0111
+1 : 0b1111 = -1

There should be a warning comment about that:
```Nim
proc flipSign(value: var UInt256) =
  ## ⚠ Warning: low(Int256) (binary repr 0b1000...0000) cannot be negated, please handle this special case
  value = not value
  value += 1.u256
```

PR #76, opened 2018-07-19
Only write to child computation if data present

coffeepots:
If the child computation is empty, we currently call `write` with `toOpenArray(0, 0)`, which fails with an index error into the empty array.

I'm not sure if `toOpenArray` should be able to handle this as a special case, but that's a separate issue.

### DIFF
--- a/nimbus/vm/interpreter/opcodes_impl.nim
+++ b/nimbus/vm/interpreter/opcodes_impl.nim
@@ -738,10 +738,9 @@ template genCall(callName: untyped): untyped =
 
     if not childComputation.shouldEraseReturnData:
       let actualOutputSize = min(memOutLen, childComputation.output.len)
-      if actualOutputSize > 0:
-        computation.memory.write(
-          memOutPos,
-          childComputation.output.toOpenArray(0, actualOutputSize))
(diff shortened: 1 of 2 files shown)


nimbus/vm/interpreter/opcodes_impl.nim
```diff
@@ -738,9 +738,10 @@ template genCall(callName: untyped): untyped =
 
     if not childComputation.shouldEraseReturnData:
       let actualOutputSize = min(memOutLen, childComputation.output.len)
-      computation.memory.write(
-        memOutPos,
-        childComputation.output.toOpenArray(0, actualOutputSize))
+      if actualOutputSize > 0:
+        computation.memory.write(
+          memOutPos,
+          childComputation.output.toOpenArray(0, actualOutputSize))
```
yglukhov: Should not it be `actualOutputSize-1` here? `toOpenArray` takes an inclusive range, while len is usually an out-of-bounds index.

coffeepots: Changed it to `actualOutputSize - 1`. Looks better than with the `if` :)

PR #84, opened 2018-07-26
Workaround for nim issue 7833

yglukhov:


nimbus/config.nim
```diff
@@ -255,6 +255,7 @@ proc setNetwork(conf: var NetConfiguration, network: NetworkFlags,
   conf.flags.excl({MainNet, MordenNet, RopstenNet, RinkebyNet, KovanNet,
                    CustomNet})
   conf.flags.incl(network)
+  assert(not conf.bootNodes.isNil) # Nim bug #7833
```
cheatfate: This is will never happens if `result.net.bootNodes = @[]` will be executed.

yglukhov: This is added because the initialization happens many lines away from this line. Mostly for "documentation" purposes. And then, no `assert` should ever happen, but we still put them here and there for some reason :)

yglukhov:
Got approve from cheatfate in a privately held top secret conversation. Merging.

PR #91, opened 2018-08-01
Don't require mocking/faking special case BLOCKHASH instruction for VMTests

tersec:
They were removed a couple months ago in the official Ethereum tests repo for this reason.

https://github.com/ethereum/tests/pull/444 and https://github.com/ethereum/tests/issues/415 discuss reasons for this, e.g.,
> Perhaps the VMTests that only deal with VM-related stuff could remain (eg. arithmetic, local memory ops, program counter ops), and anything that deals with Ethereum-specific state could be translated into GeneralStateTests instead?
>
>I've become interested in this because I'm in the middle of refactoring the KEVM control-flow to make it more uniform to make doing proofs with KEVM easier (and also to support more uniform typed exceptions), but there are several spots where special cases are made for VMTests mode which are causing me headache.

Properly supporting any/all of these tests (the 257/258/outofrange/etc ones sort of conveniently fail, but that's because their BLOCKHASH isn't supposed to actually function at all -- barring a real "VMTest mode" as alluded to, they don't actually distinguish between failure modes of desired failure not to return a hash more than 256 blocks old and undesired returning an invalid block hash because they don't actually exist) would appear to require a similarly architecturally problematic VMTest mode in Nimbus due to, essentially, layering violations/inversions in the test fixtures.

I agree with the consensus in the Ethereum tests repo discussions that while testing BLOCKHASH and the like is, obviously, worthwhile, the VM tests should be VM tests, not VM state tests in disguise.

In particular, block hashes are currently calculated by directly calling rlpHash on the block, which seems fine. There's no way short of colliding `rlpHash` or, e.g., making `getAncestorHash` notice it's running VMTests (e.g., by noticing that it doesn't have ancestor blocks it theoretically would given its current block number, because in the VMTests, its current block number is completely artificial).


zah:
Why don't we sync to the latest test suite instead of applying changes on a case-by-case basis?

tersec:
That's next. The immediate goal was to determine which of the existing VMTests were worth continuing to get working, versus which would have introduced architectural changes the need for which would have been obviated by the coming, larger test suite sync.

PR #92, opened 2018-08-01
 Don't require mocking/faking special case BLOCKHASH instruction for VMTests, part 2

tersec:
Same idea as https://github.com/status-im/nimbus/pull/91 -- follow [Remove VM tests using the mock BLOCKHASH/BALANCE instruction just for VM tests](https://github.com/ethereum/tests/pull/444) upstream, but mostly finish the job.


tersec:
So far, all the removed ones I've checked also show up in modified form in https://github.com/ethereum/tests/tree/develop/GeneralStateTests -- I'll be checking if all of these show up in some form there and/or have good reason not to, and if so, I'll consider this worth merging.

coffeepots:
LGTM too.

PR #93, opened 2018-08-01
Added genesis initialization

yglukhov:
Note: There's still an issue with genesis state root. It can't be calculated for Ropsten and Rinkeby (see https://github.com/status-im/nim-eth-trie/issues/17), and it is incorrect for main net.


nimbus/config.nim
```diff
@@ -138,6 +156,47 @@ var nimbusConfig {.threadvar.}: NimbusConfiguration
...
+      homesteadBlock: 1.u256,
+      daoForkSupport: true,
+      eip150Block:    2.u256,
+      eip150Hash:     toDigest("9b095b36c15eaf13044373aef8ee0bd3a382a5abb92e402afa44b8249c3a90e9"),
+      eip155Block:    3.u256,
+      eip158Block:    3.u256,
+      byzantiumBlock: 1035301.u256
+    )
+  else:
+    raise newException(Exception, "No chain config for " & $id)
```
cheatfate: Nimbus is end product we are making, and generating exceptions for end-user is not a good way of showing them errors.

nimbus/config.nim
```diff
@@ -243,42 +302,43 @@ proc processRpcArguments(key, value: string): ConfigStatus =
...
-  for item in (nodes):
+  onodes = newSeqOfCap[ENode](nodeUris.len)
+  for item in nodeUris:
     doAssert(processENode(item, node) == Success)
-    (onodes).add(node)
+    onodes.add(node)
 
-proc setNetwork(conf: var NetConfiguration, network: NetworkFlags,
-                id: uint = 0) =
+proc toPublicNetwork*(id: uint): PublicNetwork {.inline.} =
```
cheatfate: Way on how this procedure works is currently not working in Nim.
This is example:
```nim
type
  PublicNetwork* = enum
    CustomNet = 0
    MainNet = 1
    MordenNet = 2
    RopstenNet = 3
    RinkebyNet = 4
    KovanNet = 42

proc toPublicNetwork*(id: uint): PublicNetwork {.inline.} =
  for i in PublicNetwork.low .. PublicNetwork.high:
    if uint(i) == id:
      result = i
      break

when isMainModule:
  echo toPublicNetwork(0)
  echo toPublicNetwork(1)
  echo toPublicNetwork(2)
  echo toPublicNetwork(3)
  echo toPublicNetwork(4)
  echo toPublicNetwork(5)
  echo toPublicNetwork(6)
```

nimbus/config.nim
yglukhov: Should I add a `TODO` to the error message?

mratsim: This is redundant with https://github.com/status-im/nimbus/blob/d94dfbdb432feafd029d681406502bcf36957e8e/nimbus/vm/interpreter/vm_forks.nim#L23-L31

You can move vm_forks.nim at a higher folder along with the [forks_list.md](https://github.com/status-im/nimbus/blob/d94dfbdb432feafd029d681406502bcf36957e8e/nimbus/vm/interpreter/forks_list.md) and forks_list.png file if needed.

nimbus/genesis.nim
```diff
@@ -0,0 +1,112 @@
...
+    coinbase*: EthAddress
+    alloc*: GenesisAlloc
+
+  GenesisAlloc = TableRef[EthAddress, GenesisAccount]
+  GenesisAccount = object
+    code*: seq[byte]
+    storage*: Table[UInt256, UInt256]
+    balance*: UInt256
+    nonce*: UInt256
+    #privateKey: seq[byte] # For tests?
```
mratsim: behind a when nimbusdebug or nimbustest flag then.

We shouldn't carry private keys in types except when required to reduce the attack surface.

nimbus/genesis.nim
```diff
@@ -0,0 +1,112 @@
...
+
+proc toAddress(n: UInt256): EthAddress =
+  let a = n.toByteArrayBE()
+  result[0 .. ^1] = a[12 .. ^1]
```
mratsim: toOpenArray to avoid copy

mratsim: Also where possible we should use `func` I think.

nimbus/config.nim
cheatfate: If you raising exceptions, then other parts of code needs to catch it, so simple `TODO` will be not enough.

yglukhov: @mratsim, I agree we'll have to make the VM use the config at some point.

yglukhov: @cheatfate This exception doesn't have to be caught, it's more like a panic. The missing blocks can be filled out trivially, just don't want to waste time on currently irrelevant stuff.

cheatfate: For me its not trivial, i even don't know where to find this information, so if this so trivial for you is it possible to fill the gap?

cheatfate:
Table[T] is very memory hungry structure, do we really need its capabilities?

nimbus/config.nim
yglukhov: It's trivial but time consuming, as I also have to lookup this info, and let me repeat it once again: we don't need it right now.

yglukhov:
> Table[T] is very memory hungry structure

We don't really care, this code is executed once per DB init, the table is freed afterwards.

yglukhov:
Hey guys, please review one more commit (Slightly reworked ChainDB interface to more consistent). The main idea is to have throwing and non-throwing overloaded getters. As you can some code in this particular file already benefits from those.

PR #94, opened 2018-08-02
fix assorted test fixtures which have been un-skipped

tersec:
These are all, effectively, typos, except `fromHex(UInt256, fenv{"currentDifficulty"}.getStr),`, which avoids an overflow in extremely high currentDifficulty settings in those random test fixtures.


coffeepots:
Bizarrely Appveyor is failing in 32 bit for the RocksDB checks with this commit.

tersec:
Yes, should that failing 32-bit RocksDB check be taken as blocking this merge, or go ahead with merge and for that to be diagnosed separately?

tersec:
Double-checked and it was failing at the storage tests. Just not really plausible causation.

PR #95, opened 2018-08-03
Eth get balance

coffeepots:
This is currently still a work in progress as it will be reliant on sync being implemented, however shouldn't touch any other parts of the system for now.

Currently the initialisation elements are lacking, however #93 adds a lot of the code required for this, so I'm putting this up now in order to merge in with that PR.

### DIFF
--- a/nimbus/rpc/hexstrings.nim
+++ b/nimbus/rpc/hexstrings.nim
@@ -90,14 +90,14 @@ proc fromJson*(n: JsonNode, argName: string, result: var HexQuantityStr) =
   n.kind.expect(JString, argName)
   let hexStr = n.getStr()
   if not hexStr.validateHexQuantity:
-    raise newException(ValueError, "Parameter \"" & argName & "\" value is not valid as an Ethereum hex quantity \"" & hexStr & "\"")
-  result = hexStr[2..hexStr.high].hexQuantityStr
+    raise newException(ValueError, "Parameter \"" & argName & "\" is not valid as an Ethereum hex quantity \"" & hexStr & "\"")
+  result = hexStr.hexQuantityStr
 
 proc fromJson*(n: JsonNode, argName: string, result: var HexDataStr) =
   # Note that '0x' is stripped after validation
   n.kind.expect(JString, argName)
   let hexStr = n.getStr()
   if not hexStr.validateHexData:
-    raise newException(ValueError, "Parameter \"" & argName & "\" value is not valid as a Ethereum data \"" & hexStr & "\"")
-  result = hexStr[2..hexStr.high].hexDataStr
+    raise newException(ValueError, "Parameter \"" & argName & "\" is not valid as a Ethereum data \"" & hexStr & "\"")
+  result = hexStr.hexDataStr
 
--- a/nimbus/rpc/hexstrings.nim
+++ b/nimbus/rpc/hexstrings.nim
@@ -1,0 +1,103 @@
+# Nimbus
+# Copyright (c) 2018 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
+## This module implements the Ethereum hexadecimal string formats for JSON
+## See: https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding
+
+type
+  HexQuantityStr* = distinct string
+  HexDataStr* = distinct string
+
+# Hex validation
+
+template stripLeadingZeros(value: string): string =
+  var cidx = 0
+  # ignore the last character so we retain '0' on zero value
+  while cidx < value.len - 1 and value[cidx] == '0':
+    cidx.inc
+  value[cidx .. ^1]
+
+proc encodeQuantity*(value: SomeUnsignedInt): string =
+  var hValue = value.toHex.stripLeadingZeros
+  result = "0x" & hValue
+
+template hasHexHeader*(value: string): bool =
+  if value != "" and value[0] == '0' and value[1] in {'x', 'X'} and value.len > 2: true
+  else: false
+
+template isHexChar*(c: char): bool =
+  if  c notin {'0'..'9'} and
+      c notin {'a'..'f'} and
+      c notin {'A'..'F'}: false
+  else: true
+
+proc validateHexQuantity*(value: string): bool =
+  if not value.hasHexHeader:
+    return false
+  # No leading zeros
+  if value[2] == '0': return false
+  for i in 2 ..< value.len:
+    let c = value[i]
+    if not c.isHexChar:
+      return false
+  return true
+
+proc validateHexData*(value: string): bool =
+  if not value.hasHexHeader:
+    return false
+  # Must be even number of digits
+  if value.len mod 2 != 0: return false
+  # Leading zeros are allowed
+  for i in 2 ..< value.len:
+    let c = value[i]
+    if not c.isHexChar:
+      return false
+  return true
+
+# Initialisation
+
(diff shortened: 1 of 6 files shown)


nimbus/rpc/hexstrings.nim
```diff
@@ -0,0 +1,111 @@
...
+
+template hasHexHeader*(value: string): bool =
+  if value != "" and value[0] == '0' and value[1] in {'x', 'X'} and value.len > 2: true
+  else: false
+
+template isHexChar*(c: char): bool =
+  if  c notin {'0'..'9'} and
+      c notin {'a'..'f'} and
+      c notin {'A'..'F'}: false
+  else: true
```
mratsim: All those hex stuff already exists in byteutils: https://github.com/status-im/nim-byteutils/blob/master/byteutils.nim

nimbus/rpc/hexstrings.nim
```diff
@@ -0,0 +1,111 @@
...
+
+proc validateHexQuantity*(value: string): bool =
```
mratsim: I would rename `isValidHexQuantity`

nimbus/rpc/hexstrings.nim
```diff
@@ -0,0 +1,111 @@
...
+    return false
+  # No leading zeros (but allow 0x0)
+  if value.len > 3 and value[2] == '0': return false
+  for i in 2 ..< value.len:
+    let c = value[i]
+    if not c.isHexChar:
+      return false
+  return true
+
+proc validateHexData*(value: string): bool =
```
mratsim: `isValidHexData`

nimbus/rpc/hexstrings.nim
```diff
@@ -0,0 +1,111 @@
...
+# Hex validation
+
+template stripLeadingZeros(value: string): string =
+  var cidx = 0
+  # ignore the last character so we retain '0' on zero value
+  while cidx < value.len - 1 and value[cidx] == '0':
+    cidx.inc
+  value[cidx .. ^1]
+
+proc encodeQuantity*(value: SomeUnsignedInt): string =
```
mratsim: `func` if there is no global.

If we have weird to reproduce error, we will know that unless if we use unsafe cast or pointer, the issue is not there

nimbus/rpc/hexstrings.nim
mratsim: and use `func`

nimbus/rpc/hexstrings.nim
```diff
@@ -0,0 +1,111 @@
...
+const
+  SInvalidQuantity = "Invalid hex quantity format for Ethereum"
+  SInvalidData = "Invalid hex data format for Ethereum"
+
+proc validateRaiseHexQuantity*(value: string) =
+  if not value.validateHexQuantity:
+    raise newException(ValueError, SInvalidQuantity & ": " & value)
+
+proc validateRaiseHexData*(value: string) =
+  if not value.validateHexData:
```
mratsim: Nitpicking:
it should be an {.inline.} func

and the line should be: `if unlikely(not value.validateHexData):`

The compiler rearranges the if branches to please the CPU branch predictor (though GCC also says that sometimes it's placebo)

nimbus/rpc/hexstrings.nim
```diff
@@ -0,0 +1,111 @@
...
+    let c = value[i]
+    if not c.isHexChar:
+      return false
+  return true
+
+const
+  SInvalidQuantity = "Invalid hex quantity format for Ethereum"
+  SInvalidData = "Invalid hex data format for Ethereum"
+
+proc validateRaiseHexQuantity*(value: string) =
```
mratsim: once the other proc is renamed `isValidHexQuantity`, you can rename this one `validateHexQuantity`

nimbus/rpc/hexstrings.nim
```diff
@@ -0,0 +1,111 @@
...
+  result = %(value.string)
+
+proc `%`*(value: HexDataStr): JsonNode =
+  result = %(value.string)
+
+proc fromJson*(n: JsonNode, argName: string, result: var HexQuantityStr) =
+  # Note that '0x' is stripped after validation
+  n.kind.expect(JString, argName)
+  let hexStr = n.getStr()
+  if not hexStr.validateHexQuantity:
```
mratsim: if unlikely(...)

nimbus/rpc/hexstrings.nim
```diff
@@ -0,0 +1,111 @@
...
+    raise newException(ValueError, "Parameter \"" & argName & "\" is not valid as an Ethereum hex quantity \"" & hexStr & "\"")
+  result = hexStr.hexQuantityStr
+
+proc fromJson*(n: JsonNode, argName: string, result: var HexDataStr) =
+  # Note that '0x' is stripped after validation
+  n.kind.expect(JString, argName)
+  let hexStr = n.getStr()
+  if not hexStr.validateHexData:
```
mratsim: if unlikely(...)

nimbus/rpc/p2p.nim
```diff
@@ -6,10 +6,42 @@
...
+proc headerFromTag(chain:BaseChainDB, blockTag: string): BlockHeader =
+  let tag = blockTag.toLowerAscii
+  case tag
+  of "latest": result = chain.getCanonicalHead()
+  of "earliest": result = chain.getCanonicalBlockHeaderByNumber(GENESIS_BLOCK_NUMBER)
+  of "pending":
+    #TODO
+    raise newException(ValueError, "Pending tag not yet implemented")
+  else:
+    # Raises are trapped and wrapped in JSON when returned to the user.
```
mratsim: I suppose this is missing a TODO tag.

tests/rpcclient/test_hexstrings.nim
```diff
@@ -0,0 +1,67 @@
+# Separated from main tests for brevity
+
+import unittest, ../../nimbus/rpc/hexstrings, json
+
+proc doHexStrTests* =
+  suite "Hex quantity":
```
mratsim: Can you prefix with `[RPC] Hex quantity`

tests/rpcclient/test_hexstrings.nim
```diff
@@ -0,0 +1,67 @@
...
+    test "Empty string":
```
mratsim: I think actually that we should have:

- "Remote procedure Call" suite
- Hex Quantity / Hex data tests
- everything else in blocks

nimbus/rpc/hexstrings.nim
coffeepots: I don't actually want to do the conversion here, only validate it.

Could put `isHexChar` into `byteutils`? For now I've removed export as it seems unlikely to be used from outside of `hexstrings`.

coffeepots:
@yglukhov @mratsim 

>  It says it works with "quantity" whatever that means, the file name has smth to do with hex, and the things that are done there look pretty like what toHex, parseHex do. Could we reuse/extend existing hex encoding routines instead of inventing the same thing but slightly (if any) different?

The hex string processing is the validation in the RPC spec here: https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding

`byteutils` doesn't quite cover these specifications as it's intended more for actual encoding rather than string verifications. The actual encoding in this PR is parsed by `stint.fromHex`.

There are two different types of hex strings being passed to the RPC calls, and they're treated slightly differently. Differences between `hexstrings` processing and `byteutils` in terms of the spec:

*Quantities*

* The spec allows odd length hex strings (whereas data is only allowed to be even length). `byteutils` asserts the length must be even.
* There must be at least one zero after the prefix (ie; zero = "0x0")

*Data*

* Data must be an even length, which `byteutils` covers but does so using an `assert`, so this would need to be changed to `doAssert` as we want to fail on incorrect data,
* There must be no leading zeros, which `byteutils` will ignore within `skip0xPrefix`.

Additionally, it's a fail if either hex type doesn't start with the "0x" prefix, where `byteutils` simply skips them.

I can always update `byteutils` with these constraints, however they seem rather Ethereum & RPC specific.

PR #97, opened 2018-08-03
Sanitize memory addresses and lengths

tersec:
https://github.com/status-im/nimbus/pull/86 and https://github.com/status-im/nim-stint/issues/58#issuecomment-410215957 have background.

Hopefully the comment blocks in ``cleanMemRef`` explain it adequately. I did check that each replacement is associated with either a memory address or length.

With this fix, every VMTest it runs passes, the only skipped ones being those benchmarking-type performance tests.

### DIFF
--- a/VMTests.md
+++ b/VMTests.md
@@ -295,15 +295,15 @@ OK: 5/5 Fail: 0/5 Skip: 0/5
 + calldatacopyUnderFlow.json                                      OK
 + calldatacopyZeroMemExpansion.json                               OK
 + calldatacopyZeroMemExpansion_return.json                        OK
-- calldatacopy_DataIndexTooHigh.json                              Fail
-- calldatacopy_DataIndexTooHigh2.json                             Fail
-- calldatacopy_DataIndexTooHigh2_return.json                      Fail
-- calldatacopy_DataIndexTooHigh_return.json                       Fail
++ calldatacopy_DataIndexTooHigh.json                              OK
++ calldatacopy_DataIndexTooHigh2.json                             OK
++ calldatacopy_DataIndexTooHigh2_return.json                      OK
++ calldatacopy_DataIndexTooHigh_return.json                       OK
 + calldatacopy_sec.json                                           OK
 + calldataload0.json                                              OK
 + calldataload1.json                                              OK
 + calldataload2.json                                              OK
-- calldataloadSizeTooHigh.json                                    Fail
++ calldataloadSizeTooHigh.json                                    OK
 + calldataloadSizeTooHighPartial.json                             OK
 + calldataload_BigOffset.json                                     OK
 + calldatasize0.json                                              OK
@@ -313,20 +313,20 @@ OK: 5/5 Fail: 0/5 Skip: 0/5
 + callvalue.json                                                  OK
 + codecopy0.json                                                  OK
 + codecopyZeroMemExpansion.json                                   OK
-- codecopy_DataIndexTooHigh.json                                  Fail
++ codecopy_DataIndexTooHigh.json                                  OK
 + codesize.json                                                   OK
 + extcodecopy0.json                                               OK
 + extcodecopy0AddressTooBigLeft.json                              OK
 + extcodecopy0AddressTooBigRight.json                             OK
 + extcodecopyZeroMemExpansion.json                                OK
-- extcodecopy_DataIndexTooHigh.json                               Fail
++ extcodecopy_DataIndexTooHigh.json                               OK
 + extcodesize0.json                                               OK
 + extcodesize1.json                                               OK
 + extcodesizeUnderFlow.json                                       OK
 + gasprice.json                                                   OK
 + origin.json                                                     OK
 ```
-OK: 44/51 Fail: 7/51 Skip: 0/51
+OK: 51/51 Fail: 0/51 Skip: 0/51
 ## vmIOandFlowOperations
 ```diff
 + BlockNumberDynamicJump0_AfterJumpdest.json                      OK
--- a/nimbus/vm/interpreter/opcodes_impl.nim
+++ b/nimbus/vm/interpreter/opcodes_impl.nim
@@ -224,6 +224,22 @@ proc writePaddedResult(mem: var Memory,
   # Note, we don't need to write padding bytes
   # mem.extend already pads with zero properly
 
+func cleanMemRef(x: UInt256): int {.inline.} =
+  # https://github.com/status-im/nim-stint/issues/58
+  # Ensures that only 31 lower bits of x are used to
+  # keep its signed interpretation non-negative, per
+  # Ethereum yellow paper. In general, addresses and
+  # lengths >= 2^31 are beyond what the VM should be
+  # running due to gas costs, block sizes, etc.
+  #
+  # If an opcode uses mem.extend, writePaddedResult,
+  # or an m_handler for gas costs, it should ensure,
+  # either via this proc or some other mechanism, it
+  # checks for .toInt calls it otherwise get checked
+  # for spurious negative interpretations for bounds
+  # checking purposes.
+  x.toInt and (-1.int32 shr 1)
+
 op address, inline = true:
   ## 0x30, Get address of currently executing account.
   push: computation.msg.storageAddress
@@ -249,7 +265,7 @@ op callValue, inline = true:
 op callDataLoad, inline = false, startPos:
   ## 0x35, Get input data of current environment
   # TODO simplification: https://github.com/status-im/nimbus/issues/67
-  let dataPos = startPos.toInt
+  let dataPos = startPos.cleanMemRef
   if dataPos >= computation.msg.data.len:
     push: 0
     return
@@ -278,7 +294,7 @@ op callDataCopy, inline = false, memStartPos, copyStartPos, size:
   ## 0x37, Copy input data in current environment to memory.
   # TODO tests: https://github.com/status-im/nimbus/issues/67
 
-  let (memPos, copyPos, len) = (memStartPos.toInt, copyStartPos.toInt, size.toInt)
+  let (memPos, copyPos, len) = (memStartPos.cleanMemRef, copyStartPos.cleanMemRef, size.cleanMemRef)
 
   computation.gasMeter.consumeGas(
     computation.gasCosts[CallDataCopy].m_handler(computation.memory.len, memPos, len),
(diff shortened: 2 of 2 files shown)


tersec:
I agree with you: this doesn't belong (anywhere near as written, at least) in stint, because it depends on several EVM-specific contextual constraints. For example, yes, stripping the sign bit from -3 results in 1. That's true, and mathematically it's not a particularly correct transformation for most notions of correct, but not a problem in this scenario. Crucially, the end result whenever this proc would have an effect is that opcode failing. It's just a convenient way to catalyze failure detection in a semi-centralized way, rather than a principled transformation of `foo.toInt`.

Concretely, several test cases (all these ones here) push some `0xFFFF_FFFF_..._FFFF_FFFA` or similar bit pattern as a memory address or length, which `.toInt` interprets as (in all the specific cases in VMTests) as -6, which then undermines the gas calculations and memory/array indexing and bypasses bounds checks. These happen to all live near the very upper bound of the Uint256 space, which is why they cause problems.

More broadly, this and your int3 analogy demonstrates that the Nimbus opcode memory address/length handling is even more broken than I've been discussing, with more edge cases that don't happen to appear in test suites that it's been exposed to, but which could occur: it still at best would accidentally correctly handle cases where the least significant limb of the stint was a reasonable value, but the more significant limbs are not. To be clear, it does not fix all the bounds-checking and gas calculation holes, just the 0 +/- high(int32)/2 or so of them, but it does so adding just one additional, branchless, AND instruction.

Translating this into int3-world, in a 1 element array or memory, -3 and 1 (along with -1 and 4, and -2 and -2) are meaningfully equivalent: in each case, both fail. As you point out, what it's doing isn't mathematically correct; it's a heuristic that doesn't, that I can see, have false positives, and has a few true positives that align with the test cases, which tend to trigger these with very high, near-wrap-around addresses and lengths.

I'm suggesting a partial fix which is low-risk, addresses genuine issues, doesn't complicate point-of-call code, and doesn't add ongoing tech debt that's not trivial to remove (intentionally same signature as `stint.toInt`), but the underlying issue lies in using ``foo.toInt`` at all in those contexts.

A better fix looks at the whole stint, checks whether it's greater than, say, 2^31-1, or some other threshold that could be defined in constants.nim, and clamps incoming stints used as memory addresses or lengths to that value. Other options exist too, but 2^31-1 is convenient because in both 32-bit and 64-bit signed integers, those are interpreted as positive, not negative, so it doesn't require changing types anywhere else.

mratsim:
I see, then if the goal is to have the opcode fail fast it would be best to just raise a VMError exception:

https://github.com/status-im/nimbus/blob/6a3fa4cb224ce0734bb65fe11ebb927f3e0d6e0d/nimbus/errors.nim#L24-L27

https://github.com/status-im/nimbus/blob/6a3fa4cb224ce0734bb65fe11ebb927f3e0d6e0d/nimbus/vm/interpreter_dispatch.nim#L223-L231

```Nim
func cleanMemRef(x: UInt256): int {.inline.} =
  ## Sanitize memory addresses, catch negative or impossibly big offsets
  # See https://github.com/status-im/nimbus/pull/97 for more info
  result = x.toInt
  if result > high(int32).int: # This is 0b01111111_...._11111111
    raise newException(VMError, "Invalid memory offset: read " & $x & " (Uint256). " &
      " Interpreted as " & $result & " (int)."
```

VM errors are properly caught during opcode execution and will fail the test.

tersec:
Throwing that exception doesn't work, because I misremembered about the transaction succeeding/failing. It has to succeed, for some definition of success; at least, it can't throw a VMError.

For any of these opcodes which have memory expansion calculations, they must not try to coerce a negative to a Natural, and therefore the least significant limb (the one that in all these examples gets interpreted as -6) has to become positive, and in general, a large-ish positive number.

I go through this case by case below, but I'll summarize:

(1) Any memory-expansion opcode needs to disallow negative `.toInt` or equivalent values because the gas cost calculation involves coercion to `Natural`, which fails via range error.

(2) These are all supposed to work, by virtue of zero-padding copies from invalid address ranges via `mem.extend` via `writePaddedResult`.

(3) As a result of (2), the exact value returned by this sanitization function doesn't much matter, as long as it's large enough to avoid being part of actual address space, but it cannot, ever, be negative, in signed `int` terms.

This is what all of my variations on this theme have accomplished.

I also checked that
```nim
func cleanMemRef(x: UInt256): int {.inline.} =
  ## Sanitize memory addresses, catch negative or impossibly big offsets
  # See https://github.com/status-im/nimbus/pull/97 for more info
  result = x.toInt
  const upperBound = high(int32).int
  const upperBoundFoo = upperBound.u256
  if x > upperBoundFoo:
    return upperBound
```
and
```nim
func cleanMemRef(x: UInt256): int {.inline.} =
  ## Sanitize memory addresses, catch negative or impossibly big offsets
  # See https://github.com/status-im/nimbus/pull/97 for more info
  result = x.toInt
  if result < 0 or result > high(int32).int: # This is 0b01111111_...._11111111
    result = high(int32).int
```
work with current test cases. However,
```nim
func cleanMemRef(x: UInt256): int {.inline.} =
  ## Sanitize memory addresses, catch negative or impossibly big offsets
  # See https://github.com/status-im/nimbus/pull/97 for more info
  result = x.toInt
  ## What you wrote above, without explicit < 0 check
  if result > high(int32).int: # This is 0b01111111_...._11111111
    result = high(int32).int
```
Does not work, because, again, it must not, ever, return a negative number, and by the time it sees `result`, that's already -6, which is less than `high(int32).int`. And `abs` still won't work, because it has to return a high integer.)

The problem is the obfuscatory nature of the illusory/spurious negative interpretations coming from `stuint.toInt`. It's possible to add bounds checks vs 0 everywhere, but the whole point is to centralize handling this condition.

calldatacopy_DataIndexTooHigh:
```
[1] PUSH1 0xff
[34] PUSH32 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffa
[36] PUSH1 0x00
[37] CALLDATACOPY
[39] PUSH1 0x00
[40] MLOAD
[42] PUSH1 0x00
[43] SSTORE
```
This should run to completion. `writePaddedResult` just pads this one out to all zeros due to (if, and only if, that copy position isn't spuriously viewed as a negative number, given a reasonable length) `dataPos + len - 1 < data.len` and `data.len - dataPos > 0` effectively being checked to figure out there's no overlap with the input data.

It would also fail on gas cost calculation where trying to coerce that copy position from -6 (`0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffa.toInt`, sort of) to a Natural fails. So it should succeed, just without mattering much precisely what the copyPos is, i.e. what this cleanMemRef returns, so long as it's large enough.

calldatacopy_DataIndexTooHigh2:
```

[1] PUSH1 0x09
[34] PUSH32 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffa
[36] PUSH1 0x00
[37] CALLDATACOPY
[39] PUSH1 0x00
[40] MLOAD
[42] PUSH1 0x00
[43] SSTORE 
```
Same idea, same success criteria, same detection of lack of overlap, so just padding out with all zeros via `mem.extend`.

calldatacopy_DataIndexTooHigh2_return:
```
[1] PUSH1 0x09
[34] PUSH32 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffa
[36] PUSH1 0x00
[37] CALLDATACOPY
[39] PUSH1 0x00
[40] MLOAD
[42] PUSH1 0x00
[43] SSTORE
[44] MSIZE
[46] PUSH1 0x00
[47] RETURN 
```
Ditto. Zero-padded copy from invalid address range that's not properly detected due to being viewed as negative number. Fails on gas calculation due to Natural coercion.

calldatacopy_DataIndexTooHigh_return:
```
[1] PUSH1 0xff
[34] PUSH32 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffa
[36] PUSH1 0x00
[37] CALLDATACOPY
[39] PUSH1 0x00
[40] MLOAD
[42] PUSH1 0x00
[43] SSTORE
[44] MSIZE
[46] PUSH1 0x00
[47] RETURN
```
More of the same. Zero-padded copy from invalid address range, if `writePaddedResult` properly picks it up. Should work. Fails becaue of gas cost calculation type coercion.

calldataloadSizeTooHigh:
```
[32] PUSH32 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffa
[33] CALLDATALOAD
[35] PUSH1 0x00
[36] SSTORE 
```
Because its bounds checking only checks for `dataEndPosition < computation.msg.data.len` i.e. `dataPos + 31 < computation.msg.data.len`, i.e. `-6 + 31 < computation.msg.data.len`, it thinks it can go down the non-error path. It's confused by the spurious negative, because none of these numbers are ever really negative except as interpreted through Nim's signed int types via `stuint.toInt`. It's supposed to succeed, and just zero-pads the complete result (pre-existing code in callDataLoad) rather than read out-of-bounds. To trigger this, one just has to make sure that cleanMemRef returns a non-negative-interpreted number higher than computation.msg.data.len.

codecopy_DataIndexTooHigh:
```
[1] PUSH1 0x08
[34] PUSH32 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffa
[36] PUSH1 0x00
[37] CODECOPY
[39] PUSH1 0x00
[40] MLOAD
[42] PUSH1 0x00
[43] SSTORE
```
Another zero-padded copy from an absurd address range, one which gets interpreted, totally spuriously, as starting with -6 right now, thus just cutting off the sign bit. Again, in Ethereum terms, that's a large positive integer, not a negative number, and the type system-created confusion here isn't helping. Even viewing this as an unsigned int needs wraparound detection (implemented in a previous PR which had other issues) if the top part of the range isn't truncated to perform that check automatically.

extcodecopy_DataIndexTooHigh:
```
[1] PUSH1 0x08
[34] PUSH32 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffa
[36] PUSH1 0x00
[37] ADDRESS
[38] EXTCODECOPY
[40] PUSH1 0x00
[41] MLOAD
[43] PUSH1 0x00
[44] SSTORE 
```
Same idea as for codecopy_DataIndexTooHigh.

mratsim:
Thanks for the deep investigation. This sounds good to me:

```Nim
func cleanMemRef(x: UInt256): int {.inline.} =
  ## Sanitize memory addresses, catch negative or impossibly big offsets
  # See https://github.com/status-im/nimbus/pull/97 for more info
  const upperBound = high(int32).u256
  if x > upperBound:
    return high(int32)
  return x.toInt
```

I can't help but feel that `cleanMemRef` and `writePaddedResult` are overcomplicated though.
In the future I think cleanMemRef should be called inside writePaddedResult as it's special logic only related to that proc. And hopefully we can sanitize writePaddedResult as well: https://github.com/status-im/nimbus/issues/67

tersec:
@mratsim the reason `cleanMemRef` can't currently be effectively part of `writePaddedResult` is that gas cost calculation runs first, and has to (to catch real gas cost violations). It coerces, via its own type signature, the `cleanMemRef`-relevant values to `Natural`, resulting in a `RangeError`. Thus, if instead of `toInt`, those opcode functions which now call `cleanMemRef` called a function in `stint` which never returned a negative number, then this logic could, indeed, be localized in `writePaddedResult`.

@coffeepots This is more correct, yes.

PR #99, opened 2018-08-07
Run VMTests during CI and drop VMTests.md

tersec:


coffeepots:
The issue blocking CI has been fixed now here :) https://github.com/nim-lang/Nim/pull/8567

PR #100, opened 2018-08-08
Fix the genesis state hash computation

zah:
The account database needed to use the secure flavour of the Hexary
Trie. This means all keys are SHA3 hashed before being inserted.

### DIFF
--- a/nimbus/genesis.nim
+++ b/nimbus/genesis.nim
@@ -1,6 +1,6 @@
 import db/[db_chain, state_db], genesis_alloc, eth_common, tables, stint,
     byteutils, times, config, rlp, ranges, block_types, eth_trie,
-    eth_trie/memdb, constants, nimcrypto, chronicles
+    eth_trie/memdb, account, constants, nimcrypto, chronicles
 
 type
   Genesis* = object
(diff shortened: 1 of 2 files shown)


nimbus/genesis.nim
```diff
@@ -70,22 +70,17 @@ proc toBlock*(g: Genesis): BlockHeader =
...
-  block tempRootHashStub: # TODO: Remove this block when we calculate the root hash correctly
-    if g.config.chainId == 1:
-      const correctMainnetRootHash = toDigest("d7f8974fb5ac78d9ac099b9ad5018bedc2ce0a72dad1827a1709da30580f0544")
-      if root != correctMainnetRootHash:
-        error "Root hash incorrect. Stubbing it out."
-        root = correctMainnetRootHash
-      else:
-        error "Yay! Root hash is correct. Please remove the block where this message comes from."
+  var root = sdb.rootHash
+  doAssert $root == "D7F8974FB5AC78D9AC099B9AD5018BEDC2CE0A72DAD1827A1709DA30580F0544"
```
yglukhov: The assertion should be done only if `g.config.chainId == 1`

nimbus/genesis.nim
```diff
@@ -70,22 +70,17 @@ proc toBlock*(g: Genesis): BlockHeader =
 
   for address, account in g.alloc:
     sdb.setBalance(address, account.balance)
-    sdb.setCode(address, account.code.toRange)
-    sdb.setNonce(address, account.nonce)
 
-    for k, v in account.storage:
-      sdb.setStorage(address, k, v)
+    when false:
```
yglukhov: I'd prefer keeping it enabled (optionally add empty checks), because we'll definitely have custom genesis support at some point.

nimbus/genesis.nim
yglukhov: Actually why not just remove it. We have a test for this.

nimbus/genesis.nim
zah: The extra DB operations were making me nervous, but I've added guards now that will make the extra calls benign.

zah:
Review comments addressed

Issue #41, opened 2018-05-25
Support Gas costs and balance >= 2^31 on 32-bit


mratsim:
#40 is introduce native integer for gas prices and balances via int64.

Internally it uses [`toInt` proc from stint](https://github.com/status-im/nim-stint/blob/8e25b30209e951bcabfc2d1cd97ef0b971554648/stint/io.nim#L64-L66) implemented as follow:

```Nim
func toInt*(num: Stint or StUint): int {.inline.}=
  # Returns as int. Result is modulo 2^(sizeof(int)
  num.data.least_significant_word.int
```

On 32-bit, int32 would be returned instead of int64.

TODO: introduce a `toInt64` conversion in Stint

Optionally, `toInt64` might be wrapped in Nimbus in a `toGasInt` proc that will add logging and range check for example to make sure a gas value on the stack (uint256) is not over 2^63 - 1 before conversion.

jangko:
this has been long fixed. closing

PR #101, opened 2018-08-08
Fix for VMTest parseHex in 32 bit

coffeepots:
Fixes some of the failed VM tests.
Put simply, parseUtils.parseHex cannot handle large hex values when running in 32 bit as it returns an `int`.

This was causing the gas values to be zero and therefore the test to fail.

There are more fixes to come in this PR for 32 bit due to `int` overflow.


coffeepots:
This PR relies on the fix in https://github.com/status-im/nim-stint/pull/60

tersec:
While it seems unobjectionable as a fix for this issue, as such, I wonder more broadly whether what
```nim
      prev_words: int64 = currentMemSize.wordCount
      new_words: int64 = (memOffset + memLength).wordCount
```
fixes should even be allowed, especially on mobile devices. Probably better to enable the code to support it, as this patch does, and disallow it elsewhere, however.

To be clear, not an objection to this patch as such, but it reveals assumptions baked into Nimbus which might not be ideal.

mratsim:
Good catch.

Maybe like GasInt https://github.com/status-im/nim-eth-common/blob/master/eth_common/eth_types.nim#L19
```Nim
type
  GasInt* = int64
```

We should introduce a `WordInt` or `Int` type that is forced to int64 and used everywhere.
This way it's easy to switch between int and int64 when needed.

Linked to #41 and https://github.com/status-im/nim-stint/issues/48

coffeepots:
If I recall correctly, the fix @tersec mentions above is actually only 'needed' to pass some of the out of range tests that are designed to fail. Because they stuff large numbers into `memOffset` & `memLength`, the calculation overflows and raises a different error to the one expected. So it's failing "wrong" even though it is supposed to fail. Of course, we don't actually need >`int32` to function correctly.

Now that status-im/nim-stint#60 is merged, this PR should make win32 pass all VM tests again.

tersec:
I'll suggest another approach which might keep 32-bit code away from some of these edge cases, upstream:
```nim
--- a/nimbus/vm/interpreter/opcodes_impl.nim
+++ b/nimbus/vm/interpreter/opcodes_impl.nim
@@ -235,9 +235,9 @@ proc writePaddedResult(mem: var Memory,
 func cleanMemRef(x: UInt256): int {.inline.} =
   ## Sanitize memory addresses, catch negative or impossibly big offsets
   # See https://github.com/status-im/nimbus/pull/97 for more info
-  const upperBound = high(int32).u256
+  const upperBound = (high(int32) shr 5).u256
   if x > upperBound:
-    return high(int32)
+    return high(int32) shr 5
   return x.toInt
```

Bound `cleanMemRef`'s results (which are upstream of data flow for that `int64` change) cleanly within signed int32's positive range, and then adding two int32's won't overflow. I've tested `shr 1`, `shr 2`, `shr 3`, `shr 4`, and `shr 5`. It breaks in a couple of test cases on `shr 6`, so that's too far.

(Code above is strawman way of doing that, et cetera.)

coffeepots:
@tersec It would be good to ensure things don't break upstream with unexpected `int32` problems by constraining `cleanMemRef`. Some thoughts: 

* Curious to know which tests break with `shr 6`,
* If `high(int32) shr 6` (`33,554,431`) breaks tests, perhaps we should anticipate the limits might be more expansive in future and with this we're reducing possible bandwidth. Of course, having an offset of over `33,554,431` seems unlikely, although apparently some tests do require it.
* I wonder if this is better covered with additional tests and enforced specific `int` size as @mratsim mentions above with `GasInt`/`WordInt`, my thought being that going forward we may even need to support smaller architectures such as 16 bit where even `shr 5` fix may still fail if we're using `int`.

tersec:
@coffeepots 
```
vmIOandFlowOperationscalldatacopyMemExp.json
    $HOME/nimbus/tests/test_vm_json.nim(112, 23): Check failed: computation.isError
  [FAILED] tests/fixtures/VMTests/vmIOandFlowOperations/calldatacopyMemExp.json
```

```
vmIOandFlowOperationsmstoreMemExp.json
    $HOME/nimbus/tests/test_vm_json.nim(112, 23): Check failed: computation.isError
  [FAILED] tests/fixtures/VMTests/vmIOandFlowOperations/mstoreMemExp.json
```

```
vmIOandFlowOperationscalldatacopyMemExp.json
    $HOME/nimbus/tests/test_vm_json.nim(112, 23): Check failed: computation.isError
  [FAILED] tests/fixtures/VMTests/vmIOandFlowOperations/calldatacopyMemExp.json
```

```
vmIOandFlowOperationscodecopyMemExp.json
    $HOME/nimbus/tests/test_vm_json.nim(112, 23): Check failed: computation.isError
  [FAILED] tests/fixtures/VMTests/vmIOandFlowOperations/codecopyMemExp.json
```

There's an clear theme.

I agree that limits in the future might be more expansive. For 16-bit support, at some point some beyond-16-bit-arithmetic will be quite helpful, to the extent that I'm not sure how much it makes sense to adapt this to int types smaller than `int32`, even on 16-bit platforms.  Regardless, yes, having tests for `int` behavior specifically rather than relying on random VMTests fixtures to reveal indirectly them would help in maintaining portability.

One middle ground is to just limit things to, say high(int32) shr 1 or shr 2, if that helps for now. That's quite a bit of expansion of limits before it becomes a problem.

tersec:
@coffeepots looks good to me; `shr 2` is a nice balance.

PR #104, opened 2018-08-14
Initial EVMC message type API matching

tersec:
https://github.com/status-im/nimbus/issues/61

-reorder overlapping fields to match EMVC;
-rename "to" to destination;
-replace isStatic (not in EVMC) with more general flags (with only one non-zero value);
-remove superfluous-seeming shouldTransferValue (not in EVMC);
-keep internalFoo fields, which aren't in PyEVM either, so weren't per se, it looks like, part of PyEVM matching

It appears to have been all plumbing, no real implementation of the modified fields (e.g., isStatic was never checked), so it doesn't change much executionwise at the moment.

I tried to balance specificity of names/types (e.g., "MsgFlags") with matching EMVC more verbatim.

### DIFF
--- a/nimbus/vm/message.nim
+++ b/nimbus/vm/message.nim
@@ -23,16 +23,14 @@ proc newMessageOptions*(
     depth: int = 0,
     createAddress = ZERO_ADDRESS,
     codeAddress = ZERO_ADDRESS,
-    shouldTransferValue: bool = true,
-    isStatic: bool = false): MessageOptions =
+    flags: MsgFlags = static(emvcNoFlags)): MessageOptions =
 
   result = MessageOptions(
     origin: origin,
     depth: depth,
     createAddress: createAddress,
     codeAddress: codeAddress,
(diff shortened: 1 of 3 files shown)


nimbus/vm/message.nim
```diff
@@ -23,16 +23,14 @@ proc newMessageOptions*(
     depth: int = 0,
     createAddress = ZERO_ADDRESS,
     codeAddress = ZERO_ADDRESS,
-    shouldTransferValue: bool = true,
-    isStatic: bool = false): MessageOptions =
+    flags: MsgFlags = static(emvcNoFlags)): MessageOptions =
```
zah: Just curious, why did you put `static` in here? `static` can be used to force the compile-time evaluation of an expression, but that's not necessary here.

tersec: Initially, it had been a slightly more complicated expression with explicit type coercision. It was an oversight, which I agree isn't particularly what was intended.

PR #105, opened 2018-08-14
Raise KeyError when key not found

yglukhov:


zah:
If this case must be discriminated somehow, wouldn't it be better to return an `Option` instead? Are you handling this `KeyError` as a recoverable error somewhere?

yglukhov:
Yes, there needs to be a nothrow variant ideally. The interface goes a long way through `TrieDBRef` so this will need to be addressed separately eventually.

PR #116, opened 2018-08-21
travis: use ccache

arnetheduck:
Changes the travis and appveyor build to:
* use HEAD from https://github.com/status-im/Nim - the idea is that we keep it locked at a working nim version
* use ccache to build dependencies
* doesn't use Docker any more - this means the build on OSX and linux builds work the same way

Downsides:
* docker image contained a single source of truth of how to set up a "nimbus" build env.. with many repos, this needs to be replicated somehow
  * ideally, build stuff such as the rocksdb build in the linux env would be handled by the wrapper and transitively applied by nimble to everything that uses rocksdb - this would allow nimble to become the single source of truth. we're not there yet I think - an alternative might be to do some script hackery, but that's going to be fragile..

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -54,7 +54,7 @@ install:
   - koch nimble
 build_script:
   - cd C:\projects\%APPVEYOR_PROJECT_SLUG%
-  - nimble install -y
+  - nimble install -dy
 test_script:
   - nimble test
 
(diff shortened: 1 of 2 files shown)


cheatfate:
@arnetheduck how it works with cache when build/test process gets an error, is cache become invalidated or no?

arnetheduck:
there's a cache per branch/pr and it uses the latest working one.. lacking that, it uses the cache from master: https://docs.travis-ci.com/user/caching/

PR #118, opened 2018-08-24
Whisper stubs, fleshing out p2p RPCs

coffeepots:
This PR adds stubs for the whisper RPCs and their types, and fleshes out some more details that were TODO in other RPC calls such as fetching sender addresses for transactions and receipts.

### DIFF
--- a/nimbus/rpc/hexstrings.nim
+++ b/nimbus/rpc/hexstrings.nim
@@ -17,6 +17,7 @@
     * ref EthAddress
     * Hash256
     * UInt256
+    * seq[byte]
     * openArray[seq]
     * ref BloomFilter
 ]#
@@ -28,11 +29,10 @@ type
   HexDataStr* = distinct string
   EthAddressStr* = distinct string  # Same as HexDataStr but must be less <= 20 bytes
   EthHashStr* = distinct string     # Same as HexDataStr but must be exactly 32 bytes
+  WhisperIdentity* = distinct string # 60 bytes
+  HexStrings = HexQuantityStr | HexDataStr | EthAddressStr | EthHashStr | WhisperIdentity
 
-func len*(quantity: HexQuantityStr): int = quantity.string.len
-func len*(data: HexDataStr): int = data.string.len
-func len*(data: EthAddressStr): int = data.string.len
-func len*(data: EthHashStr): int = data.string.len
+func len*(value: HexStrings): int = value.string.len
 
 # Hex validation
 
@@ -91,11 +91,17 @@ func isValidEthHash*(value: string): bool =
   # TODO: Allow shorter hashes (pad with zeros) for convenience?
   result = value.len == 66 and value.isValidHexData
 
+func isValidWhisperIdentity*(value: string): bool =
+  # 60 bytes for WhisperIdentity plus "0x"
+  # TODO: Are the HexData constratins applicable to Whisper identities?
+  result = value.len == 122 and value.isValidHexData
+
 const
   SInvalidQuantity = "Invalid hex quantity format for Ethereum"
   SInvalidData = "Invalid hex data format for Ethereum"
   SInvalidAddress = "Invalid address format for Ethereum"
   SInvalidHash = "Invalid hash format for Ethereum"
+  SInvalidWhisperIdentity = "Invalid format for whisper identity"
 
 proc validateHexQuantity*(value: string) {.inline.} =
   if unlikely(not value.isValidHexQuantity):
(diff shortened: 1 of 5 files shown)


nimbus/rpc/p2p.nim
```diff
@@ -192,12 +246,17 @@ proc setupP2PRPC*(node: EthereumNode, rpcsrv: RpcServer) =
...
     ## Returns the code from the given address.
     let
-      accountDb = accountDbFromTag(quantityTag)
+      accountDb = getAccountDbFromTag(quantityTag)
       addrBytes = strToAddress(data.string)
       storage = accountDb.getCode(addrBytes)
     # Easier to return the string manually here rather than expect ByteRange to be marshalled
     result = byteutils.toHex(storage.toOpenArray).HexDataStr
 
+  template sign(privateKey: PrivateKey, message: string): string =
```
zah: This should be a `func` because it use the `message` parameter twice.

nimbus/rpc/p2p.nim
```diff
@@ -33,6 +33,53 @@ func toHash(value: array[32, byte]): Hash256 {.inline.} =
...
+  return TransHashObj(
+    accountNonce: transaction.accountNonce,
+    gasPrice: transaction.gasPrice,
+    gasLimit: transaction.gasLimit,
+    to: transaction.to,
+    value: transaction.value,
+    payload: transaction.payload
+    ).rlpHash
+
+template toSignature(transaction: Transaction): Signature =
```
zah: This should be a `func` because it uses the `transaction` parameter multiple times (if you a pass a call with side effects, it will be executed multiple times).

nimbus/rpc/hexstrings.nim
```diff
@@ -26,13 +27,12 @@ import eth_common/eth_types, stint, byteutils, nimcrypto
...
+  EthAddressStr* = distinct string        # Same as HexDataStr but must be less <= 20 bytes
+  EthHashStr* = distinct string           # Same as HexDataStr but must be exactly 32 bytes
+  WhisperIdentityStr* = distinct string   # 60 bytes
+  HexStrings = HexQuantityStr | HexDataStr | EthAddressStr | EthHashStr | WhisperIdentityStr
 
-func len*(quantity: HexQuantityStr): int = quantity.string.len
-func len*(data: HexDataStr): int = data.string.len
-func len*(data: EthAddressStr): int = data.string.len
-func len*(data: EthHashStr): int = data.string.len
+func len*(value: HexStrings): int = value.string.len
```
zah: This may be a template, because the body is so simple and the `HexStrings` param is a generic type which will activate a more expensive compilation machinery in the compiler that will need to instantiate the function for each unique distinct type.

nimbus/rpc/p2p.nim
```diff
@@ -260,7 +321,8 @@ proc setupP2PRPC*(node: EthereumNode, rpcsrv: RpcServer) =
       startIdx += 32    
     result.sha3Uncles = keccak256.digest(rawData)
 
-    result.logsBloom = nil  # TODO: Create bloom filter for logs
+    result.logsBloom = new BloomFilter
```
zah: Why is this kept by reference btw? Pretty good chance it's a legacy from the Python code.

nimbus/rpc/p2p.nim
coffeepots: Unfortunately `sign` too can have side effects so cannot be a `func`.

nimbus/rpc/p2p.nim
zah: The `Option` type can be supported with similar semantics. You woudn't have to allocate separate objects then, but these are minor points (not that important to change).

nimbus/rpc/p2p.nim
zah: Well, ok, it could be a `proc` then?

nimbus/rpc/p2p.nim
coffeepots: Good idea, I'll look at doing this for the next PR.

nimbus/rpc/p2p.nim
coffeepots: `toSignature` is now a `proc` :)

Do you want me to do this for the other templates too?
Namely `balance` and `sign`? I'm assuming `chain` is probably okay as a template?

nimbus/rpc/p2p.nim
coffeepots: @zah Option support added to `nim-json-rpc`: https://github.com/status-im/nim-json-rpc/pull/32

coffeepots: @zah I have updated this PR locally to support Option[T], will commit it here if the nim-json-rpc PR is accepted.

coffeepots:
@zah updated options stuff here now that nim-json-rpc PR is merged.

PR #119, opened 2018-08-24
Remove BaseTransaction object, add transaction.getSender()

coffeepots:


nimbus/transaction.nim
```diff
@@ -6,38 +6,62 @@
...
+  bytes[32..63] = transaction.S.toByteArrayBE()
+  #[
+    TODO: In the yellow paper:
+      It is assumed that v is the ‘recovery id’, a 1 byte value
+      specifying the sign and finiteness of the curve point; this
+      value is in the range of [27,30].
+    Does this need to be checked that it is [0, 1] and inc by 27?
+  ]#
+  # TODO: V will become a byte or range soon.
+  bytes[64] = (cast[uint64](transaction.V.data.lo) and 0xff'u64).uint8
```
mratsim: `0xff'u64` is not needed, converting to uint8 will take care of that already

I think V should be `range[byte 0 .. 1]` or `range[byte 27 .. 28]`

and R, S: `array[32, bytes]`

coffeepots: Agreed, that would be a separate PR though as it's a different repo (nim-eth-common).

coffeepots: Removed `and 0xff`. We need to fix `V` in `Transaction` ASAP if we are to test.

nimbus/transaction.nim
```diff
@@ -6,38 +6,55 @@
...
+  bytes[64] = cast[uint64](transaction.V.data.lo).uint8
+  initSignature(bytes)
+
+proc getSender*(transaction: Transaction, output: var EthAddress): bool =
+  ## Find the address the transaction was sent from.
+  let
+    txHash = transaction.hash # hash without signature
+    sig = transaction.toSignature()
+  var pubKey: PublicKey
+  if recoverSignatureKey(sig, txHash.data, pubKey) != EthKeysStatus.Success:
```
yglukhov: `==`?

coffeepots: Haha omg! Fixed :)

PR #124, opened 2018-08-29
Separate Chain from ChainDB

yglukhov:
* Subj
* Dry up RPC code, decouple it from `AbstractChainDB`, use `BaseChainDB` instead. CC @coffeepots 
* Always START_SYNC

Depends on https://github.com/status-im/nim-eth-p2p/pull/34

### DIFF
--- a/nimbus/rpc/p2p.nim
+++ b/nimbus/rpc/p2p.nim
@@ -47,14 +47,16 @@ func headerFromTag(chain:BaseChainDB, blockTag: string): BlockHeader =
     let blockNum = stint.fromHex(UInt256, tag)
     result = chain.getBlockHeader(blockNum)
 
-proc setupP2PRPC*(node: EthereumNode, rpcsrv: RpcServer) =
-  template chain: untyped = BaseChainDB(node.chain) # TODO: Sensible casting
-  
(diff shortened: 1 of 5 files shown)


nimbus/rpc/p2p.nim
```diff
@@ -47,15 +47,17 @@ func headerFromTag(chain:BaseChainDB, blockTag: string): BlockHeader =
...
+proc setupEthRpc*(node: EthereumNode, chain: BaseChainDB, rpcsrv: RpcServer) =
   proc accountDbFromTag(tag: string, readOnly = true): AccountStateDb =
     let
       header = chain.headerFromTag(tag)
       vmState = newBaseVMState(header, chain)
     result = vmState.chaindb.getStateDb(vmState.blockHeader.hash, readOnly)
 
+  proc getBlockBody(hash: KeccakHash): BlockBody =
+    if not chain.getBlockBody(hash, result):
+      raise newException(ValueError, "Cannot find hash")
```
coffeepots: Maybe this can be `"Cannot find block for hash " & $hash` to be a bit more specific. Originally I'd intended to make each message unique, but there is no need and your approach is a lot cleaner.

Issue #133, opened 2018-09-04
[Info] Contract address collision


mratsim:
cc @coffeepots 

So that this is not lost within our sea of discussions: https://github.com/status-im/nimbus/blob/e1bc482f398db393b6f0674fdc349d4970808807/nimbus/vm/interpreter/opcodes_impl.nim#L553-L560

EIP-684

![2018-09-04_11-50-29](https://user-images.githubusercontent.com/22738317/45024342-1236b600-b039-11e8-957d-2054d495d4f0.png)

![2018-09-04_11-50-50](https://user-images.githubusercontent.com/22738317/45024350-1531a680-b039-11e8-9447-ca08e7f3d103.png)

coffeepots:
The account nonce is now touched on contract address creation, so even if the new address ends up as a collision, the nonce gets incremented.

PR #134, opened 2018-09-04
Add to Create Op, generateAddress

coffeepots:
This PR adds the `generateAddress` function and fills in contract address generation for the Create op code.

### DIFF
--- a/nimbus/utils/addresses.nim
+++ b/nimbus/utils/addresses.nim
@@ -1,0 +1,4 @@
+import nimcrypto, eth_common, rlp
+
+proc generateAddress*(address: EthAddress, nonce: AccountNonce): EthAddress =
+  result[0..19] = keccak256.digest(rlp.encodeList(address, nonce).toOpenArray).data[12..31]
--- a/nimbus/utils/addresses.nim
+++ b/nimbus/utils/addresses.nim
@@ -1,4 +1,4 @@
 import nimcrypto, eth_common, rlp
 
 func generateAddress*(address: EthAddress, nonce: AccountNonce): EthAddress =
-  result[0..19] = keccak256.digest(rlp.encodeList(address, nonce).toOpenArray).data[12..31]
+  result[0..19] = keccak256.digest(rlp.encodeList(address, nonce).toOpenArray).data.toOpenArray(12, 31)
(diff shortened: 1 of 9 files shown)


mratsim:
Link #133

nimbus/utils/addresses.nim
```diff
@@ -0,0 +1,4 @@
+import nimcrypto, eth_common, rlp
+
+func generateAddress*(address: EthAddress, nonce: AccountNonce): EthAddress =
+  result[0..19] = keccak256.digest(rlp.encodeList(address, nonce).toOpenArray).data[12..31]
```
mratsim: ah you can use `.data.toOpenArray(12, 31)` here

coffeepots: Hmm I tried this but its longer and doesn't compile on my end:

    func generateAddress*(address: EthAddress, nonce: AccountNonce): EthAddress =
      result[0..19] = keccak256.digest(rlp.encodeList(address, nonce).toOpenArray).data.toOpenArray(12..31)

`addresses.nim(4, 84) Error: type mismatch: got <array[0..31, byte], HSlice[system.int, system.int]>`

coffeepots:
On further investigation:

### Nonce definition:

> A scalar value equal to the number of transactions  sent  from  this  address  or, in  the  case of accounts with associated code, the **number of contract-creations** made by this account. Note we use one fewer than the sender’s nonce value; we assert that we have incremented the sender account’s nonce **prior** to this call, and so the **value used is the sender’s nonce at the beginning of the responsible transaction or VM operation**. The account’s nonce is initially defined as one

In other words, nonce incrementation preceeds what is used as the 'active' value.

    let currentNonce = db.getNonce(storageAddress)
    db.setNonce(storageAddress, currentNonce + 1)

However, we currently assume Nonce starts at zero: https://github.com/status-im/nimbus/blob/9e1be6438e724231da9b2079b8c951b79e28dbe1/nimbus/account.nim#L18

nimbus/utils/addresses.nim
cheatfate: @coffeepots but in your attempt `toOpenArray(12..31)` while @mratsim ask you to call `toOpenArray(12, 31)`.

coffeepots: Ah I missread.

What is the advantage of using `toOpenArray` here? Hiding implementation details of `data`?

    func generateAddress*(address: EthAddress, nonce: AccountNonce): EthAddress =
      result[0..19] = keccak256.digest(rlp.encodeList(address, nonce).toOpenArray).data.toOpenArray(12, 31)

PR #135, opened 2018-09-04
Assorted fixes towards GeneralStateTests working

tersec:
1. fix overflow/signed confusion when reading large-valued storage entries
2. adjust test fixture validity for GeneralStateTests
3. remove incorrect usage of state trie for code (not codehash) storage
4. add ability to display statedb information for an account;
5. avoid double-hashing keys for already-hashed SecureHexaryTrie for storage;
6. add intentionally local workaround for empty code in GeneralStateTest fixtures lacking 0x prefix
7. fix nonce parsing, which is always hex string in test fixtures

This doesn't fix everything -- there are a couple of additional items. It fixes most of the outstanding issues blocking GeneralStateTests usage.

### DIFF
--- a/nimbus/db/state_db.nim
+++ b/nimbus/db/state_db.nim
@@ -6,7 +6,7 @@
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
 import
-  sequtils, tables,
+  sequtils, strformat, tables,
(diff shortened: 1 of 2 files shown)


tersec:
https://github.com/status-im/nimbus/issues/113

PR #138, opened 2018-09-06
Various fixes towards state reconstruction

yglukhov:
Now we consistently reach 46147, first block with a transaction :)

Notable changes:
- DB backends return empty seqs if key not found. Hopefully its the last time i'm changing this :) CC @coffeepots 
- Switched db backend to rocksdb, sqlite is barely usable because it is too slow
- Changed `Message.code` type from `string` to `seq[byte]`
- "Fixed" (or should I say mocked) `Transaction.toSignature` so that it works with frontier transactions CC @coffeepots 
- Various fixes here and there.

### DIFF
--- a/nimbus/p2p/chain.nim
+++ b/nimbus/p2p/chain.nim
@@ -1,5 +1,5 @@
-import ../db/db_chain, eth_common, chronicles, ../vm_state, ../vm_types, ../transaction,
-  ../vm/[computation, interpreter_dispatch, message]
+import ../db/[db_chain, state_db], eth_common, chronicles, ../vm_state, ../vm_types, ../transaction, ranges,
+  ../vm/[computation, interpreter_dispatch, message], ../constants
 
 
 type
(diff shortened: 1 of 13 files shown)


nimbus/p2p/chain.nim
```diff
@@ -35,17 +35,28 @@ method persistBlocks*(c: Chain, headers: openarray[BlockHeader], bodies: openarr
...
+      if bodies[i].transactions.len != 0:
+        # echo "block: ", headers[i].blockNumber
+        for t in bodies[i].transactions:
+          var sender: EthAddress
+          if t.getSender(sender):
+            echo "Sender: ", sender
+            let code = stateDb.getCode(sender)
+            debug "Transaction", sender, to = t.to, value = t.value, hasCode = code.len != 0
+            let msg = newMessage(t.gasLimit, t.gasPrice, t.to, sender, t.value, t.payload, code.toSeq)
+      assert(false, "Dont know how to persist transactions")
```
coffeepots: Is this meant to be behind an `else`, or just as a todo to say more work is required?

yglukhov: Yes, that's a todo i'm working on.

PR #139, opened 2018-09-07
Flesh out call op

coffeepots:
This PR updates the call op:

* Now checks balance against funds,
* Fetches code instead of empty seq,
* Generates the child computation (still some work to do here as forks are handled differently).

### DIFF
--- a/nimbus/vm/computation.nim
+++ b/nimbus/vm/computation.nim
@@ -168,6 +168,10 @@ proc generateChildComputation*(fork: Fork, computation: BaseComputation, childMs
       computation.vmState,
       computation.vmState.blockHeader.blockNumber,
       childMsg)
+  
+  # Copy the fork op code executor proc (assumes child computation is in the same fork)
+  childComp.opCodeExec = computation.opCodeExec
+
   if childMsg.isCreate:
     fork.applyCreateMessage(childComp)
   else:
--- a/nimbus/vm/computation.nim
+++ b/nimbus/vm/computation.nim
@@ -85,7 +85,7 @@ proc prepareChildMessage*(
     code,
     childOptions)
 
-proc applyFrontierMessage(computation: var BaseComputation) =
+proc applyMessage(computation: var BaseComputation) =
   # TODO: Snapshots/Journaling, see: https://github.com/status-im/nimbus/issues/142
   #let snapshot = computation.vmState.snapshot()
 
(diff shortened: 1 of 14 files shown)


nimbus/vm/computation.nim
```diff
@@ -70,19 +84,46 @@ proc prepareChildMessage*(
...
     childOptions)
 
-func output*(c: BaseComputation): seq[byte] =
-  if c.shouldEraseReturnData:
-    @[]
-  else:
-    c.rawOutput
+proc applyCreateMessage(computation: BaseComputation): BaseComputation =
+  # TODO: This needs to be different depending on fork.
+  raise newException(NotImplementedError, "Apply create message not implemented")
```
zah: Do you plan to work on these stubs? I guess they can be finished before this is merged.

It's easy to detect the correct fork by the `computation.vmState.blockHeader.blockNumber` variable.

The forks and their starting block numbers are summarized here:
https://ethereum.stackexchange.com/questions/13014/please-provide-a-summary-of-the-ethereum-hard-forks

You've probably already noticed how `apply_create_message` exists in several copies in `py-evm` defined in a separate file per fork.

nimbus/vm/computation.nim
```diff
@@ -70,19 +84,46 @@ proc prepareChildMessage*(
...
+  raise newException(NotImplementedError, "Apply message not implemented")
 
-proc outputHex*(c: BaseComputation): string =
-  if c.shouldEraseReturnData:
-    return "0x"
-  c.rawOutput.bytesToHex
+proc generateChildComputation*(computation: BaseComputation, childMsg: Message): BaseComputation =
+  new result
+  if childMsg.isCreate:
+    result = newBaseComputation(
```
zah: You could extract `newBaseComputation(computation.vmState, computation.vmState.blockHeader.blockNumber, childMsg)` in a variable to make the code prettier.

nimbus/vm/computation.nim
coffeepots: I'll finish the stubs before merging.

nimbus/db/backends/caching_backend.nim
```diff
@@ -0,0 +1,44 @@
+import ranges, eth_trie, tables, sets
+import ../storage_types
+
+type
+  CachingDB* = ref object of RootObj
+    backing: TrieDatabaseRef
+    changed: Table[seq[byte], seq[byte]]
+    deleted: HashSet[seq[byte]]
```
VoR0220: I was looking for this. Cool. Now it's implemented :)

coffeepots: Yes but I haven't actually used it in this PR yet :)

coffeepots:
@mratsim Can you check that the gas calculation is done correctly here? https://github.com/status-im/nimbus/pull/139/commits/c7a014bb102d78657c09a72909d4ecb713a5890d

zah:
@mratsim, I can answer these questions, because Ryan discussed with me some problems he ran into.

`opCodeExec` was added to the computation object in order to resolve a cyclic module dependency between `interpreter_dispatch` and `opcodes_impl` (`CALL` needs `executeOpcodes` which must be generated in `interpreter_dispatch`, which in turn must generate calls to `CALL`). 

The alternative approach would be to merge these modules into a single larger module that relies on forward declarations. 

Semantically, `opCodeExec` means "execute the bytecode using the correct dispatching logic for the current block"

With that said, `applyFrontierMessage` can be named just `applyMessage` because we are supposed to handle the fork-specific logic inside the procs using `if` and `when` statements. Also, `applyMessage` is not really fork-dependent - there are differences between the forks are only in `applyCreateMessage`.

nimbus/vm/computation.nim
```diff
@@ -114,36 +114,43 @@ proc applyMessage(computation: var BaseComputation) =
...
       if fork >= FkSpurious and contractCode.len >= EIP170_CODE_SIZE_LIMIT:
-        # TODO: Revert snapshot
         raise newException(OutOfGas, &"Contract code size exceeds EIP170 limit of {EIP170_CODE_SIZE_LIMIT}.  Got code of size: {contractCode.len}")
 
       try:
         computation.gasMeter.consumeGas(
           computation.gasCosts[Create].m_handler(0, 0, contractCode.len),
           reason = "Write contract code for CREATE")
+
+        let storageAddr = computation.msg.storage_address
```
zah: I had to move this code, because the translation from Python was not quite correct. You may have been confused by the unusual semantics of the Python's `try/except/else` construct which is used in the py-evm code. Read about it here:

https://stackoverflow.com/questions/855759/python-try-else

zah:
@coffeepots, I didn't have a lot of time to test delve deeper and test the `apply*` procs. I hope you'll be able to take the code from here and debug any issues with the transactions API if such arise.

zah:
To use the code, you'll have to switch to the `transactions` branch of the `eth-trie` repo pushed here:
https://github.com/status-im/nim-eth-trie/pull/21

PR #143, opened 2018-09-11
skip 1000+ tests which either are slow (quadratic tests) or don't support Homestead; add ~50 new working tests

tersec:
766 failing tests, down from almost 2000, remain.

An entire category of tests, stArgsZeroOneBalance, looks easy enough for Nimbus to pass, but BALANCE needs to cost 400, not 20, for it to do so, for example -- that's 46 tests in that folder alone.


tersec:
Add back-ref from https://github.com/status-im/nimbus/issues/113

PR #144, opened 2018-09-12
Caching DB backend

yglukhov:


arnetheduck:
nice! I saw @zah call this a `JournalDB` which seems closer to its intended use?

yglukhov:
I would gladly rename it, "caching" is indeed not the right term for this. JournalDB sounds closer, but still not ideal, because there's no journal actually. I'll wait for a couple of days, maybe we'll come up with a better name :)

zah:
I don't think It's a good idea to rely on the DB's transactions. The memory layers are required for other capabilities too such as executing transactions in the stateless clients, where you'll get tree branches from the network that will be only shortly inserted into a memory layer during the execution.

zah:
The C++ client uses `OverlayDB` which is also a suitable name.

PR #146, opened 2018-09-14
48 new working GeneralStateTests

tersec:
Towards https://github.com/status-im/nimbus/issues/113.
Is a regardless-necessary refactor away from https://github.com/status-im/nimbus/issues/130.

via
- creating a rudimentary balance commit/revert scheme (https://github.com/status-im/nimbus/issues/142 would be a more general/well-architected/factored form of this functionality).
- fixing some overflows/underflows
- and implementing account deletion for the selfdestruct instruction

Current progress:
- 711 working tests
- 1011 skipped tests, nearly all because they don't support Homestead
- 612 failing tests
- 91 failing tests with no `CREATE` (`0xf0`), `CALL` (`0xf1`), `CALLCODE` (`0xf2`), `DELEGATECALL` (`0xf4`), or `STATICCALL` (`0xfa`) opcodes in `transaction.to`'s code; `CALL` instructions are coming in via https://github.com/status-im/nimbus/issues/121 and https://github.com/status-im/nimbus/pull/139
- 40 failing tests with neither `CREATE` nor `CALL` instructions, as well as not creating contracts in other ways, e.g, `transaction.to` being empty.

### DIFF
--- a/nimbus/vm/computation.nim
+++ b/nimbus/vm/computation.nim
@@ -96,12 +96,14 @@ proc addLogEntry*(c: var BaseComputation, account: EthAddress, topics: seq[UInt2
 
 # many methods are basically TODO, but they still return valid values
 # in order to test some existing code
-proc getAccountsForDeletion*(c: BaseComputation): seq[(string, string)] =
+func getAccountsForDeletion*(c: BaseComputation): seq[EthAddress] =
   # TODO
   if c.isError:
     result = @[]
   else:
     result = @[]
+    for account in c.accountsToDelete.keys:
+        result.add(account)
 
 proc getLogEntries*(c: BaseComputation): seq[(string, seq[UInt256], string)] =
   # TODO
--- a/tests/test_generalstate_json.nim
+++ b/tests/test_generalstate_json.nim
@@ -29,6 +29,7 @@ proc stringFromBytes(x: ByteRange): string =
     result[i] = char(x[i])
 
 proc testFixture(fixtures: JsonNode, testStatusIMPL: var TestStatus) =
+  # XXX: this is becoming a mess. refactor.
   var fixture: JsonNode
   for label, child in fixtures:
     fixture = child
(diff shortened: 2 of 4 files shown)


tersec:
`computation.gasMeter.refundGas(24_000 * deletedAccounts.len)` should pull from `RefundSelfDestruct` in https://github.com/status-im/nimbus/blob/master/nimbus/vm/interpreter/gas_costs.nim -- my plans's to move all that gas refunding into the actual self-destruct instruction, rather than just blindly copy Py-EVM's design, where it's natural to get at the constant properly, so this is a temporary kludge.

nimbus/vm/computation.nim
```diff
@@ -96,12 +96,14 @@ proc addLogEntry*(c: var BaseComputation, account: EthAddress, topics: seq[UInt2
...
 # in order to test some existing code
-proc getAccountsForDeletion*(c: BaseComputation): seq[(string, string)] =
+func getAccountsForDeletion*(c: BaseComputation): seq[EthAddress] =
   # TODO
   if c.isError:
     result = @[]
   else:
     result = @[]
+    for account in c.accountsToDelete.keys:
+        result.add(account)
```
mratsim: You can use `toSeq c.accountsToDelete.keys`

tests/test_generalstate_json.nim
```diff
@@ -48,28 +49,44 @@ proc testFixture(fixtures: JsonNode, testStatusIMPL: var TestStatus) =
...
-  let currentCoinbase = fenv["currentCoinbase"].getStr.ethAddressFromHex
-
   vmState.mutateStateDB:
+    # TODO: combine some of these
+    # Also, in general, map out/etc the whole vmState.mutateStateDB flow set
     db.setBalance(sender, db.getBalance(sender) - gas_cost)
     db.deltaBalance(currentCoinbase, gas_cost)
     db.setNonce(sender, db.getNonce(sender) + 1)
+    db.deltaBalance(transaction.to, transaction.value)
+    db.setBalance(sender, db.getBalance(sender) - transaction.value)
```
zah: isn't this also a `deltaBalance`?

tests/test_generalstate_json.nim
```diff
@@ -48,28 +49,44 @@ proc testFixture(fixtures: JsonNode, testStatusIMPL: var TestStatus) =
...
-  doAssert vmState.readOnlyStateDB.getBalance(sender) >= gas_cost
+  let currentCoinbase = fenv["currentCoinbase"].getStr.ethAddressFromHex
 
-  # TODO: implement other sorts of transctions
+  # XXX: https://github.com/status-im/nimbus/issues/35#issuecomment-391726518
+  # TODO: put yellow paper ref here from that link justifying the limit (1 shl 34 is stand-in)
+  if transaction.gasLimit < transaction.getFixtureIntrinsicGas or
+     transaction.gasPrice > (1 shl 34) or
+     transaction.accountNonce != vmState.readOnlyStateDB.getNonce(sender) or
+     vmState.readOnlyStateDB.getBalance(sender) < gas_cost:
```
coffeepots: Whilst this isn't going to impact performance much for the tests, it's worth noting that `readOnlyStateDb` does allocate memory and creates a new trie each time it's called, so might be worth putting the result in a var and calling `getNonce` and `getBalance` on that.

tests/test_generalstate_json.nim
tersec: `deltaBalance` needs a nonnegative `UInt256`. It cannot (modulo 2's complement gimmickry) subtract.

tersec: It's just a copy of Py-EVM, where it works because the bigint representation is signed.

nimbus/vm/computation.nim
tersec: I'd been looking for how to do that. I much prefer the `toSeq` version and have implemented it locally, for the next commit.

tests/test_generalstate_json.nim
zah: I see, `increaseBalance` would be a more honest name then.

tersec: Yes, it would. It looks like my code's the only caller of `deltaBalance`, so I've changed the `proc` name in `state_db` and the half-dozen or so call sites in the GeneralStateTest harness.

tests/test_generalstate_json.nim
tersec: Definitely an optimization worth making. Done locally.

tersec:
@mratsim yes, that's what I mean -- that embedded clause confuses things.

PR #148, opened 2018-09-18
20 new working GeneralStateTests

tersec:
In no particular order:
- further bounds-checked the SHA3 opcode, which avoids a gas calculation overflow.
- rewrote `writePaddedResult` to always write padding for non-extended sections of memory up to the total number of bytes requested written, regardless of input data size. The correct semantics are not to leave them alone, as the previous `writePaddedResult` did. I find this version a bit simpler, too -- no branches, for example.
- switched the `LOG1`, `LOG2`, `LOG3`, and `LOG4` opcodes to use `cleanMemRef`, to solve the usual issues that solves. This required moving `cleanMemRef` to `nimbus/vm/interpreter/utils/utils_numeric.nim`, where it belongs regardless.
- removed another couple pointless (because `UInt256`) `validateGte` commented lines from the auto-conversion.
- moved `db.increaseBalance(currentCoinbase, gas_cost)` to the end of each codepath (yet to be re-merged; e.g., the two error paths are intentionally identical, and most of them will collapse entirely once transactional database usage comes in), separate from its pair `db.setBalance(sender, db.getBalance(sender) - gas_cost)`. This has created a bit more duplication temporarily, but solved around a dozen `GeneralStateTests`.
- to begin addressing the suicide coinbase cases, I ensure that coinbases don't spuriously get credited if they're going to be destroyed regardless in the main codepath.

Progress:
- 758 working tests
- 1011 skipped tests
- 565 failing tests, of which 54 don't `CALL`, or `CREATE`, and of which 3 (`tests/fixtures/GeneralStateTests/stSystemOperationsTest/suicideCoinbase.json`, `tests/fixtures/GeneralStateTests/stTransactionTest/TransactionFromCoinbaseHittingBlockGasLimit1.json`,  and `tests/fixtures/GeneralStateTests/stTransactionTest/SuicidesMixingCoinbase.json`) have neither those nor an empty `transaction.to` field.

### DIFF
--- a/nimbus/vm/interpreter/opcodes_impl.nim
+++ b/nimbus/vm/interpreter/opcodes_impl.nim
@@ -6,7 +6,7 @@
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
 import
-  strformat, times, ranges,
+  strformat, times, ranges, sequtils,
   chronicles, stint, nimcrypto, ranges/typedranges, eth_common,
   ./utils/[macros_procs_opcodes, utils_numeric],
   ./gas_meter, ./gas_costs, ./opcode_values, ./vm_forks,
@@ -193,7 +193,7 @@ op sha3, inline = true, startPos, length:
   ## 0x20, Compute Keccak-256 hash.
   let (pos, len) = (startPos.toInt, length.toInt)
 
-  if pos < 0 or len < 0:
+  if pos < 0 or len < 0 or pos > int32.high:
     raise newException(OutOfBoundsRead, "Out of bounds memory access")
 
   computation.gasMeter.consumeGas(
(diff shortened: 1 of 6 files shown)


tersec:
The Travis CI build is failing because:
```
Hint:transaction [Processing]
Hint:eth_keys [Processing]
Hint:secp256k1 [Processing]
Error: unhandled exception: cannot open: /home/travis/.nimble/pkgs/secp256k1-0.1.0/secp256k1_wrapper/secp256k1/src/secp256k1.c [IOError]
       Tip:157 messages have been suppressed, use --verbose to show them.
     Error:Execution failed with exit code 1
```

I've mentioned this last week or earlier, someone came in to Slack mentioning it, et cetera. My local build environment setup script has a workaround:
```
git clone --recurse https://github.com/status-im/nim-secp256k1.git
...
cp -npvr ~/nim-secp256k1/secp256k1_wrapper/ ~/.nimble/pkgs/secp256k1-0.1.0/
```

coffeepots:
LGTM. `writePaddedResult` definitely looks better now.
Interested to see how this runs with #139 merged, we got a few extra tests passing with that already, but not sure if they're extras compared to this PR.

tersec:
At least 3 or 4 of them (`callcall_`, et cetera) are distinct.

nimbus/vm/interpreter/opcodes_impl.nim
```diff
@@ -212,34 +212,20 @@ op sha3, inline = true, startPos, length:
...
-  if x > upperBound:
-    return high(int32) shr 2
-  return x.toInt
+  let sourceBytes = data[min(dataPos, data.len) .. min(data.len - 1, dataEndPosition)]
+
+  mem.write(memPos, sourceBytes)
+
+  # Don't duplicate zero-padding of mem.extend
+  let paddingOffset = memPos + sourceBytes.len
+  mem.write(paddingOffset, repeat(paddingValue, max(prevLen - paddingOffset, 0)))
```
zah: Can we avoid the call to `repeat` here? It performs a memory allocations that's not really necessary.

The previous code is more convoluted for sure, but it actually seems more economical to me when it comes to branching. There is a potential hidden branch in every call to `min` and `max` and you would still call `mem.write` even when `repeat` returns a zero-len sequence.

tersec: @zah yes, I'll look into a better approach than `repeat`. I share your concern here. In this version I went for readability, but to be fair, this code's seems plausibly more performance-sensitive than most in Nimbus. It's worth optimizing.

@coffeepots yes, that `min` and `max` are often branchless was part of my thought process. The issue I had with the explicit code was that it created multiple subcases that I found harder to verify were all correct, created some duplication, and obscured the degree to which the different nominally disjoint code flows were similar in practice. I'm not strongly opinionated about this though.

PR #149, opened 2018-09-18
More state restoring

yglukhov:


nimbus/db/state_db.nim
```diff
@@ -64,6 +64,12 @@ proc setBalance*(db: var AccountStateDB, address: EthAddress, balance: UInt256)
 proc increaseBalance*(db: var AccountStateDB, address: EthAddress, delta: UInt256) =
```
zah: I guess `increaseBalance` can be removed now in favor of `addBalance`.

nimbus/genesis.nim
```diff
@@ -63,8 +63,9 @@ proc defaultGenesisBlockForNetwork*(id: PublicNetwork): Genesis =
     Genesis()
   result.config = publicChainConfig(id)
 
-proc toBlock*(g: Genesis): BlockHeader =
-  let tdb = trieDB(newMemDB())
+proc toBlock*(g: Genesis, db: BaseChainDB = nil): BlockHeader =
+  let tdb = if db.isNil: trieDB(newMemDB())
```
zah: Since yesterday, the new way to create an in-memory database is with `newMemoryDB()`. This replaces the whole call to `trieDB(newMemDB())`

PR #150, opened 2018-09-18
Homestead contract creation gas cost

mratsim:
See EIP-2: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2.md
and https://etherscan.io/tx/0x31ded263506ea36e6ea777efc2c39a999e6fba4f4d338c7313af6aac6d9bf3e3

Contract creation was 21k + 0 instead of the current 21k + 32k.

### DIFF
--- a/nimbus/vm/interpreter/gas_costs.nim
+++ b/nimbus/vm/interpreter/gas_costs.nim
@@ -536,7 +536,7 @@ const
     GasExp:             10,
     GasExpByte:         10,     # Changed to 50 in Spurious Dragon (EIP160)
     GasMemory:          3,
-    GasTXCreate:        32000,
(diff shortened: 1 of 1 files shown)


yglukhov:
Appveyor failed. Is that ok?

mratsim:
It failed for 32-bit only, 64-bit worked. And it was out of bounds memory access.

I don't think it's related but I'm surprised that didn't come up earlier.

coffeepots:
I was just looking at that, the out of bounds error doesn't seem to happen on master for test_vm_json.

coffeepots:
LGTM. The out of bounds error means the index `<0` or `>Int32`. Unlikely to be anything to do with this PR.

PR #153, opened 2018-09-19
integrate GeneralStateTest runner into CI with 775 running tests

tersec:
It does this by preventing specific ~500 specific allowed tests to fail without causing nonzero error codes. The `GeneralStateTests.md` file will still accurately show which tests work, fail, and are skipped, but for those specific named tests, it won't cause the `unittest` module to exit such that the CI sees an error.

It also iterates over all GeneralStateTest indexes per JSON fixture, rather than just using the first; this caused a loss of one of the 776 otherwise working (@coffeepots's `CALL` work netted about 20 immediately upon merging) tests, `RETURN_Bounds.json `, but everything else that had been working with just that first index set turned out to have been working for all of them. Most fixtures only have the one (0, 0, 0) index tuple.

I'm also starting to catch up with the tech debt I've been making for myself with the hitherto awful code quality with a substantial refactoring and cleanup. It's not particularly good yet, but it's better, with a goal towards closing out the `GeneralStateTest` part of https://github.com/status-im/nimbus/issues/113 and getting it in shape for https://github.com/status-im/nimbus/issues/130.


coffeepots:
LGTM, tests look tidier now :)

tersec:
@zah Right, https://github.com/status-im/nimbus/commit/75f973e654f0c2c8ef2f994535f45285791d733d breaks this for no reason I can tell and with no evidence of testing.

PR #160, opened 2018-09-25
Better cmdline flags handling

yglukhov:
Depends on https://github.com/status-im/nim-eth-p2p/pull/39
* support `--nodiscover` switch
* `--discport` defaults to `--port` value
* `--bootnodes` flag disables network default bootnodes

### DIFF
--- a/nimbus/config.nim
+++ b/nimbus/config.nim
@@ -367,7 +367,7 @@ proc processNetArguments(key, value: string): ConfigStatus =
   let config = getConfiguration()
   let skey = key.toLowerAscii()
   if skey == "bootnodes":
-    result = processENodesList(value, config.net.bootnodes)
+    result = processENodesList(value, config.net.bootNodes)
   elif skey == "bootnodesv4":
     result = processENodesList(value, config.net.bootNodes)
   elif skey == "bootnodesv5":
@@ -511,7 +511,7 @@ NETWORKING OPTIONS:
   --bootnodesv4:<value>   Comma separated enode URLs for P2P v4 discovery bootstrap (light server, full nodes)
   --bootnodesv5:<value>   Comma separated enode URLs for P2P v5 discovery bootstrap (light server, light nodes)
   --port:<value>          Network listening TCP port (default: 30303)
-  --discport:<value>      Network listening UDP port (default: 30303)
+  --discport:<value>      Network listening UDP port (defaults to --port argument)
   --maxpeers:<value>      Maximum number of network peers (default: 25)
   --maxpendpeers:<value>  Maximum number of pending connection attempts (default: 0)
   --nodiscover            Disables the peer discovery mechanism (manual peer addition)
@@ -538,7 +538,11 @@ LOGGING AND DEBUGGING OPTIONS:
 
 proc processArguments*(msg: var string): ConfigStatus =
   ## Process command line argument and update `NimbusConfiguration`.
-  discard getConfiguration()
+  let config = getConfiguration()
+  var tempBootNodes: seq[ENode]
+  swap(tempBootNodes, config.net.bootNodes)
+  config.net.discPort = 0
+
   var opt = initOptParser()
   var length = 0
   for kind, key, value in opt.getopt():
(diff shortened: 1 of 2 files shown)


nimbus/config.nim
```diff
@@ -538,7 +538,11 @@ LOGGING AND DEBUGGING OPTIONS:
 
 proc processArguments*(msg: var string): ConfigStatus =
   ## Process command line argument and update `NimbusConfiguration`.
-  discard getConfiguration()
+  let config = getConfiguration()
+  var tempBootNodes: seq[ENode]
+  swap(tempBootNodes, config.net.bootNodes)
```
zah: I could use a clarification for what's going on here. The original value of the `bootNodes` setting (which seem to be the mainnet bootnotes) is restored if the cmd line end setting a bootNodes list of zero len, but how could that happen?

yglukhov: Added a comment. Hopefully it clears things up

zah: But wouldn't this code be simpler if we just start with an empty list in `initConfiguration`? At the end we can check if the list is still empty and then assign the default value. And actually, why is overriding the default value during the processing of the cmd params considered a problem? Is it because there is no way to detect conflicting settings (--network and --bootnodes specified separately)?

yglukhov: > But wouldn't this code be simpler if we just start with an empty list

That surely would be simpler, but I kinda tried to keep the API more user friendly so that config is usable without `processArguments`, and `processArguments` doesn't blindly overwrite whatever defaults you set upfront. Ideally I would prefer a somewhat different interface for this, but that's surely not critical enough. I can also simplify the behavior if you still suggest.

> why is overriding the default value during the processing of the cmd params considered a problem

It's the other way around. This PR enables us to e.g. connect to main net without using main net public boot nodes. Otherwise we are doomed to always ping those.

zah: Considering that Nimbus should also get a config file eventually, and that it would be nice to be able to detect situations with conflicting settings, perhaps we can extend the settings-holding structures a bit to remember the origin of each setting (i.e. default value/value from default config file/value from config file override/value from command-line/etc). This can be a `SettingsValue[T]` type holding an extra enum value for example.

yglukhov: Conflicts is something completely orthogonal to this discussion, and I'm really not sure if we need to care much about settings conflict resolution.

Now, as i'm seeing the API, the settings can be defined at 3 levels, prioritized from least to most important: (1) defaults, (2) persistent storage (e.g. config file, windows registry, ios user defaults, etc), command line (4).

Once all of the settings sources are read and settings are constructed from those, there's little reason to know which origin a particular setting is coming from. Here you could argue that potentially conflicting settings should be reported to the user with their respective origins, but additional complexity is not worth it imo. Besides nimbus is likely going to be used as a lib, so it becomes even less critical. All of that, given setting conflicts exist in reality at all :).

As such the api as i'm seeing it would be smth like:
```nim
proc newConfig(): Config
proc overrideWithCommandLine(c: Config, args: openarray[string])
proc overrideWithJson(c: Config, j: JsonNode)
proc overrideWithWindowsRegistry(c: Config, keyPath: string) # windows only
proc overrideWithUserDefaults(c: Config, keyPath: string) # macos/ios
# ... etc ...

proc main() =
  let c = newConfig()
  c.overrideWithJson(readJson("config.json"))
  c.overrideWithCommandLine()
  maybeValidateItIfNeeded(c)
  setSharedConfig(c)
```

The advantage of this is:
- the semantics is straightforward
- the config is composable from unlimited amount of layers
- override functions can be written in a generic way to accept any object structure
- the `Config` is a simple aggregation (POD, strings, seqs, objects) type, that can incorporate subconfigs from its dependencies, e.g. nimbus `Config` could contain `p2p: P2PConfig` from eth_p2p repo, without eth_p2p knowing too much about the config structure (aggregation types instead of `SettigValue[T]`)
- the overall complexity is minimal given that initial versions of `override*` don't have to be generic.

Now given all of the above I'm seeing my changes as a compromise (slightly ugly, true) between my desired behavior and the current interface, which I'm hoping will be changed eventually.

zah: Tracking the origin may be helpful for the user though.

Consider the situation where Nimbus cannot open a certain port, because another application is listening. It may be nice to report something like: 

```
Cannot open RPC port 30303, specified in ~/.config/nimbus.ini:12
```

yglukhov: > Tracking the origin may be helpful for the user though

I'm not saying it's useless. I'm saying it is not useful enough to sacrifice simplicity.
- 99% I already know where the port is specified, if it is not default. If i don't its trivial to find.
- If it is default your system will not tell me how I override it, which is more important.
- nimbus as a lib is unlikely to report meaningful end-user messages

But don't you agree that we need the override semantics for settings that I'm suggesting?

zah: The functions you outlined are great. I'm just still wondering if defining the `SettingsValue[T]` is worth it. I don't imagine it to be a lot of extra complexity.

Anyway, let's merge this for now and revisit is soon.

PR #161, opened 2018-09-27
Get blockchain reading working through block 48680

tersec:
- refactor aspects of transaction processing out of GeneralStateTests handler
- get p2p/chain runnning through block 48680
- combine/deduplicate two different but equivalent methods of caulculating transaction intrinsic gas cost
- fix BaseTransaction references to just Transaction references in hitherto unreferenced stubbed out source so it builds
- clean up some imports from refactoring

### DIFF
--- a/nimbus/p2p/chain.nim
+++ b/nimbus/p2p/chain.nim
@@ -1,6 +1,8 @@
 import ../db/[db_chain, state_db], eth_common, chronicles, ../vm_state, ../vm_types, ../transaction, ranges,
   ../vm/[computation, interpreter_dispatch, message], ../constants, stint, nimcrypto, ../utils/addresses,
-  eth_trie/memdb, eth_trie, rlp
+  ../vm_state_transactions,
+  eth_trie/memdb, eth_trie, rlp,
+  sugar
 
 
 type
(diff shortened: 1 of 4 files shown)


nimbus/p2p/chain.nim
```diff
@@ -65,26 +63,53 @@ proc processTransaction(db: var AccountStateDB, t: Transaction, sender: EthAddre
...
+          gasRefund = min(gasRefunded, gasUsed2 div 2)
+          gasRefundAmount = (gasRefund + gasRemaining) * t.gasPrice.u256
+        #echo "gasRemaining is ", gasRemaining, " and gasRefunded = ", gasRefunded, " and gasUsed2 = ", gasUsed2, " and gasRefund = ", gasRefund, " and gasRefundAmount = ", gasRefundAmount
+        let codeCost = (200 * c.output.len).u256
+        db.addBalance(sender, (t.gasLimit.u256 - gasUsed2 - codeCost)*t.gasPrice.u256)
+        echo c.output.len, " bytes: ", c.output
+        return (gasUsed2 + codeCost) * t.gasPrice.u256
+
+      else:
+        # FIXME: don't do this revert, but rather only subBalance correctly
```
zah: You can also use a Transaction here I suppose.

tersec: Yes, I've played with those unsuccessfully in the GST runner. Definitely my plan, since I've seen the alternative once there's a need to deal with rollbacks/reverts. Right now it works so little that it would be a triumph to simply getting it running correct, straight-line EVM code.

PR #162, opened 2018-09-27
Get blockchain reading through block 49017

tersec:
This reads through block 49017 by properly setting createAddress/storageAddress of the computation message and detecting whether the computation has the gas required to pay 200*code length; also, clean up pointless commented nonnegativity assertions for nonnegative GasInt type.

This is as far as I intend to go with blockchain reconstruction until it doesn't need `del` from `rocksdb_backend` commented out to even get this far (as advised by @yglukhov). I've checked that `eth.getBalance` on the coinbase, on the contract, and on the sender all match what Nimbus sets, and similarly with `eth.getCode`. The remaining checkable item is `eth.getStorage`, which just commenting out such a routine as `del` could do interesting things with, that I'm not going to waste time hoping disabling that and actual state calculations don't interact unfortunately.

The deleted comments in `gas_meter` are left over from pre-`GasInt` conversion.

### DIFF
--- a/nimbus/p2p/chain.nim
+++ b/nimbus/p2p/chain.nim
@@ -64,7 +64,6 @@ proc processTransaction(db: var AccountStateDB, t: Transaction, sender: EthAddre
     return
 
   var gasUsed = t.payload.intrinsicGas.GasInt
-  #echo "gasUsed is ", gasUsed
 
   if t.isContractCreation:
     # gasUsed += 32000 # This appears in Homestead.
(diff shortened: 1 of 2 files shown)


zah:
Hrm, while the disabled `del` will be fixed eventually, it really shouldn't interact with the consensus rules in any way. Without `del`, we are merely failing to delete the no longer necessary records from the persistent storage. The Trie structures are still properly maintained and their root hashes will still be computed properly (please correct me if I'm wrong). The root hash is the only piece of data that will be examined from the point of view of the consensus rules.

PR #163, opened 2018-09-28
Refactor contract creation into transaction code from P2P code; get 14 new GSTs working

tersec:
This doesn't handle any of the sorts of more interesting edge cases that GSTs contain in any detail, but represents the first time two parts of Nimbus are using the same transaction processing code (for contract creation at least), with all the usual upsides.

Overall progress is that 809 GeneralStateTests pass and 514 GeneralStateTests fail; in the future, improvement on this should be directly linked, via running the same code, to blockchain ingestion, so the GST runner's not off in its own world.


tests/test_generalstate_json.nim
```diff
@@ -47,10 +47,25 @@ proc testFixtureIndexes(header: BlockHeader, pre: JsonNode, transaction: Transac
   let gas_cost = transaction.gasLimit.u256 * transaction.gasPrice.u256
   vmState.mutateStateDB:
     db.setNonce(sender, db.getNonce(sender) + 1)
-    db.addBalance(transaction.to, transaction.value)
     db.subBalance(sender, transaction.value + gas_cost)
 
+  if transaction.to == CREATE_CONTRACT_ADDRESS and transaction.payload.len > 0:
```
yglukhov: It's better to check for `transaction.isContractCreation`, and delete `CREATE_CONTRACT_ADDRESS` constant once and for all. `getFixtureTransaction` should be fixed accordingly.

tersec: Doesn't work yet, since it's only set when it's read in via whatever P2P or RPC mechanism (didn't check details) that you or someone else created.

tersec: I tried that first.

tersec: Done, https://github.com/status-im/nimbus/pull/163/commits/7ae0607e8bb4fdc34808989efc12e134c0473bc4

nimbus/vm_state_transactions.nim
```diff
@@ -60,6 +60,58 @@ proc execComputation*(computation: var BaseComputation, vmState: BaseVMState): b
   except ValueError:
     result = false
 
+proc applyCreateTransaction*(db: var AccountStateDB, t: Transaction, head: BlockHeader, vmState: var BaseVMState, sender: EthAddress, useHomestead: bool = false): UInt256 =
```
zah: This is not very different from `applyCreateMessage` defined in `computation.nim`. Any chance we can adapt the code to cover both cases with a single definition?

tersec: Indeed, they definitely need merging. One of my next couple of refactorings will do so.

PR #167, opened 2018-10-05
Populate the persistent databases with the empty RLP key.

zah:
Also implements transactional block persistence. Two issues
in the transaction processing code have been discovered that
might affect other usages such as the CALL instruction.

The main fix gets us past block 49000.

You may need to clean up your database.

### DIFF
--- a/nimbus/db/backends/rocksdb_backend.nim
+++ b/nimbus/db/backends/rocksdb_backend.nim
@@ -1,4 +1,4 @@
-import os, rocksdb, ranges, eth_trie/db_tracing
+import os, rocksdb, ranges, eth_trie/[db_tracing, constants]
 import ../storage_types
 
 type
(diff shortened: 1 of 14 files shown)


zah:
Merging, because the change is relatively trivial, but the build will be broken without it (due to merged changes in nim-eth-trie).

nimbus/db/backends/rocksdb_backend.nim
```diff
@@ -46,3 +34,18 @@ proc del*(db: ChainDB, key: openarray[byte]) =
...
+    dataDir = basePath / "data"
+    backupsDir = basePath / "backups"
+
+  createDir(dataDir)
+  createDir(backupsDir)
+
+  let s = result.store.init(dataDir, backupsDir)
+  if not s.ok: raiseStorageInitError()
+
+  put(result, emptyRlpHash.data, emptyRlp)
```
yglukhov: IMO the backends are better not to depend on eth-trie. It's better to do trie related init elsewhere, e.g. in nimbus.nim upon the db creation, or in the trie itself, if possible.

zah: Dunno, the Trie code is written assuming its back-end stores has this emptyRlp value (this is done to prevent unnecessary checks in the code). If the back-end here is consider a backing store for a Trie, it's logical to assume that this value is a mandatory part of it. 

You can argue that the back-end is "purpose-agnostic", but I think this would be a different object and we already have it with the `RocksDBInstance` type.

yglukhov: > purpose-agnostic

Well we are using the backends now not only for tries, but for assorted stuff like storing block headers, mapping block numbers to block hashes, storing head hash, etc.

> this is done to avoid some unnecessary checks in the code

Good argument, so we could do this step in `initializeEmptyDb()`.

zah: OK, it's reasonable for this to be moved in `initializeEmptyDb` indeed.

PR #173, opened 2018-10-28
Initial TransactionTracer impl

yglukhov:
This partially implements the transaction trace in geth format. Currently the traces are disabled, intended to be used (and enabled) in traceTransaction rpc call.


coffeepots:
Looks good to me. As @tersec mentions this will be invaluable when working on 1.0 and immediately useful for state test execution. Next step, automatic diffs on state failure? :)

PR #175, opened 2018-11-09
[WIP] block validation debugging tool

jangko:
- [x] Implement the per-transaction state root checking in `persistBlocks` (in `chain.nim`). 

This is done by comparing the Nimbus computed state root against the state root indicated in the block's transaction receipts.

The implementation is deviated a bit. During block validation, transaction receipt is produced and later they will be compared to valid receipts from geth.

------------------------

- [x] Enhance the `HexaryTrie` from the `eth_trie` package to support enumerating all the leaf keys starting from a given root.

This is required for the next point.

------------------------

- [x] Implement the `debug_traceTransaction` RPC call in Nimbus. The implementation for this is already underway, you can see the TODO item left in the code below:

https://github.com/status-im/nimbus/commit/bac452f589a1928a6966dc5b8c96f0e2b483144e#diff-01829f818cf11c3fd12b02e5f3c2bc3aR35

See a transaction trace from Geth to see the desired output.

------------------------

- [x] Upon detecting a failure for some transaction. Produce a Nimbus trace for it and store in a file along any other required metadata for the next steps. 
- [x] Launch the premix tool.

The code for this should be enabled only in debug builds.

------------------------

- [x] Implement the premix tool as a separate executable.

- [x] It will consume the file produced in the previous step,
- [x] it will contact a local copy of Geth (or a public JSON-RPC point such as [Infura](https://infura.io/docs)) to get a correct trace. 
- [x] It will produce a JSON test case and a report page.
- [x] It will print instructions for running the test case
- [x] and it will launch the default browser to see the report page.

- [x] pull blocks and transactions from geth
- [x] get transaction trace information from both geth and nimbus
- [x] implement configuration
- [x] rpc stuff
- [x] implement web tool
- [x] implement test case gen tool
- [x] Enhance the [test runner](https://github.com/status-im/nimbus/blob/master/tests/test_generalstate_json.nim) with command line options for running individual test case files


yglukhov:
This is fancy, but i'd propose to start with implementing `debug_traceTransaction` rpc call first :)

jangko:
yeah, I will put that into another PR

zah:
@jangko, I took the liberty to modify and clarify your to-do items a bit. As discussed on Gitter, I think it's possible to skip the generation of the CSV-like database and to go straight to creating a test case JSON file that includes the same information. Here is an example of a JSON test case including storage keys:

https://github.com/ethereum/tests/blob/develop/GeneralStateTests/stSStoreTest/sstore_Xto0.json#L202

stefantalpalaru:
>   [ ]  Enhance the [test runner](https://github.com/status-im/nimbus/blob/master/tests/test_generalstate_json.nim) with command line options for running individual test case files

This is already possible:

```
nim c -r -o:build/test_generalstate_json tests/test_generalstate_json.nim tests/fixtures/GeneralStateTests/stAttackTest/ContractCreationSpam.json
```

jangko:
finally, premix debugging tool is ready for review. the only obstacle left is status-im/nim-json-rpc#46

zah:
Thanks @jangko, this looks great. I have two follow up ideas that were partly inspired by your code.

1) When Nimbus encounters a problematic block, would it be possible to effectively skip it by getting the correct execution from Geth and by storing its side effects in the Nimbus DB? I assume that this will allow Nimbus to continue syncing and to find even more problematic blocks. We'll be able to produce multiple test cases like this that can be attacked in parallel by different members of the team.

2) Once I have a failing test case, how do I iterate on the fix? It would be very useful if I can see the updated trace comparison page after each change I introduce in the code. Perhaps this is already possible?

jangko:
Great ideas! they would speed up syncing progress significantly.

1. The post state tracer in `js_tracer.nim` was actually a modified prestate tracer from geth. With further modification we can reconstruct prestate needed to validate a block without entire block history following. I imagine the implementation would allow us to enter `hunting mode` where we use memoryDB only to find problematic block and discard any unwanted side effects from polluting the disk db. Although it need more works, I believe it will not to much code to add, perhaps only a combination between `premix.nim` and `prestate.nim` and a little bit more accounts storing code. The possibility to make it real is very high.

2. The second idea already implemented partially in `debug.nim` and `premix.nim`. We only need to move some code around, all the needed code already available.

jangko:
@zah, I've implemented both of your ideas, we can start hunting more problematic blocks with `hunter` tool.

I also enhance `debug` tool to allow it to update report page at each fix iteration.

zah:
Amazing work, @jangko, thanks for implementing the additional ideas so quickly.

The best next steps would be to put this tool to good use and to make some progress on the blockchain sync. Please get in touch with @stefantalpalaru in our Gitter channel to coordinate who will be investigating and fixing the issues affecting a particular block.

@Swader, can you review and improve the added [README files](https://github.com/status-im/nimbus/blob/master/premix/readme.md)?

Swader:
Absolutely. It's perfectly suited for the status update too, well done @jangko, superb work! 🤘

zah:
OK, finishing the `contractCall` implementation would be a very good next goal. I think it would also fix many of the currently failing state tests. Would you like to work on this, @jangko?

jangko:
Ok @zah, it also will give me opportunity to test and fix entire debugging tool set more thoroughly with real case.

PR #176, opened 2018-11-13
Pass CI fixes

coffeepots:
For some reason this is not passing CI in 32 bit Windows, even though the output should provide a valid exit code. I ran the state tests locally on 32 bit and there was no obvious issue.

@tersec do you have any input into why this isn't going green?


arnetheduck:
My guess is build timeout - it's taking 1hr to run the tests..

coffeepots:
@arnetheduck Timeout makes a lot of sense actually, I did wonder if that was the case but didn't see any obvious timeout information in Appveyor. I do wonder why x86 takes an hour whereas x64 takes 16 minutes.
Assuming it is a timeout, that places a constraint on CI being accurate for 32 bits and will mean a 'Failed' state for Windows builds likely every time.

tersec:
If there's some specifically slow test, `slowTest` in `test_helpers.nim` can semantically meaningfully exclude it. I'm sifting through out of interest, but so far it's not obvious to me if/which there are specific tests. It's a bit trickier because it runs these tests in different orders each time. Obviously, the question of why 32-bit only is relevant.

tersec:
Looking at the `DBG` timestamps, the CI is running the 32-bit version several times slower, so that it didn't need to be any one particular test, but just that the 64-bit version in CI is running at usually several state tests per second, while the 32-bit version's running a lot slower than that, just in general.

It would be potentially useful to test this in isolation from the CI -- if this is an AppVeyor CI issue, then that's not worth spending too much time, necessarily, on, but if the 32-bit builds are really that slow, that's a real problem.

coffeepots:
It's almost like the 32 bit instance is running a much slower CPU entirely, or there's some kind of bottleneck.

As this seems to be a 32 bit only CI issue and considering this is a light on code PR, I vote for merge to at least pass the Linux/Mac/x64 CI.

PR #180, opened 2018-11-21
enable global gitignore rules

stefantalpalaru:
Looks like "\*.swp" in a file pointed at by core.excludesFile is not able to override "!\*.\*" in the project's .gitignore file.

Original patterns:
```
$ git check-ignore -v nimbus/.nimbus.nim.swp
.gitignore:4:!*.*	nimbus/.nimbus.nim.swp
```
Problem block commented out in .gitignore:
```
$ git check-ignore -v nimbus/.nimbus.nim.swp
/home/stefan/.gitignore:3:*.swp	nimbus/.nimbus.nim.swp
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,18 +1,21 @@
+### the following pattern disables global gitignore rules (the ones from core.excludesFile)
 # Ignore all files without extensions (unix executable files)
-*
-!*.*
-!*/
-!LICENSE*
+#*
+#!*.*
+#!*/
+#!LICENSE*
 
 nimcache/
 
 # Executables shall be put in an ignored build/ directory
-# Ignore dynamic, static libs and libtool archive files
 build/
+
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,18 +1,21 @@
+### the following pattern disables global gitignore rules (the ones from core.excludesFile)
 # Ignore all files without extensions (unix executable files)
-*
-!*.*
-!*/
-!LICENSE*
+#*
+#!*.*
+#!*/
+#!LICENSE*
```
mratsim: I think you disabled those rules by mistake, but it's true that if everyone is using the `build/` folder this is not needed anymore.

stefantalpalaru: > I think you disabled those rules by mistake

No, disabling them is the whole point of this PR. Those rules prevent any other external .gitignore override, as illustrated in the first comment.

PR #182, opened 2018-11-26
Rpc eth call

coffeepots:
Implements `eth_call`.

### DIFF
--- a/tests/test_rpc.nim
+++ b/tests/test_rpc.nim
@@ -1,7 +1,7 @@
 import
-  unittest, json, strformat, nimcrypto, rlp,
+  unittest, json, strformat, nimcrypto, rlp, options,
   json_rpc/[rpcserver, rpcclient],
-  ../nimbus/rpc/[common, p2p, hexstrings],
+  ../nimbus/rpc/[common, p2p, hexstrings, rpc_types],
   ../nimbus/constants,
   ../nimbus/nimbus/[vm_state, config],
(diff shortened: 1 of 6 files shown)


coffeepots:
Note: Requires https://github.com/status-im/nim-json-rpc/pull/43

coffeepots:
There's still some work required to test `eth_call`, which is currently on another branch, however this is bashing into other PRs so needs manual merging before it gets too far out of sync.

tests/test_rpc.nim
```diff
@@ -68,6 +68,13 @@ proc doTests =
   waitFor client.connect("localhost", Port(8545))
 
   suite "Remote Procedure Calls":
+    # TODO: Currently returning 'block not found' when fetching header in p2p, so cannot perform tests
+    test "eth_call":
```
stefantalpalaru: This test fails with:

```
Exception message: {"code":-32000,"id":1,"message":"eth_call raised an exception","data":"Parameter \"value\" is not valid as a UInt256 \"0x100\""}
```
Is it supposed to fail like that?

Also, is the comment above supposed to be moved to the next test, "eth_getBalance"? That's the one failing with:

```
Exception message: {"code":-32000,"id":3,"message":"eth_getBalance raised an exception","data":"No block hash for number 0"}
```

stefantalpalaru: Great.

PR #184, opened 2018-11-29
Update rpcutils

coffeepots:
A small update to `p2p` and `debug` to update the utility names.

As the `toHash` now accepts a `EthHashStr` directly, I took the liberty of changing the input parameter to the `debug_traceTransactions` RPC to accept it. This will mean an error will be raised if the input string isn't a hash - for example if it's too long - which doesn't happen with `HexDataStr`.


coffeepots:
@jangko I can't ask for your review directly, but can you comment on this as it touches your `traceTransaction` prototype.

jangko:
so far it doesn't have any unwanted side effect to traceTransaction implementation.

PR #186, opened 2018-11-30
add prune cli switch and implement transaction tracer

jangko:
this is needed for upcoming traceTransaction

### DIFF
--- a/nimbus/db/storage_types.nim
+++ b/nimbus/db/storage_types.nim
@@ -8,6 +8,7 @@ type
     blockHashToScore
     transactionHashToBlock
     canonicalHeadHash
+    slotHashToSlot
 
   DbKey* = object
     # The first byte stores the key type. The rest are key-specific values
@@ -40,6 +41,12 @@ proc blockNumberToHashKey*(u: BlockNumber): DbKey {.inline.} =
 proc canonicalHeadHashKey*(): DbKey {.inline.} =
   result.data[0] = byte ord(canonicalHeadHash)
   result.dataEndPos = 1
+
+proc slotHashToSlotKey*(h: openArray[byte]): DbKey {.inline.} =
+  assert(h.len == 32)
+  result.data[0] = byte ord(slotHashToSlot)
+  result.data[1 .. 32] = h
+  result.dataEndPos = uint8 32
 
 const hashHolderKinds = {genericHash, blockHashToScore, transactionHashToBlock}
 
--- a/nimbus/vm/computation.nim
+++ b/nimbus/vm/computation.nim
@@ -286,11 +286,14 @@ proc getGasRemaining*(c: BaseComputation): GasInt =
   else:
     result = c.gasMeter.gasRemaining
 
-template tracingEnabled*(c: BaseComputation): bool =
+proc tracingEnabled*(c: BaseComputation): bool =
   c.vmState.tracingEnabled
 
-template traceOpCodeStarted*(c: BaseComputation, op: string) =
+proc traceOpCodeStarted*(c: BaseComputation, op: string) =
   traceOpCodeStarted(c.vmState.tracer, c, op)
 
 proc traceOpCodeEnded*(c: BaseComputation) =
   c.vmState.tracer.traceOpCodeEnded(c)
+
+proc traceError*(c: BaseComputation) =
+  c.vmState.tracer.traceError(c)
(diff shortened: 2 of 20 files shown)


jangko:
waiting for status-im/nim-eth-trie#28

nimbus/vm/computation.nim
```diff
@@ -284,3 +285,12 @@ proc getGasRemaining*(c: BaseComputation): GasInt =
     result = 0
   else:
     result = c.gasMeter.gasRemaining
+
+template tracingEnabled*(c: BaseComputation): bool =
+  c.vmState.tracingEnabled
+
+template traceOpCodeStarted*(c: BaseComputation, op: string) =
```
yglukhov: It's better (safer) to make these `proc`s, not `template`s. Otherwise you evaluate `c` twice here.

jangko: thanks, I will change it to proc

nimbus/vm/transaction_tracer.nim
```diff
@@ -1,40 +1,77 @@
...
+    let mem = newJArray()
+    const chunkLen = 32
+    let numChunks = c.memory.len div chunkLen
+    for i in 0 ..< numChunks:
+      mem.add(%c.memory.bytes.toOpenArray(i * chunkLen, (i + 1) * chunkLen - 1).toHex())
+    j["memory"] = mem
+
+  # TODO: this seems very inefficient
+  # could we improve it?
+  # TODO: figure out how to get storage
```
zah: This TODO item is no longer relevant :)

nimbus/vm/transaction_tracer.nim
```diff
@@ -1,40 +1,77 @@
 import
```
zah: Why do we have `transaction_tracer.nim` as a separate file from `tracer.nim`? Then seem quite complimentary to me and both are tightly coupled to the expected json output.

zah:
Closed and re-opened to re-trigger the CI tests now that the `eth_trie` changes are merged.
The branch needs to be rebased before we can merge it.

nimbus/vm/transaction_tracer.nim
jangko: It is not relevant if the opcode finished execution. But in a buggy/failed contract, the opcode will raise exception, and `traceOpCodeEnded` never be called. In that situation, tracing information will be incomplete, at least compared to geth output. looks like geth allow opcode execution to be finished up to the problematic instruction and then report it as an error + possible expected outcome.

nimbus/vm/transaction_tracer.nim
jangko: perhaps I need to rename `tracer.nim` to something else. I envisioned it will contains not only `traceTransaction`, but also all functionality exposed by `debug_` rpc namespace.

jangko:
sorry, I forget to fix the test after I changed some of the procs signature

jangko:
This PR is ready, I will move to next part - exporting metadata to debugging tool.

nimbus/db/storage_types.nim
```diff
@@ -41,6 +42,12 @@ proc canonicalHeadHashKey*(): DbKey {.inline.} =
   result.data[0] = byte ord(canonicalHeadHash)
   result.dataEndPos = 1
 
+proc slotHashToSlotKey*(h: openArray[byte]): DbKey {.inline.} =
```
zah: @coffeepots, this is a scheme for a reverse map able to turn the "secure" storage hashed keys back to their original human-readable form. It's needed in some of the JSON-RPC procs and I remember that you planned to implement it.

I'm mentioning this in case you already have an alternative scheme or in case you could benefit from this implementation.

nimbus/vm/transaction_tracer.nim
```diff
@@ -60,6 +57,9 @@ proc traceOpCodeEnded*(tracer: var TransactionTracer, c: BaseComputation) =
 
   j["gasCost"] = %(tracer.gasRemaining - c.gasMeter.gasRemaining)
 
+  if c.lastOpCodeHasRetVal:
```
zah: Can't we inspect the last opcode by looking into the BaseComputation's CodeStream (`c.code`)? It would be nice to avoid updating the `lastOpCodeHasRetVal` on every instruction even when tracing is not used.

zah:
LGTM, but I'll let @coffeepots merge it. @jangko, can you please rebase the branch first?

coffeepots:
Really nice set of functionalities included in this PR.

nimbus/vm/transaction_tracer.nim
jangko: looks like we can do that once I get a reliable information on which opcodes actually return something

PR #187, opened 2018-12-04
Modexp fixes

mratsim:
This fixes all the modular exponentiation precompiles tests.

Highlights:
  - The tests are testing up to 8192-bit integers

Potential issues:
  - There is no gas tests
  - The very last test is very very slow `"nagydani-5-pow0x10001"`, the other two uint8192 tests are somewhat OK
  - The specs mentions that the result should be the same format as the modulus. The case where base and/or exponents are bigger than the modulus are not covered for example if base and exponents are Uint512 and modulus is uint256, the current code will return an uint512.


tersec:
`slowTest` from https://github.com/status-im/nimbus/blob/master/tests/test_helpers.nim#L20 can help with truly slow tests.

coffeepots:
LGTM :)

nimbus/vm/precompiles.nim
```diff
@@ -145,28 +138,62 @@ proc modExp*(computation: var BaseComputation) =
...
-    # Start with EVM special cases
+    # TODO: specs mentions that we should return in "M" format
+    #       i.e. if Base and exp are uint512 and Modulo an uint256
+    #       we should return a 256-bit big-endian byte array
 
     # Force static evaluation
-    func zero256(): static array[32, byte] = discard
-    func one256(): static array[32, byte] =
+    func zero(): static array[T.bits div 8, byte] = discard
+    func one(): static array[T.bits div 8, byte] =
```
coffeepots: Since the EVM is always big endian, do you need to do this here?

coffeepots: Since the EVM is always big endian, do you need to distinguish the endian here?

PR #189, opened 2018-12-05
Whisper rpc types

coffeepots:
Small PR that:

* Updates the whisper types and `FilterOptions` to be optional
* Removs the duplicated `WhisperIdentity` from `rpc_types.nim` as it's already present in `eth_common.eth_types`.
* Adds `WhisperFilterOptions`
* Updates the signature for `shh_newFilter` to use `WhisperFilterOptions` rather than the eth `FilterOptions`

### DIFF
--- a/nimbus/rpc/rpc_types.nim
+++ b/nimbus/rpc/rpc_types.nim
@@ -119,6 +119,10 @@ type
     toBlock*: Option[string]              # (optional, default: "latest") integer block number, or "latest" for the last mined block or "pending", "earliest" for not yet mined transactions.
     address*: Option[EthAddress]          # (optional) contract address or a list of addresses from which logs should originate.
     topics*: Option[seq[FilterData]]      # (optional) list of DATA topics. Topics are order-dependent. Each topic can also be a list of DATA with "or" options.
+  
+  WhisperFilterOptions* = object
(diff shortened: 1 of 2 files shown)


zah:
@kdeme, Ryan (@coffeepots) wasn't aware that you'll be also working on this. He'll be happy to leave the rest of the integration work to you.

coffeepots:
Yes, these changes are just updates to data structures really. In particular `shh_newFilter` was incorrect so that had to change anyway.

kdeme:
OK, that's fine.

I was however planning of changing the RPC calls to the list here: https://github.com/ethereum/go-ethereum/wiki/Whisper-v6-RPC-API

I assumed that this is the latest version and that https://github.com/ethereum/wiki/wiki/JSON-RPC is not up to date with regards to Whisper v6.

Or where should I be looking for the latest correct Whisper RPC API?

There is [this](https://github.com/ethereum/EIPs/pull/1474) EIP in the making, but Whisper is not included there.

zah:
The v6 RPCs would indeed be the most useful ones. It's what the Status app relies on.

coffeepots:
Yes I was working with https://github.com/ethereum/wiki/wiki/JSON-RPC so it looks like all of the RPC stubs for whisper are out of date now.

PR #191, opened 2018-12-06
Add separate badges for appveyor jobs

Swader:
Appveyor does not support badges per job yet, so this is a workaround. Badges much less customizable and using 3 proxy services, but it's a solution unless we want to roll our own server with little things like these. Had to list the windows jobs because the badges look the same, and pure text with no list doesn't look so good. Ideas for improving visuals?


stefantalpalaru:
Looks good to me.

PR #193, opened 2018-12-07
general-state testsuite changes and better fork selection

stefantalpalaru:
- skipped the tests in allowedFailInCurrentBuild()
- replaced doAssert() with check() in testFixtureIndexes() so we can see both hash values on failure
- checking filename extension for JSON tests to avoid editor swap files
- replaced the duplicated block values in the main net's ChainConfig with values from forkBlocks
- allowed overriding the current fork in computations, because the old strategy of only looking at the block number doesn't work with JSON tests where the block number is usually 1
- explicitly pass the fork to gasCosts() and use it for conditional cost calculation
- fixed a logic error in the CREATE opcode
- fixed VM selection based on current fork in updateOpcodeExec()
- single point of control for supported forks in tests (just one fork, at the moment)
- 44 new test failures (that were probably passing for the wrong reasons)


nimbus/vm/interpreter/opcodes_impl.nim
```diff
@@ -522,7 +522,7 @@ op create, inline = false, value, startPosition, size:
       computation.vmState.blockHeader.rlphash, false).
       getBalance(computation.msg.sender)
 
-  if senderBalance >= value:
+  if senderBalance > value:
```
coffeepots: Shouldn't this be
`if senderBalance < value:`?

I tried this change out but it leads to a stack overflow for test `recursiveCreateReturnValue` at call depth `328`, where the current `MaxCallDepth` is 1024. Changing `MaxCallDepth` to allow this to run, however, it still fails the test.

stefantalpalaru: No. That value is transferred from the sender to the newly created account.

stefantalpalaru: Good catch! It's fixed now, with another previously-passing test failing as a result.

PR #194, opened 2018-12-07
attempt to fix block 49018 sync

jangko:
I just can't help myself not to fix it, but I want to give it a try because it really annoying when developing debugging tool with this kind of bug.

another layer of indirection to accountStateDb. perhaps this is completely wrong approach, but it prevent multiple instance of statedb, while maintaining ergonomics.

along the way I also find numerous small bugs related to account state db usage.


stefantalpalaru:
You might want to rebase on top of the master branch, because we've been changing the same code.

jangko:
bad approach, closed

PR #196, opened 2018-12-11
more VM tracing

stefantalpalaru:
The existing vmState tracing is plugged into chronicles, at the TRACE level, to facilitate state test debugging.

Some useful chronicles defines are added to "tests/nim.cfg" to simplify the compile-and-run command for individual tests.


PMunch:
Looks good to me. Only thing is the imports which I've seen a lot of various styles for in the existing code-base. We seem to mix and match imports between stdlib, nimble packages, and local modules. It would be nice to have them sorted by type. But that's not specific to this issue.

stefantalpalaru:
When I notice that different import lines are used for different module types, I try to maintain the pattern.

Issue #188, opened 2018-12-05
first fruit of debugging tool, block 49018 statediff result


jangko:
```json
{
  "before": [
    {
      "name": "contract",
      "address": "630ea66c8c5dc205d45a978573fa86df5af1fe7a",                  
      "nonce": "0000000000000000",
      "balance": "0",
      "codeHash": "",
      "storage": {}
    },
    {
      "name": "sender",
      "address": "3d0768da09ce77d25e2d998e6a7b6ed4b9116c2d",
      "nonce": "000000000000000A",
      "balance": "43f0cf1107e922ce",
      "codeHash": "",
      "storage": {}
    },
    {
      "name": "miner",
      "address": "bb7b8287f3f0a933474a79eae42cbca977791171",
      "nonce": "0000000000000003",
      "balance": "47d269fc4069450d050",
      "codeHash": "",
      "storage": {}
    }
  ],
  "after": [
    {
      "name": "contract",
      "address": "630ea66c8c5dc205d45a978573fa86df5af1fe7a",
      "nonce": "0000000000000000",
      "balance": "0",
      "codeHash": "",
      "storage": {}  //// HERE IS THE PROBLEM
    },
    {
      "name": "sender",
      "address": "3d0768da09ce77d25e2d998e6a7b6ed4b9116c2d",
      "nonce": "000000000000000B",
      "balance": "43d6743654e7182e",
      "codeHash": "",
      "storage": {}
    },
    {
      "name": "miner",
      "address": "bb7b8287f3f0a933474a79eae42cbca977791171",
      "nonce": "0000000000000003",
      "balance": "47d6c1db0638c46daf0",
      "codeHash": "",
      "storage": {}
    }
  ]
}

//// IT SHOULD BE LIKE THIS
{
  "name": "contract",
  "address": "630ea66c8c5dc205d45a978573fa86df5af1fe7a",
  "nonce": "0000000000000000",
  "balance": "0",
  "codeHash": "",
  "storage": {
    "b10e2d527612073b26eecdfd717e6a320cf44b4afac2b0732d9fcbe2b7fa0cf6": "48656c6c6f20576f726c64210000000000000000000000000000000000000000",
    "0000000000000000000000000000000000000000000000000000000000000001": "000000000000000000000000000000000000000000000000000000000000000c",
    "0000000000000000000000000000000000000000000000000000000000000000": "0000000000000000000000003d0768da09ce77d25e2d998e6a7b6ed4b9116c2d"
  }
}
```

[...]

PR #197, opened 2018-12-12
Tracer part4: implement traceBlock, dumpBlockState, transaction statediff

jangko:
I hope this is the last missing piece of debugging tool infrastructure. 
after this, I will start to assemble debugging tool.
review please, also review #195 please.


tests/tracerTestGen.nim
```diff
@@ -0,0 +1,73 @@
...
+  ../nimbus/[tracer, vm_types]
+
+const backEnd {.strdefine.} = "lmdb"
+
+when backEnd == "sqlite":
+  import ../nimbus/db/backends/sqlite_backend
+elif backEnd == "rocksdb":
+  import ../nimbus/db/backends/rocksdb_backend
+else:
+  import ../nimbus/db/backends/lmdb_backend
```
zah: Is the `lmdb_backend` commited elsewhere? I don't see it here.

jangko: sqlite is vey slow even on SSD. I successfully build rocksdb but it crash at runtime, I don't want to build it again, it took very looong time.

I decided to write my own backend using lmdb, and disable the auto-commit during sync and manually commit synced block every 128 blocks. It prove to be faster than sqlite altough only using HDD.

but I doubt the [license](https://www.openldap.org/software/release/license.html) is compatible with nimbus license(or to be precisely, I don't understand they are compatible or not).

if someone can help me to decode the licenses and what to do with that, perhaps then I can add it to nimbus db backends.

zah: The OpenLDAP license seems to be a permissive BSD-style license, so using such software in Nimbus should be fine.
But please change the define name to something like `nimbus_db_backend`, set the default to "rocksdb" and use a local `nim.cfg` file specifying the lmdb option locally.

jangko:
review please, I have more PR to come based on this PR.

stefantalpalaru:
What' weird is that "~/.cache/nimbus/db/nimbus.db" keeps growing while the db seems stuck - 12 MB after 3.5 minutes of no debugging output.

I see the same lack of output with `-d:nimbus_db_backend=sqlite`. Only when using the RocksDB backend the output is back.

stefantalpalaru:
My bad. Debugging output eventually appears, after a much longer time than with RocksDB.

jangko:
both sqlite and lmdb are slow. lmdb will only become fast if you manually call `txBegin ... txCommit` to avoid auto commit.
I suspect we can also improve sqlite speed if we switch to manual commit.

with manual commit on a HDD, lmdb only takes about 10 minutes when sync to block 49018.

btw, I already have a solution to fix #188. I alread succesfully sync to block 49439 and located a bug in VM opcode `CALL` using tracing result, the screenshot is [here](https://github.com/jangko/nimbus/tree/debugging_tool/premix). 

the only thing hindered me to push more PRs is they depend on this PR.

zah:
I've reviewed this few times already, so I'm merging it now due to lack of activity of other reviewers.

jangko:
thank you @zah

nimbus/p2p/chain.nim
```diff
@@ -71,87 +36,27 @@ method persistBlocks*(c: Chain, headers: openarray[BlockHeader], bodies: openarr
...
-    assert(headers[i].bloom == bloom)
+    when not defined(release):
+      if validationResult == ValidationResult.Error:
+        dumpDebuggingMetaData(c.db, headers[i], bodies[i], vmState.receipts)
 
-    let receiptRoot = calcReceiptRoot(receipts)
-    if headers[i].receiptRoot != receiptRoot:
-      debug "wrong receiptRoot in block", blockNumber = headers[i].blockNumber, actual=receiptRoot, expected=headers[i].receiptRoot
-    assert(headers[i].receiptRoot == receiptRoot)
+    assert(validationResult == ValidationResult.OK)
```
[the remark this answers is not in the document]
stefantalpalaru: (I actually wrote this comment yesterday, but it seems I had to submit the review for it to appear. It's still a valid problem - we should also be validating in release builds)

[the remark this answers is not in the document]
jangko: > Why are we only doing this validation in debug builds?

sorry, that was part of my reflexes to type assert rather than throw a proper error

PR #199, opened 2018-12-14
better error handling

stefantalpalaru:
[this is in relation to https://github.com/status-im/nimbus/issues/198]

- added some useful `chronicles` defines in the top-level nim.cfg
- converted some assert() calls to doAssert() to avoid a reported
  segfault with -d:release
- the exceptions raised by doAssert() are being recovered from, in
  eth_p2p
- the only case that's a show stopper exits the program explicitly


arnetheduck:
don't sweep errors under the rug - if a network packet is able to cause a random failure in the app, that's a security bug or dos attack waiting to happen. fix the code so it de-taints the data that causes the issue to happen explicitly. 

asserts are for checking assumptions you've made - ie "this list should have been sorted" - they indicate bugs in the code

stefantalpalaru:
> don't sweep errors under the rug

I log them and, when possible, I recover from them. The previous behaviour was to have the program raise an uncatched exception in a debug build or wait until a further problem down the road triggered a segfault in a release build.

> asserts are for checking assumptions you've made - ie "this list should have been sorted" - they indicate bugs in the code

That's not how they've been used in here. From a pragmatic point of view, I'd rather get an incremental improvement and stability in a couple of days instead of rewriting a significant chunk of this exceptions-and-async bundle of joy in who knows how many months.

Let's make it work first and we'll do the rewrite(s) later.

arnetheduck:
> raise an uncatched exception in a debug build

yeah, so that's intentional - it means "around this code, we need to gain better understanding - for now, let's crash"

you're not making progress by creating another problem (that of hiding errors with catch-alls) and kicking this can down the road - we don't need to sync the eth1 chain any further until these issues have been dealt with.. 

if it's a genuine issue, the improvement here is to turn this into an exception and catch exceptions.. 
this is also in line with what's happening in nim, where exceptions are being moved towards defects and errors - the former are not meant to be handled.. 

keep in mind as well that many such defect-level errors are allowed to be turned into `quit` by the compiler, to avoid the exception overhead, per the nim manual

stefantalpalaru:
So the progress you're looking for is replacing `doAssert(...)` with `if ...: raise(ValidationError, "...")`?

yglukhov:
Those asserts are merely mocks of proper error handling, and replacing them with doAsserts doesn't make the code any more correct. Also the `quit` you added doesn't play well conceptually with @jangko's work, when we want to see the traces of the failed transactions.

I would recommend either not touching this for now, or instead of asserts return an error (`false`?) where appropriate.

arnetheduck:
> So the progress you're looking for is replacing

yep, if that's appropriate - or finding the root cause and introducing the fix there - or keep crashing until it's fixed..

this way we also have a more honest progress bar on how far along we are towards actually achieving sync. it's incredibly hard to remove these catch-alls later on when the corpus of code that depends upon them has grown.

mratsim:
I suggest we start to add `proc checkFoo` in one place like there: https://github.com/status-im/nimbus/blob/f8d3d0e721f836048719e216745091053af33c39/nimbus/errors.nim

Those `checkFoo` should raise the proper error and be in charge of logging as well.
I agree that asserts were used to detect bugs during rapid development and do not tag recoverable errors.

This is mentioned in this wiki page: https://github.com/status-im/nimbus/wiki/Exceptions-and-errors



I.e. once we are ready to recover from an error or exception, it should not be an assert anymore.

stefantalpalaru:
> Also the `quit` you added doesn't play well conceptually with @jangko's work, when we want to see the traces of the failed transactions.

Are you sure the RPC code hits that part and deals properly with the previous assert()?

yglukhov:
> Are you sure the RPC code hits that part and deals properly with the previous assert()?

No, I think this code will have to be changed differently from what you're suggesting to support the rpc thing.

stefantalpalaru:
Raising an exception that is not going to be caught works as a quit() replacement for me.

nimbus/db/db_chain.nim
```diff
@@ -250,7 +250,7 @@ proc persistBlockToDb*(self: BaseChainDB; blk: Block) =
 
   if blk.uncles.len != 0:
     let ommersHash = self.persistUncles(blk.uncles)
-    assert ommersHash == blk.header.ommersHash
+    doAssert ommersHash == blk.header.ommersHash
```
zah: Since this malformed data is coming from the network, this is a recoverable error. See my comments below regarding the proper usage of `doAssert`:

https://github.com/status-im/nim-eth-p2p/pull/55/files#r241794489

zah: Ah, I hadn't noticed the whole discussion about this :)

stefantalpalaru: Converted to error code.

stefantalpalaru:
Most explicit and implicit exceptions were converted to error codes and the error recovery is more clear now.

PR #200, opened 2018-12-24
new build system

stefantalpalaru:
- isolate dependencies (both Git repos and ".nimble" dir) in "vendor"
- ensure that everything is compiled using our version of Nim with
  something as simple as "make"
- pull the latest changes with "make update"
- re-create Nimble's package dir and re-build the Nim compiler automatically when needed
- ~~Nimble deps now point to Git HEAD~~
- "env.sh" script can be used to prefix any command that needs to use
  "nim" or "nimble" from our Nim fork
- "fetch-dlls" `make` target for Windows to get precompiled Snappy, RocksDB and SQLite DLLs

Test this in a clean directory, like a new user would:
```bash
git clone https://github.com/status-im/nimbus.git
cd nimbus
git checkout build
# build nimbus
make
./build/nimbus --help
# run the tests
make test
# some repo was updated upstream?
make update
# delete all the binaries and the .nimble dir, so we can start again
make clean
```
I tested it on Linux, macOS Sierra 10.12.5 (inside VMware Workstation, ~~where Nimble got confused about dependency resolution on the first "make" run, but managed to work on the second~~), 32-bit Windows 7, 64-bit Windows 10.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -101,8 +101,10 @@ install:
 
 build_script:
   - cd C:\projects\%APPVEYOR_PROJECT_SLUG%
-  - nimble install -y
+  - nimble install -y --depsOnly
+  - nimble nimbus
 test_script:
+  - build\nimbus.exe --help
   - nimble test
 
 deploy: off
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,13 @@
-nimcache/
+/nimcache
 
 # Executables shall be put in an ignored build/ directory
-build/
+/build
+
+# external packages
+/vendor
+
+# ntags/ctags output
+/tags
 
 # Ignore dynamic, static libs and libtool archive files
 *.so
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,5 +51,9 @@ install:
   - cd ../..
 
 script:
-  - nimble install -y
+  # fail fast
+  - set -e
+  - nimble install -y --depsOnly
+  - nimble nimbus
+  - ./build/nimbus --help
   - nimble test
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,8 @@ REPOS := $(addprefix $(REPOS_DIR)/, \
 	status-im/nim-eth-keyfile \
 	status-im/nim-eth-bloom \
 	status-im/nim-bncurve \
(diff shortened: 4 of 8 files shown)


arnetheduck:
a few questions:

* does it support windows?
* does it scale with transitive dependencies? (aka what to do when someone adds a dep to `nim-eth-p2p`)
* what does the `#head` change do, compared to what we have now?
* the standard way of choosing nim version is through `choosenim` - how does this compare and integrate?

stefantalpalaru:
>     * does it support windows?

Yes. I added instructions for building it with MinGW and "Git for Windows". It would be nice to find a better Rocksdb installation scenario than this: https://github.com/facebook/rocksdb/wiki/Building-on-Windows
 
> does it scale with transitive dependencies? (aka what to do when someone adds a dep to `nim-eth-p2p`)

Yes, because Nimble is still used for dependency management.

> what does the `#head` change do, compared to what we have now?

It requests the latest commit, not the latest release tag for a package.
 
> the standard way of choosing nim version is through `choosenim` - how does this compare and integrate?

It doesn't and it shouldn't, because we want everything to be compiled using our Nim fork.

Developers wanting to test other Nim versions, can do so by not prepending `./env.sh` to their Nimble commands (i.e.: run `nimble nimbus` and `nimble test` which will use the system's Nim).

stefantalpalaru:
I had to give up on the "...#head" dependency versioning because of https://github.com/nim-lang/nimble/issues/432, but it turns out I didn't actually need it to avoid .nimble/pkgs/* duplication.

cheatfate:
This is instructions how to build rocksdb on windows much easier https://github.com/status-im/nimbus-deps.

stefantalpalaru:
I've implemented a "fetch-dlls" `make` target for Windows based on ".appveyor.yml".

zah:
This is working nicely and I've now integrated the nim-beacon-chain repo as well.
My only suggestions are to:

1) Shorten the paths to the inner repos
  I find myself typing these paths quite often in the terminal or vim and I'll appreciate if the paths were shorter (e.g. nimbus/deps/nim-beacon-chain)

2) Don't use nimble when building nimbus.
    This slows things down a bit and it's not necessary for now IMO.

PR #217, opened 2019-01-23
Premix-related changes

stefantalpalaru:
- build all tools using the Makefile and place them in ./build
- add copyright header to Makefile
- premix/readme.md edited to fix some errors, improve descriptions and
  document Makefile usage
- link the Premix documentation in the main README.md
- also build `hunter` in the nimble tests
- refactored the file and directory search so all debugging tools can be run
  from the top-level dir like this: `./build/<tool> ...`
- write all JSON debugging data in the current directory
- add JSON files generated in the top-level dir to .gitignore
- Nimbus now exits with an exception after dumping debug data and
  running `premix` on it

This PR depends on: https://github.com/status-im/nim-rlp/pull/25, https://github.com/status-im/nim-eth-trie/pull/36

Since I don't have any fully synced blocks from `geth` right now (looks like it will spend days doing a fast sync before getting to the full sync part) I need a tester who does. Please run this from the top-level dir:

```bash
# in one terminal:
geth --rpc --rpcapi eth,debug --syncmode full --gcmode=archive --maxpeers 0
# in another terminal
make
./build/nimbus --prune:archive --port:30304
```
After a few minutes, you should see a debugging report opened in a new browser tab. Let me know if it has actual debugging data in it.


zah:
How does this depend on the RLP changes btw?

stefantalpalaru:
Without those changes, Nimbus crashes during the debug info generation step due to an RLP instance that gets reduced to nothing through recursive merges.

Since this has not been reported before, it is possible that it gets triggered by data from those specific Ethereum nodes I connect with.

jangko:
This PR looks good and perhaps not depend on RLP changes, see my comment at status-im/nim-eth-trie#36. They are separate issues.

jangko:
> Since I don't have any fully synced blocks from geth right now (looks like it will spend days doing a fast sync before getting to the full sync part) I need a tester who does.

Although I don't understand what have caused your geth behaviour, I have tested your changes and It produce nice debugging data on my machine(windows 10 64 bit)

stefantalpalaru:
You can see what I did to get the crash before a dump in the first comment of this PR.

How is that supposed to be a good thing, if I don't get any debugging output to help me debug this RLP that gets merged to nothing?

jangko:
I cannot reproduce it, the only difference is, my copy of geth doing full sync since block 1 onwards.

stefantalpalaru:
We're obviously connecting to different peers and getting different data. I'm using geth-1.8.21.

stefantalpalaru:
I found and documented my `geth` problem - at some point in the past it started to do a fast sync, so it will keep doing that regardless of the command line option telling it otherwise. I deleted its db directory and it started doing a full sync from the start, as advertised.

I also reworded the HTML help and got rid of that fake menu that was supposed to be an "image" of the real menu. Yes, I tried clicking on it because it was the most visible menu there :-)

This PR is ready to merge.

P.S.: I also tried getting Premix to work with Parity, leading to a nim-json-rpc fix that's coming in another PR. Unfortunately, Parity doesn't have a "debug_traceTransaction" RPC method in its API. The closest is "trace_rawTransaction", but it has different response fields.

PR #223, opened 2019-02-01
[WIP] Fix CALL-family Opcodes

jangko:


nimbus/db/state_db.nim
```diff
@@ -171,13 +171,26 @@ proc getCode*(db: AccountStateDB, address: EthAddress): ByteRange =
...
-proc hasCodeOrNonce*(account: AccountStateDB, address: EthAddress): bool {.inline.} =
-  account.getNonce(address) != 0 or account.getCodeHash(address) != EMPTY_SHA3
+proc hasCodeOrNonce*(db: AccountStateDB, address: EthAddress): bool {.inline.} =
+  db.getNonce(address) != 0 or db.getCodeHash(address) != EMPTY_SHA3
 
 proc dumpAccount*(db: AccountStateDB, addressS: string): string =
   let address = addressS.parseAddress
   return fmt"{addressS}: Storage: {db.getStorage(address, 0.u256)}; getAccount: {db.getAccount address}"
 
+proc accountExist*(db: AccountStateDB, address: EthAddress): bool =
```
yglukhov: `accountExists`? Note the trailing `s`.

jangko: Thank you, I will fix it at next commit.

jangko:
some notes for next week goals:

From Yellow Paper page 4 point 13 & 14:
> An account is **empty** when it has no code, zero nonce and zero balance
> An account is **dead** when its account state is **non-existent** or **empty**
> Even callable precompiled contracts can have an **empty** account state. This is because their account states do not usually contain the code describing its behavior.

I will add something:
> Precompiled contracts can have an **empty** account state but **should not dead**.

I found the culprit in `applyMessage` proc in `computation.nim`. Need to resurrect the dead account.
- [ ] resurrect the dead precompiled account.
- [ ] completing `call` gas cost equation.
- [ ] completing `call` gas params.

zah:
@jangko, I've merged this because we are making some restructuring of the repos (most of the small libraries are now combined in a single `eth` package and asyncdispatch2 is being renamed to `chronos`). I didn't notice the failing VM test cases, I hope you'll be able to address them soon.

PR #226, opened 2019-02-07
Fix call part2

jangko:
summary of this PR:
- various fixes of debugging tools
- fixes regression of previous PR
- trade 7 failed GST with 107 OK GST
- sync up to block 50295 -> hit weird bug in Nimbus, related to `accountDB.setStorage`, unable to produce debugging data because it crash.


jangko:
this PR is ready.

premix/dumper.nim
```diff
@@ -26,6 +27,9 @@ proc dumpDebug(chainDB: BaseChainDB, blockNumber: Uint256) =
 
   captureChainDB.setHead(parent, true)
   let validationResult = processBlock(captureChainDB, parent, header, body, vmState)
+
+  # success or not dispose it
+  transaction.dispose()
```
zah: It's better if you place `transaction.dispose()` in `finally` or `defer` blocks.

zah: I tried address this here:
https://github.com/status-im/nimbus/commit/48cdc6da0b8998e453371c56bc53f1732bca3c47

I was a little bit confused by the code at one of the places. Please note that if you want the transaction to be applied to the database, you must explicitly call `transaction.commit()` at some point. `dispose` is something like a destructor for the transaction object and it's safe to call even on commited transactions. After we upgrade to Nim 0.20, we'll use a real destructor.

jangko: actually, my intention is to discard any changes to the db, can `dispose` do that?

zah: If you don't `commit`, no changes will be ever written. Calling `dispose` just throws away the transaction in this case.

jangko: Thanks. Perhaps i misused `dispose`.
How to prevent any changes on DB from `processBlock` entering `dumpDebuggingMetaData`?
I want both `processBlock` and `dumpDebuggingMetaData` have same starting state.

zah: Yes, calling `dispose` also removes the "layer" of visible changes that were written to memory since the start of the transaction, so I think your usage is correct here.

PR #229, opened 2019-02-12
fixes #228, hush little baby bug

### LINKED ISSUE #228: investigating block 50295 crash
how to reproduce:
using `nimbus` or `persist`. sync until block 50295. it will crash with:

```text
persist.nim(104)         persist
persist.nim(76)          main
eth_types.nim(377)       persistBlocks
chain.nim(46)            persistBlocks
executor.nim(176)        processBlock
executor.nim(9)          processTransaction
executor.nim(27)         contractCall
vm_state_transactions.nim(50) execComputation
interpreter_dispatch.nim(255) executeOpcodes
interpreter_dispatch.nim(243) updateOpcodeExec
interpreter_dispatch.nim(210) frontierVM
opcodes_impl.nim(452)    sstore
state_db.nim(104)        setStorage     /// <---- interesting line
hexary.nim(632)          put
hexary.nim(624)          put
hexary.nim(610)          mergeAt
hexary.nim(529)          mergeAtAux
hexary.nim(519)          mergeAt
rlp.nim(67)              rawData
rlp.nim(236)             currentElemEnd
system.nim(3790)         failedAssertImpl
system.nim(3783)         raiseAssert
system.nim(2830)         sysFatal
```

`state_db.setStorage` crash only happened if the db is corrupted or the state trie is incomplete.

When using `hunter`, it pass block validation without any issue. This means the bug is not inside the VM but outside VM.

It involves contract address `0x109c4f2ccc82c4d77bde15f306707320294aea3f`

Looking at etherscan.io history, right before block 50295, there is a failed OOG contract call at block 50294. My current suspicion is when the OOG happened, the state trie is not updated properly, but where? can some transaction.dispose() call at wrong place? or we already need some journalDB?

jangko:
small bug it is, but it well hidden in the forest of tries.
this bugfix allow us to sync until block 55296
battle with contract call bug continues....

### DIFF
--- a/nimbus/db/state_db.nim
+++ b/nimbus/db/state_db.nim
@@ -80,7 +80,10 @@ template createTrieKeyFromSlot(slot: UInt256): ByteRange =
   # morally equivalent to toByteRange_Unnecessary but with different types
 
 template getAccountTrie(stateDb: AccountStateDB, account: Account): auto =
-  initSecureHexaryTrie(HexaryTrie(stateDb.trie).db, account.storageRoot)
+  # non-pruning trie at other places is optional
+  # but storage trie seems prefer non-pruning trie
+  # see bug #228
+  initSecureHexaryTrie(HexaryTrie(stateDb.trie).db, account.storageRoot, false)
 
 # XXX: https://github.com/status-im/nimbus/issues/142#issuecomment-420583181
 proc setStorageRoot*(db: var AccountStateDB, address: EthAddress, storageRoot: Hash256) =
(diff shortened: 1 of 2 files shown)


stefantalpalaru:
Could it be a deeper problem with the way we prune our tries?

jangko:
Digging deeper, I finally understand the problem. And we can leave the trie as the last suspect in this case.

The problem involves two or more accounts with the same `storageRoot`. To simplify the illustration, I will use only two accounts.

During transaction `U` execution, account `A` was modified and it has `storageRoot=0xABCDEF`
During transaction `V` execution, account `B` was modified and also arrive at `storageRoot=0xABCDEF`

Later on, during transaction `X`, either account `A` or `B` is modified again and it has `storageRoot=0xDEADBEEF`.
Here the problem arise, during transaction `Y` execution, account with `storageRoot=0xABCDEF` will have corrupted/incomplete trie. The other account that previously have the same `storageRoot` already doing damage on the trie if the pruning is turned on.

Perhaps this is more of design issue.

@zah, you know the hexary trie better than me, do you have any idea how to approach this situation? although my little fix works, but conceptually it doesn't remove the problem.

jangko:
closed in favor of #232

PR #231, opened 2019-02-13
Stefan's submodules work

zah:
This still feels as an improvement over our current situation, so I'm willing to give it a try for few weeks. Does anyone have strong objections?

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -25,12 +25,13 @@ build_script:
   - git submodule update --init --recursive --depth 10
   # the 32-bit build is done on a 64-bit image, so we need to override the architecture
   - mingw32-make ARCH_OVERRIDE=%PLATFORM% fetch-dlls
-  - mingw32-make -j2 ARCH_OVERRIDE=%PLATFORM% V=0 nimbus
+  - dir C:\mingw-w64
 
 test_script:
-  - dir C:\mingw-w64
+  - mingw32-make -j2 ARCH_OVERRIDE=%PLATFORM% V=0 nimbus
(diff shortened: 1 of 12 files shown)


arnetheduck:
technically, the instead of a lock file, `nimble` could spit out a directory with vendor submodules.. I suspect we'll quickly run into severe limitations though. 

I have no objections to try this out, but the approach is inherently limited / does not scale, as is.
as a concrete example, it does not really help us build `nim-beacon-chain`, unless we fold it into the nimbus repo.

In general, the repos are not getting any easier to build and navigate, and this is something to think carefully about, both for the average nim developer, and for people trying to ease into nim from a different language - use nimble! no, use nix! no use a makefile with submodules! no, go to the nim-beacon-chain repo! no, that code is in some small one-file repo that will be grand 5 years from now! no, back to nimbus to run beacon_node because else things will break! no, use choosenim! no, use nimble develop! or maybe it was nimble install! or...

It's easy to explain away all these things when you're a seasoned nim developer that has learned all the excuses for why things are so complicated and all the workarounds for how to fix things when they break ("oh, when merging submodules, must take care to update the submodule ref to the merge commit of the dependent repo, and not just the commit on the branch that's in the PR")

zah:
Yes, to clarify, I'm still advocating for [Nimble lock files](https://github.com/nim-lang/nimble/issues/127) to be implemented as soon as possible (we'll be supporting a long-term external contributor for this work). What I meant is that Stefan approach improves on the status quo and trying it now will give us one more data point for comparison in the future.

PR #232, opened 2019-02-15
various contract call fixes

jangko:
summary of this PR:
* allow us to sync until block 78458
* pass additional 48 GST
* each commit consists of single topic, make review easier


nimbus/vm/computation.nim
```diff
@@ -216,17 +217,24 @@ proc applyCreateMessage(fork: Fork, computation: var BaseComputation, opCode: st
...
-    fork.applyCreateMessage(childComp, opCode)
-  else:
-    applyMessage(childComp, opCode)
+  try:
+    if childMsg.isCreate:
+      fork.applyCreateMessage(childComp, opCode)
+    else:
+      applyMessage(childComp, opCode)
+  except VMError:
+    # weird Nim bug
```
zah: What is the Nim bug here? A more detailed comment would help.

nimbus/vm_state.nim
```diff
@@ -64,12 +66,15 @@ method getAncestorHash*(vmState: BaseVMState, blockNumber: BlockNumber): Hash256
   if ancestorDepth >= constants.MAX_PREV_HEADER_DEPTH or ancestorDepth < 0:
     return
 
-  let idx = ancestorDepth.toInt
-  if idx >= vmState.prevHeaders.len:
-    return
+  result = vmState.chainDB.getBlockHash(blockNumber)
+  #TODO: should we use deque here?
+  when false:
```
zah: You should either erase old code like this or provide a comment on why it's currently disabled.

nimbus/vm/computation.nim
jangko: the bug is static param vs `Error` object construction clash. I will add comment later

nimbus/vm_state.nim
jangko: ok, I will add comment later

stefantalpalaru:
> So, how was the pruning tree issue resolved in the end?

Looks like the same workaround of not pruning the trie.

jangko:
the pruning trie problem will be solved  by adding another layer of `child db` or `prefix db` only for storage trie like ethereumj, but other ethereum client such as ruby-ciri and geth seems only turn off pruning.

because the `prefix-db` solution will take quite alot disk space, I will compare it with non-pruning trie, which one consume less space.

in this PR, I only turned off pruning. the permanent solution will come later

PR #233, opened 2019-02-17
Fix block 78458 crash

jangko:
summary of this PR:
* allow us to sync until block 81383
* pass additional 122 GST


zah:
Will merge once the CI is green.

jangko:
> Let's make the snapshot API and usages a bit more exception-safe.

agree, it should be our next priority.

jangko:
why the Appveyor often failed with this kind of error now?

> Error downloading remote file: One or more errors occurred. Inner Exception: Unable to connect to the remote server. Command exited with code 2

cheatfate:
Mostly because `sourceforge.net` from where we trying to download `mingw` distro are not available 100% of time.

stefantalpalaru:
> Mostly because `sourceforge.net` from where we trying to download `mingw` distro are not available 100% of time.

Did you know that multiple Mingw-w64 versions are already available in AppVeyor images? That's what I ended up using after running into the same problem: https://github.com/status-im/nimbus/blob/5ade849e176725f2e1fa16a35e91b56bbe3fc1bd/.appveyor.yml

PR #237, opened 2019-02-19
Fix block 81383 and block 81666 problem

jangko:
summary of this PR:
* allow us to sync until block 85858
* pass additional 29 GST


jangko:
with this additional commits, we are able to reach block 196647, almost 200K

jangko:
now we are at block 226522, over 226K

zah:
This looks great, @jangko, kudos for the progress.
There is a strange RPC test failure though, I struggle to see how it could be related to the changes here, but can you please take a look? I'll merge everything once the CIs are green.

jangko:
if we compile and run `test_rpc` alone, the crash not appear. it only appear if we compile it together with `all_tests.nim`.
Also, appveyor 32 bit not failed. don't know why.

PR #242, opened 2019-02-23
Fix block 231501 problem

jangko:
* this PR allow us to reach block 248032, almost 250K.
* pass additional 32 GST
* enable all_tests again


jangko:
we are at block 299804, almost 300K

zah:
Superb, the jumps forward are getting longer and longer :)

PR #247, opened 2019-02-25
Fix block 512335 problem

jangko:
* we are at block 652.148 -- lots of heavy computation.
* pass additional 12 GST


stefantalpalaru:
Some tracer tests are failing.

nimbus/vm/transaction_tracer.nim
```diff
@@ -24,13 +24,16 @@ proc initTracer*(tracer: var TransactionTracer, flags: set[TracerFlags] = {}) =
   tracer.accounts = initSet[EthAddress]()
   tracer.storageKeys = @[]
 
-proc rememberStorageKey(tracer: var TransactionTracer, compDepth: int, key: Uint256) =
+proc prepare*(tracer: var TransactionTracer, compDepth: int) =
   if compDepth >= tracer.storageKeys.len:
     let prevLen = tracer.storageKeys.len
     tracer.storageKeys.setLen(compDepth + 1)
-    for i in prevLen ..< tracer.storageKeys.len:
+    for i in prevLen ..< tracer.storageKeys.len - 1:
```
stefantalpalaru: You're introducing an off-by-one error here, leaving the last element of "tracer.storageKeys" uninitialised. At some point you had to go back and initialise it separately on line 34. It's easier to get rid of both the "- 1" and line 34.

Remember that "..<" already excludes the value on its right.

jangko: that is intentional, I want to initialize *or* clear previous storage keys.

stefantalpalaru: Read it again. In the loop, you initialise keys 0 to (len -2). Outside the loop you initialise key (len-1) in exactly the same way.

jangko: inside the loop: initialize.
outside the loop(line 34): initialize *or* clear previous _depth_ keys.
they are not exactly the same.

the key to understand this is, `compDepth` is not increasing sequentially like 1,2,3,4,...
`compDepth` can be something like 1,2,3,4,3,3,2,2

jangko: line 34 is not always equal to initialize last element

stefantalpalaru: It doesn't matter what value `compDepth` has. The length of `tracer.storageKeys` is set to `compDepth + 1`.

> inside the loop: initialize.

Like I said, inside the loop you initialise keys with indices 0 through (len-2).

> outside the loop(line 34): initialize _or_ clear previous _depth_ keys.

No. It's the same exact `initSet[Uint256]()` as inside the loop, but for the element with the index `compDepth` - which is to say `len-1`, because the length was defined as `compDepth+1`.

jangko: but it works. and it works well. it clears the storage keys like I want, and i don't see anything to change.
did you missed the `if compDepth >= tracer.storageKeys.len:`, storageKeys.len does not always comptDepth+1

jangko: if i remove the `-1` and remove line 34. it produce different result. I will stick with current implementation.
you can prove your idea with different implementation.
please try it with block 512335.

stefantalpalaru: OK, I'll merge it as is, but sooner or later you'll forget why that unusual-looking code is the way it is and that initialising `tracer.storageKeys[compDepth]` clears an existing set if `compDepth < tracer.storageKeys.len` or completes the initialisation of new seq fields otherwise.

PR #253, opened 2019-02-27
new member of premix tool set

jangko:


jangko:
Using this tool I'm able to resync from block 46000 to 668910 relatively faster than persist tool.
recent contract call and contract creation refactor already verified: no regression.

zah:
Quite nice :) 

How big is your local database after reaching block 668910?

jangko:
windows rocksdb 64bit: 10.7GB, perhaps no compression. 
geth with snappy compression: 1.2M blocks -> 11,8GB

PR #264, opened 2019-03-07
CI: don't set a depth for submodule cloning

stefantalpalaru:
because the build will fail when we're too far behind upstream and our
target commit is not included in that depth any more


stefantalpalaru:
The build failure appeared here: https://github.com/status-im/nimbus/pull/263

PR #269, opened 2019-03-11
Makefile: specify the Chronicles log level

stefantalpalaru:


zah:
How do we make release builds with the Makefile?

stefantalpalaru:
I added and documented the EXTRA_NIM_PARAMS variable that's meant to be overridden on the command line.

PR #270, opened 2019-03-12
ci: make everything

arnetheduck:


stefantalpalaru:
That would be an unnnecessary duplication, since all those other tools besides "nimbus" are being built by the "test" target.

arnetheduck:
I'd be more comfortable having it there though because that's what we recommend people type, in the readme. things change.

stefantalpalaru:
Works for me. Just remove the "$(TOOLS)" dependency from the "test" target in the Makefile.

arnetheduck:
is that a good idea though? won't they be stale in case someone runs just `make test`?

stefantalpalaru:
Look at the CI output and you'll see you're building those tools twice for no reason: https://travis-ci.org/status-im/nimbus/jobs/505241330

arnetheduck:
well, generally, if make was smart about it, it would notice that and not rebuild the parts that are not needed.. what I'm saying is that someone else using the makefile (outside of the CI) might run `make test` - if I remove `$tools`, will they have a good experience?

stefantalpalaru:
> if make was smart about it, it would notice that and not rebuild the parts that are not needed.. 

Telling make which \*.nim files to look at to decide whether the resulting binary is older than any of them is not trivial, so I didn't even try to implement it.

> someone else using the makefile (outside of the CI) might run make test - if I remove $tools, will they have a good experience?

Yes, of course. Those tools are not actually used to run the test suite. We want to build them during CI just to make sure they keep building. That's it.

PR #273, opened 2019-03-13
add missing Chain.getStateDB() method

stefantalpalaru:
that's actually called from getStorageNode()

### DIFF
--- a/nimbus/p2p/chain.nim
+++ b/nimbus/p2p/chain.nim
@@ -64,6 +64,6 @@ method persistBlocks*(c: Chain, headers: openarray[BlockHeader], bodies: openarr
 
   transaction.commit()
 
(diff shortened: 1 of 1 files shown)


jangko:
in `vm_state.nim` we also have `getStateDb(): AccountStateDb`, now a new proc also named `getStateDb()`, but it return `TrieDatabaseRef`, shouldn't we give it another name to avoid confusion?

stefantalpalaru:
> now a new proc also named getStateDb()

I'll rename it but, for the record, this is not a new proc. It was already defined as a base method for AbstractChainDB in vendor/nim-eth/eth/common/eth_types.nim:350.

stefantalpalaru:
Renamed to `getTrieDB()`.

PR #283, opened 2019-03-22
Docker builds for beacon_node and nimbus

zah:


docker/nimbus/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM statusteam/nim-base
+MAINTAINER Zahary Karadjov <zahary@status.im>
+
+RUN apt-get update && apt-get install -y build-essential make librocksdb-dev && \
+    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+RUN git clone https://github.com/status-im/nimbus.git && \
+    cd nimbus && \
+    make update deps
```
stefantalpalaru: Not needed here, because you run `make nimbus` later on.

zah: The split here is good in general for faster iterations when you make tweaks to the second build step. You know how Docker caches the result of each `RUN` step and skips the execution if the commands didn't change.

stefantalpalaru: OK

PR #284, opened 2019-03-22
add Makefile for beacon_node Dockerfile, make image smaller

jakubgs:
Just some small improvements:

* Added `Makefile` for easier image creation
* Added small `README` explaining the purpose
* Split `Dockerfile` into two stages to slim it down

Question:
__What should be the default value for `NETWORK_BACKEND`, `SHARDS_COUNT`, and `SLOTS_PER_EPOCH`?__

Status: __WIP__

### DIFF
--- a/docker/beacon_node/Dockerfile
+++ b/docker/beacon_node/Dockerfile
@@ -1,26 +1,41 @@
-FROM statusteam/nim-base
-MAINTAINER Zahary Karadjov <zahary@status.im>
+FROM statusteam/nim-base AS build
 
-ARG network
-ARG network_backend
+RUN apt update \
+ && apt install -y build-essential make \
+ && apt clean \
+ && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-RUN apt-get update && apt-get install -y build-essential make librocksdb-dev && \
-    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+RUN git clone https://github.com/status-im/nimbus.git \
+ && cd nimbus \
+ && make update deps
 
-RUN git clone https://github.com/status-im/nimbus.git && \
-    cd nimbus && \
-    make update deps
+ARG NETWORK
+ARG NETWORK_BACKEND
 
-RUN cd nimbus && \
-    set -a && \
-    . vendor/nim-beacon-chain/scripts/${network}.env && \
-    ./env.sh nim \
+RUN cd nimbus \
+ && set -a \
+ && . vendor/nim-beacon-chain/scripts/${NETWORK}.env \
+ && ./env.sh nim \
       -o:/usr/bin/beacon_node \
       -d:release \
-      -d:with$network_backend \
-      -d:SHARD_COUNT=$SHARD_COUNT \
-      -d:SLOTS_PER_EPOCH=$SLOTS_PER_EPOCH \
+      -d:with${NETWORK_BACKEND} \
+      -d:SHARD_COUNT=${SHARD_COUNT} \
+      -d:SLOTS_PER_EPOCH=${SLOTS_PER_EPOCH} \
       c vendor/nim-beacon-chain/beacon_chain/beacon_node.nim
 
-ENTRYPOINT ["beacon_node"]
+# --------------------------------- #
+# Starting new image to reduce size #
+# --------------------------------- #
+FROM debian:9-slim
 
+RUN apt update \
+ && apt install -y librocksdb-dev \
+ && apt clean \
+ && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+COPY --from=build /usr/bin/beacon_node /usr/bin/beacon_node 
+
+MAINTAINER Zahary Karadjov <zahary@status.im>
+LABEL description="Nimbus installation that can act as an ETH2 network bootstrap node."
+
+ENTRYPOINT ["/usr/bin/beacon_node"]
--- a/docker/beacon_node/Dockerfile
+++ b/docker/beacon_node/Dockerfile
@@ -5,8 +5,11 @@ RUN apt update \
  && apt clean \
  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
+ARG GIT_REVISION
+
 RUN git clone https://github.com/status-im/nimbus.git \
  && cd nimbus \
+ && git reset --hard ${GIT_REVISION} \
  && make update deps
 
 ARG NETWORK
@@ -18,6 +21,7 @@ RUN cd nimbus \
  && ./env.sh nim \
       -o:/usr/bin/beacon_node \
       -d:release \
+      --lineTrace:on \
       -d:with${NETWORK_BACKEND} \
       -d:SHARD_COUNT=${SHARD_COUNT} \
       -d:SLOTS_PER_EPOCH=${SLOTS_PER_EPOCH} \
@@ -29,7 +33,7 @@ RUN cd nimbus \
 FROM debian:9-slim
 
 RUN apt update \
- && apt install -y librocksdb-dev \
+ && apt install -y librocksdb-dev curl \
  && apt clean \
  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
--- a/docker/beacon_node/Makefile
+++ b/docker/beacon_node/Makefile
@@ -1,0 +1,15 @@
+# These default settings can be overriden by exporting env variables
+NETWORK         ?= testnet0
+NETWORK_BACKEND ?= rlpx
+
+IMAGE_TAG       ?= $(NETWORK)-$(NETWORK_BACKEND)
+IMAGE_NAME      ?= statusteam/nimbus_beacon_node:$(IMAGE_TAG)
+
+build: $(NIX_INSTALL)
+	docker build \
+		--build-arg="NETWORK=$(NETWORK)" \
+		--build-arg="NETWORK_BACKEND=$(NETWORK_BACKEND)" \
+		-t $(IMAGE_NAME) .
+
+push: build
+	docker push $(IMAGE_NAME)
--- a/docker/build_beacon_node.sh
+++ b/docker/build_beacon_node.sh
@@ -2,21 +2,17 @@
 
 set -euv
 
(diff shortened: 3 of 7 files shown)


jakubgs:
@zah okay, I'll remove those arguments, but what about `NETWORK_BACKEND`?

jakubgs:
It's either `rlpx` or `libp2p` right?

jakubgs:
Though truth be told this would make much more sense if one could specify these parameters at runtime rather than at compile time. That's quite an anti-pattern to make different images with different configurations.

jakubgs:
>Is this intended?

No? Why would it be intended?
Repos start without any groups added by default. You should have access now.

jakubgs:
Hmmm, but with `-d:chronicles_sinks=json` it seems to fail to build:
```
spec/beaconstate.nim(456, 9) template/generic instantiation from here
../../nim-chronicles/chronicles/log_output.nim(37, 63) template/generic instantiation from here
../../nim-chronicles/chronicles/log_output.nim(527, 30) template/generic instantiation from here
../../nim-chronicles/chronicles/log_output.nim(503, 42) Error: type mismatch: got <Slot>
but expected one of: 
proc `%`(formatstr, a: string): string
proc `%`(formatstr: string; a: openArray[string]): string
proc `%`[T](elements: openArray[T]): JsonNode
proc `%`(o: ref object): JsonNode
proc `%`(s: string): JsonNode
proc `%`(o: enum): JsonNode
proc `%`(b: bool): JsonNode
template `%`(j: JsonNode): JsonNode
proc `%`(n: BiggestInt): JsonNode
proc `%`(o: object): JsonNode
proc `%`(n: float): JsonNode
proc `%`(keyVals: openArray[tuple[key: string, val: JsonNode]]): JsonNode

expression: %va
```

docker/beacon_node/Makefile
```diff
@@ -0,0 +1,15 @@
+# These default settings can be overriden by exporting env variables
+NETWORK         ?= testnet0
+NETWORK_BACKEND ?= rlpx
+
+IMAGE_TAG       ?= $(NETWORK)-$(NETWORK_BACKEND)
```
jakubgs: This way we always have the tag formatted from those two arguments.

zah:
Feel free to disable `-d:chronicles_sinks=json` for now. I'll enable it later.

jakubgs:
I'm also getting this when running `buildAndPush testnet0 libp2p`:
```
Hint: tables_shims [Processing]
beacon_node_types.nim(32, 12) Error: ambiguous identifier: 'KeyPair' --use crypto.KeyPair or keys.KeyPair
```
Not sure what that is about.

zah:
The `libp2p` builds are also work-in-progress. Let's stick to `rlpx` builds for now.

jakubgs:
Okay, will comment them out for now.

docker/beacon_node/Dockerfile
```diff
@@ -1,26 +1,33 @@
...
-      -d:SLOTS_PER_EPOCH=$SLOTS_PER_EPOCH \
+      -d:with${NETWORK_BACKEND} \
+      -d:SHARD_COUNT=${SHARD_COUNT} \
+      -d:SLOTS_PER_EPOCH=${SLOTS_PER_EPOCH} \
       c vendor/nim-beacon-chain/beacon_chain/beacon_node.nim
 
-ENTRYPOINT ["beacon_node"]
+FROM statusteam/nim-base
+
+COPY --from=build /usr/bin/beacon_node /usr/bin/beacon_node
```
zah: Using this as an optimisation is pretty nice, but if I understand correctly, it won't preserve the first layer created with `apt-get install -y build-essential make librocksdb-dev`. It's important for us to package rocksdb as a run-time dependency in the container as well.

jakubgs:
Btw @zah , can you explain to me this:
> * 1 role application with testnet1, rlpx, 5 nodes, port 9100

What is the point of running 5 containers with the same configuration on the one host?
Also, they cannot all run on the same port `9100`.

docker/beacon_node/Dockerfile
jakubgs: Oh, that's good to know, then I'll re-add it there.

zah:
> What is the point of running 5 containers with the same configuration on the one host?
Also, they cannot all run on the same port 9100

The role is written in a way, such that each container is using a consecutive port starting from 9100 (i.e. 9100, 9101, 9102, and so on).

Also, each container uses a different directory from the host for its run-time data. So far, my plan is to manipulate this data directory with other scripts that I'll execute over SSH. I need to be able to perform quick experiments that modify the run-time data of the containers and restart them, so I can see how the network will behave.

jakubgs:
>Also, each container uses a different directory from the host for its run-time data. So far, my plan is to manipulate this data directory with other scripts that I'll execute over SSH. I need to be able to perform quick experiments that modify the run-time data of the containers and restart them, so I can see how the network will behave.

I see, that makes sense. Thanks for the explanation.

docker/beacon_node/Dockerfile
```diff
@@ -1,26 +1,38 @@
...
-      -d:with$network_backend \
-      -d:SHARD_COUNT=$SHARD_COUNT \
-      -d:SLOTS_PER_EPOCH=$SLOTS_PER_EPOCH \
+      -d:with${NETWORK_BACKEND} \
+      -d:SHARD_COUNT=${SHARD_COUNT} \
+      -d:SLOTS_PER_EPOCH=${SLOTS_PER_EPOCH} \
       c vendor/nim-beacon-chain/beacon_chain/beacon_node.nim
 
-ENTRYPOINT ["beacon_node"]
+FROM statusteam/nim-base
```
zah: We can go further and start with a smaller base image as well. Nim is not needed as a run-time dependency. I'm not sure what would be the most appropriate base image, but the dependencies are pretty minor:

```
$ objdump -p build/beacon_node | grep NEEDED
  NEEDED               librt.so.1
  NEEDED               librocksdb.so.5.8
  NEEDED               libc.so.6
```

docker/build_beacon_node.sh
```diff
@@ -1,22 +1,20 @@
 #!/bin/bash
 
+GIT_ROOT=$(git rev-parse --show-toplevel)
```
zah: Wouldn't it be simpler to use something like:

```
cd $(dirname "$0")
```

jakubgs: I guess.

docker/beacon_node/Dockerfile
jakubgs: I tried using `alpine:3.6`:
```
FROM alpine:3.6

RUN echo '@testing http://dl-cdn.alpinelinux.org/alpine/edge/testing' >> /etc/apk/repositories \
 && apk add --no-cache rocksdb-dev@testing

COPY --from=build /usr/bin/beacon_node /usr/bin/beacon_node 

MAINTAINER Zahary Karadjov <zahary@status.im>
LABEL description="Nimbus installation that can act as an ETH2 network bootstrap node."

ENTRYPOINT ["/usr/bin/beacon_node"]
```
But not sure if it works, it doesn't crash at least:
```
 $ d run --rm -it statusteam/nimbus_beacon_node:latest
Error: unhandled exception: Failed to download URL: https://serenity-testnets.status.im/testnet0/rlpx-network.json [IOError]
```
But the image is tiny:
```
 $ d images | grep testnet0-rlpx        
statusteam/nimbus_beacon_node   testnet0-rlpx       63cc86cf5af2        2 minutes ago       16.9MB
```

jakubgs: Are you fine with `rocksdb-dev` in `5.15.10-r4`, or do you need some specific version?
https://pkgs.alpinelinux.org/package/edge/testing/x86/rocksdb-dev
Or would you prefer Ubuntu?

zah: @stefantalpalaru, could you help with selecting the base image here?

stefantalpalaru: I would have said "Clear Linux" for their insane optimisations, but they forgot to update RocksDB since 5.4.6. Pick whatever you're familiar with.

By the way, what's the procedure for Docker base image upgrade?

jakubgs: I tested the image on a server and I got back:
```
standard_init_linux.go:207: exec user process caused "no such file or directory"
```
I've made another image using `debian:9-slim` as a base and it works better:
```
Error: unhandled exception: Failed to download URL: https://serenity-testnets.status.im/testnet0/rlpx-network.json [IOError]
```
Which is an improvement I guess...

jakubgs: Though I don't get why I'm getting an error downloading the JSON since it is available:
```
 $ curl -s https://serenity-testnets.status.im/testnet0/rlpx-network.json 
{
  "networkId": 1000000,
  "genesisRoot": "12C3679F608C24897D87282CF38D179D2173BB23713F97112ED4FD8C2C98EF8F",
  "bootstrapNodes": [
    "enode://e7b5e57da97b753bcb8a3dbb46f66e7b8ef53bcc700451a5835406cbe758be6377f02e9bffbc65f870842f5cbac6ada45d41c610341ee0bcb122f77601f2d806@188.166.38.238:9631"
  ],
  "numShards": 8,
  "slotDuration": 6,
  "slotsPerEpoch": 8,
  "totalValidators": 400,
  "firstUserValidator": 340
}
```

jakubgs: I settled on using `debian:9-slim` as base image, gives us ~75 MB images.

zah: These image sizes are fine and we don't need to spend more time on this, but have you explored the idea of using Nix to create very minimal images? (this could benefit the core app as well)

http://lethalman.blogspot.com/2016/04/cheap-docker-images-with-nix_15.html

jakubgs: Yeah, I've heard of people doing that, but I've never done it myself. We can try that out later. Right now the issue is the JSON download error. Can you look into that? The site is up and available as I showed above.

zah: I have an idea what might be wrong. I'll run some tests locally

zah:
Everything seems to be working fine after the latest changes, so I'm merging this.

jakubgs:
Nice, thanks!

PR #285, opened 2019-03-22
WIP push all containers to a single repo / use docker tags

zah:


jakubgs:
This is a really bad antipattern, making different images just to have different configurations for a service. Could those not be included in the binary as a command line option? Or is that much too much work compared to the benefits?

jakubgs:
I've incorporated your changes into https://github.com/status-im/nimbus/pull/284.

zah:
Yes, in this particular case, we have technical reasons to do it this way. The beacon node has many run-time settings, but the ones here have to be selected at compile-time.

PR #286, opened 2019-03-23
[WIP] Initial implementation of Whisper RPC

kdeme:
Alright, first implementation of Whisper RPC. Next to the available test this has been also somewhat tested manually with curl. Work in progress...

TODO:

- [x] Need to get https://github.com/status-im/nim-eth/pull/23 accepted ;)
- [x] General review of return values (for sure there are some differences with Spec with regards to e.g. bool false versus error returns. Might not always follow spec though, as it is not so "finished" as it should be..
- [ ] Implement shh_subscribe, but I think I need IPC or WS transport here. Not sure what we have there already.
- [x] Implement `shh_generateSymKeyFromPassword`
- [x] Smaller TODOs mentioned in code
- [x] Fix tests in test_rpc_whisper (issue with `shh_getFilterMessages`) and perhaps add more tests
- [x] Perhaps some more cleanup such as the string manipulations

Eventually I want to have also tests with several fully running nodes and whisper RPC interaction/testing via web3 or others, but I don't see that in scope of this PR.

### DIFF
--- a/nimbus/rpc/whisper.nim
+++ b/nimbus/rpc/whisper.nim
@@ -15,6 +15,8 @@ type
     asymKeys*: Table[string, KeyPair]
     symKeys*: Table[string, SymKey]
 
+  KeyGenerationError = object of CatchableError
+
 proc newWhisperKeys*(): WhisperKeys =
   new(result)
   result.asymKeys = initTable[string, KeyPair]()
@@ -52,8 +54,13 @@ proc setupWhisperRPC*(node: EthereumNode, keys: WhisperKeys, rpcsrv: RpcServer) 
     ## pow: The new PoW requirement.
     ##
     ## Returns true on success and an error on failure.
-    # TODO: is asyncCheck here OK?
-    asyncCheck node.setPowRequirement(pow)
+    # Note: If any of the `peer.powRequirement` calls fails, we do not care and
+    # don't see this as an error. Could move this to `setPowRequirement` if
+    # this is the general behaviour we want.
+    try:
+      waitFor node.setPowRequirement(pow)
+    except:
+      trace "setPowRequirement error occured"
     result = true
 
   # TODO: change string in to ENodeStr with extra checks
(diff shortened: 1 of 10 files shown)


kdeme:
This is ready to be reviewed.

I leave subscriptions out of this PR as it will require some investigation on how to implement with json-rpc.

The `WhisperKeys` part could still be abstracted to another file, but it is currently so basic that I didn't do so. However, I do not see it as part of the `whisper_protocol` thus do not want to add it there.

nimbus/rpc/whisper.nim
```diff
@@ -1,74 +1,356 @@
...
-    ## Returns the address of the new identiy.
-    discard
+    ## pow: The new PoW requirement.
+    ##
+    ## Returns true on success and an error on failure.
+    # Note: If any of the `peer.powRequirement` calls fails, we do not care and
+    # don't see this as an error. Could move this to `setPowRequirement` if
+    # this is the general behaviour we want.
+    try:
+      waitFor node.setPowRequirement(pow)
```
zah: The `rpc` macro makes the handler procs async, so you can use `await` here.

arnetheduck: the `rpc` macro would arguably be easier to use if it didn't hide this transformation, but rather required an annotation on the proc - you wouldn't have to read code or docs to find out, it would give an error

kdeme: I was initially not sure but suspected it to be `async`, and then confirmed with `-d:nimDumpAsync` as documentation also indicates (which is then also one of the reasons why I did not want to use `poll()` in minePow to make it unblock, but that is another story).

That being said, I was then surprised that `await` would actually not work, and thus opted for`waitFor`.
With `await` i got `Error: undeclared identifier: 'await'`
With `yield` i got `Error: 'yield' only allowed in an iterator`
Typically what you get in non async?

kdeme: The RPC proc actually gets wrapped with an `async` proc so it looks normal that this doesn't work:
https://github.com/status-im/nim-json-rpc/blob/master/json_rpc/router.nim#L232

PR #288, opened 2019-03-25
Nimbus: runtime log level selection and logfile option

stefantalpalaru:
(and some drive-by fixes and improvements in some argument-related code)

### DIFF
--- a/nimbus.nimble
+++ b/nimbus.nimble
@@ -36,5 +36,5 @@ task test, "Run tests":
   test "test_rpc"
 
 task nimbus, "Build Nimbus":
-  buildBinary "nimbus", "nimbus/"
+  buildBinary "nimbus", "nimbus/", "-d:chronicles_log_level=TRACE"
 
--- a/nimbus/nim.cfg
+++ b/nimbus/nim.cfg
@@ -1,5 +1,5 @@
 -d:chronicles_line_numbers
--d:"chronicles_sinks=textblocks[file]"
+-d:"chronicles_sinks=textlines[file]"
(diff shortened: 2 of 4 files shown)


nimbus/nim.cfg
```diff
@@ -1,5 +1,6 @@
 -d:chronicles_line_numbers
--d:"chronicles_sinks=textblocks"
+-d:"chronicles_sinks=textblocks[file]"
```
zah: I like the `textblocks` format, but the team decided that we'll be using `textlines` by default.

nimbus.nimble
```diff
@@ -36,5 +36,5 @@ task test, "Run tests":
   test "test_rpc"
 
 task nimbus, "Build Nimbus":
-  buildBinary "nimbus", "nimbus/"
+  buildBinary "nimbus", "nimbus/", "-d:chronicles_log_level=TRACE"
```
zah: The idea we agreed upon was that `trace` can be used so often that it creates performance issues (even with filtering enabled). Most of the time it should be selectively activated it in a particular topic/module you are working on. Perhaps we don't have enough tracing to be a problem yet, but `DEBUG` might be a better default for a release build.

stefantalpalaru: It's not a performance issue right now. We can revisit this when it becomes a measurable problem or when we produce a release build and decide on the compile time flags for it.

nimbus/nim.cfg
stefantalpalaru: Changed.

PR #298, opened 2019-03-28
[WIP] Nimbus CI: add nim-beacon-chain tests

stefantalpalaru:
We need to make sure that bumping Git submodules doesn't break nim-beacon-chain here, since this is the recommended setup for using that repo.


zah:
I think all of the sub-projects should be tested here. At least until we have an alternative solution for pinning inside the sub-repos.

stefantalpalaru:
I hear you, but it would take too long for a single CI iteration. Testing the two projects that end up using all those other dependencies will have to do as a sanity check.

Issue #244, opened 2019-02-25
every EVM opcode has potential to OOG and ruin tracing result


jangko:
current implementation:
If an opcode trigger OOG, it will raise exception, there is no chance for the tracer to record the gas consumption.

The problem become worse if the computation have more than one depth. There will be many opcode with blank `gasCost` field in the tracing result.

Besides making EVM debugging more difficult, the incorrect tracing result also does not conform to the `standard` tracing result.

example block: 420301

PR #301, opened 2019-04-02
[WIP] EVM transformation

jangko:


jangko:
this PR is ready. review please.
summary:
* GST +57, only 2 left from Frontier and Homestead test suite.
* discover an important fact related to long standing stack overflow issue: `{.computedGoto.}` pragma can eat a lot of stack space. I will transform the EVM into jump table based in another PR.
* we are able to sync pass DAO, I will tell the exact number later, right now I still syncing.

nimbus/vm/computation.nim
```diff
@@ -142,65 +153,68 @@ proc writeContract*(computation: var BaseComputation, fork: Fork): bool =
     if fork < FkHomestead: computation.output = @[]
     result = false
 
-proc transferBalance(computation: var BaseComputation, opCode: static[Op]): bool =
+proc transferBalance(computation: BaseComputation, opCode: static[Op]) =
   if computation.msg.depth > MaxCallDepth:
-    debug "Stack depth limit reached", depth=computation.msg.depth
-    return false
+    computation.setError(&"Stack depth limit reached depth={computation.msg.depth}")
```
stefantalpalaru: Why is it necessary to copy the string and where are these errors logged through chronicles?

jangko:
The prefix '&' is not a concat operator, it is a strfmt string interpolation shortcut.
The error msg should be appear at transaction tracer, we can easily log it with chronicles at the end of 'executeOpcodes'.
I will add the logging code later when I fix #244

nimbus/vm/computation.nim
```diff
@@ -142,65 +153,68 @@ proc writeContract*(computation: var BaseComputation, fork: Fork): bool =
...
 
-  result = true
+template continuation*(comp: BaseComputation, body: untyped) =
+  # this is a helper template to implement continuation
+  # passing and convert all recursion into tail call
+  var tmpNext = comp.nextProc
+  comp.nextProc = proc() =
+    body
+    tmpNext()
+
```
stefantalpalaru: There's no continuation being passed here, is it? Just chaining blocks of code.
By the way, is it intentional that the last added block is executed first?

jangko: > There's no continuation being passed here, is it? Just chaining blocks of code.

traditional cps passing the continuation via proc param, but it doesn't have to. you can pass it via closure too.

> By the way, is it intentional that the last added block is executed first?

yes, it emulates stack behaviour, last in first out, implicit stack. but still follow the principles of continuation passing.

PR #309, opened 2019-04-12
Updated nim to 0.19.4

yglukhov:


stefantalpalaru:
Now `make update` knows to rebuild the Nim compiler if its submodule was updated.

stefantalpalaru:
With this Nim version, we should be able to get coloured output from Chronicles on Windows.

PR #313, opened 2019-04-15
fixes #244

### LINKED ISSUE #244: every EVM opcode has potential to OOG and ruin tracing result

jangko:


stefantalpalaru:
Test failure: https://travis-ci.org/status-im/nimbus/jobs/520119429#L1963

PR #333, opened 2019-06-05
 Forward maxPeers config to EthereumNode minPeers 

kdeme:
`maxPeers` config is currently not used and thus we would always be stuck on 10 peers, this changes that.
The change is not 100% correct, as it is basically "minPeers before stopping to connect". The listener still stays active for incoming connections. (And if you provide enough `--staticnodes` you could also go over the max).
Could change this in nim-eth though. Or could alter the meaning of `maxPeers` here in the documentation.

### DIFF
--- a/nimbus/nimbus.nim
+++ b/nimbus/nimbus.nim
@@ -94,7 +94,8 @@ proc start(): NimbusObject =
 
   nimbus.ethNode = newEthereumNode(keypair, address, conf.net.networkId,
                                    nil, nimbusClientId,
-                                   addAllCapabilities = false)
+                                   addAllCapabilities = false,
+                                   minPeers = conf.net.maxPeers)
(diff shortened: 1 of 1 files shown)


stefantalpalaru:
How about renaming the configuration option to "minPeers"?

kdeme:
Yes, that's an option. But a bit of a weird setting imo, might be confusing also.
I'd rather leave it like this and if deemed necessary, we can adjust the behaviour it in nim-eth.

stefantalpalaru:
I looked at the code and it really is used as a minimum number of peers to connect to. Why call it "max" on the command line?

kdeme:
Yes, that's true.

The maxPeers was probably just taken over from geth. Typically, if max peers are reached this information is send in a disconnect message to the peer.

Issue #338, opened 2019-06-19
eth_blockNumber is base10 but outputs as hex


Swader:
From file https://github.com/status-im/nim-stint/blob/6853ebe97c21426952ce64c2df6278514797dc3f/stint/io.nim#L183

RPC call:

```bash
curl --header "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}' 127.0.0.1:8545
{"jsonrpc":"2.0","id":1,"result":"0x731712","error":null}
```

kdeme:
Fixed now

PR #339, opened 2019-06-19
Fix uint256 rpc overload to show correct hex value. #338

kdeme:


kdeme:
Too quick on the merge, looks like it breaks rpc tests. :eyes: 

I'll have a look

Swader:
This is for #338?

PR #342, opened 2019-06-21
drop name of network from docker image tag

jakubgs:
We no longer support different flavors of images depending on protocol used.


zah:
Thanks @jakubgs. This was already done in the libp2p-testnet branch here:
https://github.com/status-im/nimbus/blob/libp2p-testnet/docker/beacon_node/Makefile

The branch will be merged soon.

PR #344, opened 2019-06-26
periodically log internal statistics

stefantalpalaru:
and bump submodules


arnetheduck:
er thought we decided to drop this approach in favour of something more generic? looks like quite the hack to log things periodically..

kdeme:
I accepted this, while I agree that we need another approach, as for now it is still better than nothing.
But if you do not want to risk this growing without the other approach being implemented, I don't mind that it doesn't get added.

If it is just about logging this specific info, you could just add a simple log in the nim-eth peer_pool run loop.

PR #346, opened 2019-07-08
Add skip parameter in getSuccessorHeader + add getAncestorHeader

kdeme:
Goes with https://github.com/status-im/nim-eth/pull/78

### DIFF
--- a/nimbus/p2p/chain.nim
+++ b/nimbus/p2p/chain.nim
@@ -23,8 +23,14 @@ method getBlockHeader*(c: Chain, b: HashOrNum, output: var BlockHeader): bool {.
 method getBestBlockHeader*(c: Chain): BlockHeader {.gcsafe.} =
   c.db.getCanonicalHead()
 
-method getSuccessorHeader*(c: Chain, h: BlockHeader, output: var BlockHeader): bool {.gcsafe.} =
-  let n = h.blockNumber + 1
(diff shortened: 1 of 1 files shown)


zah:
What happens when the requesting party asks for a range that's too wide (e.g. it can send you in the negative block numbers)? The database look-up will return false and everything will be fine?

kdeme:
Yes, I've checked for that and it wraps around the number and just doesn't find that block. Looked acceptable to me for not having to do a check each time there.

I added a simple check in the nim-eth code, but it is not exhaustive as it only checks the negative bounds. Well, actually also only for the first "next value" now that I look at it again, oops.
Hmmm, could add some better checking there so the look-ups are not actually performed in every wrong case, or do no check at all.. :eyes:

kdeme:
Alright, added bounds check not to depend on the current behaviour that uint or stuint wraps around.
Could also do these checks in nim-eth `getBlockHeaders` but that makes it a bit more ugly/complex (however probably more performant if done before the while loop there, taking maxResults into account).

I understood from @zah that perhaps the idea of https://github.com/status-im/nim-eth/blob/master/eth/common/eth_types.nim#L174 is to actually migrate to using int64. If that is the case, I can do the changes in another PR to make both `int64` and `UInt256` an option, although then I would probably rather opt then for `uint64`

These added bound checks are currently not good for int64, as the `not 0.toBlockNumber` will not be OK. Would need `high(uint64)` there I guess.

PR #347, opened 2019-07-10
add nim-metrics

stefantalpalaru:


nimbus/vm/interpreter/opcodes_impl.nim
```diff
@@ -596,14 +596,14 @@ proc setupCreate(computation: BaseComputation, memPos, len: int, value: Uint256,
     some(computation.getFork))
 
 template genCreate(callName: untyped, opCode: Op): untyped =
-  op callName, inline = false, value, startPosition, size:
+  op callName, inline = false, val, startPosition, size:
```
kdeme: Are these changes because there was some conflict when compiling?

stefantalpalaru: Yes. The `metrics` module is now exported by `eth/common` and, since `metrics.value` is a template, the corresponding NimNode doesn't have an `ident` field, breaking the `op` macro.

zah: Hrm, `genCreate` is really a `dirty` template and it could have been marked as such. Then the above wouldn't be a problem AFAICT. But this is just sharing tips, no need to rework this.

PR #348, opened 2019-07-10
Use of BlockNumber to be able to switch between UInt256 and int64

kdeme:
Not sure if this is really needed but considering some sort of effort was once started here: https://github.com/status-im/nim-eth/blob/master/eth/common/eth_types.nim#L171
I decided to do an attempt to complete it. Will require https://github.com/status-im/nim-eth/pull/81

Should definitely be reviewed by someone that knows/created the code as I'm not sure this was the intention and is even OK to begin with.
Anyhow, as `BlockNumber` is currently still set to `UInt256` in https://github.com/status-im/nim-eth/pull/81, these changes should not actually do anything different.

### DIFF
--- a/nimbus/config.nim
+++ b/nimbus/config.nim
@@ -206,24 +206,24 @@ proc publicChainConfig*(id: PublicNetwork): ChainConfig =
   of RopstenNet:
     ChainConfig(
       chainId:        RopstenNet.uint,
-      homesteadBlock: 0.u256,
+      homesteadBlock: 0.toBlockNumber,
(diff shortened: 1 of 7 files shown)


zah:
These are fine changes and if they don't produce any issues when you test Nimbus, please go ahead and merge them.

kdeme:
Well, tested both as UInt256 and int64 and did not notice anything out of the ordinary. 
So I'll merge https://github.com/status-im/nim-eth/pull/81 and bump nim-eth + merge

PR #352, opened 2019-07-11
opt-in metrics HTTP server and periodic "stats" logging

stefantalpalaru:
...and example "prometheus.yml" file to use this server as a Prometheus scraping target.

(copy it in an empty dir and run `prometheus` from there, start a Nimbus sync with something like `./build/nimbus --prune:archive --log-level:info --log-file:output.log`, then go to http://localhost:9090/graph?g0.range_input=15m&g0.stacked=1&g0.expr=connected_peers&g0.tab=0)

Example graph:

### DIFF
--- a/nimbus/config.nim
+++ b/nimbus/config.nim
@@ -497,9 +497,9 @@ proc processNetArguments(key, value: string): ConfigStatus =
     result = processInteger(value, res)
     if result == Success:
       config.net.discPort = uint16(res and 0xFFFF)
-  elif skey == "metricsserver":
+  elif skey == "metricsserver" and defined(insecure):
     config.net.metricsServer = true
-  elif skey == "metricsserverport":
+  elif skey == "metricsserverport" and defined(insecure):
     var res = 0
     result = processInteger(value, res)
     if result == Success:
@@ -602,6 +602,8 @@ proc processDebugArguments(key, value: string): ConfigStatus =
     result = processInteger(value, res)
     if result == Success:
       config.debug.logMetricsInterval = res
+  else:
+    result = EmptyOption
 
 proc dumpConfiguration*(): string =
   ## Dumps current configuration as string
@@ -680,6 +682,14 @@ proc getHelpString*(): string =
       continue
     logLevels.add($level)
 
+  when defined(insecure):
+    let metricsServerHelp = """
+
+  --metricsServer         Enable the metrics HTTP server
+  --metricsServerPort:<value> Metrics HTTP server port on localhost (default: 9093)"""
+  else:
+    let metricsServerHelp = ""
+
   result = """
 
 USAGE:
@@ -696,9 +706,7 @@ NETWORKING OPTIONS:
   --bootnodesv5:<value>   Comma separated enode URLs for P2P v5 discovery bootstrap (light server, light nodes)
   --staticnodes:<value>   Comma separated enode URLs to connect with
   --port:<value>          Network listening TCP port (default: 30303)
-  --discport:<value>      Network listening UDP port (defaults to --port argument)
-  --metricsServer         Enable the metrics HTTP server
-  --metricsServerPort:<value> Metrics HTTP server port on localhost (default: 9093)
+  --discport:<value>      Network listening UDP port (defaults to --port argument)$7
   --maxpeers:<value>      Maximum number of network peers (default: 25)
   --maxpendpeers:<value>  Maximum number of pending connection attempts (default: 0)
   --nat:<value>           NAT port mapping mechanism (any|none|upnp|pmp|<external IP>) (default: "any")
@@ -737,7 +745,8 @@ LOGGING AND DEBUGGING OPTIONS:
     $defaultLogLevel,
     strip($defaultProtocols, chars = {'{','}'}),
     $defaultMaxMsgSize,
-    $defaultMinPow
+    $defaultMinPow,
+    metricsServerHelp,
   ]
 
 proc processArguments*(msg: var string): ConfigStatus =
@@ -783,6 +792,7 @@ proc processArguments*(msg: var string): ConfigStatus =
           processArgument processDebugArguments, key, value, msg
           if result != Success:
             msg = "Unknown option: '" & key & "'."
+            break
     of cmdEnd:
       doAssert(false) # we're never getting this kind here
 
--- a/nimbus/config.nim
+++ b/nimbus/config.nim
@@ -135,7 +135,8 @@ type
     staticNodes*: seq[ENode]      ## List of static nodes to connect to
     bindPort*: uint16             ## Main TCP bind port
     discPort*: uint16             ## Discovery UDP bind port
-    metricsPort*: uint16          ## metrics HTTP server port
+    metricsServer*: bool           ## Enable metrics server
+    metricsServerPort*: uint16    ## metrics HTTP server port
     maxPeers*: int                ## Maximum allowed number of peers
     maxPendingPeers*: int         ## Maximum allowed pending peers
     networkId*: uint              ## Network ID as integer
@@ -150,6 +151,8 @@ type
     flags*: set[DebugFlags]       ## Debug flags
     logLevel*: LogLevel           ## Log level
     logFile*: string              ## Log file
+    logMetrics*: bool             ## Enable metrics logging
+    logMetricsInterval*: int      ## Metrics logging interval
 
   PruneMode* {.pure.} = enum
     Full
@@ -494,11 +497,13 @@ proc processNetArguments(key, value: string): ConfigStatus =
     result = processInteger(value, res)
     if result == Success:
       config.net.discPort = uint16(res and 0xFFFF)
-  elif skey == "metricsport":
+  elif skey == "metricsserver":
+    config.net.metricsServer = true
+  elif skey == "metricsserverport":
     var res = 0
     result = processInteger(value, res)
     if result == Success:
-      config.net.metricsPort = uint16(res and 0xFFFF)
+      config.net.metricsServerPort = uint16(res and 0xFFFF)
   elif skey == "maxpeers":
     var res = 0
     result = processInteger(value, res)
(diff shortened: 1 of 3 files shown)


nimbus/config.nim
```diff
@@ -675,6 +682,7 @@ NETWORKING OPTIONS:
   --staticnodes:<value>   Comma separated enode URLs to connect with
   --port:<value>          Network listening TCP port (default: 30303)
   --discport:<value>      Network listening UDP port (defaults to --port argument)
+  --metricsport:<value>   Metrics HTTP server port on localhost (defaults to 9093, set to 0 to disable)
```
kdeme: It's a bit weird to turn it on/off by setting the port. Also, should we not default turn it off?

nimbus/nimbus.nim
```diff
@@ -136,14 +136,11 @@ proc start() =
       result = "EXITING"
     nimbus.rpcServer.start()
 
-  # periodically log internal statistics
```
kdeme: Not sure here, but perhaps we can still keep a log with some very basic information?

nimbus/config.nim
Swader: +1 it should default to off and value should have no impact on toggle

stefantalpalaru: You want a new option for starting the metrics server?

kdeme: Yeah, I think a `--metrics` to enable (default off) and `--metricsport` to set the port is more clear.

stefantalpalaru: Done.

nimbus/nimbus.nim
stefantalpalaru: Added. Also opt-in.

cheatfate:
Sorry, but metrics is using `asyndispatch/asynchttpserver`. And because both are unmaintained and not reliable i don't see any reasons to accept this PR.
Also i want to show this quote:
https://github.com/nim-lang/Nim/blob/devel/lib/pure/asynchttpserver.nim#L12-L15

stefantalpalaru:
Good thing it's only used locally, then.

Since there is no alternative HTTP server, I see no reason to reject this PR.

cheatfate:
1. Because there no alternative doesn't mean that you need to use a toy server in production software. I don't see any warnings or cautions in the README about usage of toy http server.
2. There was alternative method to gather metrics and you know it, but you have made choice to use toy http server.
3. Also because this toy http server using asyncdispatch, by accepting this PR we are adding this layer to our project too.

stefantalpalaru:
> Because there no alternative doesn't mean that you need to use a toy server in production software.

Unlike this production-ready version of it? https://github.com/status-im/nim-json-rpc/blob/master/json_rpc/servers/httpserver.nim

cheatfate:
Its HTTP/1.0 server and its support timeouts and limits.

stefantalpalaru:
> Its HTTP/1.0 server and its support timeouts and limits. So yeah it is much more safer to use it in production.

Tell me more about the safety of an unprotected HTTP server executing all sorts of expensive, sensitive and Whisper-impersonating RPC calls.

It's a toy server and it needs to be completely removed, or it's the same level of security as the one from this PR and it's enough to have it disabled by default.

cheatfate:
The server embedded to `nim-json-rpc` is not a target of this PR, but anyway.

1. It uses much more sane and strict http request/response parser.
2. It has timeouts on headers and body receive operation.
3. It has size limits on both headers and body receive operation.
4. It is properly supports only HTTP/1.0 protocol and do not have proper support of HTTP/1.1.
5. It is based on `chronos`.

While asynchttpserver:

1. Has simple request parser, which only depends on CRLF.
2. Do not have timeouts at all.
3. It has size limits on headers receive operation.
4. It tends to support not only HTTP/1.0 but also HTTP/1.1.
5. Its networking based on `asyncdispatch`.

stefantalpalaru:
> The server embedded to nim-json-rpc is not a target of this PR

Yet it's a very similar security vulnerability to this one (even more so, I'd argue, because it offers sensitive account information, plenty of DoS opportunities and the ability to impersonate the node over Whisper, all on 127.0.0.1 without any authentication).

Timeouts and size limits are great, but they don't change the nature of the beast. So, if such a vulnerable piece of code is allowed in Nimbus because it's disabled by default, what's the argument against adding a similar one under the same conditions?

If it's some sort of "chronos is better than asyncdispatch" thing, it's invalid, because it has no bearing on the most important security aspects of running HTTP servers locally.

cheatfate:
On HTTP server level it do not have DoS opportunities, if its possible to achieve DoS on application level but its not a problem of `nim-json-rpc` HTTP server.

> Timeouts and size limits are great, but they don't change the nature of the beast. So, if such a vulnerable piece of code is allowed in Nimbus because it's disabled by default, what's the argument against adding a similar one under the same conditions?

And common stop repeating that something vulnerable without any vulnerabilities disclosed. Prove your words with technical details how to break server which is embedded to `nim-json-rpc`. And we not talking about application level, we talking exactly about HTTP server.

> If it's some sort of "chronos is better than asyncdispatch" thing, it's invalid, because it has no bearing on the most important security aspects of running HTTP servers locally.

We made choice to start `chronos` not because it is better, but because we can fix all the bugs much more easily and faster and not wait till somebody will accept changes to asyncdispatch.

stefantalpalaru:
> On HTTP server level it do not have DoS opportunities

Of course it has. You can trigger expensive hash calculations in a single-threaded application. You're not only hogging the CPU, but you can significantly delay any other I/O operation - all that thanks to the RPC HTTP server.

> We made choice to start chronos not because it is better, but because we can fix all the bugs much more easily and faster and not wait till somebody will accept changes to asyncdispatch.

Is that why there's no generic HTTP server in it for me to use instead of the one from the standard library?

cheatfate:
You manipulating, there no DoS opportunities on HTTP server level. `nim-json-rpc` http server can do nothing with your application level...

```nim
import json_rpc/rpcserver

when isMainModule:
  while true:
    echo "endless cycle"
```

Here is DoS, but can you blame the `nim-json-rpc` server for this? Because looks like you trying...

stefantalpalaru:
> there no DoS opportunities on HTTP server level. nim-json-rpc http server can do nothing with your application level...

Here are all the wonderful RPC commands that little HTTP server can execute after a simple GET request:

https://github.com/status-im/nimbus/blob/master/nimbus/rpc/p2p.nim

https://github.com/status-im/nimbus/blob/master/nimbus/rpc/whisper.nim

cheatfate:
You started to blame HTTP server which was embedded inside of `nim-json-rpc`, when i proved that you are wrong you started to blame whole `nim-json-rpc`. But `json-rpc` is **offtopic** for both topic and discussion. 

We are talking here exactly and only about HTTP server functionality which you trying to push to Nimbus. We are not talking here about `json-rpc` or `metrics` functionality.

stefantalpalaru:
> But json-rpc is offtopic for both topic and discussion.

The topic is Nimbus' security profile, not personal grudges against contributors to Nim's standard library.

Do you honestly think that this PR changes in any way, shape or form Nimbus' security profile, as defined by the types of attacks it's vulnerable to?

cheatfate:
When you adding weak httpserver (which is based on not reliable networking framework) you increasing the number of attack vectors (+1), but also you increasing number of potential bugs which can be happen inside of it, and number of already open bugs (there no timeout handling).

stefantalpalaru:
> When you adding weak httpserver (which is based on not reliable networking framework) you increasing the number of attack vectors (+1)

That's not how it works. No new vulnerabilities are added.

> there no timeout handling

Not much of a problem, since it runs in its own thread, unlike the RPC HTTP server.

cheatfate:
This is exactly how it works.
You can search definition for term **attack vector** (i have not talked about vulnerabilities) and when you adding network server to your software you increasing number of **attack vectors**.

stefantalpalaru:
> when you adding network server to your software you increasing number of attack vectors.

You probably mean attack surfaces, not attack vectors, but it's still wrong. It doesn't matter if you start the RPC HTTP server or the metrics HTTP server, your attack surface still includes an HTTP server.

You can add 10 more HTTP servers and have the same attack surface, if those are all vulnerable to the same vectors.

cheatfate:
Please https://en.wikipedia.org/wiki/Attack_surface.
When you have one HTTP server software which is already embedded and working you have 1 attack vector, when you adding one more HTTP server software (which is different from previous one) to implement metrics you adding 1 more attack vector.

But if you have 10 equal HTTP servers on different ports you still have 1 attack vector.

zah:
Clearly, the long-term solution here is the development of a HTTP server that we'll be happy to ship to our end-users and that will cover both JSON-RPC and Prometheus as end-points.

Until we have such a server, I think it's reasonable for us to take advantage of Prometheus as a development tool, as long as we take the necessary steps to not expose the insecure server on a public host within our testnet.

cheatfate:
@zah please consider, that Prometheus can be supported without implementation and/or usage of HTTP server, by using PUSH model.

arnetheduck:
Alright, so I suggest we do this for now: 

enable the code to be compiled in with a special compile flag only: -d:insecure - this allows us to develop the feature conveniently and incrementally and thus not blocking anything, while not exposing the code in any builds that users "normally" create. 

From a decision point of view, this means we have to make an active decision down the road about http servers - by that time we should have more attractive options available - for example that nim-metrics will also support statsd or chronos will be up to speed (it's still an open question if we should run a separate thread/event loop for this server even if we have chronos support - that would allow it to stay more insulated from the rest of the application which is probably good, with only the metrics serving as touch points between the two layers.

stefantalpalaru:
HTTP metrics server now under `-d:insecure`. You'll have to `make NIMFLAGS="-d:insecure" nimbus` to get it.

zah:
@cheatfate, even though sticking to the push-model is technically possible, the market might demand the added convenience of having a stand-alone solution for Prometheus monitoring. For example, Geth 1.9 just added such a built-in server and other clients are likely to copy it:

https://blog.ethereum.org/2019/07/10/geth-v1-9-0/#metrics-collection

PR #354, opened 2019-07-15
Add txRoot check to avoid assertions in tracer + don't throw exception

kdeme:
I would run into this assertion when running Nimbus in debug mode:
```
Error: unhandled exception: /home/deme/repos/nimbus/nimbus/tracer.nim(202, 11) `calcRootHash(body.transactions) == header.txRoot`
```
This already gets reported before in `processBlock` : see also log statement:
```
DBG 2019-07-12 11:02:01+02:00 Mismatched txRoot tid=2989 file=p2p/executor.nim:122 blockNumber=1353812
```

So I added a check before dumping the debug information, to avoid the assertions. Because I don't think we need to raise (anything) and also as I didn't know for sure if other code (premix) requires/assumes these assertions.

I've also removed the raised exception, as I think it is/was not the intention to raise this all the way up (and potentially crash on it, if not catched in nimbus.nim).

I also checked a bit the history and I noticed that originally we would only raise here:  https://github.com/status-im/nimbus/commit/048a43b2f1cc8835f3556147842618dd8c29f765#diff-04e63af065108e405533c68136d6ef55R137
And then that also got removed: https://github.com/status-im/nimbus/blob/master/nimbus/p2p/executor.nim#L168
That's OK, I guess? As there is the `dumpDebuggingMetaData` now?

### DIFF
--- a/nimbus/p2p/chain.nim
+++ b/nimbus/p2p/chain.nim
@@ -52,9 +52,10 @@ method persistBlocks*(c: Chain, headers: openarray[BlockHeader], bodies: openarr
     let validationResult = processBlock(c.db, headers[i], bodies[i], vmState)
 
     when not defined(release):
-      if validationResult == ValidationResult.Error:
+      if validationResult == ValidationResult.Error and
+          bodies[i].transactions.calcTxRoot == headers[i].txRoot:
         dumpDebuggingMetaData(c.db, headers[i], bodies[i], vmState)
-        raise newException(Exception, "Validation error. Debugging metadata dumped.")
+        warn "Validation error. Debugging metadata dumped."
 
     if validationResult != ValidationResult.OK:
       return validationResult


kdeme:
hmm, you're right.

However, isn't the release code wrong then?

stefantalpalaru:
The most likely reason it fails validation right now is an incomplete/buggy VM, so it makes no sense to retry downloading and processing the block - hence the raised exception stopping everything.

In the off chance we really got a corrupt block over the network, we should probably retry a few more times before bailing.

kdeme:
I was confused there with `defer` for a moment, the storing part looks fine.

But, as mentioned, this error is not related to VM but txRoot trie / hash calculation. And originally it was not being raised on, neither in `processBlock`  nor in `persistBlocks`. 
Now nothing gets raised in `processBlock`, but on any error there is this `dumpDebuggingMetaData` + raise `Exception`, atleast in `debug` mode.

So what I did is:
1. remove the possible assertions in `dumpDebuggingMetaData`
2. remove the general raise in case of debug.

Reasoning is:
a. I don't like the fact that there is different error flow between release and debug (the `dumpDebuggingMetaData` is fine ofcourse)
b. If we want Nimbus to "die" in case of a wrong state `rootHash` (which I think is a good idea for now), we should only do so on that occasion. I believe there are still other possible errors not state related.

PR #357, opened 2019-07-19
Ecrecover index error

kdeme:
When syncing Nimbus there was a segfault occuring (release mode) at block number 1352922.

The problem was invalid access to `data[63]`,  in `getSignature` used by `ecRecover`.
The data provided there does not necessarily have 64 bytes, and can in fact be completely empty.

The reason as to why this happens at block 1352922, is because it is the first TX to 0x0...01 (ecrecover precompile) with empty input data, see:
https://etherscan.io/txs?a=0x0000000000000000000000000000000000000001&p=166
https://etherscan.io/tx/0x4127eb13b273120f81267673dc89a158cae570215eb06e795ab0636ee7f7f015

The first commit  fixes this. The second commit is a rework of the code in hope to make it more clear + do certain actions in a more reasonable order.

Additionally, this bug would only be noticed in `release` mode, without much debug data to work from.
`IndexError` would be raised in `debug` mode, but quickly catched by the catch all. The last commit should make debug mode also crash on this type of errors. This could create some new crashes, I haven't seen any yet but I have not yet tested to sync again from start. 

I did run the general state tests, and have same results, skipping the 2 known failed and slow tests (slow tests seem to not give reliable results, with or without this change)

### DIFF
--- a/nimbus/errors.nim
+++ b/nimbus/errors.nim
@@ -6,7 +6,7 @@
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
 type
-  EVMError* = object of Exception
+  EVMError* = object of CatchableError
     ## Base error class for all evm errors.
 
   BlockNotFound* = object of EVMError
--- a/nimbus/vm/precompiles.nim
+++ b/nimbus/vm/precompiles.nim
@@ -16,34 +16,37 @@ type
     paEcMul,
     paPairing = 8
 
-proc getSignature*(computation: BaseComputation): (array[32, byte], Signature) =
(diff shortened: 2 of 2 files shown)


zah:
Something went wrong for the CI. Closed/reopened to retrigger it.

zah:
The CI fails with:

```
nimbus.nim(143, 19) Error: undeclared identifier: 'peerGauge'
```

Probably something is wrong with the pinned modules.

kdeme:
Yes, I see nim-eth got bumped, but https://github.com/status-im/nimbus/pull/352 isn't merged yet.

PR #360, opened 2019-07-31
refactor the C and Go wrapper build system

stefantalpalaru:
- moved "nimbus/api" to "wrappers"
- renamed files
- replaced the build scripts with Makefile targets
- set the rpath relative to the test binary's location so it can look
  for libnimbus.so there at runtime
- libnimbus.so.0 required on Linux, apparently
- compiled all the Nimbus code with `--app:lib`, not just one file (this
  required skipping a proc in "nimbus/config.nim" because it uses an API
  that's unavailable in libraries)
- removed static linking from the Go wrapper. It doesn't make sense at a
  global level, when using a shared Nimbus library. To selectively link
  static libraries, we should probably be specifying them as "*.a". I did
  build a static "libnimbus.a", as a test, but it insisted on dlopen-ing a
  shared version of itself which looked too ugly to continue.


nimbus/config.nim
```diff
@@ -752,60 +752,61 @@ LOGGING AND DEBUGGING OPTIONS:
...
-          if result != Success:
-            msg = "Unknown option: '" & key & "'."
-            break
-    of cmdEnd:
-      doAssert(false) # we're never getting this kind here
-
-  if config.net.bootNodes.len == 0:
-    # No custom bootnodes were specified on the command line, restore to
-    # previous values
+when declared(os.paramCount): # not available with `--app:lib`
```
kdeme: But I don't think that currently the config is used for anything in the whisper/status wrapper. And I was planning on moving the bootnodes from the config.nim to the nim-eth repo, so it ain't needed for that either then.

Of course this could be useful for other wrappers, but whisper doesn't rely on Nimbus in general.

stefantalpalaru: It doesn't matter. It won't compile with `--app:lib` because `initOptParser()` is not defined for this compilation target: https://github.com/nim-lang/Nim/blob/803406d07ce95189b08a16c7f51e2bf91f9c0c12/lib/pure/parseopt.nim#L196 and https://github.com/nim-lang/Nim/blob/803406d07ce95189b08a16c7f51e2bf91f9c0c12/lib/pure/os.nim#L2491

kdeme: Yeah, I understood that. Sorry, it was not clear as should have put my comment not at this line I guess,
My point is that the config part, and actually nimbus as a whole is not a dependency on the current wrapper. So think this is not really needed at the moment

stefantalpalaru: Think long-term. It would be nice to have most of Nimbus' functionality exported to C from a library. At that point, it's hard to decouple config.nim from it all.

arnetheduck:
as far as naming goes, `wrapper.*` is a bit too generic. something more contextual would be more appropriate..

nim.cfg
```diff
@@ -17,3 +17,7 @@
 --excessiveStackTrace:on
 -d:metrics # enable metric collection
 
+# Required to make up for the compiler's inability to invalidate the C file
+# (and object) cache on different command line arguments.
+--forceBuild
```
zah: This is not acceptable. You'll be slowing down all the daily builds of everyone in every sub project.

Instead, aim to use a configuration-specific nimcache folder. To achieve this, you have to combine two techniques:

1) Conditional statements in the config:

```
@if debug:
   ...
@end
```

2) Variable interpolation in the setting value:

```
-d:"nimcache=root_dir/some_config/$projectName
```

zah:
> I did build a static "libnimbus.a", as a test, but it insisted on dlopen-ing a
shared version of itself which looked too ugly to continue.

I'm curious what happened here. Are you aware of the `--dynlibOverride` option of the compiler. It's described here:

https://nim-lang.org/docs/nimc.html

stefantalpalaru:
> as far as naming goes, `wrapper.*` is a bit too generic.

I was more worried about correctness. "API" is a misnomer here, because the main goal is wrapping parts of Nimbus to expose them to other programming languages.

The technical name for "wrapper.nim" and "wrapper.h" would be a "thin wrapper". I'm open to other names, as long as they are not misleading. Something mentioning "libnimbus" would also be appropriate, because there is no generic library without exporting Nim procedures to C.

nim.cfg
stefantalpalaru: > This is not acceptable.

Acceptable or not, it's what you need in order to alternate between stand-alone binary and shared library targets on the same Nim codebase. The original solution was to stick a "rm -rf nimcache" in the build script. That's yet another ugly band-aid on a small part of the wound.

I've been bitten by the Nim compiler's failure to invalidate the cache on different command line parameters in the past, so I think it's a great time to just give up on caching completely at take the compile time penalty.

> aim to use a configuration-specific nimcache folder

We already do: https://github.com/status-im/nimbus/blob/78714295dab4ead5e56c0048c46ae07d8bbed91a/nim.cfg#L1

Doesn't help, because the relevant compiler flags changing from one compilation to the next are not just `-d:release` and `-d:debug`.

So we either fix the compiler's cache invalidation logic or we stop relying on it. Guess which is easier.

stefantalpalaru:
> I'm curious what happened here.

A bit of confusion. I thought this warning message came from Nim, but it comes from glibc:

```
/usr/lib/gcc/x86_64-pc-linux-gnu/9.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: build/libnimbus.a(stdlib_system.c.o): in function `nimLoadLibrary':
stdlib_system.c:(.text+0xfcb7): warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
```

> Are you aware of the `--dynlibOverride` option of the compiler?

Yes. Doesn't apply.

I added "libnimbus.a" and "wrappers-static" Makefile targets. They work as expected, with that link-time warning about dlopen.

kdeme:
re correctness on naming. This is actually an API + a wrapper :). As it are not directly Whisper calls wrapped, but the Status part (or a very very basic version of it) added on top also.

Maybe this says something on how we should (re)organise this code?

nim.cfg
zah: I don't follow. You control the compilation command when you build the library. You can add custom defines in there, you can change the nimcache dir, what's the problem?

Considering that libnimbus would be built relatively rarely, just about any solution is better than placing `--forcebuild` in the top-level nim.cfg.

stefantalpalaru:
> This is actually an API + a wrapper

The main API is formed by Nim procs marked with an asterisk. This is a wrapper that adds some `{.exportc.}` procs around that existing Nim API.

nim.cfg
stefantalpalaru: > You control the compilation command when you build the library. You can add custom defines in there, you can change the nimcache dir, what's the problem?

The nimcache dir getting corrupt afterwards. You just did `rm -rf nimcache; make libnimbus.so` and now you want to do `make nimbus`? You can't, because the Nim cache still holds C files that exclude pieces of code incompatible with the library target. You actually need `rm -rf nimcache; make nimbus`.

Do you see where this is going? Manual cache invalidation. We might as well face the music and disable it globally.

> you can change the nimcache dir

Didn't we do just that for release vs. debug builds? It's a futile attempt to delay the inevitable and a source of future time wasting, now with more confusion regarding which nimcache dir you need to delete with the hope that your unexplainable compiler error will just go away.

This is not acceptable. We need to reduce these voodoo programming practices, not facilitate them some more.

wrappers/wrapper_example.go
```diff
@@ -6,8 +6,8 @@ import (
 	"time"
 )
 
-// #cgo LDFLAGS: /home/oskarth/git/nimbus/nimbus/libnimbus_api.so -lm
-// #include "libnim.h"
+// #cgo LDFLAGS: -Wl,-rpath,'$ORIGIN' -L${SRCDIR}/../build -lnimbus -lm
+// #include "wrapper.h"
```
oskarth: Actually this might be a bit too generic considering it is going to be used in other code bases. What's the main objection against something like `libnim`?

nim.cfg
zah: There is nothing wrong with using a different nimcache dir when you build `libnimbus.so`. Please do that.

oskarth:
I compiled a rough list of packages/functions/types etc we currently rely on from geth and status-go in status-console-client. Perhaps this is useful in terms of coming up with a plan for getting Nimbus into the app? I suspect a lot of these already exist, and in some cases (util fns) we can do it gradually. Just to get a rough overview. https://gist.github.com/oskarth/bf3a523df8cef353990a4e500a8dde03

nim.cfg
stefantalpalaru: > to make it easier for the user to wipe out everything in case of confusion

So to make voodoo programming easier and encourage magic rituals instead of a debugging session based on relevant error messages. Fine, I'll do it.

stefantalpalaru: Done.

wrappers/wrapper_example.go
stefantalpalaru: Renamed to "libnimbus.h".

stefantalpalaru:
Changes:

- re-enable Nim's cache globally
- add a new "nimcache" subdir for the libraries
- make sure Go doesn't try to link libnimbus.a in the dynamically linked wrapper example
- always delete the static archive before recreating it
- rename wrapper.nim/.h to libnimbus.nim/.h
- better hygiene in libnimbus.h (include guards and C++ support)
- remove MainnetBootnodes copy, since we do include "config.nim" after all

nim.cfg
zah: Thank you. The nimcache issues are not that common in my experience and using a separate dir should mitigate them. Nim incremental compilation is not yet perfect, but `--forcebuild` would have added up to 30 seconds to every compile-test cycle in the largest projects.

arnetheduck: fwiw they're more common when external c files are involved. I suspect there are bugs where nim confuses freshness of the `.nim` file with that of the corresponding `.c` file - I encountered this regularly with both this wrapper and `nlvm`.. a workaround will be to add a make target that runs forcebuild and depends on the wrapper objects.. or you can start trying to pinpoint the exact conditions that this happens and upstream bugreport.

arnetheduck: and it's pretty bad because those bit by it in this case are people coming from other languages trying to integrate nimbus - their first experience becomes one of hitting randomly happening nim bugs. doesn't instil confidence.

zah: Let's not get ahead of ourselves too much. "random people from the internet trying to integrate Nimbus" is something that's expected to happen in the very distant future. Our teammates at Status, we can support. If rebuilding `libnimbus` is specifically broken at the moment, `--forcebuild` can be added only to that target.

Otherwise, I've fixed few such cache invalidation issues in the compiler myself and they are usually quite trivial (both to figure out what happens and to implement the solution). So, it's indeed a good idea to try to report the problem upstream.

PR #375, opened 2019-08-14
Add PersistBlockTest for block 1352922, fixes #363

### LINKED ISSUE #363: block 1.352.922 doesn't have a test case
PR #357 doesn't include a test case. In the past, the problematic code already fixed numerous times. Probably because of premature optimization while the code was written first time.

Better to have a test case to prevent further regression. To generate a test case, a database contains information about the block should available. Then use `persistBlockTestGen` tool in folder `tests/` to generate a json test case.

kdeme:


kdeme:
Build failures because of missing nim-eth bump. 
But it already got bumped here: https://github.com/status-im/nimbus/pull/373
So I'll wait till that one gets merged.

jangko:
@zah already bump nim-eth

jangko:
#376 fixes CI failures problem. merging

Issue #177, opened 2018-11-19
Current TODO items for 1.0


coffeepots:
# Nimbus 1.0 TODO

## Summary

Meta-issue of the currently known work remaining to reach 1.0.

### Chain, State & VM
- [ ] Missing functionality:
  - [x] Fetch block bodys (See #126)
  - [ ] Fetch proofs
  - [ ] Fetch settings
- [x] Gas costs intricacies and estimation
- [x] Fork differences in VM & edge cases (particularly with `Call` & `Create`) (See #121 and #123)
- [x] Missing `transaction pending` pool
- [x] Transaction logging and retrieval
- [x] Mining
- [ ] Filtering functionality for working with RPC filters
- [x] Journal DB integration in chain (handled in `Call`/`Create`, needs to be integrated for other VM components)

### Networking, P2P & LES

- [ ] Low number of blocks parsed before failure (block number `49,439`)
- [ ] RPC internal integration where appropriate & comprehensive test suite
- [ ] Small additions to sync and LES
- [ ] IP 6 support

### Misc

- [x] Whisper
- [x] EVMC support (See #114)
- [ ] Truffle support (See #103)
- [ ] Sharable debugging and testing environment (See #172)
- [x] Block validation tool (See #171)
- [ ] Fuzzy testing & Sanitisation (See #164)

# Missing Important Functionality

- [x] DB_Chain:
  - [x] Stub: `getBlockBody`
  - [x] Stub: `getBlockTransactionHashes`
- [x] Chain:
  - [x] `processTransaction` must run VM
- [x] VM_State_Transitions:
  - [x] Stub: `executeTransaction`
  - [x] Stub: `addTransaction`
- [x] Header:
  - [x] `generateHeaderFromParentHeader` difficulty calculation & data
- [ ] P2P:
  - [ ] Stub: `handleLostPeer`
  - [x] Stub: Handshake response (mock peers)
- [ ] LES
  - [ ] Announce must handle a new block
- [x] Eth_Common
  - [x] Many stubbed procedures

# TODO by unit

### nimbus.nim

- [ ] Networking: IP6, multiple bind addresses
- [ ] TODO mentions "No database support"? What about rocksdb? Or is this referencing journalling?
- [ ] Use CLI/RPC when ready

### sqlite_backend.nim

- [ ] Implement schema versioning

### db_chain.nim

- [x] Journal DB integration
- [x] Stub: `getBlockBody` (reads block body from hash)
[...]

PR #380, opened 2019-08-17
fixes #379, reactivates GST and fixed 32 bit problem

### LINKED ISSUE #379: missing generalStateTest from CI
during march 19-20 features addition of `test_generalstate_json.nim`, I accidentally made GST disappeared from CI log. And because we are not aware of this situation, many new 32 bit GST failed during byzantium and constantinople implementation.

I already know whats wrong. but I open this issue to remind myself of related TODOs:
- [x] make GST appear in CI again
- [x] remove old 32 bit failed GST from skip list if it already fixed
- [x] include new 32 bit failed GST into skip list

jangko:


nimbus/vm/transaction_tracer.nim
```diff
@@ -104,7 +104,7 @@ proc traceOpCodeEnded*(tracer: var TransactionTracer, c: BaseComputation, op: Op
         storage[key.dumpHex] = %(value.dumpHex)
       j["storage"] = storage
 
-  let gasRemaining = j["gas"].getInt()
+  let gasRemaining = j["gas"].getBiggestInt()
```
arnetheduck: aren't we simply going from one "random" int size to another here?

jangko: well yeah, too bad std-lib json library doesn't offer better alternative.

zah: As a long-term to-do item, you can try to define the Tracer json structures as a strongly-typed values in Nim and then rely on `nim-json-serialization` to save and load them. This will give you slightly more control over the data types.

jangko: all-right, I will add it to #177 todo list

Issue #67, opened 2018-07-05
Tests & implementation of CallDataLoad and CodeCopy


mratsim:
Extracting discussion from https://github.com/status-im/nimbus/pull/59/files#r200313762

## CallDataLoad
The current master implementation for calldataload is (from #59):

```Nim
proc callDataLoad*(computation: var BaseComputation) =
  # Load call data into memory
  let origDataPos = computation.stack.popInt
  if origDataPos >= computation.msg.data.len:
    computation.stack.push(0)
    return

  let
    dataPos = origDataPos.toInt
    dataEndPosition = dataPos + 31

  if dataEndPosition < computation.msg.data.len:
    computation.stack.push(computation.msg.data[dataPos .. dataEndPosition])
  else:
    var bytes: array[32, byte]
    var presentBytes = min(computation.msg.data.len - dataPos, 32)

    if presentBytes > 0:
      copyMem(addr bytes[0], addr computation.msg.data[dataPos], presentBytes)
    else:
      presentBytes = 0

    for i in presentBytes ..< 32: bytes[i] = 0
      computation.stack.push(bytes)
```

With a potential simpler solution (from #65)

```Nim
op callDataLoad, inline = false, startPos:
  ## 0x35, Get input data of current environment
  let start = startPos.toInt

  # If the data does not take 32 bytes, pad with zeros
  let endRange = min(computation.msg.data.len - 1, start + 31)
  let padding = start + 31 - endRange
  var value: array[32, byte] # We rely on value being initialized with 0 by default
  value[padding ..< 32] = computation.msg.data.toOpenArray(start, endRange)

  push: value # TODO, with the new implementation we can delete push for seq[byte]
```

## CodeCopy
The current master implementation for codecopy is (from #59):

```Nim
proc writePaddedResult(mem: var Memory,
                       data: openarray[byte],
                       memPos, dataPos, len: Natural,
                       paddingValue = 0.byte) =
  mem.extend(memPos, len)

  let dataEndPosition = dataPos + len - 1
  if dataEndPosition < data.len:
    mem.write(memPos, data[dataPos .. dataEndPosition])
  else:
    var presentElements = data.len - dataPos
    if presentElements > 0:
      mem.write(memPos, data.toOpenArray(dataPos, data.len - 1))
    else:
      presentElements = 0

    mem.writePaddingBytes(memPos + presentElements,
                          len - presentElements,
                          paddingValue)

proc codeCopy*(computation: var BaseComputation) =
[...]

PR #381, opened 2019-08-20
Simplify calldataload

jangko:
this is the first part of #67

### DIFF
--- a/nimbus/vm/interpreter/opcodes_impl.nim
+++ b/nimbus/vm/interpreter/opcodes_impl.nim
@@ -253,26 +253,19 @@ op callValue, inline = true:
 
 op callDataLoad, inline = false, startPos:
   ## 0x35, Get input data of current environment
-  let dataPos = startPos.cleanMemRef
-  if dataPos >= computation.msg.data.len:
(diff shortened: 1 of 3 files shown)


zah:
I'm surprised to see the changed test cases. Was the old implementation producing wrong results as well? Otherwise, we should be careful to produce an optimal solution, not just a solution that uses less lines of code.

jangko:
this test case was deactivated because at that time I'm not sure if the result should be taken from parent block or current block.

also at that time, the implementation of vm_state.nim was wrong. the new values in the test case is taken from etherscan.io and it pass tests. opcodes that has new value in the test case is not related to 'calldataload' opcode. they are related to parent/current block details.

both old and new implementation of 'calldataload' pass test cases taken from java-ethereum test suite. I only add one more test case to cover special case that is not available in java-ethereum test case.

jangko:
I remember now, those old values was taken from java-ethereum too, but it test them against block number '0x01'. when `macro_assembler.nim` still being developed, I didn't use block '0x01' but use whatever available at that time -- block number '47205'.

that is why we need new numbers in the test case, they are tied to block '47205'.

PR #382, opened 2019-08-20
add 'trace' switch to tester configuration

jangko:
this changes let the 'test_generalstate_json.nim" to turn tracing on or off.


stefantalpalaru:
How about documenting it somewhere? Maybe here: https://github.com/status-im/nimbus/wiki/Understanding-and-debugging-Nimbus-EVM-JSON-tests

jangko:
great idea, that wiki page is totally outdated. it's time to fix it

PR #384, opened 2019-08-20
use nimbus-build-system

stefantalpalaru:
Also remove the "nim-beacon-chain" submodule and its specific dependencies.


zah:
We'll also need to refactor the docker build scripts in the nimbus/docker directory. Currently, both nimbus and the testnet beacon node containers are built from there.

README.md
```diff
@@ -109,7 +109,7 @@ make update
 
 To run a command that might use binaries from the Status Nim fork:
 ```bash
-./env.sh bash
+vendor/nimbus-build-system/scripts/env.sh bash
```
zah: I have to type this all the type. Let's keep a short env.sh file at the root of the repo as well.

README.md
```diff
@@ -281,8 +282,8 @@ convinced not to run a dependency check, thus clashing with our jury-rigged
 "vendor/.nimble/pkgs"):
 
 ```bash
-cd vendor/nim-blscurve
-../../nimble.sh test
+cd vendor/nim-rocksdb
+../nimbus-build-system/scripts/nimble.sh test
```
zah: Perhaps the README should recommend the use of `env.sh` in general and all the commands such as `add_submodule` and `nimble` can work as aliases or shell functions defined there.

README.md
stefantalpalaru: It would have to be duplicated in all repos using this build system. Maybe if it's just a `vendor/nimbus-build-system/scripts/env.sh "$@"` it won't need many changes in the future.

stefantalpalaru: Added.

README.md
stefantalpalaru: Implemented as shell functions and documented these alternatives in the README.

PR #386, opened 2019-08-23
add diffficulty test

jangko:


stefantalpalaru:
If "difficulty.nim" is only used in the test suite, shouldn't it be in the "tests/" directory?

jangko:
It will be used by util/header.nim to construct a block header later, after we have implemented block chain test

stefantalpalaru:
OK

stefantalpalaru:
You saw the 32-bit failures, right?

jangko:
it was status-im/nim-beacon-chain#353. it should be green after you merge them

jangko:
but i'll will rebase and activates the test first

PR #394, opened 2019-09-05
tests: reduce RAM usage

stefantalpalaru:


nimbus.nimble
```diff
@@ -26,10 +26,11 @@ proc buildBinary(name: string, srcDir = "./", params = "", lang = "c") =
   var extra_params = params
   for i in 2..<paramCount():
     extra_params &= " " & paramStr(i)
-  exec "nim " & lang & " --out:./build/" & name & " " & extra_params & " " & srcDir & name & ".nim"
+  exec "nim " & lang & " --out:build/" & name & " " & extra_params & " " & srcDir & name & ".nim"
```
arnetheduck: wow, is this really the best way to exec? seems so.. awkward to be concatenating strings (that might need quoting)

stefantalpalaru: The resulting string is passed to a shell, so quoting is up to you. There are no user-generated paths in here, so we won't get surprise spaces in them.

arnetheduck: yeah, that's a lucky coincidence though - it's just a fragile way of doing things in general, was mainly wondering if the bar is really so low.. letting the shell quote means having to deal with platforms etc

jangko:
looks like this PR only contribute to a small amount of RAM usage reduction.
Sure this PR reduce the RAM used by the compiler, but it does not reduce the RAM usage of the test suite in `all_tests.nim`. I think the problem is `all_tests` process reach the limit of a process max allowed memory usage, and it is not equal to max OS memory.
perhaps we should break up `all_tests` into separate executables or write a simple CLI to select which sub test we want to execute

stefantalpalaru:
> write a simple CLI to select which sub test we want to execute

This is better, because you get to keep the lower compile time by still aggregating all tests in one binary.

PR #395, opened 2019-09-11
Improve and add on Nimbus Status/Whisper API + wrapper

kdeme:


wrappers/libnimbus.h
```diff
@@ -23,20 +24,22 @@ typedef struct {
...
   uint8_t topic[4];
   int allowP2P;
 } filter_options;
 
 typedef struct {
-  const char* symKeyID;
-  uint8_t* pubKey; // 64 bytes public key
-  const char* sourceID;
+  const char* symKeyID; /* Identifier for symmetric key, set to nil if none */
+  uint8_t* pubKey; /* 64 bytes public key, set to nil if none */
```
PombeirP: @kdeme I guess this can also be `const`?

kdeme: done.

kdeme:
This is ready for merge/review.
The current version has been used in the PoC with status-console-client.

wrappers/libnimbus.h
```diff
@@ -2,45 +2,124 @@
...
  */
 void nimbus_poll();
 
-void nimbus_post(const char* channel, const char* payload);
-void nimbus_subscribe(const char* channel, received_msg_handler msg);
+/** Asymmetric Keys API */
+
+/** It is important that caller makes a copy of the returned strings before
+ * doing any other API calls. */
+const char* nimbus_new_keypair();
```
zah: The practice of returning pointers to Nim-managed buffers is the most disputable aspect of the current API. Let's try to avoid it as much as possible. Here, my understanding is that the output buffer has a fixed size. Why don't we use a regular output parameter instead? (i.e. this will require the user to pass in the address of a buffer of the right size where the result will be stored)

zah: Also, if the keys are fixed size, why are all APIs here working with `const char*`? Why don't we pass pointers to fixed-size arrays?

wrappers/libnimbus.nim
```diff
@@ -1,106 +1,425 @@
...
+
+proc nimbus_unsubscribe_filter(id: cstring): bool {.exportc, raises: [].} =
+  doAssert(not id.isNil, "Filter id cannot be nil.")
+
+  result = node.unsubscribeFilter($id)
+
+proc nimbus_get_min_pow(): float64 {.exportc, raises: [].} =
+  result = node.protocolState(Whisper).config.powRequirement
+
+proc nimbus_get_bloom_filter(bloom: ptr Bloom) {.exportc, raises: [].} =
```
zah: It's OK to assume that Nim's `var` is compiled to a pointer type.

wrappers/wrapper_example.go
```diff
@@ -24,37 +30,40 @@ func poll() {
...
 	peer1 := "enode://2d3e27d7846564f9b964308038dfadd4076e4373ac938e020708ad8819fd4fd90e5eb8314140768f782db704cb313b60707b968f8b61108a6fecd705b041746d@192.168.0.33:30303"
 	peer2 := "enode://4ea35352702027984a13274f241a56a47854a7fd4b3ba674a596cff917d3c825506431cf149f9f2312a293bb7c2b1cca55db742027090916d01529fe0729643b@206.189.243.178:443"
-
 	peer3 := "enode://94d2403d0c55b5c1627eb032c4c6ea8d30b523ae84661aafa18c539ce3af3f114a5bfe1a3cde7776988a6ab2906169dca8ce6a79e32d30c445629b24e6f59e0a@0.0.0.0:30303"
-	fmt.Println(C.nimbus_add_peer(C.CString(peer1)))
-	fmt.Println(C.nimbus_add_peer(C.CString(peer2)))
-
-	fmt.Println(C.nimbus_add_peer(C.CString(peer3)))
-
+	C.nimbus_add_peer(C.CString(peer1))
```
zah: This code is just an example, but it will leak memory, so perhaps we shouldn't be teaching people bad practices. According to the [docs](https://golang.org/cmd/cgo/), Go's `C.CString` function allocates memory using C's `malloc` and this memory has to be freed manually afterwards.

wrappers/libnimbus.h
kdeme: Yes, that practice is disputable and that is why I wrote that exactly in the readme file under caveat. The solution of passing that id into an output parameter is also given  there as possible alternative but I thought it might be more cumbersome to use for the calling library.

wrappers/wrapper_example.go
kdeme: Yes agreed, I created and cleaned up the wrapper_whisper_example.go, which doesn't do this faulty practice anymore (unless I forgot some), so that would be the better example to follow.
I will clean-up this example file also.

wrappers/libnimbus.h
zah: I've read the README, yes. I believe some time will be spent in creating a Go-specific wrapper, so most inconveniences can be smoothed out.

wrappers/libnimbus.nim
kdeme: Move to use of `var` instead of `ptr` where possible

wrappers/libnimbus.h
kdeme: Changed the API so that now pointers to 32 raw byte arrays are used for the IDs. 
Hex conversion will need to happen (if wanted) on the Go/C side.

wrappers/wrapper_whisper_example.go
```diff
@@ -60,23 +60,45 @@ func StatusListenAndPost(channel string) {
 	channelC := C.CString(channel)
 	defer C.free(unsafe.Pointer(channelC))
 
-	symKeyId := C.GoString(C.nimbus_add_symkey_from_password(channelC))
-	asymKeyId := C.GoString(C.nimbus_new_keypair())
+	tmp := C.malloc(C.size_t(32))
```
zah: There is a memory leak here if the panicking branch below is taken. In general, let's use `defer free` immediately after every `malloc`

wrappers/wrapper_example.go
```diff
@@ -24,37 +32,44 @@ func poll() {
...
-	fmt.Println(C.nimbus_add_peer(C.CString(peer1)))
-	fmt.Println(C.nimbus_add_peer(C.CString(peer2)))
-
-	fmt.Println(C.nimbus_add_peer(C.CString(peer3)))
+	fmt.Println("[nim-status] Start Nimbus")
+	if C.nimbus_start(30306, true, false, 0.002, nil, false) == false {
+		panic("Can't start nimbus")
+	}
 
+	peer1 := C.CString("enode://2d3e27d7846564f9b964308038dfadd4076e4373ac938e020708ad8819fd4fd90e5eb8314140768f782db704cb313b60707b968f8b61108a6fecd705b041746d@192.168.0.33:30303")
```
zah: `defer free` can be used after each `C.CString` conversion as well.

wrappers/wrapper_whisper_example.go
kdeme: Yeah, I wanted to explicitly show at least once that the free can be done right after the `GoBytes` as it does a copy. Hence the difference with the usual usage of defer in the rest of the example.
No real memory leak as we are panicking anyway, but yes, not the right example as in a usual case you will not panic there.

PR #397, opened 2019-09-21
attempt to add simple cli to all_tests

jangko:


jangko:
Although not very ideal, this PR tries to overcome appveyor 32 bit oom issue. The approach taken is implementing a simple cli for all_tests.nim. If no cli param used when all_tests executed, it will call itself repeatedly until all tests module were executed.
This approach requires a test module to define an entry poin a.ka. a 'Main' proc instead of relying on the compiler to generate and call it.
An ideal solution would be the Nim GC fixed.

arnetheduck:
shouldn't the `unittest` module already have capabilities like this, ie filter by test name?

jangko:
That's right, but it also cannot be automated to execute all of the tests without knowing the test's name.
On the other hand, this PR tries to solve different issues. Maybe the title I give is a bit misleading.
The actual purpose of this PR is to overcome the Appveyor 32 bit OOM problem while at the same time avoiding to modifies the makefiles and friends.
We aggregate all the tests in one file is to reduce compiling time during CI run, and this PR keep that as a priority.

stefantalpalaru:
That `cliBuilder` macro is very ugly. Can we get away with something more readable, like a template?

jangko:
> That cliBuilder macro is very ugly

Agree, now it should be more readable, please check it out.

> Can we get away with something more readable, like a template?

No. Template is not powerful enough to to this kind of trick.

stefantalpalaru:
Yes, it's better now.

tests/all_tests.nim
```diff
@@ -5,24 +5,90 @@
...
+# proc mytestMain*() =
+#   # put anything you want here
+# and then give it a name `test_mytest.nim`
+# the `mytest` part should match between
+# the proc name and the module name
+
+# if this executable called without any params
+# it will execute each of the test by executing itself
+# repeatedly until all sub-tests are executed.
+# you can execute the sub-test by a number start from zero.
```
stefantalpalaru: Not exactly user friendly. You'd expect to be able to pass the test name.

On the other hand, you can always compile and run individual tests without using this binary.

jangko: >On the other hand, you can always compile and run individual tests without using this binary.

I can add a template to enable that

stefantalpalaru: You can delay that until there's a real need.

PR #398, opened 2019-10-02
example Grafana dashboard + nim-metrics bump [skip ci]

stefantalpalaru:
Dashboard exported from Grafana 6.3.5 running locally, with Prometheus as a data source. Once imported, it would look almost like this:
screenshot

("almost", because importing is a bit broken and Grafana fills in missing gauge graph limits with a default value)

You'll notice that the metrics have been manually grouped on two different vertical axes, based on the range of their values. No, there's no apparent way to set a display gauge's maximum value to a metric's historical max. It wants fixed numbers there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,42 @@ cd vendor/nim-rocksdb
 ../../env.sh nimble test
 ```
 
+### Metric visualisation
+
+Install Prometheus and Grafana. On Gentoo, it's `emerge prometheus grafana-bin`.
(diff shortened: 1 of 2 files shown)


zah:
The team (for now) and our users (in the future) could benefit from a readme listing the steps you need to do in order to have a local Grafana setup working.

stefantalpalaru:
Instructions added to the main README.

PR #400, opened 2019-10-02
Switch to Nim v1.0.2

zah:
This is not working yet. Various compilation issues are described here:
https://github.com/status-im/nimbus/issues/399

The currently tracked Nim version is a recent Nim devel commit, which
aims to approximates a release candidate for Nim 1.0.2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ install:
 
 script:
   - set -e # fail fast
-  - make -j2 update # allow detecting and building a new Nim compiler
+  - make -j${NPROC} NIMFLAGS="--parallelBuild:${NPROC}" update # to allow a newer Nim version to be detected
   - make -j${NPROC} NIMFLAGS="--parallelBuild:${NPROC}"
   - build/nimbus --help
   - make -j${NPROC} NIMFLAGS="--parallelBuild:${NPROC}" test test-reproducibility
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: c
 
+dist: bionic
+
 # https://docs.travis-ci.com/user/caching/
 cache:
   directories:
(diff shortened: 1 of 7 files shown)


narimiran:
> The currently tracked Nim version is a recent Nim devel commit, which
aims to approximates a release candidate for Nim 1.0.2.

It is better to use one of the commits from [this branch](https://github.com/nim-lang/Nim/commits/version-1-0) as it contains only what will become 1.0.2; `devel` is ahead of that.

stefantalpalaru:
You're cherry-picking commits from the "devel" branch instead of having a stabilisation period?

narimiran:
> You're cherry-picking commits from the "devel" branch instead of having a stabilisation period?

Devel has features that cannot be part of 1.0.x, no matter of stabilization period.

jangko:
it should pass the test if we merge and bump status-im/nim-stint#96

jangko:
looks like only Arm64 CI failed. all others are OK.

zah:
@jangko, @stefantalpalaru, I've merged the Nim 1.0 branch in the beacon-chain repo. Can you rebase and merge this one as well?

.travis.yml
```diff
@@ -42,6 +42,7 @@ install:
 
 script:
   - set -e # fail fast
+  - make -j2 update # allow detecting and building a new Nim compiler
```
stefantalpalaru: Not needed. I manually deleted the Travis cache for this PR. That should fix it.

stefantalpalaru: My bad, the explicit "update" target is needed to detect new Nim compiler versions. Do use the same `-j${NPROC} NIMFLAGS="--parallelBuild:${NPROC}"` params instead of `-j2` so we can benefit from any elasticity in the number of cores in the ARM LXC container.

jangko: does it harmful to keep it around? actually I don't know whats going on with this command. I seldom look into makefile.

jangko: okay. will change it

stefantalpalaru: Thanks.

PR #410, opened 2019-11-06
Travis: use distro's librocksdb-dev on AMD64 and ARM64

stefantalpalaru:


tersec:
This would be quite useful.

PR #415, opened 2019-11-11
Implement istanbul hard fork - EIP 1679

jangko:
Work in progress towards https://github.com/status-im/nimbus/issues/412


nimbus/vm/blake2b_f.nim
```diff
@@ -0,0 +1,141 @@
+import nimcrypto/utils
+
+# Blake2 `F` compression function
```
zah: Shouldn't this change be contributed to NimCrypto? cc @cheatfate

jangko: initially I have the same idea: cheatfate/nimcrypto#32. but I doubt there will be another use case.

jangko:
The implementation of EIP 1679 (Istanbul HF) part 1 is done.
part 2: refactor `stateDB` in order to implement `getComittedStorage` will come next.
part 3: update test suite also will come in separate PR.

PR #419, opened 2019-11-13
[RDY] fix bc/gst tester for istanbul

jangko:
waiting for #417 and #418 merged

### DIFF
--- a/tests/test_blockchain_json.nim
+++ b/tests/test_blockchain_json.nim
@@ -434,7 +434,7 @@ proc validateUncles(chainDB: BaseChainDB, currBlock: PlainBlock, checkSeal: bool
     raise newException(ValidationError, "Header suggests block should have uncles but block has none")
 
   # Check for duplicates
-  var uncleSet = initSet[Hash256]()
+  var uncleSet = initHashSet[Hash256]()
   for uncle in currBlock.uncles:
     let uncleHash = uncle.hash
     if uncleHash in uncleSet:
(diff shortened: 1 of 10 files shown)


jangko:
this PR is ready for review.

tests/test_generalstate_json.nim
```diff
@@ -175,15 +179,17 @@ proc generalStateJsonMain*(debugMode = false) =
   if paramCount() == 0 or not debugMode:
     # run all test fixtures
     suite "generalstate json tests":
-      jsonTest("GeneralStateTests", testFixture)
+     jsonTest("GeneralStateTests", testFixture)
```
stefantalpalaru: Wrong indentation here?

tests/test_blockchain_json.nim
```diff
@@ -663,14 +669,16 @@ proc main() =
...
+      jsonTest("newBlockchainTests", testFixture)
   else:
     # execute single test in debug mode
     let config = getConfiguration()
     if config.testSubject.len == 0:
       echo "missing test subject"
       quit(QuitFailure)
 
-    let path = "tests" / "fixtures" / "BlockChainTests"
+    let path = "tests" / "fixtures" / "newBlockChainTests"
```
stefantalpalaru: This means that when in debug mode you only run half the tests? I saw the same being done in "tests/test_generalstate_json.nim".

tests/test_generalstate_json.nim
jangko: fixed

tests/test_blockchain_json.nim
jangko: in debug mode, only one json file is executed, one EVM fork selected and one specific subtest selected. perhaps we can use cli to configure the input path or reorganize test fixtures folders.

tests/test_generalstate_json.nim
stefantalpalaru: But why didn't it crash at compilation time?

jangko: a block with single line body is allowed to have any indentation GTE one. or in other words the first line of the block body will set the valid indentation for the whole body.

PR #427, opened 2019-11-26
Fix static wrapper

kdeme:
+ work around for wrapper-static build
+ add whisper example in static wrapper
+ attempt to add these builds to CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,11 +93,14 @@ wrappers: | build deps libnimbus.so go-checks
 libnimbus.a: | build deps
 	echo -e $(BUILD_MSG) "build/$@" && \
 		rm -f build/$@ && \
-		$(ENV_SCRIPT) nim c --app:staticlib --noMain --nimcache:nimcache/libnimbus $(NIM_PARAMS) -o:build/$@ wrappers/libnimbus.nim
+		$(ENV_SCRIPT) nim c --app:staticlib --noMain --nimcache:nimcache/libnimbus $(NIM_PARAMS) wrappers/libnimbus.nim && \
+		mv libnimbus build/$@ # workaround for https://github.com/nim-lang/Nim/issues/12745
 
 wrappers-static: | build deps libnimbus.a go-checks
 	echo -e $(BUILD_MSG) "build/C_wrapper_example_static" && \
 		$(CC) wrappers/wrapper_example.c -static -pthread -Lbuild -lnimbus -lm -ldl -lpcre -g -o build/C_wrapper_example_static
 	echo -e $(BUILD_MSG) "build/go_wrapper_example_static" && \
 		go build -ldflags "-linkmode external -extldflags '-static -ldl -lpcre'" -o build/go_wrapper_example_static wrappers/wrapper_example.go wrappers/cfuncs.go
+	echo -e $(BUILD_MSG) "build/go_wrapper_whisper_example_static" && \
+		go build -ldflags "-linkmode external -extldflags '-static -ldl -lpcre'" -o build/go_wrapper_whisper_example_static wrappers/wrapper_example.go wrappers/cfuncs.go
 
--- a/Makefile
+++ b/Makefile
@@ -93,8 +93,8 @@ wrappers: | build deps libnimbus.so go-checks
 libnimbus.a: | build deps
 	echo -e $(BUILD_MSG) "build/$@" && \
 		rm -f build/$@ && \
-		$(ENV_SCRIPT) nim c --app:staticlib --noMain --nimcache:nimcache/libnimbus $(NIM_PARAMS) wrappers/libnimbus.nim && \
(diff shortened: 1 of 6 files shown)


kdeme:
Hmm, wondering if the parallel builds for wrappers and wrappers-static in travis CI is OK as they share nimcache folders currently. Not sure...

Makefile
```diff
@@ -93,11 +93,14 @@ wrappers: | build deps libnimbus.so go-checks
 libnimbus.a: | build deps
 	echo -e $(BUILD_MSG) "build/$@" && \
 		rm -f build/$@ && \
-		$(ENV_SCRIPT) nim c --app:staticlib --noMain --nimcache:nimcache/libnimbus $(NIM_PARAMS) -o:build/$@ wrappers/libnimbus.nim
+		$(ENV_SCRIPT) nim c --app:staticlib --noMain --nimcache:nimcache/libnimbus $(NIM_PARAMS) wrappers/libnimbus.nim && \
+		mv libnimbus build/$@ # workaround for https://github.com/nim-lang/Nim/issues/12745
```
stefantalpalaru: How about this instead:
```sh
[[ -e "$@" ]] && mv "$@" build/
```

It will still work when upstream fixes the bug and it uses the target's full name. (I'm curious how dropping the suffix worked on your system)

kdeme: Yes, I'll change to include the check.
Dropping the suffix works because I don't provide any output anymore. So it just gets the name of the nim file minus the .nim extension.

stefantalpalaru: Why not let the `-o` option as is?

kdeme: Well, I removed it with the idea of not having this fail when it gets fixed upstream.

But perhaps it is better to actually allow it to fail so the workaround gets removed.

stefantalpalaru: Perfect.

stefantalpalaru:
"-linkshared not supported on darwin/amd64" - https://travis-ci.org/status-im/nimbus/jobs/617432992#L7298

Looks like an old Go problem: https://www.mail-archive.com/golang-nuts@googlegroups.com/msg16154.html

kdeme:
Yeah, but I don't think we need that -linkshared option. Unless I misunderstand something, but from that link it sounds that this is to create a shared lib from go?

kdeme:
@cheatfate Judging from the appveyor failures. It seems that nimcrypto `randomBytes` has a different signature on Windows than on Linux. That is, it is possible to raise `Exception`, as an empty raises pragma fails: https://ci.appveyor.com/project/nimbus/nimbus/builds/29142199/job/aoaq5ootqu87uccc#L11616 

Can we get that equal, or at least get something more specific than `Exception` ?

stefantalpalaru:
Yes, `-linkshared` is an useless option here, so go ahead and remove it from:

```text
Makefile:89:		go build -linkshared -o build/go_wrapper_example wrappers/wrapper_example.go wrappers/cfuncs.go
Makefile:91:		go build -linkshared -o build/go_wrapper_whisper_example wrappers/wrapper_whisper_example.go wrappers/cfuncs.go
nix/nimbus-wrappers.nix:68:      ${go}/bin/go build -linkshared -o build/go_wrapper_example wrappers/wrapper_example.go wrappers/cfuncs.go
nix/nimbus-wrappers.nix:70:      ${go}/bin/go build -linkshared -o build/go_wrapper_whisper_example wrappers/wrapper_whisper_example.go wrappers/cfuncs.go
```

stefantalpalaru:
Shouldn't the Nix file just call `make` instead of duplicating the Makefile?

kdeme:
Regarding Nix calling make. Sure, that would certainly de-duplicate things. Although some might say there is a benefit in keeping Nix build independent from make. Either way, not scope of this PR for me.

Next, I disabled the wrappers-static for OSX builds as static system libraries are not available. Other option perhaps could be to only static link to libnimbus and dynamically with the rest. But I'm not sure that this kind of build is worth adding. (see also https://stackoverflow.com/questions/844819/how-to-static-link-on-os-x ).

Next, Windows builds still fail after the [fix](https://github.com/cheatfate/nimcrypto/commit/4402703da85279133ac4b87f70ae261e59f2ac9b) in nimcrypto (sorry, should have tested this locally first on a full build), this time similar issue, but now in `Nim\lib\pure\terminal.nim`, coming from a chronicles log I assume.

kdeme:
Disabled wrappers for Windows for now until that Nim issue is fixed. After that the libraries should build however the wrappers and wrappers-static builds (examples) will need some work still.
Will need to add .dll build target instead of .so, else linking will fail (or adjust the command, see: http://www.mingw.org/wiki/Specify_the_libraries_for_the_linker_to_use). And also in the static build there is an issue similar like on osx I believe (but with pcre lib instead).

Anyhow, I'd like to merge as is now because the libraries themselves will build and are currently not even used on Windows.

Issue #424, opened 2019-11-22
test failure with Nim-1.0.4 release candidate


stefantalpalaru:
Using the "nim1.0.4" branch:

```text
  [OK] tests/fixtures/newGeneralStateTests/stArgsZeroOneBalance/xorNonConst.json
subtest no: 7 failed
```

There's also a problem with marking that failure, because there's no [FAILED] label.

jangko:
I get worse error messages:
```
nimcache/release/test_generalstate_json/utils.nim.c.o: In function `parseAddress__trh1CSgdG9bVndhQ0q3oWww':
C:/Users/Jangko/.nimble/pkgs/eth-1.0.0/eth/common/utils.nim:8: multiple definition of `parseAddress__trh1CSgdG9bVndhQ0q3oWww'
nimcache/release/test_generalstate_json/utils.nim.c.o:C:/Users/Jangko/.nimble/pkgs/eth-1.0.0/eth/common/utils.nim:8: first defined here
nimcache/release/test_generalstate_json/utils.nim.c.o: In function `toHex__3p6VYpmDb6FsTe2ZIUcuNQbyteutils':
C:/Users/Jangko/.nimble/pkgs/stew-0.1.0/stew/byteutils.nim:125: multiple definition of `dollar___Kaj5xasOxq8cJRx5ksLlmQ'
nimcache/release/test_generalstate_json/utils.nim.c.o:C:/Users/Jangko/.nimble/pkgs/stew-0.1.0/stew/byteutils.nim:125: first defined here
nimcache/release/test_generalstate_json/utils.nim.c.o: In function `eth_utilsInit000':
C:/Users/Jangko/.nimble/pkgs/eth-1.0.0/eth/common/utils.nim:12: multiple definition of `eth_utilsInit000'
nimcache/release/test_generalstate_json/utils.nim.c.o:C:/Users/Jangko/.nimble/pkgs/eth-1.0.0/eth/common/utils.nim:12: first defined here
nimcache/release/test_generalstate_json/hash.nim.c.o: In function `burnMem__WQXOjiuy7r6qJs6Q2oUAcgutils':
C:/Users/Jangko/.nimble/pkgs/nimcrypto-0.4.5/nimcrypto/utils.nim:217: undefined reference to `burnMem__haAOjkJjOuRy8ux9bV57dXw'
nimcache/release/test_generalstate_json/hash.nim.c.o: In function `burnMem__3kHY0BwnzMmCH3xJFhlgPAutils':
[...]

PR #429, opened 2019-11-29
Nim 1.0.4

stefantalpalaru:


jangko:
I could not reproduce the failure using both nim 1.0.4 and nim devel branch.

stefantalpalaru:
I restarted one of the AppVeyor jobs and it passed this time. What's weird about this spurious failure is that it happened at the same spot in both jobs, the first time.

Have you tried running that test in a loop until it fails?

stefantalpalaru:
It also failed 7 days ago, in the same test, on both architectures, in one or two subtests (those are the forks?):

https://ci.appveyor.com/project/nimbus/nimbus/builds/29055190/job/nlxex8eulb4guocw#L11578

https://ci.appveyor.com/project/nimbus/nimbus/builds/29055190/job/9muxpekct4qj3ncg#L11592

jangko:
> It also failed 7 days ago, in the same test, on both architectures, in one or two subtests (those are the forks?):

subtest no 7 #424 already fixed by #425. I can reproduce that one.
but this one seems like a mysterious hiccup, outdated cache maybe.

PR #435, opened 2019-12-09
enable EVM sha3_bigsize test on 32 bits

jangko:


jangko:
need to rebase on #428

PR #437, opened 2019-12-12
Start implementation of waku node

kdeme:
This is still WIP, issue: https://github.com/status-im/nimbus/issues/426#event-2868770644

First question would be if this code belongs in this repository or should be moved?

Point for now is to have a Node that implements Waku protocol to be able to test:
- Simulations, see : https://github.com/vacp2p/simulations/issues/1
- Waku - Whisper bridge, see also: https://github.com/vacp2p/pm/issues/4

Still requires quite some work, but for now it can be used by launching a bootnode and several nodes connecting to that bootnode. Next, if RPC is enabled one can subscribe to topics and post message, e.g. via web3, as shown in this quick test: https://gist.github.com/kdeme/17006627154cbb796df845a0b7a14fb2

When https://github.com/status-im/nim-web3 gets updated with that RPC interface we can write an application in nim that controls e.g. the simulation process.


nimbus/rpc/waku.nim
```diff
@@ -0,0 +1,355 @@
...
+    ##
+    ## pow: The new PoW requirement.
+    ##
+    ## Returns true on success and an error on failure.
+    # Note: If any of the `peer.powRequirement` calls fails, we do not care and
+    # don't see this as an error. Could move this to `setPowRequirement` if
+    # this is the general behaviour we want.
+    try:
+      waitFor node.setPowRequirement(pow)
+    except CatchableError:
```
zah: Well, this doesn't return error on failure as documented above.
It should carry at least a TODO item.

nimbus/rpc/waku.nim
```diff
@@ -0,0 +1,355 @@
+import
+  json_rpc/rpcserver, rpc_types, hexstrings, tables, options, sequtils,
+  eth/[common, rlp, keys, p2p], eth/p2p/rlpx_protocols/waku_protocol,
+  nimcrypto/[sysrand, hmac, sha2, pbkdf2]
+
+from stew/byteutils import hexToSeqByte, hexToByteArray
```
zah: Is this expected to differ considerably from Whisper? Perhaps the RCP service can use a `wakuMode` flag enabling the different behavior instead of duplicating the entire implementation.

zah:
I think it's fine for this to remain in the nimbus repo for now.

nimbus/rpc/waku.nim
kdeme: Yes, such thing has crossed my mind, as there is indeed no real difference right now. 
However, the best I could come up with for now is some compile time flag to use either waku or whisper protocol.

nimbus/rpc/waku.nim
kdeme: Yeah, I guess that comment got out of date, I'll resolve this.

nimbus/rpc/waku.nim
kdeme: I've decided to leave it as is. I don't think it is worth it to rework this with either compile time define or some templated / generic version. In the end I believe it will make it only more difficult when changes need to be made on Waku without touching Whisper.

kdeme:
So, ready for review/merge

waku/wakunode.nim
```diff
@@ -0,0 +1,147 @@
...
+    if config.metricsServer:
+      let
+        address = config.metricsServerAddress
+        port = config.metricsServerPort + config.portsShift
+      info "Starting metrics HTTP server", address, port
+      metrics.startHttpServer($address, Port(port))
+
+  if config.logMetrics:
+    proc logMetrics(udata: pointer) {.closure, gcsafe.} =
+      {.gcsafe.}:
```
zah: Why is this override necessary? Is it because reading a metric is considered non gcsafe?

kdeme: Yes, initially I tried removing that but without it would error with something like `is not GC-safe as accesses global using GC'ed memory`. 
I mostly just copied this from here: https://github.com/status-im/nimbus/blob/master/nimbus/nimbus.nim#L65 , but I didn't like the default over verbose logs.

PR #438, opened 2019-12-13
implement and integrate evmc host with nimbus evm

jangko:
- [x] implement evmc host
- [x] integrate evmc host with nimbus evm

### DIFF
--- a/nimbus/vm/computation.nim
+++ b/nimbus/vm/computation.nim
@@ -131,7 +131,7 @@ proc generateContractAddress(c: Computation, salt: Option[Uint256]): EthAddress 
     let creationNonce = c.vmState.readOnlyStateDb().getNonce(c.msg.sender)
     result = generateAddress(c.msg.sender, creationNonce)
   else:
-    result = generateSafeAddress(c.msg.sender, salt.get(), c.msg.code)
+    result = generateSafeAddress(c.msg.sender, salt.get(), c.msg.data)
 
 proc newComputation*(vmState: BaseVMState, message: Message, salt=none(Uint256)): Computation =
   new result
@@ -142,10 +142,13 @@ proc newComputation*(vmState: BaseVMState, message: Message, salt=none(Uint256))
   result.gasMeter.init(message.gas)
   result.touchedAccounts = initHashSet[EthAddress]()
   result.suicides = initHashSet[EthAddress]()
-  result.code = newCodeStream(message.code)
 
   if result.msg.isCreate():
     result.msg.contractAddress = result.generateContractAddress(salt)
+    result.code = newCodeStream(message.data)
+    message.data = @[]
+  else:
+    result.code = newCodeStream(vmState.readOnlyStateDb.getCode(message.codeAddress).toSeq)
 
   when evmc_enabled:
     result.host.init(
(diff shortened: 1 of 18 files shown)


jangko:
Review please.
Summary of this PR:
* EVMC host implemented and integrated into nimbus EVM, but we still can use native host via compiler switch `evmc_enabled` or disabled.
* EVMC `call/create` and `vm interface` will be implemented in next PR because of the complexity.

nimbus/vm/evmc_host.nim
```diff
@@ -0,0 +1,200 @@
...
+
+  assert storageAddr == ctx.msg.contractAddress
+
+  var
+    status = EVMC_STORAGE_MODIFIED
+    gasRefund = 0.GasInt
+    origValue = 0.u256
+
+  block:
+    if newValue == currValue:
```
zah: Isn't this block equivalent to the following:

```nim
if newValue == currValue:
  status = EVMC_STORAGE_UNCHANGED
else:
  origValue = statedb.getCommittedStorage(storageAddr, slot)
  if origValue == currValue or ctx.fork < FkIstanbul:
    if currValue == 0:
      status = EVMC_STORAGE_ADDED
    elif newValue == 0:
      status = EVMC_STORAGE_DELETED
  else:
    status = EVMC_STORAGE_MODIFIED_AGAIN
  ctx.vmState.mutateStateDB:
    db.setStorage(storageAddr, slot, newValue)
 
```

nimbus/vm/evmc_api.nim
```diff
@@ -0,0 +1,111 @@
...
+    Hash256.fromEvmc ctx.host.get_block_hash(ctx.context, number.truncate(int64))
+
+proc accountExists*(ctx: HostContext, address: EthAddress): bool =
+  var address = toEvmc(address)
+  {.gcsafe.}:
+    ctx.host.account_exists(ctx.context, address.addr).bool
+
+proc getStorage*(ctx: HostContext, address: EthAddress, key: Uint256): Uint256 =
+  var
+    address = toEvmc(address)
```
zah: We can avoid a lot of the casts here if the evmc host interface was directly working with the nim-eth types. We are already baking in the assumption that the nim-evmc types match the nim-eth types precisely, so we can go one step further and just use them outright. This would introduce a dependency on nim-eth in nim-evmc, but I think that's fine (in the future, we can create a smaller nim-eth-types package again).

nimbus/vm/computation.nim
```diff
@@ -185,7 +318,26 @@ proc addChildComputation*(c, child: Computation) =
     c.gasMeter.returnGas(child.gasMeter.gasRemaining)
 
 proc registerAccountForDeletion*(c: Computation, beneficiary: EthAddress) =
```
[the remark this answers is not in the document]
jangko: fixed

nimbus/vm/evmc_api.nim
```diff
@@ -0,0 +1,111 @@
...
+
+proc init*(x: var HostContext, host: ptr evmc_host_interface, context: evmc_host_context) =
+  x.host = host
+  x.context = context
+
+proc init*(x: typedesc[HostContext], host: ptr evmc_host_interface, context: evmc_host_context): HostContext =
+  result.init(host, context)
+
+proc getTxContext*(ctx: HostContext): evmc_tx_context =
+  {.gcsafe.}:
```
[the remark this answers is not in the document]
jangko: fixed

nimbus/vm/evmc_host.nim
jangko: fixed. thanks

nimbus/vm/evmc_api.nim
jangko: fixed, except for Uint256. EVMC using Big Endian format, while stint using Little Endian. But I'll take deeper look later because rlp encoding and the EVM stack operates using Big Endian format.

PR #442, opened 2019-12-20
attempt to remove computation children sequence PART 2

jangko:
- [x] implement AccountsCache
- [x] AccountsCache test
- [ ] AccountsCache in action
- [x] collect touched accounts and remove computation children sequence


nimbus/db/accounts_cache.nim
```diff
@@ -279,23 +287,23 @@ proc hasCodeOrNonce*(ac: AccountsCache, address: EthAddress): bool {.inline.} =
 
 proc accountExists*(ac: AccountsCache, address: EthAddress): bool {.inline.} =
   let acc = ac.getAccount(address)
-  result = IsNew notin acc.flags
+  acc.isExists()
```
zah: `isExists` doesn't read well in English. We can go for just `exists` or `doesExist`.

jangko: thanks @zah, will fix it

jangko:
review please. summary of this PR:
* accounts cache implemented but not used(need more simplification before integration)
* redundant code are removed from Base computation
* EVM Message creation are simplified and consistent(easier evmc integration)
* recursive collectTouchedAccounts transformed into 'merge after child computation execution'
* we do not keep child computation in a list anymore and this is a huge reduction in memory consumption. Previously memory usage in the range of 3-4GB but now reduced to 350+ MB.
* CI  tests not choked to death anymore because of frequent OOM or near OOM timeouts.

nimbus/db/accounts_cache.nim
```diff
@@ -0,0 +1,397 @@
...
+  initSecureHexaryTrie(db, acc.account.storageRoot, false)
+
+proc originalStorageValue(acc: RefAccount, slot: UInt256, db: TrieDatabaseRef): UInt256 =
+  # share the same original storage between multiple
+  # versions of account
+  if acc.originalStorage.isNil:
+    acc.originalStorage = newTable[UInt256, UInt256]()
+  else:
+    if slot in acc.originalStorage:
+      result = acc.originalStorage[slot]
```
zah: You can avoid the double look-up here, by using `acc.originalStorage.getOrDefault(slot)` or `acc.originalStorage .withValue(slot, val): return val[] else: discard`

nimbus/vm/computation.nim
```diff
@@ -233,25 +214,19 @@ proc addChildComputation*(computation: BaseComputation, child: BaseComputation)
...
-      "invariant:  should be impossible for an account to be " &
-      "registered for deletion multiple times")
-  # TODO: collecting touched accounts should be done in state db
-  # we should remove this accountsToDelete map delegate it to state db
-  # we will only keep suicides list here in order to make
-  # evmc integration easier
-  c.accountsToDelete[c.msg.storageAddress] = beneficiary
-  c.suicides.incl(c.msg.storageAddress)
+  c.touchedAccounts.incl beneficiary
+  c.touchedAccounts.incl beneficiary
```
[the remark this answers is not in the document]
jangko: It must be an artifact left by too smart search and replace. will fix it.

nimbus/db/accounts_cache.nim
jangko: fixed using `withValue`. thank you.

PR #443, opened 2020-01-02
Support custom networks

acolytec3:
Problem: Nimbus does not currently support custom/private networks for Ethereum 1 like Geth/Parity.

Solution: This PR will bring Nimbus one step closer to feature parity with Geth by accepting a new new `customnetwork` argument that takes a json file as a value and then loads `ChainConfig` values from the genesis block provided.  The format for the genesis block must match the format [specified in the Geth documentation](https://github.com/ethereum/go-ethereum/wiki/Private-network).  This will work out of the box with a `genesis.json` created by [puppeth](https://github.com/ethereum/go-ethereum#executables).

PR status: ready

### DIFF
--- a/nimbus/config.nim
+++ b/nimbus/config.nim
@@ -8,7 +8,7 @@
 # those terms.
 
 import
-  parseopt, strutils, macros, os, times,
+  parseopt, strutils, macros, os, times, json, stew/[byteutils, ranges],
   chronos, eth/[keys, common, p2p, net/nat], chronicles, nimcrypto/hash,
   eth/p2p/bootnodes, eth/p2p/rlpx_protocols/whisper_protocol,
   ./db/select_backend,
@@ -143,6 +143,27 @@ type
     net*: NetConfiguration        ## Network configuration
     debug*: DebugConfiguration    ## Debug configuration
     shh*: WhisperConfig           ## Whisper configuration
+    customGenesis*: CustomGenesisConfig  ## Custom Genesis Configuration
+
+  CustomGenesisConfig = object
+    chainId*: uint
+    homesteadBlock*: BlockNumber
+    daoForkBlock*: BlockNumber
+    daoForkSupport*: bool
+    eip150Block*: BlockNumber
+    eip150Hash*: Hash256
+    eip155Block*: BlockNumber
+    eip158Block*: BlockNumber
+    byzantiumBlock*: BlockNumber
+    constantinopleBlock*: BlockNumber
+    nonce*: BlockNonce
+    extraData*: seq[byte]
+    gasLimit*: int64
+    difficulty*: UInt256
+    prealloc*: JsonNode
+    mixHash*: Hash256
+    coinBase*: EthAddress
+    timestamp*: EthTime
 
 const
   defaultRpcApi = {RpcFlags.Eth, RpcFlags.Shh}
@@ -154,6 +175,22 @@ var nimbusConfig {.threadvar.}: NimbusConfiguration
 
 proc getConfiguration*(): NimbusConfiguration {.gcsafe.}
 
+proc privateChainConfig*(): ChainConfig =
+  let config = getConfiguration()
+  result = ChainConfig(
+    chainId:          config.customGenesis.chainId,
+    homesteadBlock:   config.customGenesis.homesteadBlock,
+    daoForkSupport:   config.customGenesis.daoForkSupport,
+    daoForkBlock:     config.customGenesis.daoForkBlock,
+    eip150Block:      config.customGenesis.eip150Block,
+    eip150Hash:       config.customGenesis.eip150Hash,
+    eip155Block:      config.customGenesis.eip155Block,
+    eip158Block:      config.customGenesis.eip158Block,
+    byzantiumBlock:   config.customGenesis.byzantiumBlock,
+    constantinopleBlock: config.customGenesis.constantinopleBlock,
+  )
+  trace "Custom genesis block configuration loaded", configuration=result
+  
 proc publicChainConfig*(id: PublicNetwork): ChainConfig =
   result = case id
   of MainNet:
@@ -190,6 +227,8 @@ proc publicChainConfig*(id: PublicNetwork): ChainConfig =
       eip158Block:    3.toBlockNumber,
       byzantiumBlock: 1035301.toBlockNumber
     )
+  of CustomNet:
+    privateChainConfig()
   else:
     error "No chain config for public network", networkId = id
     doAssert(false, "No chain config for " & $id)
@@ -197,6 +236,124 @@ proc publicChainConfig*(id: PublicNetwork): ChainConfig =
 
   result.chainId = uint(id)
 
+proc processCustomGenesisConfig(customGenesis: JsonNode): ConfigStatus =
+  ## Parses Custom Genesis Block config options when customnetwork option provided
+
+  template checkForFork(chain, currentFork, previousFork: untyped) =
+  # Template to load fork blocks and validate order
+    let currentForkName = currentFork.astToStr()
+    if chain.hasKey(currentForkName):
+      if chain[currentForkName].kind == JInt:
+        currentFork = chain[currentForkName].getInt().toBlockNumber
+        if currentFork < previousFork:
+          error "Forks can't be assigned out of order"
+          quit(1)
+      else:
+        error "Invalid block value provided for", currentForkName, invalidValue=currentFork
+        quit(1)
+
+  proc parseConfig(T: type, c: JsonNode, field: string): T =
+    when T is string:
+      c[field].getStr()    
+    elif T is Uint256:
+      parseHexInt(c[field].getStr()).u256
+    elif T is bool:
+      c[field].getBool()
+    elif T is Hash256:
+      MDigest[256].fromHex(c[field].getStr())
+    elif T is uint:
+      c[field].getInt().uint
+    elif T is Blocknonce:
+      (parseHexInt(c[field].getStr()).uint64).toBlockNonce
+    elif T is EthTime:
+      fromHex[int64](c[field].getStr()).fromUnix
+    elif T is EthAddress:
+      parseAddress(c[field].getStr())
+    elif T is seq[byte]:
+      hexToSeqByte(c[field].getStr())
+    elif T is int64:
+      fromHex[int64](c[field].getStr())
+    
+
+  template validateConfigValue(chainDetails, field, jtype, T: untyped, checkError: static[bool] = true) =
+    let fieldName = field.astToStr()
+    if chainDetails.hasKey(fieldName):
+      if chainDetails[fieldName].kind == jtype:
+        field = parseConfig(T, chainDetails, fieldName)
+      else:
+        error "Invalid value provided for ", fieldName
+        quit(1)
+    else:
+      when checkError:
+        error "No value found in genesis block for", fieldName
+        quit(1)
+  
+  let config = getConfiguration() 
+  result = Success
+  var 
+    chainId = 0.uint
+    homesteadBlock, daoForkblock, eip150Block, eip155Block, eip158Block, byzantiumBlock, constantinopleBlock = 0.toBlockNumber
+    eip150Hash, mixHash : MDigest[256]
+    daoForkSupport = false
+    nonce = 66.toBlockNonce
+    extraData = hexToSeqByte("0x")
+    gasLimit = 16777216.int64
+    difficulty = 1048576.u256
+    alloc = parseJson("{}")
+    timestamp : EthTime
+    coinbase : EthAddress
+
+
+  if customGenesis.hasKey("config"):
+  # Validate all fork blocks for custom genesis
+    let forkDetails = customGenesis["config"]
+    validateConfigValue(forkDetails, chainId, JInt, uint)
+    config.net.networkId = chainId
+    checkForFork(forkDetails, homesteadBlock, 0.toBlockNumber)
+    validateConfigValue(forkDetails, daoForkSupport, JBool, bool, checkError=false)
+    if daoForkSupport == true:
+      checkForFork(forkDetails, daoForkBlock, 0.toBlockNumber)
+  
+    checkForFork(forkDetails, eip150Block, homesteadBlock)
+    validateConfigValue(forkDetails, eip150Hash, JString, Hash256)
+    checkForFork(forkDetails, eip155Block, eip150Block)
+    checkForFork(forkDetails, eip158Block, eip155Block)
+    checkForFork(forkDetails, byzantiumBlock, eip158Block)
+    checkForFork(forkDetails, constantinopleBlock, byzantiumBlock)
+  else:
+    error "No chain configuration found."
+    quit(1)
+  validateConfigValue(customGenesis, nonce, JString, BlockNonce)
+  validateConfigValue(customGenesis, extraData, JSTring, seq[byte], checkError = false)
+  validateConfigValue(customGenesis, gasLimit, JString, int64, checkError = false)
+  validateConfigValue(customGenesis, difficulty, JString, UInt256)
+  if customGenesis.hasKey("alloc"):
+    alloc = customGenesis["alloc"]
+
+  validateConfigValue(customGenesis, mixHash, JString, Hash256)
+  validateConfigValue(customGenesis, coinbase, JString, EthAddress, checkError = false)
+  validateConfigValue(customGenesis, timestamp, JString, EthTime, checkError = false)
+  
+  config.customGenesis = CustomGenesisConfig(
+    chainId:          chainId,
+    homesteadBlock:   homesteadBlock,
+    eip150Block:      eip150Block,
+    eip150Hash:       eip150Hash,
+    eip155Block:      eip155Block,
+    eip158Block:      eip158Block,
+    daoForkSupport:   daoForkSupport,
+    byzantiumBlock:   byzantiumBlock,
+    constantinopleBlock: constantinopleBlock,
+    nonce:            nonce,
+    extraData:        extraData,
+    gasLimit:         gasLimit,
+    difficulty:       difficulty,
+    prealloc:         alloc,
+    mixHash:          mixHash,
+    coinbase:         coinbase,
+    timestamp:        timestamp
+  )
+  
 proc processList(v: string, o: var seq[string]) =
   ## Process comma-separated list of strings.
   if len(v) > 0:
@@ -436,6 +593,24 @@ proc processNetArguments(key, value: string): ConfigStatus =
     config.net.setNetwork(RinkebyNet)
   elif skey == "kovan":
     config.net.setNetwork(KovanNet)
+  elif skey == "customnetwork":
+    if value == "":
+      error "No genesis block config provided for custom network", network=key
+      result = ErrorParseOption
+    else:
+      try:
+        result = processCustomGenesisConfig(parseFile(value))
+      except IOError:
+        error "Genesis block config file not found", invalidFileName=value
+        result = ErrorParseOption
+      except JsonParsingError:
+        error "Invalid genesis block config file format", invalidFileName=value
+        result = ErrorIncorrectOption
+      except:
+        var exceptionType = getCurrentException()
+        var msg = getCurrentExceptionMsg()
+        error "Error loading genesis block config file", invalidFileName=msg
+        result = Error
   elif skey == "networkid":
     var res = 0
     result = processInteger(value, res)
@@ -678,6 +853,7 @@ NETWORKING OPTIONS:
   --rinkeby               Use Ethereum Rinkeby Test Network
   --ident:<value>         Client identifier (default is '$1')
   --protocols:<value>     Enable specific set of protocols (default: $4)
+  --customnetwork         Use custom genesis block for private Ethereum Network (as /path/to/genesis.json)
 
 WHISPER OPTIONS:
   --shh-maxsize:<value>   Max message size accepted (default: $5)
@@ -766,4 +942,3 @@ when declared(os.paramCount): # not available with `--app:lib`
 proc processConfiguration*(pathname: string): ConfigStatus =
   ## Process configuration file `pathname` and update `NimbusConfiguration`.
   result = Success
-
--- a/nimbus/genesis.nim
+++ b/nimbus/genesis.nim
@@ -1,5 +1,5 @@
 import
-  tables,
+  tables, json, strutils,
   eth/[common, rlp, trie], stint, stew/[byteutils, ranges],
   chronicles, eth/trie/db,
   db/[db_chain, state_db], genesis_alloc, config, constants
@@ -32,6 +32,12 @@ func decodePrealloc(data: seq[byte]): GenesisAlloc =
   for tup in rlp.decode(data.toRange, seq[(UInt256, UInt256)]):
     result[toAddress(tup[0])] = GenesisAccount(balance: tup[1])
 
+proc customNetPrealloc(genesisBlock: JsonNode): GenesisAlloc = 
+  result = newTable[EthAddress, GenesisAccount]()
+  for address, balance in genesisBlock.pairs():
+    let balance = fromHex(UInt256,balance["balance"].getStr())
+    result[parseAddress(address)] = GenesisAccount(balance: balance)
+
 proc defaultGenesisBlockForNetwork*(id: PublicNetwork): Genesis =
   result = case id
   of MainNet:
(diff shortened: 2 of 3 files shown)


acolytec3:
For any reviewers, this PR is based on a suggestion from @kdeme in the discord channel a while back and this is the first time I've had an opportunity to get something together.  Would appreciate any thoughts on whether this is a useful PR at this point or not.  It's definitely not 100% ready as I'm still trying to figure out why a Geth node running the same genesis configuration.

zah:
Thank you for working on this! We are looking forward to merging it as soon as possible.

The code looks quite reasonable and my only suggestion is to enhance the error handling paths, so the user is presented with a friendly error message where an invalid genesis json file is provided (or a wrong path to such a file).

acolytec3:
@zah Thanks for the feedback.  I'll add that into my next set of commits.

acolytec3:
@zah I added in some error scenarios for the missing file name/file not found/invalid format cases and then some default values for the `chainconfig` object should they not be provided in the genesis file.  Anything else you'd add?

nimbus/genesis.nim
```diff
@@ -66,12 +66,27 @@ proc defaultGenesisBlockForNetwork*(id: PublicNetwork): Genesis =
     )
   of CustomNet:
     let genesis = getConfiguration().genesisBlock
+    var nonce = 66.toBlockNonce
+    if genesis.hasKey("nonce"):
+      nonce = (parseHexInt(genesis["nonce"].getStr()).uint64).toBlockNonce
+    var extraData = hexToSeqByte("")
+    if genesis.hasKey("extraData"):
+      extraData = hexToSeqByte(genesis["extraData"].getStr())
```
zah: Another possible problem here is that the `extraData` field in the JSON file may have a non-string type. Then `getStr` will return a default value which can be supplied as a second argument.

nimbus/genesis.nim
```diff
@@ -66,12 +66,27 @@ proc defaultGenesisBlockForNetwork*(id: PublicNetwork): Genesis =
...
+    var gasLimit = 16777216
+    if genesis.hasKey("gasLimit"):
+      gasLimit = parseHexInt(genesis["gasLimit"].getStr())
+    var difficulty = 1048576.u256
```
zah: How were these default values chosen?

genesis.json
```diff
@@ -0,0 +1,25 @@
+{
```
zah: Was this file added only to facilitate testing with a particular custom network?

nimbus/config.nim
```diff
@@ -436,6 +453,22 @@ proc processNetArguments(key, value: string): ConfigStatus =
     config.net.setNetwork(RinkebyNet)
   elif skey == "kovan":
     config.net.setNetwork(KovanNet)
+  elif skey == "customnetwork":
+    if value == "":
+      error "No genesis block config provided for custom network", network=key
+      raise newException(Exception, "No genesis block configuration provided")
```
zah: In Nim, it's not recommended to raise `Exception` in general. You should always choose either an exception type derived from `Defect` (which should indicate a programmer error) or `CatchableError` (which indicates an error that we anticipate and can recover from at run-time). In this particular case, you can either raise `CatchableError` or a more specific type derived from it.

genesis.json
acolytec3: Yes, sorry.  Got picked up accidentally in my PR.  I'll remove it before I squash the final PR.

nimbus/genesis.nim
acolytec3: I just mirrored the other testnets for nonce, gasLimit, and difficulty.  I just set extraData to the empty string under the assumption that it should be nil if that value isn't provided in the genesis config.

nimbus/genesis.nim
acolytec3: I see your point but not sure the best way to handle it. From everything I read, extraData is just an extra field that takes up to 32 bytes of free form text/hex/whatever and then just locks it in the genesis block (maybe for use in creating the digest for the genesis block?).  Should I just use getStr, check to see if it's a Hex string, and if not, convert to Hex and then from Hex to Bytes?

nimbus/config.nim
acolytec3: I think I missed the pattern already established in `config.nim` to just return a `result == Error` where `error` is one of the ConfigStatus values in any of the argument processing functions.  I'll correct in the next commit.

genesis.json
acolytec3: Removed in latest commit.

nimbus/config.nim
acolytec3: Check the latest iteration on this.  I replaced the raiseException with log messages that explain the error and then just return a `result = Error...` where the error type corresponds to the correct enumerated ConfigStatus value.  This produces output like below:
```
$ build/nimbus --customnetwork=genesis.tan
Nimbus Version 0.0.1 [linux: amd64, rocksdb, db48adba]
Copyright (c) 2018-2020 Status Research & Development GmbH
ERR 2020-01-06 20:47:01-05:00 Invalid genesis block config file format   tid=17493 file=config.nim:468 invalidFileName=genesis.tan
Incorrect value for option 'customnetwork': 'genesis.tan'.
```

Is this user friendly enough?

nimbus/config.nim
```diff
@@ -154,6 +155,20 @@ var nimbusConfig {.threadvar.}: NimbusConfiguration
...
+  let config = getConfiguration().genesisBlock
+  result = ChainConfig(
+    chainId:          config["config"]["chainID"].getInt().uint,
+    homesteadBlock:   config["config"]["homesteadBlock"].getInt().toBlockNumber,
+    eip150Block:      config["config"]["eip150Block"].getInt().toBlockNumber,
+    eip150Hash:       toDigest("41941023680923e0fe4d74a34bdac8141f2540e3ae90623718e47d66d1ca4a2d"),
+    eip155Block:      config["config"]["eip155Block"].getInt().toBlockNumber,
+    eip158Block:      config["config"]["eip158Block"].getInt().toBlockNumber,
+    daoForkSupport:   true,
+    byzantiumBlock:   1700000.toBlockNumber
```
kdeme: This should probably also be made configurable and defaulted to 0.

I think we can also set `ConstantinopleBlock`, to check with @jangko

nimbus/config.nim
zah: Yes, I think this is quite reasonable. Thank you for providing the example.

nimbus/config.nim
```diff
@@ -154,6 +155,20 @@ var nimbusConfig {.threadvar.}: NimbusConfiguration
 
 proc getConfiguration*(): NimbusConfiguration {.gcsafe.}
 
+proc privateChainConfig*(): ChainConfig =
+  let config = getConfiguration().genesisBlock
+  result = ChainConfig(
+    chainId:          config["config"]["chainID"].getInt().uint,
+    homesteadBlock:   config["config"]["homesteadBlock"].getInt().toBlockNumber,
+    eip150Block:      config["config"]["eip150Block"].getInt().toBlockNumber,
+    eip150Hash:       toDigest("41941023680923e0fe4d74a34bdac8141f2540e3ae90623718e47d66d1ca4a2d"),
```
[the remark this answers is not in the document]
acolytec3: Hmm, good question.  I can't remember why I did that.  :grimacing:  I think I just copied it from the rinkeby values I was using as defaults.

nimbus/config.nim
acolytec3: > Thanks for this PR @acolytec3 !
> 
> In general I think it looks good, my biggest remark is regarding the parsing of the json values. Easiest way to fix is the approach of the default value on invalid parsing (which I think happens already now with defaults of `""`, `0`, etc). However, I think this might not be so user friendly and probably not sufficient.
> I think best would probably be to have all parsing on config.nim, to be able to immediately return a proper error, but that will probably take more rework due to the way the code is set up now, and thus perhaps out of the scope of this PR.
> 
> Practical question, does it work now together with geth?

At the moment, Geth rejects the connection when I try to have it pair up using the staticnode argument, though I can't currently nimbus to pair with Geth with any configuration.  I'd tried mainnet and rinkeby and it throughs a  "Genesis block mismatch" error each time so that's not conclusive unfortunately.  All that to say, I'm happy to do all the parsing in `config.nim`.   I thought about approaching it that way but just wasn't sure what the best approach was here.  I'll rework the parsing piece and move in the next commit.  I'll also add `constantinopleblock` and figure out to what to do with the block hashes I had hardcoded.

kdeme: In the past I've tested connecting with both geth and parity local node via the staticnodes param. Worked on mainnet and ropsten, never tested rinkeby  I'll try again to see if this still works.

Yeah, it's a good idea to check how geth handles the values. In general default values are fine if that specific key does not exist. But if the key exists and the value is invalid I think it is not ideal to enter a default.

acolytec3: I've moved all of the argument parsing to config.nim and added some `warn` level errors if a key isn't found in the genesis block.  It produces output like the below now:
```bash
WRN 2020-01-08 21:11:41-05:00 No daoForkSupport value found in genesis block.  Falling back to default value. tid=17104 file=config.nim:281 value=true
WRN 2020-01-08 21:11:41-05:00 No byzantium block value found in genesis block.  Falling back to default value. tid=17104 file=config.nim:285 value=1700000
WRN 2020-01-08 21:11:41-05:00 No constantiople block value found in genesis block.  Falling back to default value. tid=17104 file=config.nim:289 value=1700001
WRN 2020-01-08 21:11:41-05:00 No extraData value found in genesis block.  Falling back to default value. tid=17104 file=config.nim:297 value=@[]
WRN 2020-01-08 21:11:41-05:00 No preallocation values found in genesis block.  Falling back to default value. tid=17104 file=config.nim:309 value={}
```

For default values, I basically just set everything to 0 that's a fork block and then an arbitrary gaslimit/difficulty.  It looks like this is more or less what geth does.

nimbus/genesis.nim
```diff
@@ -58,12 +64,39 @@ proc defaultGenesisBlockForNetwork*(id: PublicNetwork): Genesis =
...
+      difficulty: difficulty,
+      alloc: alloc
+    )
   else:
     # TODO: Fill out the rest
     error "No default genesis for network", id
     doAssert(false, "No default genesis for " & $id)
     Genesis()
-  result.config = publicChainConfig(id)
+  if id == CustomNet:
```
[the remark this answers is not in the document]
acolytec3: Removed the duplicate here.

nimbus/config.nim
```diff
@@ -154,6 +155,20 @@ var nimbusConfig {.threadvar.}: NimbusConfiguration
 
 proc getConfiguration*(): NimbusConfiguration {.gcsafe.}
 
+proc privateChainConfig*(): ChainConfig =
+  let config = getConfiguration().genesisBlock
+  result = ChainConfig(
+    chainId:          config["config"]["chainID"].getInt().uint,
```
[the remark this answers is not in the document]
acolytec3: My updates still assign default values and then use `try...except..` blocks to throw warning messages if values aren't found in the genesis block.  Does this feel about right?

nimbus/config.nim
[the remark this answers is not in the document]
acolytec3: Fixed in current commit.

acolytec3:
@kdeme And to your practical question, no...still not able to sync with a local geth node running the same genesis block so more research to do.

nimbus/config.nim
kdeme: Sorry, I probably didn't explain this well in my earlier comments, but in case of an invalid value I would write a log message + `quit(1)`.
But in case of a missing value I don't think this needs to be logged necessarily.

However, I did have a look on how geth defaults, and it looks like it omits the forks as default (see https://github.com/ethereum/go-ethereum/blob/bc01593afbb2f0bdbf7f21adc790f161b2a8b41d/params/config.go#L281).
But it looks like we can't just leave a specific fork block not filled in, as that will probably make it have a block number of 0. Perhaps @jangko can elaborate how a specific fork can be omitted.

nimbus/config.nim
```diff
@@ -190,13 +223,109 @@ proc publicChainConfig*(id: PublicNetwork): ChainConfig =
...
+  var extraData = hexToSeqByte("0x")
+  var gasLimit = 16777216.int64
+  var difficulty = 1048576.u256
+  var alloc = parseJson("{}")
+  try:
+    chainId = customGenesis["config"]["chainID"].getInt().uint
+  except:
+    warn "No chain ID found in genesis block.  Falling back to default value.", value=chainId
+  try:
+    homesteadBlock = customGenesis["config"]["homesteadBlock"].getInt().toBlockNumber
```
kdeme: So when I talk about invalid values, I mean e.g. when a string gets passed here. 
So the problem I mean then is that these `getInt()`, `getStr()`, etc. will get a default value in case it is not an `int` (see also docs https://nim-lang.org/docs/json.html#getInt%2CJsonNode%2Cint).

So in these cases it would be cleaner to print an error and quit the application.

[nim-json-rpc](https://github.com/status-im/nim-json-rpc/blob/92a7b78f4a2b13bd56260694c0035f7c71f10cb5/json_rpc/jsonmarshal.nim#L40) has similar code on how to marshal this.

nimbus/config.nim
```diff
@@ -190,13 +223,109 @@ proc publicChainConfig*(id: PublicNetwork): ChainConfig =
...
+  var eip158Block = 0.toBlockNumber
+  var daoForkSupport = true
+  var byzantiumBlock = 1700000.toBlockNumber
+  var constantinopleBlock = 1700001.toBlockNumber
+  var nonce = 66.toBlockNonce
+  var extraData = hexToSeqByte("0x")
+  var gasLimit = 16777216.int64
+  var difficulty = 1048576.u256
+  var alloc = parseJson("{}")
+  try:
```
kdeme: I think `hasKey` would look cleaner than the the try/excepts, especially considering that logging is not really needed. 
Also, with `except` you would probably want to except specifically for `KeyError`.

nimbus/config.nim
acolytec3: I think you can omit forks as long as you don't include a later fork and omit an earlier (i.e. don't include `constantinopleBlock` if `byzantiumBlock` isn't included.\, at least, that's how geth does it.

I'll rework this section so that the fork blocks are all option as long as order is preserved (following the [geth ordering](https://github.com/ethereum/go-ethereum/blob/bc01593afbb2f0bdbf7f21adc790f161b2a8b41d/params/config.go#L429)).

For all of the options not under `config` in the genesis block, like gasLimit, difficulty, etc. should I make those all required fields?  I'm thinking extraData is the only optional one.

jangko: Currently fork selection mechanism is still in [primitive shape](https://github.com/status-im/nimbus/blob/3c336789494ea4594f405d55c71a8f9b5934fa7b/nimbus/vm/interpreter/vm_forks.nim). What happened in `config.nim` not always transferred to the next part.

acolytec3: @kdeme I've reworked all the `try...except...` into `if...hasKey...else...quit(1)` and also made all the forkblocks optional as long as any given forkblock isn't set to a block number less than the one before it.

Is this more what you were thinking?

acolytec3:
@zah @kdeme @jangko I've made some progress in the latest commit.  I've added in additional parsing of additional values (mixhash, coinbase, timestamp) that are included in the genesis block config when you create a new custom genesis block from puppeth.  Geth will now accept an incoming connection from Nimbus using that same custom genesis block created by puppeth.  Now, it's throwing an `rlpError`.  So, progress, but still not 100% working.

acolytec3:
@kdeme @zah @jangko Now that this is syncing with geth (when a proper `genesis.json` is provided) are there other changes you'd like to see?  I've squashed and it's ready for final review as far as I know.

nimbus/config.nim
```diff
@@ -436,6 +639,28 @@ proc processNetArguments(key, value: string): ConfigStatus =
...
+    if value == "":
+      error "No genesis block config provided for custom network", network=key
+      result = ErrorParseOption
+    else:
+      try:
+        var (dir, file, ext) = splitFile(value)
+        if dir != "":
+          setCurrentDir(dir)
+        result = processCustomGenesisConfig(parseFile(file & ext))
+        config.net.networkId = 0
```
jangko: can we move set network id from `nimbus.nim` to here?

nimbus/config.nim
```diff
@@ -190,13 +227,179 @@ proc publicChainConfig*(id: PublicNetwork): ChainConfig =
...
+    nonce:            nonce,
+    extraData:        extraData,
+    gasLimit:         gasLimit,
+    difficulty:       difficulty,
+    prealloc:         alloc,
+    mixHash:          mixHash,
+    coinbase:         coinbase,
+    timestamp:        timestamp
+  )
+  config.customGenesis = cg
```
jangko:
```Nim 
config.customGenesis = CustomGenesisConfig(
...
)
```
will be fine

nimbus/config.nim
```diff
@@ -190,13 +227,179 @@ proc publicChainConfig*(id: PublicNetwork): ChainConfig =
...
+    if chainDetails.hasKey("byzantiumBlock"):
+      if chainDetails["byzantiumBlock"].kind == JInt:
+        byzantiumBlock = chainDetails["byzantiumBlock"].getInt().toBlockNumber
+        if byzantiumBlock < eip158Block:
+          error "Forks can't be assigned out of order"
+          quit(1)
+      else:
+        error "Invalid value provided for byzantiumBlock"
+        quit(1)
+    if chainDetails.hasKey("constantinopleBlock"):
```
jangko: repetitive code like this can be simplified using template

```Nim
template templateExample(c, a, b: untyped) =
  let ablock = a.astToStr()
  if c.hasKey(ablock):
    if c[ablock].kind == JInt:
      a = c[ablock].getInt().toBlockNumber
      if a < b:
        error "Forks can't be assigned out of order"
        quit(1)
    else:
      error "Invalid value provided for " & ablock
      quit(1)

templateExample(chainDetails, constantinopleBlock, byzantiumBlock)
```

nimbus/config.nim
```diff
@@ -190,13 +227,179 @@ proc publicChainConfig*(id: PublicNetwork): ChainConfig =
...
+  if customGenesis.hasKey("gasLimit"):
+    if customGenesis["gasLimit"].kind == JString:
+      gasLimit = fromHex[int64](customGenesis["gasLimit"].getStr())
+    else:
+      error "Invalid value provided for gasLimit"
+      quit(1)
+  else:
+    error "No gas limit value found in genesis block."
+    quit(1)
+  if customGenesis.hasKey("difficulty"):
```
jangko: repetitive code like this can be simplified using template and generics too.
```Nim
proc parseExample(T: type, c: JsonNode, field: string): T =
  when T is string:
    c[field].getStr()    
  elif T is Uint256:
    parseHexInt(c[xstr].getStr()).u256
  elif T is ...:
  else:  
    ...
    
template templateExample(c, x, jtype: untyped, checkError: static[bool] = true) =
  let xstr = x.astToStr()
  if c.hasKey(xstr):
    if c[xstr].kind == jtype:
      x = parseExample(Uint256, c, xstr)
    else:
      error "Invalid value provided for " & xstr
      quit(1)
  else:
    when checkError:
      error "No difficulty value found in genesis block."
      quit(1)

templateExample(customGenesis, difficulty, JString)
```

nimbus/nimbus.nim
```diff
@@ -106,14 +106,19 @@ proc start() =
...
-  let chainDB = newBaseChainDB(trieDB,
+  var chainDB = newBaseChainDB(trieDB,
     conf.prune == PruneMode.Full,
     conf.net.networkId.toPublicNetwork())
 
   if canonicalHeadHashKey().toOpenArray notin trieDB:
     initializeEmptyDb(chainDb)
     doAssert(canonicalHeadHashKey().toOpenArray in trieDB)
 
+  # Set corret network ID in case of custom genesis block
```
jangko: can we move this set network id into `config.nim`?

nimbus/config.nim
```diff
@@ -190,13 +227,179 @@ proc publicChainConfig*(id: PublicNetwork): ChainConfig =
...
     doAssert(false, "No chain config for " & $id)
     ChainConfig()
 
   result.chainId = uint(id)
 
+proc processCustomGenesisConfig(customGenesis: JsonNode): ConfigStatus =
+  ## Parses Custom Genesis Block config options when customnetwork option provided
+  let config = getConfiguration() 
+  result = Success
+  var chainId = 0.uint
```
kdeme: Minor remark: In Nim you don't necessarily need to repeat `var` (or `let`, `const`, ...), so this can become:

```
var
  chainId = 0.uint
  homesteadBlock, daoForkblock, eip150Block, eip155Block, eip158Block, byzantiumBlock, constantinopleBlock = 0.toBlockNumber
  eip150Hash : MDigest[256]
  ...
```

nimbus/config.nim
acolytec3: Will be fixed in next commit.

nimbus/nimbus.nim
acolytec3: Yep, will be fixed in next commit.  I have it working locally.

nimbus/config.nim
acolytec3: Yes, per below.

nimbus/config.nim
acolytec3: Got it, will update in next commit.

nimbus/config.nim
acolytec3: I've got this piece working with one change.  I can't attach the `& ablock` piece to the `error...` call for nim-chronicles.  The compiler throws a `Error: 'error' can only be used in compile-time context
` message.  Any idea what this error means?  If I remove the `ablock` from the string passed to `error`, everything compiles fine.

nimbus/config.nim
acolytec3: I don't know if this one can be so cleanly broken down into a template/proc solution like the forks section of the config.  In the code, you have to typecast `JString`s into a `Hash256`, a `UInt256`, an `EthAddress`, and a Unix timestamp.  In your example, I still have to check the key in order to specify the type of value parsing/type-casting to do with the value assigned to that key, not just whether it's a `JString` or a `JInt`, etc..  I could rework this as JSON marshalling if preferred but I don't see how putting the value parsing into a template or private proc will make the code any cleaner for the pieces of the genesis config that aren't tied to the forks.

nimbus/config.nim
jangko: nim-chronicles is using macros voodoo. the correct syntax should be:
* error "some message", expr, expr, ...
* error "some message", someident = expr, someident = expr, ...

`expr` is anything stringable. `someident` is any valid Nim identifier to give the `expr` a meaningful name.

in your case it can be written like:
```Nim
error "Invalid value provided for", blockname = ablock
```

nimbus/config.nim
jangko: the template/generics solution only a suggestion. if  you have other better solution don't hesitate to use it. the idea is to reduce code repetition and let the compiler do the hard work.
I know my example is a bit broken. You can fix it by passing the final expected type to the template too and then feed it to the parser generic proc.

```Nim
template templateExample(c, x, jtype, T: untyped, checkError: static[bool] = true) =
...
x = parseExample(T, c, xstr)
...

templateExample(customGenesis, difficulty, JString, Uint256)
```

acolytec3: Thanks. This template stuff is all new to me so will have another go at it.

acolytec3: Thanks for the patience.  I've reworked everything along the guidelines provided and it looks like it is all working.  Is this what you all have in mind?

nimbus/config.nim
acolytec3: Thanks for the clarification.  I've revised accordingly.

jangko:
LGTM

nimbus/config.nim
```diff
@@ -436,6 +593,27 @@ proc processNetArguments(key, value: string): ConfigStatus =
...
     config.net.setNetwork(KovanNet)
+  elif skey == "customnetwork":
+    if value == "":
+      error "No genesis block config provided for custom network", network=key
+      result = ErrorParseOption
+    else:
+      try:
+        var (dir, file, ext) = splitFile(value)
+        if dir != "":
+          setCurrentDir(dir)
```
jangko: Sorry, I just realized calling `setCurrentDir` in `config.nim` could potentially confuse the database backend. 
Do we really need to use `splitFile` and `setCurrentDir` here instead of only using `parseFile`?
If we really need to use `setCurrentDir`, please revert it to previous current dir.

acolytec3: We only need to use the `setCurrentDir` if we want to support placing the genesis json file in any arbitrary path.  Otherwise, the json file has to be in the same directory as where you run nimbus from.  In other words, if we use `parseFile`, assume that `genesis.json` is in the user's home directory, you have to run nimbus from the user's home directory.

So, something like `~/nimbus/build/nimbus --customNetwork=genesis.json`.

I don't know that I have an opinion one way or the other as we can just document usage clearly in the notes if we want to keep it to `parseFile` and not deal with changing the directory around.

jangko: `parseFile` also works with arbitrary path, no?

acolytec3:
Let me experiment some more. I tried it a couple of times but it gave an
invalid file name error as I recall.

nimbus/config.nim
acolytec3: I'm not sure what I was doing before, but it does work as long as you specify the complete path to the config file.  It throws a `file not found` exception if you do something like `~/path/to/file` but I just tried `/home/username/path/to/file.json` and it worked fine so I've reverted the changes.

PR #444, opened 2020-01-20
Prepare EVMC integration

jangko:
summary of this PR:
* because we already departed from python-evm style, there is no point to keep 'BaseComputation' name, simplified to 'Computation'.
* removes more redundant codes from `Computation`.
* `origin`, `gasPrice`, `fork`, and `gasCosts` not replicated anymore each time we spawn new `Computation`. especially `gasCost` is a 4KB fat table.
* fixes long standing bug `create2nocash` along with 3 other bugs related to coinbase suicide.

note: EVMC integration already taken place in 'simplify_computation' branch.

### DIFF
--- a/nimbus/vm_state.nim
+++ b/nimbus/vm_state.nim
@@ -6,8 +6,9 @@
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
 import
-  macros, strformat, tables, sets,
+  macros, strformat, tables, sets, options,
   eth/common,
+  vm/interpreter/[vm_forks, gas_costs],
   ./constants, ./db/[db_chain, state_db],
   ./utils, json, vm_types, vm/transaction_tracer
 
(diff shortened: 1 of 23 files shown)


nimbus/vm_state.nim
```diff
@@ -35,13 +36,25 @@ proc init*(self: BaseVMState, prevStateRoot: Hash256, header: BlockHeader,
...
   self.accountDb = newAccountStateDB(chainDB.db, prevStateRoot, chainDB.pruneTrie)
-  self.touchedAccounts = initHashSet[EthAddress]()  
+  self.touchedAccounts = initHashSet[EthAddress]()
 
 proc newBaseVMState*(prevStateRoot: Hash256, header: BlockHeader,
                      chainDB: BaseChainDB, tracerFlags: set[TracerFlags] = {}): BaseVMState =
   new result
   result.init(prevStateRoot, header, chainDB, tracerFlags)
 
+proc txContext*(vmState: BaseVMState, origin: EthAddress, gasPrice: GasInt, forkOverride=none(Fork)) =
```
zah: This proc is named as a noun, which suggests that it returns something.
A better name would be `setupTxContext`.

PR #450, opened 2020-02-05
New nim

jangko:


stefantalpalaru:
You  can push more commits to the "nim1.0.6" branch directly.

jangko:
that's right, this was an 'accident'

PR #453, opened 2020-02-11
fixes modexp 'gasFee' bug

jangko:


zah:
This needs to be rebased now

arnetheduck:
what are the rest of the skipped tests?

jangko:
| what are the rest of the skipped tests?

Most of them are recursive stress test that consume a lot of time but actually pass the test.
3 are related to EIP2200, with high suspicion the bug is in the test data and not in Nimbus EVM.
still looking for a proof whether this assumption is correct or not.

PR #455, opened 2020-02-13
enable -march=native by default

stefantalpalaru:
plus some drive-by fixes and cosmetic changes

### DIFF
--- a/nimbus.nimble
+++ b/nimbus.nimble
@@ -10,14 +10,15 @@ skipDirs      = @["tests", "examples"]
 # bin           = @["build/nimbus"]
 
 requires "nim >= 0.19",
-         "chronicles",
-         "nimcrypto",
-         "stint",
-         "json_rpc",
-         "chronos",
-         "bncurve",
-         "eth",
-         "stew"
+  "bncurve",
+  "chronicles",
+  "chronos",
+  "eth",
+  "json_rpc",
+  "libbacktrace",
+  "nimcrypto",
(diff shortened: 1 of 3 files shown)


nimbus.nimble
```diff
@@ -26,7 +27,7 @@ proc buildBinary(name: string, srcDir = "./", params = "", lang = "c") =
   var extra_params = params
   for i in 2..<paramCount():
     extra_params &= " " & paramStr(i)
-  exec "nim " & lang & " --out:build/" & name & " " & extra_params & " " & srcDir & name & ".nim"
+  exec "nim " & lang & " --out:build/" & name & " -d:release --import:libbacktrace " & extra_params & " " & srcDir & name & ".nim"
```
zah: The parameters which always stay the same (such as `--import:libbacktrace`) should be preferably specified in cfg files.
This ensures that manual builds from the command line and in scripts such as `tests/simulation/start.sh` use them as well.

stefantalpalaru: Tried it and it doesn't work. I documented that here: https://github.com/status-im/nimbus/blob/e9e33d3a56f240b7c31f3e093fac716defff4b7c/config.nims#L31

zah: I see. Let's make sure that `tests/simulation/start.sh` uses the new flags as well then.

stefantalpalaru: Yes, looking into that.

PR #459, opened 2020-02-19
implement aleth/geth/parity compatibility mode -- 100% pass test

jangko:
## aleth/geth/parity compatibility mode:

affected test cases both in GST and BCT:
- stSStoreTest\InitCollision.json
- stRevertTest\RevertInCreateInInit.json
- stCreate2\RevertInCreateInInitCreate2.json

pyEVM sided with original Nimbus EVM

implementation difference:
Aleth/geth/parity using accounts cache. When contract creation happened on an existing but 'empty' account with non empty storage will get new empty storage root. Aleth cs. only clear the storage cache while both pyEVM and Nimbus will modify the state trie. 
During the next SSTORE call, aleth cs. calculate gas used based on this cached 'original storage value'. In other hand pyEVM and Nimbus will fetch 'original storage value' from persisted state trie.

Both Yellow Paper and EIP2200 are not clear about this  situation but since aleth/geth/and parity implement this behavior, we perhaps also need to implement it.

initially, parity also does not support this behavior. see paritytech/parity-ethereum#10065

Definitions of terms are as below(from EIP1283):

* Storage slot's original value: This is the value of the storage if a _reversion_ happens on the *current transaction*.
* Storage slot's current value: This is the value of the storage before SSTORE operation happens.
* Storage slot's new value: This is the value of the storage after SSTORE operation happens.

aleth/geth/parity revert to empty cache.
pyEVM and Nimbus revert to 'original state trie'.

### DIFF
--- a/nimbus/db/state_db.nim
+++ b/nimbus/db/state_db.nim
@@ -8,16 +8,49 @@
 import
   strformat,
   chronicles, eth/[common, rlp], eth/trie/[hexary, db, trie_defs],
-  ../constants, ../utils, storage_types
+  ../constants, ../utils, storage_types, sets
 
 logScope:
   topics = "state_db"
+
+# aleth/geth/parity compatibility mode:
+#
+# affected test cases both in GST and BCT:
+# - stSStoreTest\InitCollision.json
+# - stRevertTest\RevertInCreateInInit.json
+# - stCreate2\RevertInCreateInInitCreate2.json
+#
+# pyEVM sided with original Nimbus EVM
+#
+# implementation difference:
+# Aleth/geth/parity using accounts cache.
+# When contract creation happened on an existing
+# but 'empty' account with non empty storage will
+# get new empty storage root.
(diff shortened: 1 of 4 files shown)


zah:
I think we can get feedback from the py-evm dev team regarding this, but siding with Geth, Parity and Aleth seems like the safer bet. After all, nearly everyone running a node on mainnet is using one of those clients.

nimbus/db/state_db.nim
```diff
@@ -8,16 +8,49 @@
...
+# aleth/geth/parity compatibility mode:
+#
+# affected test cases both in GST and BCT:
+# - stSStoreTest\InitCollision.json
+# - stRevertTest\RevertInCreateInInit.json
+# - stCreate2\RevertInCreateInInitCreate2.json
+#
+# pyEVM sided with original Nimbus EVM
+#
+# implementation difference:
```
arnetheduck: +1 for documenting this here :100:

PR #478, opened 2020-03-25
Compile Nimbus with Nim 1.2

zah:


jangko:
need to bump confutils and deal with status-im/nim-confutils#14 issue.

PR #487, opened 2020-04-09
add goerli testnet genesis data[skip ci]

jangko:
requires status-im/nim-eth#217


zah:
Good for merging, but this should bump nim-eth, right?

PR #496, opened 2020-04-20
[WIP] Stateless client experiment: The Block Witness

jangko:
* [x] implement single proof witness builder.
* [x] implement single proof tree builder.
* [x] implement random roundtrip single proof test.
* [x] implement common-prefix-key witness builder, no changes to tree builder.
* [x] based on common-prefix-key, implement multiproof witness builder, no changes to tree builder.
* [x] implement multiproof(a group of common-prefix-key) roundtrip test.
* [x] implement multiproof for account storage too.
* [x] implement multiproof addresses container, along with EVM bytecode 'touched' flag and account storage multiproof storage slots/keys container.
* [x] ~ test buildForest roundtrip. ~ spec changed, no need to build forest anymore
* [x] implement random roundtrip common-prefix-key. one or more keys may be bad, return error.
* [x] implement EMV bytecode 'touched' or not flag.
* [x] implement and generate test data for malicious input, probably in json format.
* [ ] implement minimal fuzz-like test to generate malformed input.


stateless/witness_from_tree.nim
```diff
@@ -41,30 +135,55 @@ proc getBranchRecurseAux(db: DB, node: openArray[byte], path: NibblesSeq, output
...
                        "HexaryTrie node with an unexpected number of children")
 
-proc getBranchRecurse*(self: WitnessBuilder; key: openArray[byte]): string =
-  var node = self.db.get(self.root.data)
-  result.add node.toHex
-  getBranchRecurseAux(self.db, node, initNibbleRange(key), result)
+proc getBranchRecurse*(wb: var WitnessBuilder; key: openArray[byte]): seq[byte] =
+  var node = wb.db.get(wb.root.data)
+  getBranchRecurseAux(wb, node, initNibbleRange(key), 0)
+  shallowCopy(result, wb.output.getOutput(seq[byte]))
```
arnetheduck, line 181: unless there's a demonstrated and significant performance benefit, `shallowCopy` introduces referential safety issues - let's avoid it

stateless/tree_from_witness.nim
```diff
@@ -1,104 +1,242 @@
...
 
-proc treeNode(t: var TreeBuilder, depth: int = 0, accountMode = false) =
+template read(t: var TreeBuilder, len: int): auto =
+  let pos = t.pos
+  inc(t.pos, len)
+  toOpenArray(t.data, pos, pos + len - 1)
+]#
+
+proc readU32(t: var TreeBuilder): int =
+  # TODO: what if the value overflow int32.high?
```
arnetheduck, line 69: just return an `uint32` - specially considering the name of the function - then the calling code can verify that the length is reasonable and do the cast - otherwise, this introduces a way to crash

stateless/tree_from_witness.nim
```diff
@@ -1,104 +1,242 @@
...
+    if result != hash:
+      debugEcho "DEPTH: ", depth
+    doAssert(result == hash, "EXT HASH DIFF " & result.data.toHex & " vs. " & hash.data.toHex)
 
-proc accountNode(t: var TreeBuilder, depth: int) =
+proc accountNode(t: var TreeBuilder, depth: int): NodeKey =
   assert(depth < 65)
-  let nodeType = AccountType(t.readByte)
+  let len = t.readU32()
+  result = toNodeKey(t.read(len))
```
arnetheduck, line 214: where is the TreeBuilder data coming from? if someone supplies a malicious input, how much memory will be allocated here?

stateless/witness_from_tree.nim
jangko: thanks. not a problem at all.

stateless/tree_from_witness.nim
jangko: okay.

stateless/tree_from_witness.nim
jangko: from a 'fullnode' peer.

it's not part of the spec. this is my own hack to quickly test the first part of the implementation.
the next part not finished yet. but it will be replaced by a strict parsing rule combined with some limitation from several EIPs(EVM EIP, not block witness EIP).

arnetheduck:
it would be good if new modules were written with `{.push raises: [Defect].}` on top and explicitly listed what they raise so we don't expand on the untracked part of the codebase unnecessarily

stateless/tree_from_witness.nim
arnetheduck: the point here is though that someone could place a "bad" value in the length (2gb) and thus crash the reader - so every time the reader reads a length, it needs to sanity-check it against the length of the remaining bytes (it can't be longer than the number of bytes that remain in the buffer)

jangko: this code longer poses threat, it has become debugging code, not part of regular code anymore

jangko:
agree

jangko:
block witness implementation is done.
summary of this PR:
* witness builder capable to generate multiproof block witness.
* tree builder capable of parsing multiproof block witness.
* roundtrip test for both of them
* developers guide to multi keys algorithm used to generate multiproof.

the fuzz test will be added later after status-im/testutils#3

PR #527, opened 2020-08-26
make it compatible with Nim devel and future Nim 1.4

narimiran:
Based on my testing with `make USE_SYSTEM_NIM=1 test`, this is the only change required for `nimbus` to work with both the current stable version of Nim (1.2.x) and the latest devel version of Nim (1.3.x) which will become next stable version (1.4.x) at some point.


jangko:
so new `parseEnum` does not using `$` anymore in the comparison? well, it actually break more things than what you've fixed.
we also use `parseEnum` in nimbus cli, and it actually depends on the old behavior.

narimiran:
After some discussion, we'll try to fix it on our side.

PR #528, opened 2020-09-03
write `Fork` enum in a "modern way"

narimiran:
This provides the same functionality as `$` proc, but it keeps working
with Nim 1.3+, where `parseEnum` implementation has been changed to be
able to work with enums with holes (after a bugfix for them).

Note that the first character is case-sensitive and "Constantinople" !=
"constantinople".
Since the tests (`test_op_arith` and `test_op_bit`) use lower-case first
letter, the string representation is also changed to the lower-case.

### DIFF
--- a/nimbus/vm/interpreter/vm_forks.nim
+++ b/nimbus/vm/interpreter/vm_forks.nim
@@ -9,22 +9,11 @@ import stint, eth/common/eth_types
 
 type
   Fork* = enum
-    FkFrontier,
-    FkHomestead,
-    FkTangerine,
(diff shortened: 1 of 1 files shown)


narimiran:
After the https://github.com/status-im/nimbus/pull/527, we have tried to fix it on our side and the fix worked in your case (as currently written), but had some limitations that might pop-up later bringing us back to square one.
The change presented in this PR seems like the most practical fix.

jangko:
this compromise is quite acceptable.

PR #545, opened 2020-11-27
change case of `berlin`

narimiran:
In `vm_forks.nim`, `FkBerlin`'s value is (lowercase) `"berlin"`.

Another example of using lowercase in similar code:
`constantinople` in `test_op_bit.nim`


narimiran:
@stefantalpalaru I think this is the cause of the error I mentioned to you yesterday evening.

PR #546, opened 2020-12-02
Eip2537 implementation

jangko:


nimbus/vm/blscurve.nim
```diff
@@ -0,0 +1,279 @@
+import blscurve/bls_backend, stint
+
+when BLS_BACKEND == Miracl:
+  import blscurve/miracl/[common, milagro, hash_to_curve, bls_signature_scheme]
```
zah, line 4: We are trying to phase out Miracl in favour of BLST. Was it possible to use it here?

jangko: last time I compile it with newest upstream blst, the 32 bit version failed to compile.
that's why we still need miracl fallback.

jangko: but perhaps we also can try undocumented `_NO_ASM_` conditional compilation which is not activated by default bls_backend configuration.

PR #547, opened 2020-12-08
test_precompiles: fix enum case

stefantalpalaru:
This fixes compilation with Nim-1.4.2.

### DIFF
--- a/tests/test_precompiles.nim
+++ b/tests/test_precompiles.nim
@@ -65,7 +65,7 @@ template doTest(fixture: JsonNode, fork: Fork, address: PrecompileAddresses): un
 proc testFixture(fixtures: JsonNode, testStatusIMPL: var TestStatus) =
   let
     label = fixtures["func"].getStr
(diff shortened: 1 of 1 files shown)


tests/test_precompiles.nim
```diff
@@ -65,7 +65,7 @@ template doTest(fixture: JsonNode, fork: Fork, address: PrecompileAddresses): un
 proc testFixture(fixtures: JsonNode, testStatusIMPL: var TestStatus) =
   let
     label = fixtures["func"].getStr
-    fork  = parseEnum[Fork](fixtures["fork"].getStr)
+    fork  = parseEnum[Fork](fixtures["fork"].getStr.toLower)
```
jangko, line 68: Nim identifiers only allow ASCII subset chars, wouldn't it be more efficient if we use `toLowerAscii` instead of `toLower` in this case?

stefantalpalaru: Sure. Changed.

PR #568, opened 2021-03-22
generate docs (and other niceties ..) for nimbus sub-directory

mjfh:
why:
  * easy browsing of prototype docs, allows to follow links with
    a web browser on the local file system
  * some md & png files may contain additional documentation

overview:
  * separate nimbus/makefile, try "make -C nimbus" for instructions
  * running "make -C nimbus docs" will do the job
  * x-ref file in nimbus/docs/theindex.html
  * additional md and png files in nimbus/docs/ex/.. subdirectory

details:
  * a newer nim compiler provides better referencing when available, in
    particular the back link to the indices are not provided by the 1.2.10
    nim compiler (automatically handled by makefile)
  * make patterns are used to update files only when the timestamp changes
  * should provide "discount" markdown generator, otherwise fallback
    to <pre/> encapsulated text file


jangko:
can this makefile run on Windows, for example using msys2 bash+mingw32-make?

mjfh:
> can this makefile run on Windows, for example using msys2 bash+mingw32-make?
not tested it on mingw but it makes sense to check it there, will do and probably complement the makefile

mjfh:
> > can this makefile run on Windows, for example using msys2 bash+mingw32-make?
> > not tested it on mingw but it makes sense to check it there, will do and probably complement the makefile

It has definitely some path issues where NIM-doc wants windows-ish abs path. Update is ready, needs still some testing ..

jangko:
runs perfectly on my machine.

jangko (approved):

PR #570, opened 2021-03-23
[READY] Hive integration

jangko:
the `readyness` is for `rpc` and `consensus` category, not for `sync` or `p2p`.


jlokier:
As we found, this branch fails on a fresh checkout of Hive+Nimbus, following the instructions.
It needs a small change to make Hive use `hive_integration` branch of `nimbus-eth1`, and perhaps update the `README.md` to have the current list of which tests are expected to work.

Ironically, if this PR is merged, the right branch to use might change.  The right thing to do, then, depends on whether we'll continue to develop further on `hive_integration` branch.

PR #573, opened 2021-03-30
Feature/isolate evms

mjfh:
The directive "make test ENABLE_EVMC=1" fails exactly the same way here as it fails on the master branch


nimbus/vm/code_stream.nim
```diff
@@ -8,7 +8,7 @@
 import
   chronicles, strformat, strutils, sequtils, parseutils, sets, macros,
   eth/common,
-  ./interpreter/opcode_values
+  vm/interpreter/opcode_values
```
arnetheduck, line 11: we only use relative imports for various reasons - in general, it's good to look through https://status-im.github.io/nim-style-guide/ , specially before making style changes (which should in general head towards the guide)

mjfh: thank you, maybe i missed that particular section on the style guide here.

the reason for making source files relocatable is that it will be easier to re-factor and hide parts of the VM code (as partly discussed at the beginning of today's meeting)

jangko:
> The directive "make test ENABLE_EVMC=1" fails exactly the same way here as it fails on the master branch

sorry, I don't understand. what kind of failure?

mjfh:
> > The directive "make test ENABLE_EVMC=1" fails exactly the same way here as it fails on the master branch
> 
> sorry, I don't understand. what kind of failure?

compiling as "make test ENABLE_EVMC=1" crashes at test no. 7 with

     [..]
     [OK] tests/fixtures/eth_tests/GeneralStateTests/stCreate2/returndatacopy_afterFailing_create.json
     Segmentation fault
     subtest no: 7 failed: testgeneralstatejson

     [Suite] tracer json tests
       [OK] tests/fixtures/TracerTests/block48915.json
       [OK] tests/fixtures/TracerTests/block46147.json
     [..]

and can be reproduced as ./build/all_test 7, and terminates at test no. 24 as

     [...]
     [OK] 30_keys_0_storage.json

     [Suite] Misc test suite
       [OK] EthAddress to int
     /status/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/nimscript.nim(416, 18)
     /status/nimbus-eth1/nimbus.nimble(37, 8) testTask
     /status/nimbus-eth1/nimbus.nimble(34, 8) test
     /status/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/nimscript.nim(260, 7) exec
     /status/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/nimscript.nim(260, 7) Error: unhandled exception: FAILED: build/all_tests [OSError]
     make: *** [Makefile:112: test] Error 1

and this one will not be reprocuced running ./build/all_test 24

nimbus/vm/code_stream.nim
jlokier: > the reason for making source files relocatable is that it will be easier to re-factor and hide parts of the VM code (as partly discussed at the beginning of today's meeting)

It doesn't really improve relocation, imho. When relocating or renaming, you either need to leave behind a reexport file for existing users, or you have to update import paths everywhere, _except_ relative `import ./` for "internal" imports don't need to be changed.  Even something as basic as renaming the `vm` subtree to `evm`, then every `import vm/` would need updating.

Also my understanding is we use `import ./` to avoid ambiguity between files that are local to the current module, and files that belong to `vendor` submodules.

nimbus/vm/nvm_interpreter.nim
```diff
@@ -1,19 +1,131 @@
...
 # Licensed under either of
-#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or http://www.apache.org/licenses/LICENSE-2.0)
-#  * MIT license ([LICENSE-MIT](LICENSE-MIT) or http://opensource.org/licenses/MIT)
-# at your option. This file may not be copied, modified, or distributed except according to those terms.
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or
+#    http://www.apache.org/licenses/LICENSE-2.0)
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT) or
+#    http://opensource.org/licenses/MIT)
+# at your option. This file may not be copied, modified, or distributed except
+# according to those terms.
```
jlokier: I've noticed reformatting these copyright comments in a number of commits.  This seems unrelated to the actual change described in the commit - is it being committed by mistake?

jlokier: My main point is, if restyling is being done to copyright headers, it should not be mixed up with functional changes, including logical refactoring. It would make sense to have a separate PR that reformats all the copyright headers together, if that's the right thing to do.

Mixed them together effectively hides changes like that inside a PR, and it makes things a bit harder to review, cherry pick and revert later.  (It also means we end up with a random mixture.)

That's also style guide: Workflow, "One PR, one feature or fix".

A secondary point is, if you look at the rest of the Nimbus projects you'll see the 80-column rule is not applied to the copyright headers, and what was there before you reformatted is used in the majority of Nimbus files.
The headers aren't all the same, but there's only a few variations, and your formatting adds a new one.

jlokier:
> The directive "make test ENABLE_EVMC=1" fails exactly the same way here as it fails on the master branch

That's not a very helpful bug report, mainly because nobody knew about failures on the master branch.
But the details later are helpful.

> > sorry, I don't understand. what kind of failure?
> 
> compiling as "make test ENABLE_EVMC=1" crashes at test no. 7 with

No 7 is fine for me on the master branch.
However, no __9__ fails on master on Ubuntu Server 19.10 64-bit x86 with a SIGSEGV.
On MacOS, all the tests pass fine.
Reverting to before the last big submodule bump didn't change the result, so it's been around a while.

The final error message is a rather cryptic `Error: unhandled exception: FAILED: build/all_tests [OSError]`, for which the cause is a SIGSEGV a few thousand lines earlier.  I found it using `strace -f`.

I wondered why it's not getting caught in CI, and saw we don't have a Linux 64-bit EVMC CI setup.  Just Mac 64-bit (which works) and Linux 32-bit (which also works apparently).

As this is a bug on the master branch, this PR isn't the right place to discuss it further so I'll open an issue.

jangko:
> The final error message is a rather cryptic Error: unhandled exception: FAILED: build/all_tests [OSError],

Our usual configuration for testing nim project is `CI` -> `nimble+nimble script` -> `exec nim ...` -> `run the actual test`.
In nimbus-eth1 repo, the configuration is `CI` -> `make+makefile` -> `nim+nimscript` -> `exec nim ...` -> `"all_tests" calling itself with 0..24` -> `run the actual test`.

If any of the `exec nim...` fails, it will produce the above error message(produced by nimscript/nimblescript). you'll see it often and across many repos. this is an artifact we are  too lazy to remove.

jlokier:
This is just a heads up that I'm (at last) going to start deep work on the account/storage database (DB) called by the EVM, and shortly after that (the two things are related) the Sync process which calls the EVM. These changes are essential to have `nimbus-eth1` run usefully on mainnet. I will send out PRs for review and feedback in small pieces as it should be done, but the changes altogether are likely to touch many components.  In particular, the EVM, although I will try to keep this minimal.

The DB changes may have some effect on the EVM, because the DB and EVM are intimately related (the DB is what the EVM calls to get and store state and log events).  However, as the API between them can be expressed rather abstractly (EVMC for example), this might not require a much change.  The non-EVMC branch is more tightly coupled, and in fact switching over to something closer to EVMC all the time will be helpful for the DB changes.

- Change how the cache between EVM and DB operates, to hold more state in RAM across multiple EVM runs
- Pass more context about the EVM call to the DB so that its schema can use the extra context

The second stage of Sync changes require more surgery in the EVM. This is, eventually, the concurrent EVM calls, async-await thing we've been talking about, although we could still decide to run it in threads instead. However, for the first stage of Sync (concurrent sync algorithm), surgery to the EVM may or may not be required. Whether it is will depend on performance: Whether single EVM appears to be a significant networking and DB bottleneck when measured.

The DB and EVM are directly connected though the DB/EVM API, and the DB is going to accept some different parameters
Most likely the DB changes don't much affect the EVM due to the isolation provided by the DB/EVM API, of which EVMC is an example.  But if necessary, the non-EVMC branch might need updating to be more like EVMC.

nimbus/vm/state_transactions.nim
```diff
@@ -0,0 +1,153 @@
...
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT) or http://opensource.org/licenses/MIT)
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+import
+  options, sets,
+  eth/common, chronicles, ../db/accounts_cache,
+  ../transaction,
+  ./computation, ./interpreter, ./state, ./types
+
+proc validateTransaction*(vmState: BaseVMState, tx: Transaction, sender: EthAddress, fork: Fork): bool =
```
jangko, line 14: `validateTransaction` is not part of EVM, this can be grouped together with block validation.

nimbus/vm/state_transactions.nim
```diff
@@ -0,0 +1,153 @@
...
+
+  c.vmstate.status = c.isSuccess
+
+proc refundGas*(c: Computation, tx: Transaction, sender: EthAddress) =
+  let maxRefund = (tx.gasLimit - c.gasMeter.gasRemaining) div 2
+  c.gasMeter.returnGas min(c.getGasRefund(), maxRefund)
+  c.vmState.mutateStateDB:
+    db.addBalance(sender, c.gasMeter.gasRemaining.u256 * tx.gasPrice.u256)
+
+#[
```
jangko, line 92: all these commented methods also not part of EVM, can be grouped together with `validateTransaction`

jangko (changes requested):

jangko:
regarding the copyright header, although it should not be included in this PR, it is not our primary concern right now.

overall, this PR is a good step to isolate the EVM of it's own. 
I only ask to move `validateTransaction` and all those commented methods somewhere else from `state_transaction`.
after that changes, this PR can be merged.

jangko (approved):

PR #576, opened 2021-04-02
Bump nim-eth and nim-chronos and work-around exception effect

kdeme:
~~will require some changes in libnimbus still~~ done

### DIFF
--- a/nimbus/nimbus.nim
+++ b/nimbus/nimbus.nim
@@ -65,7 +65,9 @@ proc start(nimbus: NimbusNode) =
 
   # metrics logging
   if conf.debug.logMetrics:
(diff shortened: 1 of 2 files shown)


kdeme:
looks like failure https://travis-ci.org/github/status-im/nimbus-eth1/jobs/765654049#L296 is not related to this PR...?

jlokier:
The Travis failure.. might have something to do with this PR, as it doesn't happen with the immediately prior commit.  I can't see what the fault is, there's nothing useful in the error output, just an abrupt failure without explanation.  The prior commit is fine and carries on to run all the tests.

jlokier:
No, just a Travis random failure. If there was a cause, no way to see what from the messages.
Rebuilt again, passed the second time.

PR #578, opened 2021-04-08
Fix blockchain tests

jlokier:
Tests: Rename incorrect case `newBlockChainTests.md` file in repo

...This meant, on Linux, both files were created, test differences didn't make themselves visible in `git diff`, and the repo would not get updates with changed test output.  On Mac and Windows it worked.

---
Bugfix: Tests: It was skipping 4654 of the blockchain tests

...The "new block chain json tests" were being skipped on Linux, but silently so that CI didn't notice.  These are a significant part of the Ethereum test suite.  See the missing output from `make test`, also visible in CI logs for Linux targets (prior to this commit):

    [Suite] new block chain json tests
            <-- nothing here

### DIFF
--- a/tests/test_helpers.nim
+++ b/tests/test_helpers.nim
@@ -91,6 +91,8 @@ proc jsonTestImpl*(inputFolder, outputName: string, handler, skipTest: NimNode):
       if `skipTest`(last, name):
         continue
       filenames.add(filename)
+
+    doAssert(filenames.len > 0)
     for fname in filenames:
       test fname:
         {.gcsafe.}:
--- a/newBlockchainTests.md
+++ b/newBlockchainTests.md
@@ -1,6 +1,6 @@
 newBlockchainTests
 ===
-## BlockChainTests
+## BlockchainTests
 ```diff
 + randomStatetest391.json                                         OK
 ```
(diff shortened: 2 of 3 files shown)


tests/test_helpers.nim
```diff
@@ -91,6 +91,8 @@ proc jsonTestImpl*(inputFolder, outputName: string, handler, skipTest: NimNode):
       if `skipTest`(last, name):
         continue
       filenames.add(filename)
+
+    assert(filenames.len > 0)
```
jangko, line 95: the CI tests always run in release mode, better use `doAssert` here, because it will have no purpose in CI if using `assert`

jlokier: I tested the condition locally and the `assert` did trigger.  The `make test` runs in release mode too, like CI.

...Looking harder, that's because `-d:release` does not disable `assert` since Nim v0.20.0 back in 2019:
https://github.com/nim-lang/Nim/commit/721534119000c2bd53cc72b531726a6104381222

Unless the CI uses `-d:danger`, in which case we should fix the CI, it's fine.

stefantalpalaru: > Unless the CI uses -d:danger, in which case we should fix the CI, it's fine.

We've settled on always using `doAssert()` on nimbus-eth2. It's the one that's guaranteed to always run.

jlokier: More of an FYI really; I've changed the commit.
And to note that if there's any doubt that assertions or other run-time checks might be disabled in CI, that probably should be fixed in CI, so that it includes all reasonable checks.

PR #580, opened 2021-04-08
switch to chronos metrics, remove insecure

arnetheduck:
See https://github.com/status-im/nimbus-eth2/pull/2468

also fixes pcre linking for real, and adds some random build flags that
help nimbus-eth2 stay afloat

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,16 @@
-# Copyright (c) 2018-2020 Status Research & Development GmbH. Licensed under
+# Copyright (c) 2018-2021 Status Research & Development GmbH. Licensed under
 # either of:
 # - Apache License, version 2.0
 # - MIT license
 # at your option. This file may not be copied, modified, or distributed except
 # according to those terms.
 
-SHELL := bash # the shell used internally by Make
+SHELL := bash # the shell used internally by "make"
 
 # used inside the included makefiles
 BUILD_SYSTEM_DIR := vendor/nimbus-build-system
+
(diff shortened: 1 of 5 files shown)


stefantalpalaru (approved):

stefantalpalaru:
The Windows ICE is probably triggered by `-fno-omit-frame-pointer`:

https://sourceforge.net/p/mingw-w64/bugs/880/

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86593

arnetheduck:
> The Windows ICE is probably triggered by -fno-omit-frame-pointer:

odd, any idea why `nimbus-eth2` doesn't trigger it?

stefantalpalaru:
I'm thinking there's some mitigating factor, like LTO.

tersec:
Is there a reason the CIs still use a version of mingw-w64 which has this problem, as opposed to one which according to those two links doesn't, such as a mingw-w64 with gcc 9.x? For example, https://packages.ubuntu.com/focal/mingw-w64 from 20.04 LTS has gcc 9.3.

This doesn't seem like an end-user build configuration necessarily worth supporting or working around at this point.

stefantalpalaru:
Ah, the nimbus-eth2 solution was to downgrade MinGW-w64 to 6.3.0: https://github.com/status-im/nimbus-eth2/blob/8391d3f83e43e7e6d13b6bd716f06510e6ffd05c/.github/workflows/ci.yml#L182

PR #584, opened 2021-04-09
also use "-fno-omit-frame-pointer" on Windows

stefantalpalaru:
now that we no longer use the buggy gcc-8.1.0 in CI

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -34,12 +34,8 @@ else:
 
 # omitting frame pointers in nim breaks the GC
 # https://github.com/nim-lang/Nim/issues/10625
-if not defined(windows):
-  # ...except on windows where the option crashes GCC:
-  # https://sourceforge.net/p/mingw-w64/bugs/880/
-  # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86593
-  switch("passC", "-fno-omit-frame-pointer")
-  switch("passL", "-fno-omit-frame-pointer")
+switch("passC", "-fno-omit-frame-pointer")
+switch("passL", "-fno-omit-frame-pointer")
 
 --threads:on
 --opt:speed


arnetheduck:
is it possible to somehow blacklist gcc versions and give a reasonable error message?

stefantalpalaru:
> is it possible to somehow blacklist gcc versions and give a reasonable error message?

Yes, but blacklisting might be too harsh, in case one of the many GCC forks people use on Windows might happen to work at that particular version. We can always show a warning.

jlokier:
I think it's fine as long as it's a compiler crash (ICE) rather than faulty code generation. When there's faulty code generation that's when we should put in the effort to seriously blacklist or find a workaround.

Warning on a specific GCC version is probably not worth it, as figuring out the range(s) of compiler versions that crash is tricky and time consuming. Pinpointing a specific version (8.1.0) would suggest it's just this _one tricky version_, that only a few people will happen to land on. But it's not just one version, and we don't do the kind of range testing to find the range. We have a report of GCC 10.1.0 crashing too, also fixed by removing `-fno-omit-frame-pointer`. Doesn't matter that the underlying GCC bug may be a different one.

jlokier:
If this is sticking around, let's just generally recommend GCC >= 10.2.0 on Windows in build docs.

However, I'd rather we recognise `-fno-omit-frame-pointer` is a Glibc-specific workaround (and when Glibc is used on Windows it's needed there too), so gate it on Glibc. That said, even with Glibc it's not the best workaround, and there's no guarantee that it will always be enough. It's a pragmatic kludge to build the whole program differently, when only a single function needs a small tweak to fix https://github.com/nim-lang/Nim/issues/10625.

PR #594, opened 2021-04-23
Feature/re implement op handlers for vm2 updated

mjfh:
The last patch in this series sort of proves that the handle re-implementation (more a copy than nomething new) is correct and correctly assigned (fork x op-code)


nimbus/vm2/interpreter/op_handlers/oph_blockdata.nim
```diff
@@ -0,0 +1,182 @@
...
+
+import
+  ./oph_defs,
+  stint
+
+# ------------------------------------------------------------------------------
+# Kludge BEGIN
+# ------------------------------------------------------------------------------
+
+when not breakCircularDependency:
```
zah, line 27: So, what's the purpose of this exactly? I was hoping that this code will go away eventually and I wouldn't have to study and understand the problem.

nimbus/vm2/interpreter/op_handlers/oph_arithmetic.nim
```diff
@@ -0,0 +1,541 @@
...
+# ------------------------------------------------------------------------------
+# Kludge END
+# ------------------------------------------------------------------------------
+
+# ------------------------------------------------------------------------------
+# Private, op handlers implementation
+# ------------------------------------------------------------------------------
+
+const
+  addOp: Vm2OpFn = proc (k: Vm2Ctx) =
```
zah, line 79: To avoid problems with overly conservative effect types, please specify `gcsafe` and `raises: [Defect, VMError]` on the `Vm20pFn` type. This may require you to handle and remap other exceptions in various places in the code.

nimbus/vm2/interpreter/op_handlers/oph_blockdata.nim
mjfh: to be able to syntax check locally while re-factoring -- will be removed at the end

nimbus/vm2/interpreter/op_handlers/oph_arithmetic.nim
mjfh: ok, Vm2Ctx will be changed later anyway, will do that with Vm2OpFn as well (see oph_defs.nim)

nimbus/makefile
```diff
@@ -39,13 +39,18 @@ endif
 # Compat version is used with external NIM compiler
 NIM_COMPAT := --useVersion:1.2
 
-# Name of NIM compiler, test for newer version on host OS
+# Name of NIMDOC compiler, test for newer version on host OS
```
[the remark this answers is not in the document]
mjfh: should be NIMDOC as it is used for generating docs

nimbus/vm2/interpreter/op_handlers/oph_arithmetic.nim
mjfh: btw, it is declared gsave, i will add raises: [Defect, VMError]

nimbus/vm2/interpreter/op_handlers/oph_blockdata.nim
zah: But why does syntax checking require the kludge? Wasn't the goal of the whole refactoring to eliminate the problem?

nimbus/makefile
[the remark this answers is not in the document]
jlokier: @zah It may clarify that this is in a "private extra makefile" added by @mjfh, that's not really used by Nimbus e.g. to build Nimbus, or for testing, nor is it run in CI. This is not part of the regular Makefiile.

nimbus/vm2/interpreter/op_handlers/oph_blockdata.nim
jlokier: Same here. I would like to know if "syntax check locally" is supposed to still work after the refactoring and kludge is gone, otherwise what was the point?

nimbus/vm2/interpreter/op_handlers/oph_arithmetic.nim
jlokier: A comment on performance, whicih you might already have taken into account.

A proc which can raise (for example `VMError`) but doesn't usually raise is reasonably efficient.
The place which catches (`try...except`) is slowed down by the `try`, the way it's implemented in Nim.

Because op functions are called a lot in the intepreter, it's better for them to allow any exceptions raised inside the op to propagate out of the interpreter loop, _without_ transforming them into a special VM-specific error, because the `try` can then go outside the interpreter loop.

If an op function contains `try` inside the op, just to capture all errors (e.g. from database calls) and convert them to `VMError`, this will slow some ops.  Actually it will slow all ops, but of course some ops are already slow and not called often so it's fine in those.  But you probably shouldn't have `try` inside the most commonly called, "fast" ops.

mjfh: this was an inetmediate patch anyway :)

jlokier: I'm commenting on architectural direction, not a specific patch.  Particularly I'm commenting on the `{.raises: [Defect, VMError].}`, which suggests to me you are transforming exception types with `try` inside some ops.

I don't know what your eventual architecture is intended to be, so I just wanted you to know about the performance effect of `try` because it's not that well known, and it's especially relevant in a bytecode interpreter.

In my opinion there's nothing wrong with exceptions to communicate exceptional conditions in interpreter ops, and it would be marginally faster for failng transactions. So I wouldn't ask you to avoid them, just to understand their effect.

PR #597, opened 2021-04-26
Hive: Current nimbus-eth1 test results, and improve documentation

jlokier:


jangko:
there is an ambiguity here about what actually working and what not.
for example the `graphql` should not work at all because it is not enabled in the hive script.
on the other hand, the `rpc` actually works but we have bugs in nimbus, if `rpc` doesn't work at all then `consensus` should not works too because it depends on `rpc`.

perhaps we should make distinction between hive script bugs vs nimbus bugs.

PR #598, opened 2021-04-26
EVMC: Small stacks when using EVMC, closes #575 (segfaults)

### LINKED ISSUE #256: recursive EVM call trigger unrecoverable stack overflow
some GST test such as `CallRecursiveBomb0.json` will crash the test program without any stack trace, no output at all.

without any additional compiler switch, it will crash at EVM depth around 370. if I increase the stack size with `--passL:"-Wl,--stack,<size>"` and set the size to 5MB,10MB,20MB --> it will produce stack trace and print stack overflow error. But it will always fail around EVM depth 394 no matter how big the stack is.

EVM stack depth limit is 1024, and we are not even reach half of it. 

Are we using too much stack object? Any idea?

### LINKED ISSUE #575: EVM: Different segmentation faults when running the test suite with EVMC
@mjfh reported that `make test ENABLE_EVMC=1` fails. From https://github.com/status-im/nimbus-eth1/pull/573#issuecomment-810999399 @mjfh:

- compiling as `make test ENABLE_EVMC=1` crashes at test no. 7 with

       [..]
       [OK] tests/fixtures/eth_tests/GeneralStateTests/stCreate2/returndatacopy_afterFailing_create.json
       Segmentation fault
       subtest no: 7 failed: testgeneralstatejson

       [Suite] tracer json tests
         [OK] tests/fixtures/TracerTests/block48915.json
         [OK] tests/fixtures/TracerTests/block46147.json
       [..]

  and can be reproduced as `./build/all_test 7`, and terminates at test no. 24 as

       [...]
       [OK] 30_keys_0_storage.json

       [Suite] Misc test suite
         [OK] EthAddress to int
       /status/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/nimscript.nim(416, 18)
       /status/nimbus-eth1/nimbus.nimble(37, 8) testTask
       /status/nimbus-eth1/nimbus.nimble(34, 8) test
       /status/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/nimscript.nim(260, 7) exec
       /status/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/nimscript.nim(260, 7) Error: unhandled exception: FAILED: build/all_tests [OSError]
       make: *** [Makefile:112: test] Error 1

  and this one will not be reproduced running `./build/all_test 24`

Turns out it fails on different tests for different people, but not the targets in CI.  From https://github.com/status-im/nimbus-eth1/pull/573#issuecomment-811405065 @jlokier:

- No. 7 is fine for me on the master branch.
  However, no. 9 fails on master on Ubuntu Server 19.10 64-bit x86 with a SIGSEGV.
  On MacOS, all the tests pass fine.
  Reverting to before the last big submodule bump didn't change the result, so it's been around a while.

  The final error message is a rather cryptic `Error: unhandled exception: FAILED: build/all_tests [OSError]`, for which the cause is a SIGSEGV a few thousand lines earlier. I found it using `strace -f`.

  I wondered why it's not getting caught in CI, and saw we don't have a Linux 64-bit EVMC CI setup. Just Mac 64-bit (which works) and Linux 32-bit (which also works apparently).

jlokier:
This patch reduces stack space used with EVM in ENABLE_EVMC=1 mode, from 13 MB worst case to 550 kB, a 24x reduction.

This completes fixing the "stack problem" and closes #575 (`EVM: Different segmentation faults when running the test suite with EVMC`)

It also closes #256 (`recursive EVM call trigger unrecoverable stack overflow`).

After this patch, it is possible to re-enable the CI targets which had to be disabled due to #575.

This change is also a required precursor for switching over to "nearly EVMC" as the clean and focused Nimbus-internal API between EVM and sync/database processes, and is also key to the use of Chronos `async` in those processes when calling the EVM.

(The motivation is the internal interface has to be substantially changed _anyway_ for the parallel sync and database processes, and EVMC turns out to be well-designed and well-suited for this.  It provides good separation between modules, and suits our needs better than our other current interface.  Might as well use a good one designed by someone else.  EVMC is 98% done in Nimbus thanks to great work done before by @jangko, and we can use Nimbus-specific extensions where we need flexibility, including for performance.  Being aligned with the ecosystem is a useful bonus feature.)

All tests below were run on Ubuntu 20.04 LTS server, x86-64.  This matches one of the targets that has been disabled for a while in CI in EVMC mode due to stack overflow crashing the tests, so it's a good choice.

Measurements before
===================

Testing commit `e76e0144 2021-04-22 11:29:42 +0700 add submodules: graphql and toml-serialization`.

    $ rm -f build/all_tests && make ENABLE_EVMC=1 test
    $ ulimit -S -s 16384 # Requires larger stack than default to avoid crash.
    $ ./build/all_tests 9 | tee tlog
    [Suite] persist block json tests
    ...
    Stack range 38416 depthHigh 3
    ...
    Stack range 13074720 depthHigh 1024
    [OK] tests/fixtures/PersistBlockTests/block1431916.json

These tests use 13.07 MB of stack to run, and so crash with the default stack limit on Ubuntu Server 20.04 (8MB).  Exactly 12768 bytes per EVM call stack frame.

    $ rm -f build/all_tests && make ENABLE_EVMC=1 test
    $ ulimit -S -s 16384 # Requires larger stack than default.
    $ ./build/all_tests 7 | tee tlog
        ...
    Stack range 14384 depthHigh 2
        ...
    Stack range 3495456 depthHigh 457
    [OK] tests/fixtures/eth_tests/GeneralStateTests/stRandom2/randomStatetest639.json
    ...
    Stack range 3709600 depthHigh 485
    [OK] tests/fixtures/eth_tests/GeneralStateTests/stRandom2/randomStatetest458.json
        ...
    Stack range 7831600 depthHigh 1024
    [OK] tests/fixtures/eth_tests/GeneralStateTests/stCreate2/Create2OnDepth1024.json

These tests use 7.83MB of stack to run.  About 7648 bytes per EVM call stack frame.  It _only just_ avoids crashing with the default Ubuntu Server stack limit of 8 MB.  However, it still crashes on Windows x86-64, which is why the Windows CI EVMC target is currently disabled.

On Linux where this passes, this is so borderline that it affects work and testing of the complex storage code, because that's called from the EVM.

Also, this greatly exceeds the default thread stack size.

Measurements after
==================

    $ rm -f build/all_tests && make ENABLE_EVMC=1 test
    $ ulimit -S -s 600 # Because we can!  600k stack.
    $ ./build/all_tests 9 | tee tlog
    [Suite] persist block json tests
    ...
    Stack range 1936 depthHigh 3
    ...
        Stack range 556272 depthHigh 1022
        Stack range 556512 depthHigh 1023
        Stack range 556816 depthHigh 1023
        Stack range 557056 depthHigh 1024
        Stack range 557360 depthHigh 1024
        [OK] tests/fixtures/PersistBlockTests/block1431916.json

    $ rm -f build/all_tests && make ENABLE_EVMC=1 test
    $ ulimit -S -s 600 # Because we can!  600k stack.
    $ ./build/all_tests 7 | tee tlog
        ...
    Stack range 1392 depthHigh 2
        ...
    Stack range 248912 depthHigh 457
    [OK] tests/fixtures/eth_tests/GeneralStateTests/stRandom2/randomStatetest639.json
    ...
    Stack range 264144 depthHigh 485
    [OK] tests/fixtures/eth_tests/GeneralStateTests/stRandom2/randomStatetest458.json
        ...
        Stack range 557360 depthHigh 1024
    [OK] tests/fixtures/eth_tests/GeneralStateTests/stStaticCall/static_CallRecursiveBombPreCall.json

For both tests, a satisfying *544 bytes* per EVM call stack frame, and EVM takes less than 600 kB total.  With other overheads, both tests run in 600 kB stack total at maximum EVM depth.

We must add some headroom on this for database activity called from the EVM, and different compile targets.  But it means the EVM itself is no longer a stack burden.

This is much smaller than the default thread stack size on Linux (2MB), with plenty of margin.  (Just fyi, it isn't smaller than a _small_ thread stack on Linux from a long time ago (128kB), and some small embedded C targets.)

This size is well suited to running EVMs in threads.

Further reduction
=================

This patch solves the stack problem.  Windows and Linux 64-bit EVMC CI targets can be re-enabled, and there is no longer a problem with stack usage.

We can reduce further to ~340 bytes per frame and 350 kB total, while still complying with EVMC.  But as this involves changing how errors are handled to comply fully with EVMC, and removing `dispose` calls, it's not worth doing now while there are other EVMC changes in progress that will have the same effect.

A Nimbus-specific extension will allow us to avoid recursion with EVMC anyway, bringing bytes per frame to zero.  We need the extension anyway, to support Chronos `async` which parallel transaction processing is built around.

Interop with non-Nimbus over EVMC won't let us avoid recursion, but then we can't control the stack frame size either.  To prevent stack overflow in interop I anticipate using (this method in Aleth)
[https://github.com/ethereum/aleth/blob/6e96ce34e3f131e2d42f3cb00741b54e05ab029d/libethereum/ExtVM.cpp#L61].

Smoke test other versions of GCC and Clang/LLVM
===============================================

As all builds including Windows use GCC or Apple's Clang/LLVM, this is just to verify we're in the right ballpark on all targets. I've only checked `x86_64` though, not 32-bit, and not ARM.

It's interesting to see GCC 10 uses less stack.  This is because it optimises `struct` returns better, sometimes skipping an intermediate copy.  Here it benefits the EVMC API, but I found GCC 10 also improves the larger stack usage of the rest of `nimbus-eth1` as well.

Apple clang 12.0.0 (clang-1200.0.26.2) on MacOS 10.15:

- 544 bytes per EVM call stack frame

GCC 10.3.0 (Ubuntu 10.3.0-1ubuntu1) on Ubuntu 21.04:

- 464 bytes per EVM call stack frame

GCC 10.2.0 (Ubuntu 10.2.0-5ubuntu1~20.04) on Ubuntu 20.04 LTS:

- 464 bytes per EVM call stack frame

GCC 11.0.1 20210417 (experimental; Ubuntu 11-20210417-1ubuntu1) on Ubuntu 21.04:

- 8 bytes per EVM call stack frame

GCC 9.3.0 (Ubuntu 9.3.0-17ubuntu1~20.04) on Ubuntu 20.04 LTS:

- 544 bytes per EVM call stack frame

GCC 8.4.0 (Ubuntu 8.4.0-3ubuntu2) on Ubuntu 20.04 LTS:

- 544 bytes per EVM call stack frame

GCC 7.5.0 (Ubuntu 7.5.0-6ubuntu2) on Ubuntu 20.04 LTS:

- 544 bytes per EVM call stack frame

GCC 9.2.1 20191008 (Ubuntu 9.2.1-9ubuntu2) on Ubuntu 19.10:

- 528 bytes per EVM call stack frame

### DIFF
--- a/nimbus.nimble
+++ b/nimbus.nimble
@@ -31,7 +31,15 @@ proc buildBinary(name: string, srcDir = "./", params = "", lang = "c") =
 
 proc test(name: string, lang = "c") =
   buildBinary name, "tests/", "-d:chronicles_log_level=ERROR"
-  exec "build/" & name
+  when not defined(windows):
+    # Verify stack usage is kept low by setting 1024k stack limit in tests.
+    exec "ulimit -s 1024 && build/" & name
+  else:
+    # Don't enforce stack limit in Windows, as we can't control it with these tools.
+    # See https://public-inbox.org/git/alpine.DEB.2.21.1.1709131448390.4132@virtualbox/
+    # When set by ulimit -s` in Bash, it's ignored.  Also, the command passed to
+    # NimScript `exec` on Windows is not a shell script.
+    exec "build/" & name
 
 task test, "Run tests":
   test "all_tests"
--- a/nimbus/vm/computation.nim
+++ b/nimbus/vm/computation.nim
@@ -291,46 +291,61 @@ proc afterExecCreate(c: Computation) =
   else:
     c.rollback()
 
-proc beforeExec(c: Computation): bool =
+proc beforeExec(c: Computation): bool {.noinline.} =
   if not c.msg.isCreate:
     c.beforeExecCall()
     false
   else:
     c.beforeExecCreate()
 
-proc afterExec(c: Computation) =
+proc afterExec(c: Computation) {.noinline.} =
   if not c.msg.isCreate:
     c.afterExecCall()
   else:
     c.afterExecCreate()
 
-template chainTo*(c, toChild: Computation, after: untyped) =
+template chainTo*(c: Computation, toChild: typeof(c.child), after: untyped) =
   c.child = toChild
   c.continuation = proc() =
     after
 
-proc execCallOrCreate*(cParam: Computation) =
-  var (c, before) = (cParam, true)
-  defer:
-    while not c.isNil:
-      c.dispose()
-      c = c.parent
-
-  # No actual recursion, but simulate recursion including before/after/dispose.
-  while true:
-    while true:
-      if before and c.beforeExec():
-        break
+when vm_use_recursion:
+  # Recursion with tiny stack frame per level.
+  proc execCallOrCreate*(c: Computation) =
+    defer: c.dispose()
+    if c.beforeExec():
+      return
+    c.executeOpcodes()
+    while not c.continuation.isNil:
+      when evmc_enabled:
+        c.res = c.host.call(c.child[])
+      else:
+        execCallOrCreate(c.child)
+      c.child = nil
       c.executeOpcodes()
-      if c.continuation.isNil:
-        c.afterExec()
+    c.afterExec()
+
+else:
+  # No actual recursion, but simulate recursion including before/after/dispose.
+  proc execCallOrCreate*(cParam: Computation) =
+    var (c, before) = (cParam, true)
+    defer:
+      while not c.isNil:
+        c.dispose()
+        c = c.parent
+    while true:
+      while true:
+        if before and c.beforeExec():
+          break
+        c.executeOpcodes()
+        if c.continuation.isNil:
+          c.afterExec()
+          break
+        (before, c.child, c, c.parent) = (true, nil.Computation, c.child, c)
+      if c.parent.isNil:
         break
-      (before, c.child, c, c.parent) = (true, nil.Computation, c.child, c)
-    if c.parent.isNil:
-      break
-    c.dispose()
-    (before, c.parent, c) = (false, nil.Computation, c.parent)
-    (c.continuation)()
+      c.dispose()
+      (before, c.parent, c) = (false, nil.Computation, c.parent)
 
 proc merge*(c, child: Computation) =
   c.logEntries.add child.logEntries
--- a/nimbus/vm/evmc_host.nim
+++ b/nimbus/vm/evmc_host.nim
@@ -123,7 +123,7 @@ proc hostEmitLogImpl(ctx: Computation, address: EthAddress,
   log.address = address
   ctx.addLogEntry(log)
 
-template createImpl(c: Computation, m: nimbus_message, res: nimbus_result) =
+proc enterCreateImpl(c: Computation, m: nimbus_message): Computation =
   # TODO: use evmc_message to evoid copy
   let childMsg = Message(
     kind: CallKind(m.kind),
@@ -133,10 +133,9 @@ template createImpl(c: Computation, m: nimbus_message, res: nimbus_result) =
     value: Uint256.fromEvmc(m.value),
     data: @(makeOpenArray(m.inputData, m.inputSize.int))
     )
+  return newComputation(c.vmState, childMsg, Uint256.fromEvmc(m.create2_salt))
 
-  let child = newComputation(c.vmState, childMsg, Uint256.fromEvmc(m.create2_salt))
-  child.execCallOrCreate()
-
+template leaveCreateImpl(c, child: Computation, res: nimbus_result) =
   if not child.shouldBurnGas:
     res.gas_left = child.gasMeter.gasRemaining
 
@@ -153,7 +152,7 @@ template createImpl(c: Computation, m: nimbus_message, res: nimbus_result) =
       copyMem(res.output_data, child.output[0].addr, child.output.len)
       res.release = hostReleaseResultImpl
 
-template callImpl(c: Computation, m: nimbus_message, res: nimbus_result) =
+template enterCallImpl(c: Computation, m: nimbus_message): Computation =
   let childMsg = Message(
     kind: CallKind(m.kind),
     depth: m.depth,
@@ -165,10 +164,9 @@ template callImpl(c: Computation, m: nimbus_message, res: nimbus_result) =
     data: @(makeOpenArray(m.inputData, m.inputSize.int)),
     flags: MsgFlags(m.flags)
     )
+  newComputation(c.vmState, childMsg)
 
-  let child = newComputation(c.vmState, childMsg)
-  child.execCallOrCreate()
-
+template leaveCallImpl(c, child: Computation, res: nimbus_result) =
   if not child.shouldBurnGas:
     res.gas_left = child.gasMeter.gasRemaining
 
@@ -185,11 +183,22 @@ template callImpl(c: Computation, m: nimbus_message, res: nimbus_result) =
     copyMem(res.output_data, child.output[0].addr, child.output.len)
     res.release = hostReleaseResultImpl
 
-proc hostCallImpl(ctx: Computation, msg: var nimbus_message): nimbus_result {.cdecl.} =
+proc enterHostCall(c: Computation, msg: var nimbus_message): Computation {.noinline.} =
   if msg.kind == EVMC_CREATE or msg.kind == EVMC_CREATE2:
-    createImpl(ctx, msg, result)
+    enterCreateImpl(c, msg)
   else:
-    callImpl(ctx, msg, result)
+    enterCallImpl(c, msg)
+
+proc leaveHostCall(c, child: Computation, kind: evmc_call_kind): nimbus_result {.noinline.} =
+  if kind == EVMC_CREATE or kind == EVMC_CREATE2:
+    leaveCreateImpl(c, child, result)
+  else:
+    leaveCallImpl(c, child, result)
+
+proc hostCallImpl(ctx: Computation, msg: var nimbus_message): nimbus_result {.cdecl.} =
+  let child = enterHostCall(ctx, msg)
+  child.execCallOrCreate()
+  leaveHostCall(ctx, child, msg.kind)
 
 proc initHostInterface(): evmc_host_interface =
   result.account_exists = cast[evmc_account_exists_fn](hostAccountExistsImpl)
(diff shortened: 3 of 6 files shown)


nimbus.nimble
```diff
@@ -31,7 +31,8 @@ proc buildBinary(name: string, srcDir = "./", params = "", lang = "c") =
 
 proc test(name: string, lang = "c") =
   buildBinary name, "tests/", "-d:chronicles_log_level=ERROR"
-  exec "build/" & name
+  # Validate stack usage stays reasonable by setting 1024k stack limit
+  exec "ulimit -s 1024 && build/" & name
```
jangko, line 35: I guess there is no `ulimit` on Windows? hence the CI failing.

jlokier: That's not the reason.  It fails CI because on Windows, NimScript `exec` doesn't run the string as a shell script. It's given directly to WIN32 `CreateProcess{A,W}`. On non-Windows it gives the string to `system` which runs a shell, so Nim is inconsistent about this (it could use a shell on Windows, or use `execve` on other targets).

There actually is a stack limit on Windows, and we can call out to `$SHELL` to run `ulimit`.  But due to anoher quirk, Windows Bash `ulimit` has no effect, it's [accepted but quietly ignored](https://public-inbox.org/git/alpine.DEB.2.21.1.1709131448390.4132@virtualbox/). So there's no point doing it.

There's another way to set the stack limit on Windows, but it's more complicated than a one-liner and I won't implement it for this patch. The Linux/Mac stack limit is enough to keep our code validated.

Updated patch skips setting the stack limit on Windows.

jangko: I don't remember how `exec` was implemented. But on Linux, Nim's [`startProcess`](https://github.com/nim-lang/Nim/issues/8913) also have it's own quirk.

It doesn't matter, we have more important things to do.

jlokier: On Windows, basically you can either edit a field in the executable to specify what the startup stack size should be, or create a thread on startup because you can set the stack size of a new thrad, and run everything else from that thread, with the initial thread just waiting for the second thread to finish.

jlokier: Nice, thanks.

nimbus/vm/computation.nim
```diff
@@ -291,46 +291,61 @@ proc afterExecCreate(c: Computation) =
...
 
-proc afterExec(c: Computation) =
+proc afterExec(c: Computation) {.noinline.} =
   if not c.msg.isCreate:
     c.afterExecCall()
   else:
     c.afterExecCreate()
 
-template chainTo*(c, toChild: Computation, after: untyped) =
+template chainTo*(c: Computation, toChild: typeof(c.child), after: untyped) =
```
zah, line 307: You have to be careful with templates that use their parameters more than once. Nim will replace each occurrence of the parameter within the template body with the passed in expression (potentially evaluating it multiple times; sometimes described as [call-by-name](https://en.wikipedia.org/wiki/Evaluation_strategy#Call_by_name)). This might be a problem when the passed-in expression has side effects.

You can avoid the problem here by assigning `c` to a local variable before using it in the rest of the template.

nimbus/vm/evmc_host.nim
```diff
@@ -133,10 +133,9 @@ template createImpl(c: Computation, m: nimbus_message, res: nimbus_result) =
     value: Uint256.fromEvmc(m.value),
     data: @(makeOpenArray(m.inputData, m.inputSize.int))
     )
+  return newComputation(c.vmState, childMsg, Uint256.fromEvmc(m.create2_salt))
 
-  let child = newComputation(c.vmState, childMsg, Uint256.fromEvmc(m.create2_salt))
-  child.execCallOrCreate()
-
+template leaveCreateImpl(c, child: Computation, res: nimbus_result) =
```
zah, line 138: Ditto

nimbus/vm/evmc_host.nim
```diff
@@ -153,7 +152,7 @@ template createImpl(c: Computation, m: nimbus_message, res: nimbus_result) =
       copyMem(res.output_data, child.output[0].addr, child.output.len)
       res.release = hostReleaseResultImpl
 
-template callImpl(c: Computation, m: nimbus_message, res: nimbus_result) =
+template enterCallImpl(c: Computation, m: nimbus_message): Computation =
```
zah, line 155: Ditto

nimbus/vm/evmc_host.nim
```diff
@@ -165,10 +164,9 @@ template callImpl(c: Computation, m: nimbus_message, res: nimbus_result) =
     data: @(makeOpenArray(m.inputData, m.inputSize.int)),
     flags: MsgFlags(m.flags)
     )
+  newComputation(c.vmState, childMsg)
 
-  let child = newComputation(c.vmState, childMsg)
-  child.execCallOrCreate()
-
+template leaveCallImpl(c, child: Computation, res: nimbus_result) =
```
zah, line 169: Ditto

nimbus/vm/interpreter/opcodes_impl.nim
```diff
@@ -646,7 +646,8 @@ template genCreate(callName: untyped, opCode: Op): untyped =
     c.gasMeter.consumeGas(createMsgGas, reason="CREATE")
 
     when evmc_enabled:
-      let msg = nimbus_message(
+      let msg = new(nimbus_message)
+      msg[] = nimbus_message(
```
zah, line 650: You can use the following short-cut syntax here:

```nim
let msg = (ref nimbus_message)(
     kind: ...
   )
```

... or define `nimbus_message_ref` as a type and then use it directly: `nimbus_message_ref(...)`

nimbus.nimble
jlokier: It's now implemented as you suggested.

nimbus/vm/interpreter/opcodes_impl.nim
jlokier: Thanks, that's good to know.

nimbus/vm/computation.nim
jlokier: Good point, thanks for pointing it out.

nimbus/vm/evmc_host.nim
jlokier: There's a mild dilemma with this one, as the goal was to be a minimal-size "obvious" change, so that anyone could clearly see what it does. The pre-existing templates were using `c` like this already, I just split the templates up, I didn't introduce any new multiple evaluation possibilities. There's no actual bug, because we can see clearly where it's used just a few lines later. But there's an increased risk of introducing one later by mistake, I agree. Performance can also suffer from multiple evaluations.

PR #602, opened 2021-04-27
CI: Skip GitHub actions for documentation-only changes

jlokier:
The CI takes a long time (sometimes hours) and shouldn't hold up documentation-only commits.  We may also find CI for other changes starts faster, if we're not using up as many credits.  When we process documentation into something nicer (like `nimbus.guide`) we'll need to revisit this.

Syntax from [example ignoring paths](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-paths).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,9 @@
 name: nimbus-eth1 CI
-on: [push, pull_request]
+on:
+  push:
+    paths-ignore: ['doc/**', 'docs/**', '**/*.md']
+  pull_request:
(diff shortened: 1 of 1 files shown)


jangko:
anything under `hive_integration` also not tested by CI, probably need to ignore it too

jlokier:
Good point, I'll add that one. Editing `hive_integration/README.md` was what motivated this CI change in the first place: That and GHA keeping actions queued for 3 hours this morning.

There are a few other files like those, but I don't want to invalidate CI by accidentally including something the build happens to use in a very obscure and unexpected way.  Had to think about whether the test result `.md` files are used.

PR #613, opened 2021-05-03
Gather EVM-calling functions to one place

jlokier:
Start gathering the functions that call the EVM into one place, `transaction/call_evm.nim`.

This is a series of changes to gather all ways the EVM is called to one place.  Duplicate, slightly different setup functions have accumulated over time, each with some knowledge of EVM internals.  When they are brought together, these methods will be changed to use a single entry point to the EVM, allowing the entry point to be refactored, EVMC to be completed, and async concurrency to be implemented on top.  This also simplifies the callers.


jangko (approved):
Thank you, this one should be merged ASAP

PR #623, opened 2021-05-04
Assembler: Make macro_assembler tests use new function asmCallEvm

jlokier:
Move the EVM setup and call in `macro_assembler` (`runVM`) entirely to new function `asmCallEvm` in `call_evm`.  Extra return values needed for testing are returned specially from `asmCallEvm`.

### DIFF
--- a/nimbus/transaction/call_evm.nim
+++ b/nimbus/transaction/call_evm.nim
@@ -206,3 +206,37 @@ proc txCallEvm*(tx: Transaction, sender: EthAddress, vmState: BaseVMState, fork:
     return tx.gasLimit
   txRefundGas(tx, sender, c)
   return tx.gasLimit - c.gasMeter.gasRemaining
+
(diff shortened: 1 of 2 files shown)


jangko:
the main purpose of macro_assembler is to test evm op codes. I never finished the gas usage part. it is safe to pick any other evm entrace. and we can remove asmCallEvm

jlokier:
None of the other calls return the information that macro_assembler checks for.  Specifically, macro_assembler checks stack, memory and trie match the expected results.  Those are not returned by any of the functions other than `asmCallEvm`, so it needs to use `asmCallEvm` for a while.  The same is true of the fixtures tests.

jlokier:
EVMC calls don't return stack or memory either, so running these tests via pure EVMC (e.g. to another EVM) won't be possible.  We might be able to use [EIP-3155](https://eips.ethereum.org/EIPS/eip-3155) eventually for this.

PR #625, opened 2021-05-04
Fix get recipient

jangko:


jlokier:
Nice, those duplicate sender calculations were annoying.
(I assume the GraphQL changes just got tangled up in the PR.)

As this change is a mixture of performance optimsiation and bugfix to some callers, it would be good to know which ones are bugfixes and which tests should now be passing that previously failed?  But in any case, it looks good to use..

jangko:
I forget to cherry pick the test case update

jlokier (approved):

PR #626, opened 2021-05-04
Fixes graphql ethapi

jangko:


nimbus/transaction/call_evm.nim
```diff
@@ -64,9 +64,8 @@ proc rpcDoCall*(call: RpcCallData, header: BlockHeader, chain: BaseChainDB): Hex
 proc rpcMakeCall*(call: RpcCallData, header: BlockHeader, chain: BaseChainDB): (string, GasInt, bool) =
   # TODO: handle revert
   var
-    # we use current header stateRoot, unlike block validation
-    # which use previous block stateRoot
-    vmState = newBaseVMState(header.stateRoot, header, chain)
+    parent  = chain.getBlockHeader(header.parentHash)
+    vmState = newBaseVMState(parent.stateRoot, header, chain)
```
jlokier, line 68: Is this just for GraphQL's `call`, or should RPC `call` also be using the parent block `stateRoot`?

If not, why is RPC different?

nimbus/graphql/ethapi.nim
```diff
@@ -141,25 +141,25 @@ proc getBlockByNumber(ctx: GraphqlContextRef, number: Node): RespResult =
   try:
     ok(headerNode(ctx, getBlockHeader(ctx.chainDB, toBlockNumber(number))))
   except EVMError as e:
```
jlokier, line 143: Why is this catching `EVMError`, when functions like `getBlockHeader` don't call the EVM at all and won't raise `EVMError`, but they do raise other errors?

(While here, why does `headerNode` use `cast[HeaderNode]` and `cast[Node]` instead of [inheritance](https://nim-lang.org/docs/tut2.html#object-oriented-programming-inheritance)?  It seems a bit type-unsafe, especially the `cast[HeaderNode](node)` with no check.)

nimbus/graphql/ethapi.nim
```diff
@@ -141,25 +141,25 @@ proc getBlockByNumber(ctx: GraphqlContextRef, number: Node): RespResult =
...
-    err("can't get block no '$1': $2" % [number.intVal, e.msg])
+    err(e.msg)
 
 proc getBlockByNumber(ctx: GraphqlContextRef, number: BlockNumber): RespResult =
   try:
     ok(headerNode(ctx, getBlockHeader(ctx.chainDB, number)))
   except EVMError as e:
```
jlokier, line 149: Ditto about `EVMError`.

nimbus/graphql/ethapi.nim
```diff
@@ -141,25 +141,25 @@ proc getBlockByNumber(ctx: GraphqlContextRef, number: Node): RespResult =
...
-    err("can't get block no '$1': $2" % [$number, e.msg])
+    err(e.msg)
 
 proc getBlockByHash(ctx: GraphqlContextRef, hash: Node): RespResult =
   try:
     ok(headerNode(ctx, getBlockHeader(ctx.chainDB, toHash(hash))))
   except EVMError as e:
```
jlokier, line 155: Ditto about `EVMError`.

nimbus/graphql/ethapi.nim
```diff
@@ -141,25 +141,25 @@ proc getBlockByNumber(ctx: GraphqlContextRef, number: Node): RespResult =
...
-    err("can't get block with hash '$1': $2" % [hash.stringVal, e.msg])
+    err(e.msg)
 
 proc getBlockByHash(ctx: GraphqlContextRef, hash: Hash256): RespResult =
   try:
     ok(headerNode(ctx, getBlockHeader(ctx.chainDB, hash)))
   except EVMError as e:
```
jlokier, line 161: Ditto about `EVMError`.

nimbus/graphql/ethapi.nim
jangko: there is two base error, EVMError and VMError. EVMError I think has the same meaning with EVMCHostError.

about the cast, I'm still looking for the best approach to this. we can use inheritance or use a UserData Node. Inheritance is quite expensive for AST Node.

jangko: Probably we need to rename the EVMError to something else.

jlokier: I see now, you are right to use `EVMError`.

The type has an unhelpful name, but it's the root of blockchain fetch errors: `BlockNotFound`, `ParentNotFound`, `CanonicalHeadNotFound` and `ValidationError`.

nimbus/transaction/call_evm.nim
jangko: I guess it should also using parent block stateRoot. I hope there is a test case for this in hive RPC category.

nimbus/graphql/ethapi.nim
jlokier: Though, checking for `EVMError` specifically and letting other `CatchableError` leak out looks... like it leaks exceptions into the query engine?.  If there's a different error raised from any functions called from GraphQL, probably they should also turn into an error node for the response?  I don't think we should assume `EVMError` is the only error that can be raised.

If these functions had `{.raises: [Defect].}` it would be clear that calls outside the GraphQL query engine were all turned into value/error nodes without exception leakage.

jlokier: About the cast, I was wondering if there's a reason it's not `type HeaderNode = ref object of Node` and `type Node = ref object of RootObj`.

The cast doesn't bother me if you think it's the right thing to do (or are still thinking about it)). But the `headerNode(node)` which casts `Node` -> `HeaderNode` without a run-time check looks like a safety risk.

jangko: > Though, checking for EVMError specifically and letting other CatchableError leak out looks

right, all catchable error should be turned into error code

> casts Node -> HeaderNode without a run-time check looks like a safety risk.

I will use inheritance. From usage pattern, it is easier and safer to use inheritance than UserData node

Issue #636, opened 2021-05-07
CatchableError static exception tracking pollited by base class Exception


mjfh:
Are there some idea, comments on better solution for the following problem?

# Objective

The OP handlers for the re-factored VM2 should be annotated with _raises: [Defect,EVMError]_

# Problem

Some _vendor_ modules leak potential base class _Exception_  exceptions, most annoyingly from the eth/trie modules. Also, _ref object_ assignment seems to flag an _Exception_ at compile time analysis (the nimbus style guide suggests an init() function.) See https://forum.nim-lang.org/t/6783#42459 for some discussion on that topic (I also realised that the _not nil_ object attribute seems to have gone which never worked too well anyway.)

# Partial Solution Applicable to Current Code

As a partial solution without extensive code changes, I implemented the following in the branch _feature/vm2-annotate-raises-exceptions_:
 *  _Exceptions_ are treated as uncatchable, only _CatchableError_ is processed in functions calling OP handlers
 * two compile modes, _debug_ and _normal_
    + in _debug_ mode, selected functions are wrapped so that the base class _Exception_ is relayed as a defect, OP handlers are annotated with _raises: [Defect,CatchableError]_
    + in _normal_ mode, OP handlers are annorated with _raises: [Exception]_ only

So this allows a static exception analysis for the _debug_ version with comparable exception run-time behavior. BTW, running the test suite does not result in much execution time penalties (observed ~0.6%.)

zah:
> ref object assignment seems to flag an Exception at compile time analysis

Can you provide a minimal example for this? I haven't stumbled on such a problem.

In general, you shouldn't be shy about introducing changes in the vendor packages that improve the exception tracking. The infectious general `raises: [Exception]` inferred effect is most often the result of non-annotated proc pointer types or methods. In our own libraries, we can easily fix such code by adding annotations. Third-party code, we can wrap using the `raiseAssert` tricks.

mjfh:
Example:
nimbus/config.nim(~900):
 ``` 
      proc getConfiguration*(): NimbusConfiguration =
         ## ...
         if isNil(nimbusConfig):
            nimbusConfig = initConfiguration()
         result = nimbusConfig
```
Playing with that assignment and with the _initConfiguration()_ implementation led me come to that conclusion.

jlokier:
> The OP handlers for the re-factored VM2 should be annotated with `raises: [Defect,EVMError]`

[...]

PR #637, opened 2021-05-07
Feature/vm2 annotate raises exceptions

mjfh:
please see issue #636

the gist of the subject is also commented at the last commit _track and annotate exceptions for VM2_

just see that the CI bails out in i386/linux, so there will be an update needed


jangko:
from windows CI:

```
D:\a\nimbus-eth1\nimbus-eth1\premix\hunter.nim(73, 29) Error: can raise an unlisted exception: Exception
```

this inconsistency probably came from nim stdlib, it's a known old story. some of them already fixed in Nim 1.4.x, but we are still using Nim 1.2.x

nimbus/config.nim
```diff
@@ -382,6 +382,9 @@ proc processCustomGenesisConfig(customGenesis: JsonNode): ConfigStatus =
     extraData = hexToSeqByte("0x")
     gasLimit = 16777216.int64
     difficulty = 1048576.u256
+
+    # note that the following parseJson() assignment implies a possible
+    # base class `Exception`
```
zah, line 387: When you work with a third-party function that has inappropriate raises list, you can usually fix the issue by wrapping it. If the raises list was incorrectly computed and the function is not expected to raise under normal circumstances, you can turn the general `Exception` type into a `Defect` through an assertion:

```nim
let
  alloc = try: parseJson("{}")
          except Exception: raiseAssert "Parsing an empty record should work"
```

If the function raises recoverable errors that are not derived from `CatchableError`, this should be filed as a bug upstream, but you can still remap the exception to a proper `CatchableError` type in a wrapper.

nimbus/genesis.nim
```diff
@@ -118,8 +121,11 @@ proc defaultGenesisBlockForNetwork*(id: PublicNetwork): Genesis =
   result.config = publicChainConfig(id)
 
 proc toBlock*(g: Genesis, db: BaseChainDB = nil): BlockHeader =
+
+  # note: newMemoryDB() assignment may throw an Exception exception
```
zah, line 125: When you start adding raises annotations, it's worth getting to the bottom of the problems.

nimbus/vm_compile_flags.nim
```diff
@@ -0,0 +1,58 @@
...
+  vm0_enabled* = not evmc0_enabled and not vm2_activated
+
+  vm0_activated* =              ## set if either naitve VM variant is activated
+    vm0_enabled or evmc0_enabled
+
+
+  evmc_enabled* =               ## set if either evmc is activated
+    defined(evmc_enabled) or defined(evmc2_enabled)
+
+  relay_exception_base_class* = ## map selected Exception tyype exceptions to
```
zah, line 45: I'm not sure I understand the rationale for having this mode. We should strive to have the correct raises annotations enabled by default and we don't need a mode that disables them.

nimbus/vm2/stack.nim
```diff
@@ -9,6 +9,8 @@ import
   chronicles, strformat, strutils, sequtils, macros, eth/common, nimcrypto,
   ../errors, ../validation
 
+{.push raises: [Defect,FullStack,InsufficientStack,ValueError,ValidationError].}
```
zah, line 12: It's better to have a very narrow raises list specified at the top-level - something that is really true for every function in the file (e.g. `raises: [Defect]` or `raises: [VMError, Defect]`). The functions that raise different exceptions can have explicit raises list specified on them.

kdeme:
I found some of those Exception causes in hexary while working on https://github.com/status-im/nim-eth/pull/348
I'll push some in a PR soon

kdeme:
@mjfh  https://github.com/status-im/nimbus-eth1/pull/644 should hopefully remove the raw `Exception`s from trie/hexary & trie/db

mjfh:
outdated, zombie

PR #654, opened 2021-05-13
fixes crappy custom genesis and chain config parser

### LINKED ISSUE #635: investigate hive/ethereum/consensus error message `Forks can't be assigned out of order`
`Forks can't be assigned out of order config.nim:330`

~This error explains why many of Istanbul test cases failing.~

TODO: add test cases for `--customnetwork` parser and probaly rewrite of crappy customnetwork parser.

Both `0` or `high(BlockNumber)` is not ideal default number for fork block number in customnetwork.

jangko:
instead of using stdlib/json, now we switch to json_serialization
the result is much tidier code and more robust when parsing
optional fields.

fixes #635


jlokier:
Nice code.  Very clear.

Smaller than the code it replaces too. Lots of deletions, often a good sign 🙂

Why does it fix so many Hive testcases.  Was it just the #635 `Forks can't be assigned out of order` errors causing all those failures, or has the new code fixed other things as well?

Where did you get the new JSON testcase files?  We have a lot of test files that are, in effect, like compiled binaries with no source code.   (E.g. because of EVM bytecode, etc).  So I often wonder where they came from, or how to generate them.

jangko:
> Why does it fix so many Hive testcases. Was it just the #635 Forks can't be assigned out of order errors causing all those failures, or has the new code fixed other things as well?

The old code although not produce `Forks can't be assigned out of order`, it will silently make wrong fork calculation. It also can't handle missing fork number very well.

The new code will handle any optional/missing fields. it will fill those missing fields with correct number to avoid wrong fork calculation.

> Where did you get the new JSON testcase files? 

I copy them from hive output.

PR #656, opened 2021-05-13
update genesis start block and

mjfh:
update conversion to start option argument inside hive container

### DIFF
--- a/hive_integration/nimbus/nimbus.sh
+++ b/hive_integration/nimbus/nimbus.sh
@@ -53,15 +53,11 @@ if [ "$HIVE_LOGLEVEL" != "" ]; then
   FLAGS="$FLAGS --log-level:DEBUG"
 fi
 
-# Configure the chain.
-mv /genesis.json /genesis-input.json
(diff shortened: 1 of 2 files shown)


hive_integration/nimbus/nimbus.sh
```diff
@@ -53,13 +53,10 @@ if [ "$HIVE_LOGLEVEL" != "" ]; then
...
-jq -f /mapper.jq /genesis-input.json > /genesis.json
-
-# Dump genesis
+# Configure the genesis chain and use it as start block
 echo "Supplied genesis state:"
-FLAGS="$FLAGS --customnetwork:/genesis.json"
+jq -f /mapper.jq /genesis.json > /genesis-start.json
+FLAGS="$FLAGS --customnetwork:/genesis-start.json"
 
 cat genesis.json
```
jangko, line 61: I think this `cat` needs update too.

mjfh: ok, i see

PR #657, opened 2021-05-14
update hive/ethereum/consensus stat

jangko:
from: ethereum/consensus:  16556 pass, 11643 fail, 28199 total
to  : ethereum/consensus:  27353 pass,   892 fail, 28245 total


jlokier:
This is an excellent result. Thank you.

PR #658, opened 2021-05-14
update conversion to start option argument inside hive container

mjfh:
why:
  not overloading genesis.json by converted version, using new file
  with edited contents


jangko (approved):
LGTM

Issue #666, opened 2021-05-18
hive: ethereum/consensus: need to port header validation code


jangko:
the number of failing cases in `ethereum/consensus` test suite dwindling to  59.
one case (bcHomesteadToEIP150/EIP150Transition_HomesteadToEIP150At5) identified to be caused by a subtle bug in `mapper.jq` file. (#667)
but majority of them related to block header validation.

sub category such as `bcTotalDifficultyTest` and `bcInvalidHeaderTest` clearly indicate we need to port those validation code 
from `test_blockchain_json` back to nimbus.

related procs in `test_blockchain_json`:
```nim
func validateGasLimit(gasLimit, parentGasLimit: GasInt)
proc validateHeader(header, parentHeader: BlockHeader, checkSeal: bool)
func validateUncle(currBlock, uncle, uncleParent: BlockHeader)
proc validateGasLimit(chainDB: BaseChainDB, header: BlockHeader)
proc validateUncles(chainDB: BaseChainDB, currBlock: EthBlock, checkSeal: bool)
func isGenesis(currBlock: EthBlock): bool
proc validateBlock(chainDB: BaseChainDB, currBlock: EthBlock, checkSeal: bool): bool
```

`bcMultiChainTest` sub category also pass `test_blockchain_json`, I suspect we also need to port some code back to nimbus but not sure which part.

after the ported code is stable enough, we can remove those in `test_blockchain_json`

jangko:
after #668, failing test cases reduced again from 59 to 44.
and we identified the next code need to port back to nimbus

https://github.com/status-im/nimbus-eth1/blob/2539bd9904096f15cfcb47bd6181db264bc49751/tests/test_blockchain_json.nim#L262

`validateBlockUnchanged` will at least fix 4 more test cases.

jangko:
#726 further reduce failing test cases down to 10 (Berlin, without London)

jangko:
thanks to @mjfh hard work, we can close this issue now. fixed by #727

PR #668, opened 2021-05-18
Extract p2p validates

mjfh:


jangko:
The goal of #666 is to reduce the failing cases in hive/ethereum/consensus. is there more part to come?
sorry if my explanation is not clear enough.

jangko (approved):
This is great. we can add more validation engine if they become available.

nimbus/utils/lru_cache.nim
```diff
@@ -0,0 +1,212 @@
...
+
+  tabItem.value = rcValue
+  cache.tab[key] = tabItem
+  result = ok(rcValue)
+
+# ------------------------------------------------------------------------------
+# Debugging/testing
+# ------------------------------------------------------------------------------
+
+when isMainModule and isMainOK:
```
zah, line 123: We try to avoid the practice of having tests that are local to the module and not executed in CI.

PR #680, opened 2021-05-25
LRU cache tests makeover

mjfh:
why:
  source-local unit tests would hardly be triggered by github CI as rightly
  criticised with the last patch.

details:
  source-local unit tests have been moved to tests folder.

  this version also contains rlp serialisation code so rlp encode/decode
  will apply tranparently. this is not needed in p2p/validate but will be
   useful with the clique protocol.


nimbus/utils/lru_cache.nim
```diff
@@ -11,201 +11,182 @@
...
+    tuple[prv, nxt: K, value: V]
+
+  # There could be {.rlpCustomSerialization.} annotation for the tab field.
+  # As there was a problem with the automatic Rlp serialisation for generic
+  # type, the easier solution was an all manual read()/append() for the whole
+  # generic LruCacheData[K,V] type.
+  LruData[K,V] = object
+    maxItems: int                 ## Max number of entries
+    first, last: K                ## Doubly linked item list queue
+    tab: TableRef[K,LruItem[K,V]] ## (`key`,encapsulated(`value`)) data table
```
jangko, line 53: out of curiosity, why the `Table` become `TableRef`?

mjfh: Thought it might be helpful when cloning LruData descriptors -- might not make a real difference though.

PR #681, opened 2021-05-26
add assertion guard in test_blockchain_json's parseBlock function

jangko:
we don't want any surprise because of unrecognized key in test fixture
classified as `hasException`.


jlokier (approved):
I believe we prefer `doAssert` in tests.

PR #689, opened 2021-05-27
Transaction: Don't calculate `contractAddress` redundantly

jlokier:
Each place in `call_evm` that sets up an EVM call calculates the new contract address for contract creations.  But it's redundant, because `newComputation` ignores the provided value and does the calculation again.

Remove the unused address calculation.

This is also a step to merging different entry points and EVMC.  This change ends up with the same value in both `msg.contractAddress` and `msg.codeAddress` for every entry point, and this is good because it matches the EVMC message structure, where they are replaced by only one value called `msg.destination`


jangko:
right, they are artifact from the past before contract address calculation in `newComputation` is implemented, before `create2` opcode in evm.

PR #690, opened 2021-05-27
Clear up meaning of `ZERO_ADDRESS`, delete `CREATE_CONTRACT_ADDRESS`

jlokier:
Updated constants + comments:

- There is no valid `CREATE_CONTRACT_ADDRESS`.  Some places on the internet say account zero means contract creation, but that's not correct.

  Transactions to `ZERO_ADDRESS` are legitimate transfers to that account, not contract creations.  They are used to "burn" Eth.  People also send Eth to address zero by accident, unrecoverably, due to poor user interface issues.

Updated test harness to clear a TODO:

- The conditions mentioned in the old TODO comment have been checked.  All fixtures have either 40 hex digits or empty string for "to".  There is a test with all-zeros, and it means send to that account, not contract creation. Empty string means contract creation.

  This patch does not change the relaxed parsing where fewer than 40 digits is accepted.  We should probably be stricter about this.

### DIFF
--- a/nimbus/constants.nim
+++ b/nimbus/constants.nim
@@ -10,11 +10,13 @@ const
   UINT_256_MAX*: UInt256 =                  high(UInt256)
   INT_256_MAX_AS_UINT256* =                 high(Uint256) shr 1
   UINT160CEILING*: UInt256 =                2.u256.pow(160)
+
+  # Transactions to ZERO_ADDRESS are legitimate transfers to that account, not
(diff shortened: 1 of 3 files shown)


jangko:
just a nitpick request. this kind of changes is true for new test cases, how about legacy test cases?
have you tried with `--legacy:true` for both `test_generalstate_json` and `test_blockchain_json`?
because right now `--legacy` switch only change the test folder, and have no special meaning to preserve any difference in test case format.

PR #695, opened 2021-05-28
CI: enable manual running

stefantalpalaru:


jlokier:
Since we think it needs to be on master to have any effect, sending it to master to test now.

PR #702, opened 2021-05-29
RPC and GraphQL: `call` and `estimateGas` ops through unified EVM runner 

jlokier:
- The following four flags are added, to change various steps in EVM processing when a call doesn't come from a real transaction:

  - `noIntrinsic`: Don't charge intrinsic gas.
  - `noAccessList`: Don't initialise EIP2929 access list.
  - `noGasCharge`: Don't charge sender account for gas.
  - `noRefund`: Don't apply gas refund/burn rule.

  This is to support RPC and GraphQL `call` operations, which behave differently in some ways from regular transaction calls, and to support some test suites.

- In EVMC terms, all these alterations can be performed on the host side.

- The RPC/GraphQL `call` operation differs in many ways from regular transaction calls.  The above four flags are set to `true`, to disable various steps in processing.

- The RPC/GraphQL `estimateGas` operation is much more like ordinary transaction execution than `call`, and does not use the above flags, though there are still enough differences from an ordinary transaction that tx validation cannot be used.

### DIFF
--- a/nimbus/transaction/call_common.nim
+++ b/nimbus/transaction/call_common.nim
@@ -27,6 +27,10 @@ type
     value*:        HostValue            # Value sent from sender to recipient.
     input*:        seq[byte]            # Input data.
     accessList*:   AccessList           # EIP-2930 (Berlin) tx access list.
+    noIntrinsic*:  bool                 # Don't charge intrinsic gas.
+    noAccessList*: bool                 # Don't initialise EIP-2929 access list.
+    noGasCharge*:  bool                 # Don't charge sender account for gas.
+    noRefund*:     bool                 # Don't apply gas refund/burn rule.
 
   # Standard call result.  (Some fields are beyond what EVMC can return,
   # and must only be used from tests because they will not always be set).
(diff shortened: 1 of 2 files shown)


jangko:
I have a hunch this four flags actually not needed for RPC or GraphQL. we only have not enough tests coverage for them both in our own test suite and hive.

but there is no harm to keep them around until we have more proofs to remove them. it is more important to progress in unifying the evm entry point.

so lets merge this one.

jlokier:
I agree, the goal here was to keep the exact same behaviours while implementing the unified entry point.  We can change behaviours in a separate PR, and probably should do it like that even if we believe the behaviour should change.

I think at least one flag may be needed, because judging from Geth's and Infura's documentation, RPC `call` and `estimateGas` don't behave quite the same. It is hard to be sure from the different wordings. And it looks likely we've had a long-time bug in JSON-RPC `call` where it uses the wrong state root (these patches avoid changing that). So we need to think about the correct behaviour, check what Geth does by testing or reading its code, and probably include the gas cap documented by Infura because it's a good idea anyway.

At least these flags make it clear what the behaviour differences are, and they can very easily be changed.

PR #705, opened 2021-06-01
Whisper: Disable and remove all Whisper (Shh) protocol support

jlokier:
Whisper is long obsolete.  It has evolved into Waku, and Whisper itself is no longer in development.
Remove it from `nimbus-eth1`.

If someone only wants to disable Whisper without removing it, the first patch can be cherry-picked instead.

(Note, there is still code in `nim-eth` for Whisper which this does not remove.)


arnetheduck:
we can remove it - Status shut down the whisper network already, and I haven't seen any interest elsewhere, except for replacements

kdeme (approved):

PR #714, opened 2021-06-15
Add nlpn

kdeme:
Add the beginnings of the portal client to nimbus-eth1, so far only holding parts of the Portal Wire protocol.

A separate Github Actions workflow is used so that code changed for nlpn does not trigger the full nimbus-eth1 test suite.

### DIFF
--- a/.github/workflows/nlpn.yml
+++ b/.github/workflows/nlpn.yml
@@ -1,12 +1,9 @@
-name: CI
+name: nlpn CI
 on:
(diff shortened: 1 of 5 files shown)


kdeme:
You can see in the 3 different commits, the different workflows that are run.

Downside is that common steps (e.g. checkout, building Nim, etc.) are ran twice in case of a common change (e.g. vendor modules update).

PR #719, opened 2021-06-17
verify age of uncle's parent

mjfh:
why:
  parent must be older => check needed for bcFrontierToHomestead test
  cases UncleFromFrontierInHomestead and UnclePopulation

### DIFF
--- a/hive_integration/nodocker/sim_utils.nim
+++ b/hive_integration/nodocker/sim_utils.nim
@@ -19,7 +19,9 @@ template runTest*(suiteName: string, caseFolder: string, body: untyped) =
   disableParamFiltering()
   suite suiteName:
     var status = initOrderedTable[string, OrderedTable[string, Status]]()
-    for fileName {.inject.} in walkDirRec(caseFolder):
+    for fileName {.inject.} in walkDirRec(
+                    caseFolder, yieldFilter = {pcFile,pcLinkToFile}):
+
       if not fileName.endsWith(".json"):
         continue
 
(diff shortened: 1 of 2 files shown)


hive_integration/nodocker/sim_utils.nim
```diff
@@ -19,7 +19,9 @@ template runTest*(suiteName: string, caseFolder: string, body: untyped) =
   disableParamFiltering()
   suite suiteName:
     var status = initOrderedTable[string, OrderedTable[string, Status]]()
-    for fileName {.inject.} in walkDirRec(caseFolder):
+    for fileName {.inject.} in walkDirRec(
+                    caseFolder, yieldFilter = {pcFile,pcLinkToFile}):
```
jangko, line 23: why we need `pcLinkToFile` here?

nimbus/p2p/validate.nim
```diff
@@ -249,6 +249,13 @@ proc validateUncles(chainDB: BaseChainDB; header: BlockHeader;
        (uncle.parentHash == header.parentHash):
       return err("Uncle's parent is not an ancestor")
 
+    # check uncle against own parent
+    var parent: BlockHeader
+    if not chainDB.getBlockHeader(uncle.parentHash,parent):
+      return err("Uncle's parent has gone missing")
+    if uncle.timestamp <= parent.timestamp:
```
jangko, line 256: LGTM

what I don't understand is: how `tests_blockchain_json` pass all tests without this comparison, a bug in `test_blockchain_json`?

jangko (approved):

hive_integration/nodocker/sim_utils.nim
mjfh: because it is easy enough to link to a subset of rest cases when running individual tests

nimbus/p2p/validate.nim
mjfh: apparently only a cople of tests violate that condition, reversing that inequality condition you can that about 40 more tests will fail

PR #722, opened 2021-06-21
Replace unittest with unittest2

jangko:


tests/asyncunit.nim
```diff
@@ -0,0 +1,32 @@
...
+
+    template teardown(teardownBody) {.used.} =
+      teardown:
+        let asyncproc = proc {.async.} = teardownBody
+        waitFor asyncproc()
+
+    let suiteproc = proc = body # Avoids GcUnsafe2 warnings with chronos
+    suiteproc()
+
+template test*(name, body) =
```
arnetheduck, line 29: elsewhere, these have been dubbed `asyncTest` to avoid confusion (and import issues)

jangko: right, thank you.

jlokier:
After this change, can test_rpc become part of all_tests without crashing?

jangko:
it works now when I test it locally. let's see if CI tests agree with me.

PR #726, opened 2021-06-21
Re-adjust canonical head to parent of block to be inserted

mjfh:
why:
  of the failing tests that remain to be solved, 30 of those will succeed
  if the canonical database chain head is cleverly adjusted -- yes, it
  looks like a hack, indeed.

details:
  at the moment, this hack works for the non-hive tests only and is
  triggered by a boolean argument passed on to the chain.persistBlocks()
  method.

### DIFF
--- a/nimbus/p2p/chain.nim
+++ b/nimbus/p2p/chain.nim
@@ -1,7 +1,17 @@
-import ../db/db_chain, eth/common, chronicles, ../vm_state,
-  stint, nimcrypto,
-  ../utils, eth/trie/db, ./executor, ../chain_config, ../genesis, ../utils,
-  stew/endians2, ./validate, ./validate/epoch_hash_cache
+import
+  ../chain_config,
+  ../db/db_chain,
+  ../genesis,
+  ../utils,
+  ../vm_state,
+  ./executor,
+  ./validate,
+  ./validate/epoch_hash_cache,
+  chronicles,
+  eth/[common, trie/db],
+  nimcrypto,
+  stew/endians2,
+  stint
 
 when not defined(release):
   import ../tracer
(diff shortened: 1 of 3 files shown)


nimbus/p2p/chain.nim
```diff
@@ -28,6 +38,44 @@ type
...
+
+      # Restore original header only if
+      #  * the restored chain would be the lingest (i.e. the bock number
+      #    of the restored head will be unique maximum number)
+      #  * or the sibling header on the to be restored chain, shadowed by the
+      #     argument header has higher difficulty.
+      result.sibling = c.db.getBlockHeader(thisNumber)
+      if thisNumber < canoNumber or
+         header.difficulty < result.sibling.difficulty:
+        result.needRestoreOk = true
```
jlokier, line 76: What is the reason for restoring to the longer chain even if the sibling difficulty is lower?  The true canonical chain would contain the new header in this case.

jlokier (commented):
Can you explain _why_ setting the canonical head to the parent of the newly imported block is useful, and not setting the head to the new block itself?  And why is it restored after each block import, only some of the time?

This adds an alternative path in a core sync function. It creates a strange. temporary state (possibly an invalid state too, as the canonical head state is no longer the actual canonical head). And then sometimes it restores it, sometimes not.

Although passing Hive tests is an important goal, it's meant to be done by using the normal code in the regular program, rather than special alternative paths that only run during tests and don't behave the same way in "real life".  (For example: [Volkswagen emissions scandal](https://en.wikipedia.org/wiki/Volkswagen_emissions_scandal) :) )

So I've suggested moving the alternative path logic out of the core sync function `persistBlocks` and into `importRlpBlock`.

I see that `forceCanonicalParent` may be necessary to setup some tests. It's not a hack if the tests require it, although best if the logic if it has an easily explained behaviour. It would be good to understand what it is really for ("30 [tests] will succeed" doesn't really explain).

nimbus/p2p/chain.nim
```diff
@@ -28,6 +38,44 @@ type
...
     extraValidation: bool
 
+  CanonicalState = object    ## temporary state cache
+    canoHead: BlockHeader    ## current canonical head
+    restoreHead: BlockHeader ## restore value
+    sibling: BlockHeader     ## block number reference of which to be replaced
+    needRestoreOk: bool      ## need to restore canonical head
+
+
+proc updateCanonicalHead(c: Chain; header: BlockHeader;
```
jangko, line 48: `test_blockchain_json` doesn't  update the canonical head. and `setHead` is a dangerous operation. `setHead` is only used by `debug_` rpc and by premix debugging tools.

I'd prefer we stick to validation only like what is `test_blockchain_json` doing. if test_blockchain_json can pass tests without `setHead`, we should be able to do the same here.

nimbus/p2p/chain.nim
```diff
@@ -28,6 +38,44 @@ type
...
+
+    # Check whether there is something to do, at all
+    if forceUpdateOk and parentHash != canoHash and thisNumber <= canoNumber:
+      # Set new caconical head and provide restore information.
+      result.restoreHead = result.canoHead
+      result.canoHead = c.db.getBlockHeader(parentHash)
+      c.db.setHead(result.canoHead)
+
+      # Restore original header only if
+      #  * the restored chain would be the lingest (i.e. the bock number
```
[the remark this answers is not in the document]
mjfh: you're right, 30 tests succeed is just a fact that will explain nothing :)

nimbus/p2p/chain.nim
mjfh: The main difference to the `test_blockchain_json` code is that `persistBlocks()` later verifies, that the inserted headert is now canonical head. Giving  that up, the parent of the block to be inserted can be used instead of the canonical head as done in `test_blockchain_json`.

nimbus/p2p/chain.nim
```diff
@@ -140,11 +151,15 @@ method persistBlocks*(c: Chain, headers: openarray[BlockHeader], bodies: openarr
...
+  trace "Persisting blocks",
+    fromBlock = headers[0].blockNumber,
+    toBlock = headers[^1].blockNumber
+
   for i in 0 ..< headers.len:
-    let head = c.db.getCanonicalHead()
-    let vmState = newBaseVMState(head.stateRoot, headers[i], c.db)
-    let validationResult = processBlock(c.db, headers[i], bodies[i], vmState)
+    let
+      head = c.db.getBlockHeader(headers[i].parentHash)
```
jangko, line 160: Good job!

nimbus/p2p/chain.nim
```diff
@@ -167,9 +182,12 @@ method persistBlocks*(c: Chain, headers: openarray[BlockHeader], bodies: openarr
         return ValidationResult.Error
 
     discard c.db.persistHeaderToDb(headers[i])
-    if c.db.getCanonicalHead().blockHash != headers[i].blockHash:
-      debug "Stored block header hash doesn't match declared hash"
-      return ValidationResult.Error
+    # The following check makes no sense anymore if the `head` is the parent
+    # was allowed to be inserted somewhere before the canonocal head.
+    #
+    # if c.db.getCanonicalHead().blockHash != headers[i].blockHash:
```
jangko, line 188: I agree this can be removed. After some cleanup, this PR can be merged.
And I guess the rest of failing 10 is easy for you @mjfh, they are mostly block header fields comparison.

jangko (approved):

nimbus/p2p/chain.nim
mjfh: ok, so I remove obsolete code + comment

PR #731, opened 2021-06-27
Preparation for EIP-1559

jangko:


jlokier:
Thanks for taking my feedback into account, I appreciate it. Unfortunately I didn't have time to review the new version a few days ago, but I have now and the code looks great 👍

PR #739, opened 2021-06-29
[FEATURE] Stub for bridge client

KonradStaniec:
Adds stub for future bridge client which will call bridge notes for additional data capabilities

Imo it should live in `fluffy` module as it is pretty special usecase i.e this api will be used only by portal nodes

### DIFF
--- a/fluffy/conf.nim
+++ b/fluffy/conf.nim
@@ -87,7 +87,7 @@ type
       defaultValue: DefaultAdminListenAddress
       name: "rpc-address" }: ValidIpAddress
 
-    bridgeClientUri* {.
+    bridgeUri* {.
       defaultValue: none(string)
       desc: "if provided, enables getting data from bridge node"
       name: "bridge-client-uri" .}: Option[string]
--- a/fluffy/conf.nim
+++ b/fluffy/conf.nim
@@ -87,6 +87,11 @@ type
       defaultValue: DefaultAdminListenAddress
       name: "rpc-address" }: ValidIpAddress
 
+    bridgeClientUri* {.
+      defaultValue: none(string)
+      desc: "if provided, enables getting data from bridge node"
+      name: "bridge-client-uri" .}: Option[string]
+
     case cmd* {.
       command
       defaultValue: noCommand .}: PortalCmd
--- a/fluffy/rpc/bridge_client.nim
+++ b/fluffy/rpc/bridge_client.nim
@@ -1,0 +1,44 @@
+# Nimbus
+# Copyright (c) 2021 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
+
+import 
+  std/json,
+  eth/common, chronos, json_rpc/[rpcclient], httputils
+
+# Specification of api https://github.com/ethereum/stateless-ethereum-specs/blob/master/portal-bridge-nodes.md#api
+# TODO after nethermind plugin will be ready we can get few responses from those endpoints, to create mock client to test if we
+# properly parse respones
+type
+  BridgeClientRef* = ref object
+    client*: RpcClient
+
+proc init*(T: type BridgeClientRef, client: RpcClient): T = T(client: client)
+
+proc newHttpBridgeClient*(remoteAddrress: string): Future[BridgeClientRef] {.async.}=
+  var client = newRpcHttpClient()
+  client.httpMethod(MethodPost)
+  await client.connect(remoteAddrress)
+  return BridgeClientRef.init(client)
+
+# bridge_waitNewCanonicalChain
+proc waitNewCanonicalChain*(bridgeClient: BridgeClientRef): Future[void] {.async, raises: [Defect, CatchableError].} = discard
+
+#bridge_getBlockChanges
+proc getBlockChanges*(bridgeClient: BridgeClientRef, blockHash: Hash256): Future[AccessList] {.async, raises: [Defect, CatchableError].} = discard
+
+#bridge_getItemWitness"
+proc getItemWitness*(bridgeClient: BridgeClientRef, blockHash: Hash256, acctAddr: EthAddress, slotAddr: StorageKey):
+  Future[seq[seq[byte]]] {.async, raises: [Defect, CatchableError].} = discard
+
+#bridge_getNextItem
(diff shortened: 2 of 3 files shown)


fluffy/rpc/bridge_client.nim
```diff
@@ -0,0 +1,44 @@
...
+# Specification of api https://github.com/ethereum/stateless-ethereum-specs/blob/master/portal-bridge-nodes.md#api
+# TODO after nethermind plugin will be ready we can get few responses from those endpoints, to create mock client to test if we
+# properly parse respones
+type
+  BridgeClientRef* = ref object
+    client*: RpcClient
+
+proc init*(T: type BridgeClientRef, client: RpcClient): T = T(client: client)
+
+proc newHttpBridgeClient*(remoteAddrress: string): Future[BridgeClientRef] {.async.}=
```
kdeme, line 23: Personally, I find just the `init` sufficient. But some prefer to have `initNameOfType` also for making it easier to search? (I realize that in nim-json-rpc these are all over the place, but this is old code)

Anyhow, this isn't really an `init`, and rather an `init` + `connect` call. I would keep them split up in `init` and `connect` (or `start` if you prefer).

fluffy/rpc/bridge_client.nim
```diff
@@ -0,0 +1,44 @@
...
+  var client = newRpcHttpClient()
+  client.httpMethod(MethodPost)
```
kdeme, line 25: Didn't know we could actually adjust this. I see it is default already `MethodPost`. I guess this is why you have to add the `httputils` import? In theory the json_rpc httpclient should export that.

fluffy/rpc/bridge_client.nim
```diff
@@ -0,0 +1,44 @@
...
+#bridge_getItemWitness"
+proc getItemWitness*(bridgeClient: BridgeClientRef, blockHash: Hash256, acctAddr: EthAddress, slotAddr: StorageKey):
+  Future[seq[seq[byte]]] {.async, raises: [Defect, CatchableError].} = discard
+
+#bridge_getNextItem
+proc getNextItem*(bridgeClient: BridgeClientRef, blockHash: Hash256, acctAddr: EthAddress, slotAddr: StorageKey):
+  Future[(EthAddress, StorageKey)] {.async, raises: [Defect, CatchableError].} = discard
+
+proc close*(bridgeClient: BridgeClientRef): Future[void] {.async, raises: [Defect, CatchableError].} = 
+  await bridgeClient.client.close()
```
kdeme, line 44: newline

fluffy/rpc/bridge_client.nim
```diff
@@ -0,0 +1,44 @@
...
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
+
+import 
+  std/json,
+  eth/common, chronos, json_rpc/[rpcclient], httputils
```
kdeme, line 12:
```suggestion
  eth/common, chronos, json_rpc/rpcclient, httputils
```

kdeme (commented):
I know it is just stubs, but you might want to add the bridge rpc already as an option to the actual fluffly client (default not used / initialized). This way it gets compiled also at CI.

fluffy/rpc/bridge_client.nim
```diff
@@ -0,0 +1,44 @@
...
+
+# Specification of api https://github.com/ethereum/stateless-ethereum-specs/blob/master/portal-bridge-nodes.md#api
+# TODO after nethermind plugin will be ready we can get few responses from those endpoints, to create mock client to test if we
+# properly parse respones
+type
+  BridgeClientRef* = ref object
+    client*: RpcClient
```
kdeme, line 19: Considering the rest of the code only has to option for http client, can't it be set here directly to that also?

kdeme: Come to think of it, we should also support websockets eventually, so you can leave it.

Selection should be done based on url provided, for example like here: https://github.com/status-im/nim-web3/blob/master/web3.nim#L75

fluffy/rpc/bridge_client.nim
```diff
@@ -0,0 +1,44 @@
...
+
+proc init*(T: type BridgeClientRef, client: RpcClient): T = T(client: client)
```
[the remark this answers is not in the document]
KonradStaniec: I kinda got impression that `object` it for everything which supposed to have value semantics, and `ref` is broadly unerdstood resource of some sort (or something that manages those resources).  Also l thought  bridge client will be shared between other components.

Therefore T thought that putting `ref` in normal `object` is kinda misleading. Is it normal practice  to put ref in normal object ?

fluffy/rpc/bridge_client.nim
KonradStaniec: yep, my thinking there was that we may also use web sockets in the future.

fluffy/rpc/bridge_client.nim
KonradStaniec: yep you are right, so this call is not needed here.

fluffy/rpc/bridge_client.nim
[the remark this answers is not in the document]
kdeme: > I kinda got impression that `object` it for everything which supposed to have value semantics, and `ref` is broadly unerdstood resource of some sort (or something that manages those resources). Also l thought bridge client will be shared between other components.
> 
> Therefore I thought that putting `ref` in normal `object` is kinda misleading. Is it normal practice to put ref in normal object ?

I don't really mind that much here whether it is `ref` or not, as the `RpcClient` object itself is/needs to be `ref`, so it is a bit of a non discussion. And you know, you might just want to alias it as `BridgeRpcClient = RpcClient`, considering there are no additional fields right now.

However, in general I do not believe that the choice of making an object `ref` should be just because of the fact that one of its fields is a `ref`. But rather about how that object is used. As you have mentioned yourself already, for example when the the intention is to share this object around modules.

Practically here, I'm not sure yet that this will be the case. The way I see it for now (might change ;) ), is that there will be some content providing module, that will get its content either from:
- a local DB, perhaps a cache, and in case content is missing, from the network
- from this json rpc interface in case the bridge option is enabled

So only that module would use it, probably. It can always be a changed later if not, and this can also be done by just using `ref` there where you instantiate the object.

fluffy/conf.nim
```diff
@@ -87,6 +87,11 @@ type
       defaultValue: DefaultAdminListenAddress
       name: "rpc-address" }: ValidIpAddress
 
+    bridgeClientUri* {.
```
kdeme, line 90: Perhaps just call it bridgeUri? Client is confusing here, as it is actually the uri of the server you must provide

PR #740, opened 2021-06-30
[FIX] Add missing gas used validation

KonradStaniec:
While browsing codebase I have noticed that one validation is missing when validating header. 

In https://ethereum.github.io/yellowpaper/paper.pdf in equation `51`, part of equation states that `Hg ≤ Hl`, which means that header.gasUsed should always be lower than header.gasLimit.

I was considering creating and issue, but as fix is easy I decided to go with pr. If the check is implemented elswhere or there is some process for creating consensus level fixes please let me know.

### DIFF
--- a/nimbus/p2p/validate.nim
+++ b/nimbus/p2p/validate.nim
@@ -180,6 +180,9 @@ proc validateHeader(db: BaseChainDB; header, parentHeader: BlockHeader;
   if header.gasUsed == 0 and 0 < numTransactions:
     return err("zero gasUsed but tranactions present");
 
+  if header.gasUsed < 0 or header.gasUsed > header.gasLimit:
+    return err("gasUsed should be non negative and smaller or equal gasLimit")
+  
   if header.blockNumber != parentHeader.blockNumber + 1:
     return err("Blocks must be numbered consecutively")
 


jangko:
You are right, we don't have explicit validation for `header.gasUsed < header.gasLimit`, we replace it with this:

https://github.com/status-im/nimbus-eth1/blob/02afb27e4756f1e4185c1b9eafb6b39becc73d23/nimbus/p2p/validate.nim#L299

But I think there is no harm to add additional validation.

KonradStaniec:
hmm I am no sure those two are the same as from what I understand the check you linked is performed in transaction execution context i.e does we we have enough room for another transaction. 

On the other hand, checking that `gasUsed <= gasLimit`  is necessary even if we have only header, for example in fast sync when we download headers alone (without executing transactions) we should not allow headers which breaks this check.

jangko:
that's make sense. please go ahead and merge this PR.

jlokier:
The are actually separate levels of validation, and this is set out in the specifications of the store-and-forward P2P process. See the [`eth` protocol spec](https://github.com/ethereum/devp2p/blob/master/caps/eth.md), where it talks about "header validity", "data validity" and "transaction validity", which are all separate from transaction execution validity.

I belive this one comes under "data validity".

A number of things are not actually implemented in nimbus-eth1 yet, and as a result these different levels of validation have rather lax coverage and separation. There are other validation steps which are not implemented too. PoW notably.

There's enough to pass the testsuites, basically (fixtures and Hive), but things not covered by the testsuites are not.

Fixes that change a logical outcome should ideally come with a new testcase, though I know that isn't always feasible.  For validations, in particular, it appears missing validation steps can go unnoticed and don't show up in tests, but the code missing them is still insecure against attacks.

The condition in this PR is likely to be moved (and so could get lost by mistake) when we firm up the different validation levels for P2P.  So a test case for each specific validation condition that doesn't seem to be covered by other tests would be a bit more valuable, if it's feasible to make one.

PR #743, opened 2021-07-01
Feature/goerli replay clique poa

mjfh:
Prepare for PoA processing
* Replay Goerli processing as test
    + Recover the blocks of a Goerli session dumped in a compressed rlp file
    + Run against _p2p/chain.persistBlocks()_ similar to what happens in the real Goerli session
* Provide stubs for PoA processing
    + Main processing will take place in _p2p/executor.processBlock()_ (for now only stubs here)
    + Additional function prototype provided for _p2p/executor.processBlock()_ which takes PoA state descriptor as function argument
    +  _p2p/executor.processBlock()_ is typically called from _p2p/chain.persistBlocks()_ which is initialised with the _Chain_ descriptor which now also holds the PoA state
* Operational changes
    + None for non-PoA networks
    + For PoA networks (i.e. when the internal flag is set) the block processing _must_ apply the particular _p2p/executor.processBlock()_ version with the PoA desciptor argument (the other one will politely fail just by return code)


nimbus/p2p/executor/process_block.nim
```diff
@@ -0,0 +1,169 @@
...
+
+  # `applyDeletes = false`
+  # If the trie pruning activated, each of the block will have its own state
+  # trie keep intact, rather than destroyed by trie pruning. But the current
+  # block will still get a pruned trie. If trie pruning deactivated,
+  # `applyDeletes` have no effects.
+  dbTx.commit(applyDeletes = false)
+
+
+proc processBlock*(vmState: BaseVMState; poa: Clique;
```
jangko, line 146: How is the `poa` param used in this proc? can we remove it if it not actually being used? can we replace it with less confusing language feature e.g `_: type Clique`.?

nimbus/p2p/chain.nim
```diff
@@ -164,7 +166,7 @@ method persistBlocks*(c: Chain; headers: openarray[BlockHeader];
       (header, body) = (headers[i], bodies[i])
       parentHeader = c.db.getBlockHeader(header.parentHash)
       vmState = newBaseVMState(parentHeader.stateRoot, header, c.db)
-      validationResult = processBlock(c.db, header, body, vmState)
+      validationResult = vmState.processBlock(c.poa, header, body)
```
jangko, line 169: the presence of `c.poa` here is actually a bit confusing if it also able to process  PoW block.

jangko (commented):
generally this PR looks good, but please add some comments for the `poa` or `c.poa` params. so at the next iteration, other developer know what to do with them, or remove them if possible to reduce confusion.

nimbus/p2p/executor/process_block.nim
mjfh: Typically called from persistBlocks() which holds the state in the Chain type descriptor (see introductory comment for this patch)

nimbus/p2p/chain.nim
mjfh: will add some comment to avoid confusion

the c.cb parameter is passed as part of the vmState and the c.pos state descriptor is new

PR #746, opened 2021-07-05
[FEATURE] Add support for handling experimental api call

KonradStaniec:
Add support for calling:
https://github.com/ethereum/stateless-ethereum-specs/blob/master/portal-bridge-nodes.md#experimental-api-currently-implemented-by-the-nethermind-bridge-node-plugin


fluffy/rpc/bridge_client.nim
```diff
@@ -7,8 +7,8 @@
 
 {.push raises: [Defect].}
 
-import 
-  std/json,
+import
+  std/json, stew/[byteutils, results],
```
kdeme, line 11: Best to keep the std imports on one line, and the imports from other modules on the next, for visibility.

fluffy/rpc/bridge_client.nim
```diff
@@ -17,19 +17,43 @@ import
...
 proc getItemWitness*(bridgeClient: BridgeClient, blockHash: Hash256, acctAddr: EthAddress, slotAddr: StorageKey):
   Future[seq[seq[byte]]] {.async, raises: [Defect, CatchableError].} = discard
 
 #bridge_getNextItem
 proc getNextItem*(bridgeClient: BridgeClient, blockHash: Hash256, acctAddr: EthAddress, slotAddr: StorageKey):
   Future[(EthAddress, StorageKey)] {.async, raises: [Defect, CatchableError].} = discard
 
+# bridge_getBlockWitness
+# Returns a list of all RLP-encoded merkle trie values (including contract bytecode) accessed during block execution
+proc getBlockWitness*(bridgeClient: BridgeClient, blockNumber: BlockNumber): Future[Result[seq[seq[byte]], string]] {.async, raises: [Defect, CatchableError].} =
```
kdeme, line 54: This is quite the long line ;)

We (I) typically cut off at 89 lines, except from hex strings and such like in your test.

tests/all_tests.nim
```diff
@@ -117,4 +117,5 @@ cliBuilder:
           ./test_misc,
           ./test_graphql,
           ./test_lru_cache,
-          ./test_clique
+          ./test_clique,
+          ../fluffy/test_bridge_parser
```
kdeme, line 121: The fluffy tests should go in a new `all_test` file, probably best under the fluffy folder. With a new nimble task and Makefile target. So that we can run the in CI based on the paths changed.

fluffy/test_bridge_parser.nim
```diff
@@ -0,0 +1,67 @@
+import
```
kdeme, line 1: You can create a `test` folder in the fluffy folder for this.

kdeme (commented):
lgtm, we just have to separate the tests from the nimbus-eth1 tests, so that the CI can selectively run as it is done now for the compilation of the fluffy client.

tests/all_tests.nim
KonradStaniec: I wondered about that but ultimatly decided  to go the same way as `stateless` sub project went, but if fluffy should be more separate then sure :+1:

fluffy/all_fluffy_tests.nim
```diff
@@ -0,0 +1,13 @@
...
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or http://www.apache.org/licenses/LICENSE-2.0)
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT) or http://opensource.org/licenses/MIT)
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+import ../../test_macro
+
+{. warning[UnusedImport]:off .}
+
+cliBuilder:
```
kdeme, line 12: I was not aware of this code. What does it do exactly and/or why is it needed?

KonradStaniec: from what I understand with this macro it is possible to run only one selected test suite by providing index argument i.e if you have something like:
```
all_test.nim

cliBuilder:
  import  ./test_code_stream,
          ./test_gas_meter,
          ./test_memory,
          ./test_stack,
          ./test_genesis,
```

you can run only ` ./test_code_stream` by providing index argument like `make test all_test 0`

Not sure if it will be necessary in fluffy case by it seemed nice enough to reuse it.

kdeme: Ok cool, lets keep it then.

tests/all_tests.nim
```diff
@@ -117,4 +37,4 @@ cliBuilder:
           ./test_misc,
           ./test_graphql,
           ./test_lru_cache,
-          ./test_clique
+          ./test_clique
```
kdeme, line 40: Looks like your editor remove a newline here? You might want to set-up your editor to keep newlines.

kdeme (approved):

PR #747, opened 2021-07-06
[FEATURE] Use rpc proxy

KonradStaniec:
Proxy all rpc call to provided uri


fluffy/conf.nim
```diff
@@ -14,6 +14,7 @@ import
 const
   DefaultListenAddress* = (static ValidIpAddress.init("0.0.0.0"))
   DefaultAdminListenAddress* = (static ValidIpAddress.init("127.0.0.1"))
+  DefaultProxyAddrss* = (static "http://127.0.0.1:8546")
```
kdeme, line 17:
```suggestion
  DefaultProxyAddress* = (static "http://127.0.0.1:8546")
```

fluffy/conf.nim
```diff
@@ -92,6 +93,13 @@ type
       desc: "if provided, enables getting data from bridge node"
       name: "bridge-client-uri" .}: Option[string]
 
+    # it makes little sense to have default value here in final release, but until then
```
kdeme, line 96: The `http://127.0.0.1:8546` is fine I think for now. (Either that or 8545 but then we have to change our own local port).

Eventually the proxy should become optional (or might be removed completely), and should be default disabled, but only after we have implemented (most) rpc calls.

fluffy/fluffy.nim
```diff
@@ -68,11 +68,12 @@ proc run(config: PortalConf) {.raises: [CatchableError, Defect].} =
     except Exception as exc: raiseAssert exc.msg
 
   if config.rpcEnabled:
-    let
-      ta = initTAddress(config.rpcAddress, config.rpcPort)
-      rpcHttpServer = newRpcHttpServer([ta])
-    rpcHttpServer.installEthApiHandlers()
-    rpcHttpServer.start()
+    let ta = initTAddress(config.rpcAddress, config.rpcPort)
+    var rpcHttpServerWithProxy = newRpcHttpProxy([$ta])
```
kdeme, line 72: Ok for now, but we should get a `newRpcHttpProxy` which takes in the `TransportAddress`

fluffy/fluffy.nim
```diff
@@ -68,11 +68,12 @@ proc run(config: PortalConf) {.raises: [CatchableError, Defect].} =
...
+    rpcHttpServerWithProxy.installEthApiHandlers()
+    # TODO for now we can only proxy to local node (or remote one without ssl) to make it possible
+    # to call infura https://github.com/status-im/nim-json-rpc/pull/101 needs to get merged for http client to support https/
+    waitFor rpcHttpServerWithProxy.start(config.proxyUri)
```
kdeme, line 76: This could use some further error handling eventually.

kdeme (approved):

fluffy/conf.nim
KonradStaniec: yep that was my thinking about it, this is just like additional wheels in bike and ultimatly we may get rid of it

fluffy/fluffy.nim
KonradStaniec: I will make a quick pr in upstream repo to add additional constructor

fluffy/fluffy.nim
KonradStaniec: I agree, for now as nothing is implemented it makes little sense to start without without working proxy, in the future when there will some working methods, it will be better to catch error here and log thats fluffy is starting without working proxy so some methods may not work as intended.

PR #748, opened 2021-07-08
Feature/implement poa processing

mjfh:


nimbus/p2p/clique/clique_cfg.nim
```diff
@@ -48,29 +48,31 @@ type
...
     db*: BaseChainDB
-    signatures*: EcRecover  ## Recent block signatures to speed up mining
-    period*: Duration       ## time between blocks to enforce
-    prng*: Rand             ## PRNG state for internal random generator
-    bcEpoch: UInt256        ## The number of blocks after which to checkpoint
-                            ## and reset the pending votes.Suggested 30000 for
-                            ## the testnet to remain analogous to the mainnet
-                            ## ethash epoch.
+    period*: Duration      ## time between blocks to enforce
+    prng*: Rand            ## PRNG state for internal random generator
```
jangko, line 55: FYI, we have a global RNG here

https://github.com/status-im/nimbus-eth1/blob/1c5c2cba8b8286f8362e5ac8177224cd9993348f/nimbus/random_keys.nim#L3

mjfh: i know, but this is a crypto PRNG -- the one here is currently used to produce reproducable results for currently unused random time slots and testing

but it might make sense to provide a getter access method so it can be obverloaded by the real one

jangko:
LGTM

PR #757, opened 2021-07-12
Further implement Portal Nodes and FoundContent messages

kdeme:
Not complete and not fully according to spec (ContentKey), but WIP.

### DIFF
--- a/fluffy/content.nim
+++ b/fluffy/content.nim
@@ -47,6 +47,8 @@ template toSszType*(x: ContentType): uint8 =
   uint8(x)
 
 template toSszType*(x: auto): auto =
+  mixin toSszType
+
   x
 
 func fromSszBytes*(T: type ContentType, data: openArray[byte]):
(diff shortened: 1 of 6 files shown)


fluffy/content.nim
```diff
@@ -47,6 +47,8 @@ template toSszType*(x: ContentType): uint8 =
   uint8(x)
 
 template toSszType*(x: auto): auto =
+  mixin toSszType
```
KonradStaniec, line 50: just a nim question: `mixin` means that related symbol must be in scope when instatiating a template ? (manual is pretty vague here stating that `A symbol can be forced to be open by a mixin declaration` but not defining what does it mean `open`)

fluffy/network/portal_protocol.nim
```diff
@@ -41,18 +44,45 @@ proc handleFindNode(p: PortalProtocol, fn: FindNodeMessage): seq[byte] =
...
+  # TODO: Need to check networkId, type, trie path
   let
-    enrs = List[ByteList, 32](@[])
-    payload = ByteList(@[])
-  encodeMessage(FoundContentMessage(enrs: enrs, payload: payload))
+    # TODO: Should we first do a simple check on ContentId versus Radius?
+    contentId = toContentId(fc.contentKey)
+    content = p.contentStorage.getContent(fc.contentKey)
+  if content.isSome():
+    let enrs = List[ByteList, 32](@[]) # Empty enrs when payload is send
```
KonradStaniec, line 71: probably we could add some checks in `proc findContent*` that those field are empty when receiving `FoundContentMessage`  but this is probably something for other pr.

KonradStaniec (approved):
minor question/remark

fluffy/network/portal_protocol.nim
kdeme: Yes, good point, that check is missing. 
I'll probably change some of that serialization/deserialization still however to be able to convert the Node Record into the ByteLists and such in these handlers.

PR #762, opened 2021-07-14
Renamed source file clique_utils => clique_helpers

mjfh:


nimbus/p2p/clique/snapshot/snapshot_apply.nim
```diff
@@ -37,33 +37,49 @@ logScope:
...
-proc say(s: var Snapshot; v: varargs[string,`$`]) {.inline.} =
+proc say(s: Snapshot; v: varargs[string,`$`]) {.inline.} =
   # s.cfg.say v
   discard
 
+# ------------------------------------------------------------------------------
+# Private functions
+# ------------------------------------------------------------------------------
+
+iterator pairWalk(first, last: int; offTop: Positive): (int,int) {.gcsafe.} =
```
jangko, line 48: iirc, iterator with multiple yields will produce bloat code.
can we use template mixed with iterator here?

jangko (commented):
I have no real objection beside left hand mixed with right hand comparison, it's a bit confusing to read, but no problem.

nimbus/p2p/clique/snapshot/snapshot_apply.nim
mjfh: ok, will be replaced by template

PR #772, opened 2021-07-27
Tested eth/65 support

jlokier:
This is an intentionally simple PR, designed to not break, change or depend on other functionality much, so that the "old sync" methods can be run usefully again and observed.  This patch isn't "new sync" (a different set of
sync algorithms), but it is one of the foundations.

For a while now Nimbus Eth1 only supported protocol `eth/63`.  But that's obsolete, and very few nodes still support it.  This meant Nimbus Eth1 would make slow progress trying to sync, as most up to date clients rejected it.

The current specification is `eth/66`, and the functionality we really need is in `eth/64`.

So why `eth/65`?

- `eth/64` is essential because of the `forkId` feature.  But `eth/64` is on its way out as well.  Many clients, especially the most up to date Geth running the current hard-forks (Berlin/London) don't talk `eth/64` any more.

- `eth/66` is the current specification, but some clients don't talk `eth/66` yet.  We'd like to have the best peer connectivity during tests, and currently everything that talks `eth/66` also talks `eth/65`.

- Nimbus Eth1 RLPx only talks one version at a time.  (Without changes to the RLPx module.  When those go in we'll add `eth/64..eth/66` for greater peer reach and testing the `eth/66` behaviour.  For simplicity and decoupling, this patch contains just one version, the most useful.)

What are `eth/64` and `eth/65`?

- `eth/64` (EIP-2364) added `forkId` which allows nodes to distinguish between Ethereum (ETH) and Ethereum Classic (ETC) blockchains, which share the same genesis block.  `forkId` also protects the system when a new hard fork is going to be rolled out, by blocking interaction with out of date nodes.  The feature is required nowadays.

  We send the right details to allow connection (this has been tested a lot), but don't apply the full validation rules of EIP-2124/EIP-2364 in this patch. It's to keep this patch simple (in its effects) and because those rules have consequences best tested separately.  In practice the other node will reject us when we would reject it, so this is ok for testing, as long as it doesn't get seriously deployed.

- `eth/65` added more efficient transaction pool methods.

- Then a later version of `eth/65` (without a new number) added typed  transactions, described in [EIP-2976](https://eips.ethereum.org/EIPS/eip-2976).

Why it's moved to `nimbus-eth1`:

- Mainly because `eth/64` onwards depend on the current state of block synchronisation, as well as the blockchain's sequence of hard-fork block numbers, both of which are part of `nimbus-eth1` run-time state.  These aren't available to pure `nim-eth` code.  Although it would be possible to add an API to let `nimbus-eth1` set these numbers, there isn't any point because the protocol would still only be useful to `nimbus-eth1`.

### DIFF
--- a/hive_integration/nodocker/graphql/graphql_sim.nim
+++ b/hive_integration/nodocker/graphql/graphql_sim.nim
@@ -10,7 +10,7 @@
 import
   std/[os, parseopt, json],
   eth/[p2p, trie/db], ../../../nimbus/db/db_chain,
-  eth/p2p/rlpx_protocols/eth_protocol,
+  ../../../nimbus/sync/protocol_eth65,
   ../../../nimbus/[genesis, config, conf_utils],
   ../../../nimbus/graphql/ethapi, ../../../tests/test_helpers,
   graphql, ../sim_utils
--- a/nimbus/nimbus.nim
+++ b/nimbus/nimbus.nim
@@ -15,11 +15,13 @@ import
   eth/keys, db/[storage_types, db_chain, select_backend],
   eth/common as eth_common, eth/p2p as eth_p2p,
   chronos, json_rpc/rpcserver, chronicles,
-  eth/p2p/rlpx_protocols/[eth_protocol, les_protocol],
+  eth/p2p/rlpx_protocols/les_protocol,
   eth/p2p/blockchain_sync, eth/net/nat, eth/p2p/peer_pool,
+  ./sync/protocol_eth65,
   config, genesis, rpc/[common, p2p, debug, key_storage], p2p/chain,
   eth/trie/db, metrics, metrics/[chronos_httpserver, chronicles_support],
-  graphql/ethapi, utils, ./conf_utils
+  graphql/ethapi,
+  "."/[utils, conf_utils]
 
 ## TODO:
 ## * No IPv6 support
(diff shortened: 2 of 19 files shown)


zah (approved):
What was the criteria used to decide which log messages were commented out? 

EDIT: Nevermind, this was explained here:
https://github.com/status-im/nimbus-eth1/pull/772/commits/46d022f9870d69e7d26f1b94227ac4dd9948ae63

arnetheduck:
Sounds like there should be a corresponding eth/63 removal somewhere? also, we tend to squashmerge things because rolling back lots of small broken commits is messy

jlokier:
`eth/63` is in a different repo so its removal can't be part of this PR. `eth/64+` has to go here.

I try to arrange multi-repo PRs to be decoupled (when that is easy), so that the interesting place (e.g. where the new feature or bugfix arrives) can be used on its own merits without being linked quasi-atomically with a PR in other library-like repos.  In this case it means `eth/65` and things which depend on it can move forward, without having to rewrite the testsuite in `nim-eth` to get over the removal

If the result is now unused code in the library-like repos, do the cleanup there after the dust has settled. Rather like updating well-maintained libraries and applications that use them. Especially when the other, library-like repos have multiple users, as I'll want to check the change in that doesn't break any of them.

In this case I'm worried about knock on effects on Waku and any other repo I don't know about that's using RLPx, because removing `eth/63` will remove almost the last in-repo use of RLPx, which removes most or all the testsuite for RLPx+p2pProtocol in that repo.  Such cleanup is worth doing, but definitely not a priority compared with other things. (It's now become like Whisper - nobody uses it, but the tests keep RLPx honest).

PR #773, opened 2021-07-29
Add routing table and required lookup code to Portal protocol

kdeme:
Adding routing table to the Portal level.

Eventually there will be a protocol + routing table for each different content network (state, history, ...).

There is a lot of lookup overlap code with discv5. Having some of this common would be nice.


kdeme:
Not fully happy with the state of implementation, but I'm going to go ahead and merge what there is now

PR #774, opened 2021-07-30
Bump nimbus-build-system and nim-eth vendor modules

kdeme:


kdeme:
Merging this as the only CI issue seems to be oom related (`cc1.exe: out of memory allocating 65536 bytes`) which seems to have happened on other PRs too.

PR #776, opened 2021-07-30
Add monthly development updates link

kdeme:
... and few other changes

Not sure whether to keep the actual updates in a notes.status.im page or move them in an `development-updates.md` file in this repository.


kdeme:
adjusted to hackmd doc as I didn't manage to make a states note publicly available. 

I'll go ahead and merge this, as we need to share something. Can still leave remarks though if we should change it to a separate md file in the repo.

PR #777, opened 2021-08-01
add new command line options: clique-period, engine-signer, and impor…

### LINKED ISSUE #771: config: prepare PoA engine configuration for upcoming clique and rpc test in hive
hive simulators have the following flags set:

- rpc simulator:
```text
"HIVE_CLIQUE_PERIOD":     "1",
"HIVE_CLIQUE_PRIVATEKEY": "9c647b8b7c4e7c3490668fb6c11473619db80c93704c70893d3813af4090c39c",
"HIVE_MINER":             "658bdf435d810c91414ec09147daa6db62406379",
```

- smoke/clique:
```text
"HIVE_CLIQUE_PERIOD":     "2",
"HIVE_CLIQUE_PRIVATEKEY": "9c647b8b7c4e7c3490668fb6c11473619db80c93704c70893d3813af4090c39c",
"HIVE_MINER":             "658bdf435d810c91414ec09147daa6db62406379",
```

whatever it means, we need to prepare the cli parser and friends for:
- clique period -- clique only
- clique private key -- clique only, ethash cannot use insecure method to set private key.
- miner address -- common for both clique and ethash

jangko:
…t-key

new command line options:
  --clique-period:<value> Enables clique support. value is block time in seconds
  --engine-signer:<value> Enables mining. value is EthAddress in hex
  --import-key:<path>     Import unencrypted 32 bytes hex private key file

fixes #771

### DIFF
--- a/nimbus/config.nim
+++ b/nimbus/config.nim
@@ -136,6 +136,8 @@ type
     importFile*: string
     verifyFromOk*: bool           ## activate `verifyFrom` setting
     verifyFrom*: uint64           ## verification start block, 0 for disable
+    cliquePeriod*: int            ## Enables clique support, 0 for disable
+    engineSigner*: EthAddress     ## Miner account
 
 const
   # these are public network id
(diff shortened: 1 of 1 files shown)


mjfh:
ok

mjfh (approved):

jlokier:
"Enables mining."  I think this help text is unhelpful. Imagine being an end-user reading these options. It sounds like this enables a transaction pool, transaction choice and PoW mining using Ethash.  I assume it doesn't do those things. What does it do?

"Value is block time in seconds" - Is the standard 15 seconds block time not always used?

--import-key: To where, for what purpose?

jangko:
Okay, Thanks. will fix it at the next round.

PR #785, opened 2021-08-03
Transaction: Treat `CREATE2` salt as a 256-bit blob not a number

jlokier:
This changes fixes a bug in `CREATE2` ops when used with EVMC.
Because it changes the salt type, it affects non-EVMC code as well.

The salt was passed through EVMC with the wrong byte order, although this went unnoticed as the Nimbus host flipped the byte order before using it.

This was found when running Nimbus with third-party EVM, ["evmone"](https://github.com/ethereum/evmone).

There are different ways to remedy this.

If treated as a number, Nimbus EVM would byte-flip the value when calling EVMC, then Nimbus host would flip the received value.  Finally, it would be flipped a third time when generating the address in `generateSafeAddress`.  The first two flips can be eliminated by negotiation (like other numbers), but there would always be one flip.

As a bit pattern, Nimbus EVM would flip the same way it does when dealing with hashes on the stack (e.g. with `getBlockHash`).  Nimbus host wouldn't flip at all - and when using third-party EVMs there would be no flips in Nimbus.

Because this value is not for arithmetic, any bit pattern is valid, and there shouldn't be any flips when using a third-party EVM, the bit-pattern interpretation is favoured.  The only flip is done in Nimbus EVM (and might be eliminated in an optimised version).

We'll use `Hash256` because Nimbus uses this for "opaque 256 bits" and handles byte-endian conversions on this type, even though salt isn't a hash.

### DIFF
--- a/nimbus/constants.nim
+++ b/nimbus/constants.nim
@@ -16,8 +16,8 @@ const
   # address zero by accident, unrecoverably, due to poor user interface issues.
   ZERO_ADDRESS* =                           default(EthAddress)
 
-  # ZERO_HASH32 is the parent hash of genesis blocks.
-  ZERO_HASH32* =                            Hash256()
+  # ZERO_HASH256 is the parent hash of genesis blocks.
+  ZERO_HASH256* =                           Hash256()
 
   GAS_LIMIT_EMA_DENOMINATOR* =              1_024
   GAS_LIMIT_ADJUSTMENT_FACTOR* =            1_024
@@ -43,10 +43,10 @@ const
   GENESIS_BLOCK_NUMBER* =                   0.toBlockNumber
   GENESIS_DIFFICULTY* =                     131_072.u256
   GENESIS_GAS_LIMIT* =                      3_141_592
-  GENESIS_PARENT_HASH* =                    ZERO_HASH32
+  GENESIS_PARENT_HASH* =                    ZERO_HASH256
   GENESIS_COINBASE* =                       ZERO_ADDRESS
   GENESIS_NONCE* =                          "\x00\x00\x00\x00\x00\x00\x00B"
-  GENESIS_MIX_HASH* =                       ZERO_HASH32
+  GENESIS_MIX_HASH* =                       ZERO_HASH256
   GENESIS_EXTRA_DATA* =                     ""
   GAS_LIMIT_MINIMUM* =                      5000
   GAS_LIMIT_MAXIMUM* =                      high(GasInt)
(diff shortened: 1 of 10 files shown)


jangko:
And why not create a new blob type?

jlokier:
> And why not create a new blob type?

Ok, I've done that change after reconsidering.

But I'll explain why I didn't choose a new type originally.

It's a pragmatism judgement call. This value is only used for one very niche purpose, and the existing functions operate on it properly. Although technically not a hash, it is similar enough that the hash type had the right operations. I don't think it's always good to create new types for every _subtle_ difference in meaning, especially if it requires extra boilerplate, conversion functions, wrappers etc. and is only used in one place. Too many types is unhelpful.

For example, a lot of places where we use `int`, we could define a special, distinct integer type instead. But we don't because the code is clear already with `int`. Types provide a kind of documentation, but field and variable names should provide enough clarity about what it's for.  And using `int` means standard functions like `+` are available.

Extra type-checking is useful, but that's a judgement call too. After all, we don't use distinct types for each different use of a hash for different purposes, even though they are incompatible.

There have been software engineering studies on programs with too many types.  Beyond a certain amount, too many named global types makes programs harder to understand as a whole, and it's counterproductive. (See also: Enterprise software). To be fair that's more about objects, relations and complexity. But that principle is why when I'm deciding whether to create a new type or use one that already works, I try to think pragmatically, and weigh up whether the benefits are worth the loss of reuse and extra concepts people reading the code must take on board. Like bureaucracy, you need enough, but it can grow to be too much.

jangko:
I doesn't have to be `ContractSalt256`, a generic `Blob256` or `Bytes32` will do.

jlokier:
> I doesn't have to be `ContractSalt256`, a generic `Blob256` or `Bytes32` will do the job.

Ah, it would have helped if your suggestion before said that.

`Blob256` or `Bytes32` does seem to come up as something that would often be useful often in new code.

If it existed, it would be nice if it printed as hex. Perhaps when converting it to/from `Uint256` it should include endian conversion too.

Looking around, there's a trend to use named 32-byte arrays: `Topic` and `StorageKey`.  `ContractSalt` fits in with those.  For the sake of pragmatism, I'll rename to `ContractSalt` here and merge it now so EVMC can move forward.

Separately (not today), I'll look at moving this and the 32-byte arrays in `eth_types` over to using a generic `Blob256` or `Bytes32` with useful properties like hex printing.  `TrieHash` and `TriePath` also exist that may use it.

PR #787, opened 2021-08-05
config: fix config file parser

### LINKED ISSUE #786: `--customnetwork:FILE.json` does not work with test network Calaveras using the official file
Using this option `--customnetwork:calaveras.json` with or without `--networkid:123` fails to result in a successful connection ot the Calaveras network, because genesis hash, network Id and probably other values are incorrect.

The file `calaveras.json` is the JSON file in the [Calaveras specification](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/client-integration-testnets/calaveras.md).

In network logs some things are clear:

- The `networkId` we are using is zero.  It should be 123. This is visible in the `eth.Status` message, but it's a missing value in the chain config, not the network code.
- The genesis hash we are using is wrong. It should be `eb9233d066c275efcdfed8037f4fc082770176aefdbcb7691c71da412a5670f2` but we have calculated `2893a852a2fdec70cfb74307512489a2880c55703dcb623e3cf51300474af8d4` after setting up the genesis config.
- We have `poaEngine: false`, but this is a PoA network, and is configured as such in the JSON file.

Do we not support this file format?

jangko:
- allow clique period and epoch to be configured via config file
- this also activate poaEngine mode
- remove clique period configuration from cli to reduce confusion
- fix #786


jlokier (commented):
Nice work. The code looks good.  I think it's good you removed the command line option, leaving it to the config file.  (Maybe we'll add a generic command line option like OpenSSH's `-o` someday, so config file entries can be given.)

Unfortunately, it doesn't make the Calaveras network work. However this code is good, if you're happy to merge this as a config parser update only, knowing there's separate work needed to make the networking have the right values, I'm happy to approve it.

Here are the network problems: The networkid is not set (it's 0), and the calculated genesis hash is wrong. I see you do have a test for genesis hash, but it's wrong at network time (these messages are output by `protocol_eth65.nim`).  (Also trying to set the network id manually with `--networkid:123` has no effect.)

Log:
```
TRC 2021-08-05 09:42:42.136+01:00 Custom genesis block configuration loaded  tid=2204778 file=config.nim:258 conf="(chainId: 123, homesteadBlock: 0, daoForkBlock: 0, daoForkSupport: false, eip150Block: 0, eip150Hash: 0000000000000000000000000000000000000000000000000000000000000000, eip155Block: 0, eip158Block: 0, byzantiumBlock: 0, constantinopleBlock: 0, petersburgBlock: 0, istanbulBlock: 0, muirGlacierBlock: 0, berlinBlock: 0, londonBlock: 500, poaEngine: true, cliquePeriod: 30, cliqueEpoch: 30000)"
TRC 2021-08-05 09:42:42.142+01:00 DevP2P local and remote protocols          topics="rlpx" tid=2204778 file=rlpx.nim:986 selected=[eth/65] localOffer=[eth/65] remoteOffer="[eth/65, eth/66, snap/1]"
TRC 2021-08-05 09:42:42.142+01:00 >> Sending eth.Status (0x00) [eth/65]      tid=2204778 file=protocol_eth65.nim:50 peer=188.40.153.197:30303 td=131072 bestHash=2893a852a2fdec70cfb74307512489a2880c55703dcb623e3cf51300474af8d4 networkId=0 genesis=2893a852a2fdec70cfb74307512489a2880c55703dcb623e3cf51300474af8d4 forkHash=3e9144ea forkNext=500
TRC 2021-08-05 09:42:42.142+01:00 << Received eth.Status (0x00) [eth/65]     tid=2204778 file=protocol_eth65.nim:50 peer=188.40.153.197:30303 td=178375 bestHash=5cdcf9776ffbae0bcf7f68721df0b86d00c096f3e8b35a466a93d35bf1b0cbc3 networkId=123 genesis=eb9233d066c275efcdfed8037f4fc082770176aefdbcb7691c71da412a5670f2 forkHash=173783e4 forkNext=0
TRC 2021-08-05 09:42:42.142+01:00 Handshake: Local and remote networkId      tid=2204778 file=protocol_eth65.nim:83 local=0 remote=123
TRC 2021-08-05 09:42:42.142+01:00 Handshake: Local and remote genesisHash    tid=2204778 file=protocol_eth65.nim:85 local=2893a852a2fdec70cfb74307512489a2880c55703dcb623e3cf51300474af8d4 remote=eb9233d066c275efcdfed8037f4fc082770176aefdbcb7691c71da412a5670f2
TRC 2021-08-05 09:42:42.142+01:00 Handshake: Local and remote forkId         tid=2204778 file=protocol_eth65.nim:87 local=3e9144ea/500 remote=173783e4/0
TRC 2021-08-05 09:42:42.142+01:00 Peer for a different network (networkId)   tid=2204778 file=protocol_eth65.nim:92 peer=188.40.153.197:30303 expectNetworkId=0 gotNetworkId=123
TRC 2021-08-05 09:42:42.142+01:00 Disconnecting useless peer                 topics="rlpx" tid=2204778 file=rlpx.nim:1175 peer=188.40.153.197:30303
```

jangko:
I've fixed the bug in `networkId` cli. it should works now. and you also need to adjust your `calaveras.json` a bit according to #786

tests/test_misc.nim
```diff
@@ -22,5 +25,17 @@ proc miscMain*() =
       check toAddress(0x10, 0x0).toInt == 0x1000
       check toAddress(0x10, 0x0, 0x0).toInt == 0x100000
 
+    test "networkid cli":
+      var msg: string
+      const genesisFile = "tests" / "customgenesis" / "calaveras.json"
+      var opt = initOptParser("--customnetwork:" & genesisFile & " --networkid:123")
```
jlokier, line 31: Should probably test when `--networkid:123` is given first as well, i.e. two tests.

jlokier (approved):
Alright, I can confirm it connects to Calaveras ok and starts syncing!

A couple of points (not blockers):

- I'm under the impression the JSON file is in some way "standard" among other clients. Is there a reason we incompatibly require all the genesis fields under a node?
- It would be better if networkid was set automatically from the file's `chainId` when it's not set on the command line, or alternatively if it exited with an error saying you must set it.  Currently it connects to peers using networkId 0, which is a wrong default when a custom chain is used.

tests/test_misc.nim
jangko: ok. will add more test

jangko:
> I'm under the impression the JSON file is in some way "standard" among other clients. Is there a reason we incompatibly require all the genesis fields under a node?

To simplify the chain_config parser. We can gradually make it more complicated, to support geth style genesis file.

> It would be better if networkid was set automatically from the file's chainId when it's not set on the command line, or alternatively if it exited with an error saying you must set it. Currently it connects to peers using networkId 0, which is a wrong default when a custom chain is used.

Right, although they are technically different, but in reality, many networks are using the same id for `ChainId` dan `NetworkId`. We only need to make sure this additional code should never make developer assume they are the same thing. I've mix `ChainId` and `NetworkId` usage in the past and it cause a lot of confusion and bugs.

PR #791, opened 2021-08-09
Add initial implementation of overlay protocol

KonradStaniec:
There is proposal to introduce one protocol called `overlay` which will unify handling of routing table in some of the fluffy networks - https://notes.ethereum.org/tPzmxQD_S3S3uvtpUSA0-g.

This pr adds initial implementation which we will be able to discuss further.

### DIFF
--- a/fluffy/network/overlay/messages.nim
+++ b/fluffy/network/overlay/messages.nim
@@ -1,0 +1,138 @@
+# Nimbus - Portal Network- Message types
+# Copyright (c) 2021 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+# As per spec:
+# https://github.com/ethereum/stateless-ethereum-specs/blob/master/state-network.md#wire-protocol
+
+{.push raises: [Defect].}
+
+import
+  options,
+  stint, stew/[results, objects],
+  eth/ssz/ssz_serialization
+
+export ssz_serialization, stint
+
+type
+  ByteList* = List[byte, 2048]
+
+  MessageKind* = enum
+    unused = 0x00
+
+    ping = 0x01
+    pong = 0x02
+    findnode = 0x03
+    nodes = 0x04
+
+  PingMessage* = object
+    enrSeq*: uint64
+    subProtocolId*: ByteList
+    subProtocolPayload*: ByteList
+
+  PongMessage* = object
+    enrSeq*: uint64
+    subProtocolId*: ByteList
+    subProtocolPayload*: ByteList
+
+  FindNodeMessage* = object
+    subProtocolId*: ByteList
+    distances*: List[uint16, 256]
+
+  NodesMessage* = object
+    subProtocolId*: ByteList
+    total*: uint8
+    enrs*: List[ByteList, 32] # ByteList here is the rlp encoded ENR. This could
+    # also be limited to 300 bytes instead of 2048
+
+  Message* = object
+    case kind*: MessageKind
+    of ping:
+      ping*: PingMessage
+    of pong:
+      pong*: PongMessage
+    of findnode:
+      findNode*: FindNodeMessage
+    of nodes:
+      nodes*: NodesMessage
+    else:
+      discard
+
+  OverlayMessage* =
+    PingMessage or PongMessage or
+    FindNodeMessage or NodesMessage
+
+template messageKind*(T: typedesc[OverlayMessage]): MessageKind =
+  when T is PingMessage: ping
+  elif T is PongMessage: pong
+  elif T is FindNodeMessage: findNode
+  elif T is NodesMessage: nodes
+
+template toSszType*(x: UInt256): array[32, byte] =
+  toBytesLE(x)
+
+template toSszType*(x: auto): auto =
+  x
+
+func fromSszBytes*(T: type UInt256, data: openArray[byte]):
+    T {.raises: [MalformedSszError, Defect].} =
+  if data.len != sizeof(result):
+    raiseIncorrectSize T
+
+  T.fromBytesLE(data)
+
+proc encodeMessage*[T: OverlayMessage](m: T): seq[byte] =
+  ord(messageKind(T)).byte & SSZ.encode(m)
+
+proc decodeMessage*(body: openarray[byte]): Result[Message, cstring] =
+  # Decodes to the specific `Message` type.
+  if body.len < 1:
+    return err("No message data")
+
+  var kind: MessageKind
+  if not checkedEnumAssign(kind, body[0]):
+    return err("Invalid message type")
+
+  var message = Message(kind: kind)
+
+  try:
+    case kind
+    of unused: return err("Invalid message type")
+    of ping:
+      message.ping = SSZ.decode(body.toOpenArray(1, body.high), PingMessage)
+    of pong:
+      message.pong = SSZ.decode(body.toOpenArray(1, body.high), PongMessage)
+    of findNode:
+      message.findNode = SSZ.decode(body.toOpenArray(1, body.high), FindNodeMessage)
+    of nodes:
+      message.nodes = SSZ.decode(body.toOpenArray(1, body.high), NodesMessage)
+  except SszError:
+    return err("Invalid message encoding")
+
+  ok(message)
+
+template innerMessage[T: OverlayMessage](message: Message, expected: MessageKind): Option[T] =
+  if (message.kind == expected):
+    some[T](message.expected)
+  else:
+    none[T]()
+
+# All our Message variants coresponds to enum MessageKind, therefore we are able to
+# zoom in on inner structure of message by defining expected type T.
+# If expected variant is not active, retrun None
+proc getInnnerMessage*[T: OverlayMessage](m: Message): Option[T] =
+  innerMessage[T](m, messageKind(T))
+
+# Simple conversion from Option to Result, looks like somethif which coul live in
+# Result library.
+proc optToResult*[T, E](opt: Option[T], e: E): Result[T, E] =
+  if (opt.isSome()):
+    ok(opt.unsafeGet())
+  else:
+    err(e)
+
+proc getInnerMessageResult*[T: OverlayMessage](m: Message, errMessage: cstring): Result[T, cstring] =
+  optToResult(getInnnerMessage[T](m), errMessage)
--- a/fluffy/network/overlay/overlay_protocol.nim
+++ b/fluffy/network/overlay/overlay_protocol.nim
@@ -1,0 +1,446 @@
+# Nimbus - Portal Network
+# Copyright (c) 2021 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
+
+import
+  std/[sequtils, sets, algorithm, tables],
+  stew/[results, byteutils], chronicles, chronos,
+  eth/rlp, eth/p2p/discoveryv5/[protocol, node, enr, routing_table, random2],
+  ./messages
+
+export messages
+
+logScope:
+  topics = "overlay"
+
+const
+  OverlayProtocolId* = "overlay".toBytes()
+  Alpha = 3 ## Kademlia concurrency factor
+  LookupRequestLimit = 3 ## Amount of distances requested in a single Findnode
+  ## message for a lookup or query
+  RefreshInterval = 5.minutes ## Interval of launching a random query to
+  ## refresh the routing table.
+  RevalidateMax = 10000 ## Revalidation of a peer is done between 0 and this
+  ## value in milliseconds
+  InitialLookups = 1 ## Amount of lookups done when populating the routing table
+
+type
+  ByteList* = List[byte, 2048]
+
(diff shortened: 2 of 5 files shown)


fluffy/network/content/messages.nim
```diff
@@ -0,0 +1,93 @@
...
+      discard
+
+  ContentProtocolMessage* = FindContentMessage or ContentMessage
+
+template messageKind*(T: typedesc[ContentProtocolMessage]): MessageKind =
+  when T is FindContentMessage: findcontent
+  elif T is ContentMessage: content
+
+template toSszType*(x: auto): auto =
+  x
```
zah, line 61: See my other remark about `toSszType`.

fluffy/network/content/messages.nim
```diff
@@ -0,0 +1,93 @@
...
+
+proc encodeMessage*[T: ContentProtocolMessage](m: T): seq[byte] =
```
zah, line 63: See my other remark about explicit generic parameters.

fluffy/network/overlay/messages.nim
```diff
@@ -0,0 +1,128 @@
...
+  when T is PingMessage: ping
+  elif T is PongMessage: pong
+  elif T is FindNodeMessage: findNode
+  elif T is NodesMessage: nodes
+
+template toSszType*(x: auto): auto =
+  x
+
+proc encodeMessage*[T: OverlayMessage](m: T): seq[byte] =
+  ord(messageKind(T)).byte & SSZ.encode(m)
```
zah, line 78: This is the kind of copy-heavy code that can be eliminated by using `OutputStreams`:

```nim
outputStream.write ord(messageKind(T)).byte
let sszWriter = SszWriter.init(outputStream)
sszWriter.writeValue m
```

fluffy/network/overlay/overlay_protocol.nim
```diff
@@ -0,0 +1,438 @@
...
+
+proc reqResponse[Request: OverlayMessage, Response: OverlayMessage](
+      p: OverlaySubProtocol,
+      toNode: Node,
+      protocol: seq[byte],
+      request: Request
+    ): Future[SubProtocolResult[Response]] {.async.} = 
+    let respResult = await talkreq(p.baseProtocol, toNode, protocol, encodeMessage(request))
+    return respResult
+      .flatMap(proc (x: seq[byte]): Result[Message, cstring] = decodeMessage(x))
```
zah, line 119: If you like this style of programming, consider using syntax sugar such as `=>` (from the `sugar` module):

```nim
eturn respResult
      .flatMap(x => decodeMessage(x))
```

fluffy/network/overlay/overlay_protocol.nim
```diff
@@ -0,0 +1,438 @@
...
+# - setJustSeen and replaceNode on (all) message replies
+# - On incoming portal ping of unknown node: add node to routing table by
+# grabbing ENR from discv5 routing table (might not have it)?
+# - ENRs with portal protocol capabilities as field?
+
+proc subProtocolIdAsList*(p: OverlaySubProtocol): ByteList = List.init(p.subProtocolId, 2048)
+
+proc handlePing(p: OverlaySubProtocol, ping: PingMessage):
+    seq[byte] =
+  let p = PongMessage(
```
zah, line 66: Shadowing the `p` variable which originally has a different type can be quite confusing for the readers of this code in the future.

fluffy/network/overlay/overlay_protocol.nim
```diff
@@ -0,0 +1,438 @@
...
+  stew/[results, byteutils], chronicles, chronos,
+  eth/rlp, eth/p2p/discoveryv5/[protocol, node, enr, routing_table, random2],
+  ./messages
+
+logScope:
+  topics = "overlay"
+
+const
+  OverlayProtocolId* = "overlay".toBytes()
+  Alpha = 3 ## Kademlia concurrency factor
```
zah, line 21: According to NEP1, constants should use camelCase as well. PascalCase is suggested for types and compile-time variables that store types.

fluffy/network/overlay/overlay_protocol.nim
```diff
@@ -0,0 +1,438 @@
...
+
+    if pendingQueries.len == 0:
+      break
+
+    let query = await one(pendingQueries)
+    trace "Got lookup query response"
+
+    let index = pendingQueries.find(query)
+    if index != -1:
+      pendingQueries.del(index)
```
zah, line 216: You can replace the removed element with the last one in the sequence. This would be a `O(1)` algorithm.

fluffy/network/overlay/overlay_protocol.nim
```diff
@@ -0,0 +1,438 @@
...
+      if not asked.containsOrIncl(n.id):
+        pendingQueries.add(p.lookupWorker(n, target))
+      inc i
+
+    trace "Pending lookup queries", total = pendingQueries.len
+
+    if pendingQueries.len == 0:
+      break
+
+    let query = await one(pendingQueries)
```
zah, line 211: What should happen when some of the requests fail? It seems to me that you would want to ignore the error here.

fluffy/network/overlay/messages.nim
```diff
@@ -0,0 +1,128 @@
...
+    PingMessage or PongMessage or
+    FindNodeMessage or NodesMessage
+
+template messageKind*(T: typedesc[OverlayMessage]): MessageKind =
+  when T is PingMessage: ping
+  elif T is PongMessage: pong
+  elif T is FindNodeMessage: findNode
+  elif T is NodesMessage: nodes
+
+template toSszType*(x: auto): auto =
```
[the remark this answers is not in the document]
KonradStaniec: so without it, I get error: 
```
nimbus-eth1/fluffy/network/overlay/messages.nim(95, 25) template/generic instantiation of `decode` from here
nimbus-eth1/vendor/nim-serialization/serialization.nim(68, 13) template/generic instantiation of `readValue` from here
nimbus-eth1/vendor/nim-serialization/serialization.nim(29, 9) template/generic instantiation of `readValue` from here
nimbus-eth1/vendor/nim-eth/eth/ssz/ssz_serialization.nim(229, 19) template/generic instantiation of `isFixedSize` from here
nimbus-eth1/vendor/nim-eth/eth/ssz/types.nim(173, 17) Error: undeclared identifier: 'toSszType'
```

I tried to define it as 
```
template toSszType*(x: OverlayMessage): OverlayMessage =
  x
```

as ultimately each `OverlayMessage` is composed only from ssz basic types (lists and uints) but this end up with
```
template/generic instantiation of `isFixedSize` from here
/nimbus-eth1/vendor/nim-eth/eth/ssz/types.nim(173, 26) Error: type mismatch: got <uint64>
but expected one of: 
template toSszType(x: OverlayMessage): OverlayMessage
  first type mismatch at position: 1
  required type for x: OverlayMessage
  but expression 'declval T0' is of type: uint64

expression: toSszType(declval T0)
```

Ultimately in https://github.com/status-im/nimbus-eth1/blob/master/fluffy/content.nim#L50, I have found how Kim have dealt with this situation and decided go this way.

fluffy/network/overlay/messages.nim
```diff
@@ -0,0 +1,128 @@
...
+template innerMessage[T: OverlayMessage](message: Message, expected: MessageKind): Option[T] =
+  if (message.kind == expected):
+    some[T](message.expected)
+  else:
+    none[T]()
+
+# All our Message variants coresponds to enum MessageKind, therefore we are able to
+# zoom in on inner structure of message by defining expected type T.
+# If expected variant is not active, retrun None
+proc getInnnerMessage*[T: OverlayMessage](m: Message): Option[T] =
```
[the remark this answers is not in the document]
KonradStaniec: thanks for detailed info! I am definitly more accustomed to having `[T]` syntax for generic types and I assumed there is not much difference between `[T]` and `(T: typedesc)`.

fluffy/network/content/content_protocol.nim
```diff
@@ -0,0 +1,122 @@
...
+# TODO here we assume that whoever uses conten protocol, will provide not started
+# overlaySubprotocol, so start of content will equal to start of overlaySubprotocol
+# this may be error prone.
+proc start*(p: ContentSubprotocol) =
+  p.overlaySubprotocol.start()
+
+proc stop*(p: ContentSubprotocol) =
+  p.overlaySubprotocol.stop()
+
+proc messageHandler(protocol: TalkProtocol, request: seq[byte]): seq[byte] =
```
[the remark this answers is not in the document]
KonradStaniec: this would probably need to be improved upstream in discoveryv5 as this the shape of the function required by discoveryv5 for any `TalkProtocol`

fluffy/network/content/messages.nim
```diff
@@ -0,0 +1,93 @@
...
+    of content:
+      content*: ContentMessage
+    else:
+      discard
+
+  ContentProtocolMessage* = FindContentMessage or ContentMessage
+
+template messageKind*(T: typedesc[ContentProtocolMessage]): MessageKind =
+  when T is FindContentMessage: findcontent
+  elif T is ContentMessage: content
```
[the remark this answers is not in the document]
KonradStaniec: I think this is because we are not checking on enum cases but rather on types themselves which are defined as:
`ContentProtocolMessage* = FindContentMessage or ContentMessage` all cases are actually covered.

fluffy/network/overlay/messages.nim
KonradStaniec: great suggestion! although for now I would leave it as it is and leave some note in code about possible improvement as:

1. This code most probably will change several times as specs are still in flux
2. we already use similar pattern in other places, and it would probably be better to do this improvement in all places in separate pr

fluffy/network/overlay/overlay_protocol.nim
KonradStaniec: yep I am from functional programming background , so `flatMaps` are kinda most used primitive there. Thanks for suggestion about `sugar` module it looks much cleaner now.

fluffy/network/overlay/overlay_protocol.nim
KonradStaniec: hmm this is exactly what is happening here https://nim-lang.org/docs/system.html#del%2Cseq%5BT%5D%2CNatural

fluffy/network/overlay/overlay_protocol.nim
KonradStaniec: lookup procedure is carbon copy of the one used in discoveryv5 and yes for now intent here and there is to ignore the errrors. There is even todo to tackle this `# TODO: Remove node on timed-out query?` but I would tackle it in separate pr.

fluffy/network/overlay/messages.nim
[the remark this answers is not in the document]
zah: You should be able to add a single catch-all definition in `nimbus-eth1/vendor/nim-eth/eth/ssz/ssz_serialization.nim`. Once you have that definition, use `toSszType` only for types that must be remapped to one of the built-in ssz types (the ones that appear in the SSZ spec).

fluffy/network/content/messages.nim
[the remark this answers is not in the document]
zah: I know why it works now. In a `when` statement, not all branches need to produce a value. Only the one that is selected is type checked:

```nim
var x = when true:
  "test"
else:
  discard

```

fluffy/network/overlay/overlay_protocol.nim
zah: Interesting, didn't know that :)

fluffy/network/overlay/overlay_protocol.nim
zah: Just to be clear, my point was that `await` can raise an exception that you are not handling.

fluffy/network/overlay/messages.nim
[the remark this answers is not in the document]
KonradStaniec: Sounds like good plan! I have created issue in upstream repo (https://github.com/status-im/nim-eth/issues/391). I would then after this pr:
1. Add definition in upstream repo
2. Fix all catch all remapping in fluffy repo

fluffy/network/overlay/overlay_protocol.nim
KonradStaniec: My understanding here was that `one` procedure does not throw exceptions by itself, and query errors are handled in `lookupWorker` procedure by checking if query result `isOk`. 
Or is it the case that every `await` can throw and should be enclosed in `try .. except CatchableError as err..` ?

fluffy/tests/test_content.nim
```diff
@@ -0,0 +1,93 @@
+import
+  chronos, testutils/unittests, stew/shims/net,
+  eth/keys, eth/p2p/discoveryv5/routing_table,
+  eth/p2p/discoveryv5/protocol as discv5_protocol,
+  eth/ssz/ssz_serialization,
+  ../network/content/content_protocol,
+  ../network/overlay/overlay_protocol,
+  ./test_helpers
+
+procSuite "Content Tests":
```
arnetheduck, line 10: why procSuite? in `unittest2` it should generally not be needed

fluffy/network/content/messages.nim
```diff
@@ -0,0 +1,93 @@
...
+  FindContentMessage* = object
+    subProtocolId*: ByteList
+    contentKey*: ContentKey
+
+  ContentMessage* = object
+    subProtocolId*: ByteList
+    enrs*: List[ByteList, 32]
+    payload*: Content
+
+  Message* = object
```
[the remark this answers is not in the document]
pipermerriam: I think `Union` types are formally part of SSZ spec and not very hard to implement: https://ssz.dev/eth2specs/ssz/simple-serialize#vectors-containers-lists-unions

I would recommend we go that direction so that we don't have to implement the "must-only-be-exactly-one-of-these-few-fields" logic at the application layer each time.

kdeme:
Closing in favor of https://github.com/status-im/nimbus-eth1/pull/843 and https://github.com/status-im/nimbus-eth1/pull/856

I'll be taking (no overlay related) review comments into account.

PR #793, opened 2021-08-10
clique: connect period and epoch from chain_config to engine

jangko:
transfer cliquePeriod and cliqueEpoch from chain_config to
PoA engine.


mjfh (approved):
Should probably add a function comment that _period_ and _epoch_ from the argument descriptor _cfg_ are modified.

jangko:
Okay, will add more comments

PR #796, opened 2021-08-10
EVMC: Use same host interface for nested calls as top-level, drop the second copy

jlokier:
Prior to this patch, top-level EVM executions and nested EVM executions did their `getStorage` and other requests using a completely different set of host functions.  It was just unfinished, to get top-level "new" EVMC working.

This finishes the job - it stops using the old methods.  Effect:

- Functionality added at the EVMC host level will be used by all EVM calls.
  (The target here is Beam Sync).

- The old set of functions are no longer used, so they can be removed.

- When EVMC host call tracing is enabled (`showTxCalls = true`), it traces the calls from nested EVM executions as well as top-level.


jangko:
I think it's time to remove the old evm and always enable evmc mode + vm2. it will also reduce CI jobs to 1/3.

PR #798, opened 2021-08-11
code cleanup removed unused imports

bmoo0:
I removed unused imports to remove the warning when compiling and changed the 'let' to a 'const' as mentioned in the PR [here](https://github.com/status-im/nimbus-eth1/blob/87b54b33219de6555770ba9b3492038d5a411081/nimbus/vm2/interpreter/gas_costs.nim#L435)

### DIFF
--- a/nimbus/vm2/interpreter/gas_costs.nim
+++ b/nimbus/vm2/interpreter/gas_costs.nim
@@ -431,8 +431,7 @@ template gasCosts(fork: Fork, prefix, ResultGasCostsName: untyped) =
 
   # ###################################################################################################
 
-  # TODO - change this `let` into `const` - pending: https://github.com/nim-lang/Nim/issues/8015
-  let `ResultGasCostsName`*{.inject, compileTime.}: GasCosts = block:
+  const `ResultGasCostsName`*{.inject, compileTime.}: GasCosts = block:
     # We use a block expression to avoid name redefinition conflicts
     # with "fixed" and "dynamic"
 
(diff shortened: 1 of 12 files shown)


jangko:
could you please rebase. i wonder why the CI is not triggered

stefantalpalaru:
> i wonder why the CI is not triggered

This is why: https://github.com/status-im/nimbus-eth1/blob/3047c839dc5e39f2b41017c5a993f4aea89064d4/.github/workflows/ci.yml#L6

nimbus/vm2/interpreter/gas_costs.nim
```diff
@@ -431,8 +431,7 @@ template gasCosts(fork: Fork, prefix, ResultGasCostsName: untyped) =
 
   # ###################################################################################################
 
-  # TODO - change this `let` into `const` - pending: https://github.com/nim-lang/Nim/issues/8015
-  let `ResultGasCostsName`*{.inject, compileTime.}: GasCosts = block:
+  const `ResultGasCostsName`*{.inject, compileTime.}: GasCosts = block:
```
jangko, line 434: change `let` to `const` will not work. although in Nim repo claims it has been fixed and backported, it does not work when I test it locally. There are 2 `gas_cost.nim`, one in `nimbus/vm/interpreter`, and the other one in `nimbus/vm2/interpreter`.

bmoo0: Yeah, I can remove this change and look into it in more depth and keep this PR for just removing the unused imports if that makes more sense.

jangko:
closing in favor of #802

PR #805, opened 2021-08-18
Update just seen on message replies

KonradStaniec:
Mirrors logic in underlying discoveryv5 i.e updates just seen of node after successful reply or tries to replace node after failed one.


KonradStaniec:
merging as functionality is pretty small and usefull for future pr-s

PR #813, opened 2021-08-26
Remove Calaveras testnet

Zachinquarantine:
This removes the [Calaveras](https://github.com/ethereum/execution-specs/blob/master/network-upgrades/client-integration-testnets/calaveras.md) ephemeral testnet, because London has passed and IIRC Calaveras was only to be used for London testing.


jangko:
thank you, but we are not using calaveras.json for testnet. we are using it for testing the cli.

PR #814, opened 2021-08-27
Generalize network layer for portal

KonradStaniec:
Makes network layer unaware of content storage, and upper layers of network messages. 

This way we could re-use network layer primitives for other network  as long as they support the same set of messages. (it is probable)

### DIFF
--- a/fluffy/network/state/portal_network.nim
+++ b/fluffy/network/state/portal_network.nim
@@ -11,13 +11,12 @@ type PortalNetwork* = ref object
   portalProtocol*: PortalProtocol
 
 proc getHandler(storage: ContentStorage): ContentHandler =
-  result =
-    proc (contentKey: content.ByteList): ContentResult =
+    return (proc (contentKey: content.ByteList): ContentResult =
       let maybeContent = storage.getContent(contentKey)
       if (maybeContent.isSome()):
         ContentResult(kind: ContentFound, content: maybeContent.unsafeGet())
       else:
-        ContentResult(kind: ContentMissing, contentId: toContentId(contentKey))
+        ContentResult(kind: ContentMissing, contentId: toContentId(contentKey)))
 
 # Further improvements which may be necessary:
 # 1. Return proper domain types instead of bytes
--- a/fluffy/network/state/portal_network.nim
+++ b/fluffy/network/state/portal_network.nim
@@ -1,0 +1,46 @@
+import
+  std/options,
+  stew/results,
+  eth/p2p/discoveryv5/[protocol, node],
+  ./messages, ./content, ./portal_protocol
+
+# TODO expose function in domain specific way i.e operating od state network objects i.e
+# nodes, tries, hashes
+type PortalNetwork* = ref object
+  storage: ContentStorage
+  proto*: PortalProtocol
+
+proc getHandler(storage: ContentStorage): ContentHandler =
+  result =
+    proc (contentKey: ByteList): ContentResult =
+      let maybeContent = storage.getContent(contentKey)
+      if (maybeContent.isSome()):
+        ContentResult(kind: ContentFound, content: maybeContent.unsafeGet())
+      else:
+        ContentResult(kind: ContentMissing, contentId: toContentId(contentKey))
+
+# TODO temporary find content exposing Node, ulimatly this function would use
+# contentLookup, so uper layer would not need to know anything about
+# routing table nodes
+proc findContent*(p:PortalNetwork, key: ContentKey, dst: Node): Future[Option[seq[byte]]] {.async.} = 
+  var maybeContent: Option[seq[byte]] = none[seq[byte]]()
+  let keyAsBytes = encodeKeyAsList(key)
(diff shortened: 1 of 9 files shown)


kdeme (commented):
So this is mostly about moving out the content / content storage part, so that it can be specialized per content network?

I had a similar idea, however I was thinking of a ContentProvider object that could be a case object or perhaps work with inheritance.  Hadn't thought it through yet and this will work too of course.

fluffy/network/state/portal_network.nim
```diff
@@ -0,0 +1,46 @@
...
+proc getHandler(storage: ContentStorage): ContentHandler =
+  result =
+    proc (contentKey: content.ByteList): ContentResult =
+      let maybeContent = storage.getContent(contentKey)
+      if (maybeContent.isSome()):
+        ContentResult(kind: ContentFound, content: maybeContent.unsafeGet())
+      else:
+        ContentResult(kind: ContentMissing, contentId: toContentId(contentKey))
+
+# TODO temporary find content exposing Node, ulimatly this function would use
```
kdeme, line 22: Yes, if we'd have a `PortalNetwork` object, then I think it should contain that contentLookup and any of the lookup code in general.

In fact, I've thought about generalizing or at least moving that code also out of discv5 where possible

KonradStaniec: totally agree 👍  I have pr to implement a content lookup function - https://github.com/status-im/nimbus-eth1/pull/807, and it looks really simillar do node lookup. My next step was to have pr to generalize lookup procedure so it can be used in different networks

KonradStaniec:
> So this is mostly about moving out the content / content storage part, so that it can be specialized per content network?

Exactly. It is also about cleanly separating different layers as imo networking stuff should not now that much about how and where we are storing content. Next step would be to make some of the networking stuff more private as rest of the codebase does not need to know about particular messages we are using.

> I had a similar idea, however I was thinking of a ContentProvider object that could be a case object or perhaps work with inheritance. Hadn't thought it through yet and this will work too of course.

Hmm maybe you can do a quick pr just to show how what you had in mind, and we can compare it and move on with one of proposals?

kdeme:
> Hmm maybe you can do a quick pr just to show how what you had in mind, and we can compare it and move on with one of proposals?

Yeah, I can try something to show what I mean, but we don't need to block this PR for that.
I'm just slightly worried that having several these custom function handlers which might cause to code to become less readable.

Like for example, using a function handler for the talkreq/talkresp protocols, seems natural as it is really custom. Anyone can add any protocol to that. The storage and content ids here, while 'custom', are restricted to only a few versions, one for each content network (perhaps less). So maybe another mechanism might be the better choice here. Anyway, we can figure that out as we go also.

fluffy/network/state/portal_network.nim
```diff
@@ -0,0 +1,42 @@
...
+  ./content, ./portal_protocol
+
+# TODO expose function in domain specific way i.e operating od state network objects i.e
+# nodes, tries, hashes
+type PortalNetwork* = ref object
+  storage: ContentStorage
+  portalProtocol*: PortalProtocol
+
+proc getHandler(storage: ContentStorage): ContentHandler =
+  result =
```
kdeme, line 14: avoid using result

fluffy/network/state/portal_protocol.nim
```diff
@@ -96,25 +112,33 @@ proc handleFindNode(p: PortalProtocol, fn: FindNodeMessage): seq[byte] =
...
       closestNodes = p.routingTable.neighbours(
         NodeId(readUintBE[256](contentId.data)), seenOnly = true)
       payload = ByteList(@[]) # Empty payload when enrs are send
       enrs =
-        closestNodes.map(proc(x: Node): ByteList = ByteList(x.record.raw)) 
+        closestNodes.map(proc(x: Node): ByteList = ByteList(x.record.raw))
     encodeMessage(FoundContentMessage(
       enrs: List[ByteList, 32](List(enrs)), payload: payload))
 
+  of ContentKeyValidationFailure:
```
kdeme, line 135: Looks like we perhaps should allow for not responding with a talkresp message in such cases?

kdeme (approved):

fluffy/network/state/portal_protocol.nim
KonradStaniec: I think thats good idea as there is no good reason to respond to someone who breaks the protocol, although i would wait changing the logic here until this change is introduced in discoveryv5 spec

KonradStaniec:
> Yeah, I can try something to show what I mean, but we don't need to block this PR for that.
> I'm just slightly worried that having several of these custom function handlers might cause the code to become less readable.
> 

Hmm that i think about it, I probably understand more what you had in mind in case of `ContentProvider`. We will probably require more handlers at network layer level (for example to save data when receiving gossip messages), and instead of having them as separate function, it would be better to have them grouped in this separate interface which will serve as bridge between network and storage.

Then this pr is definitly a step in this direction.

PR #827, opened 2021-09-13
Implement custom distance function

KonradStaniec:
Implements custom distance function as described in: https://notes.ethereum.org/h58LZcqqRRuarxx4etOnGQ#Storage-Layout

Implementation differs slightly from spec as it takes into account that we already interpret our `content_id` and `node_id` as unsigned 256 bit integers and not some generic big ints, so we do not need to use modulo operation or `abs`.

### DIFF
--- a/fluffy/network/state/custom_distance.nim
+++ b/fluffy/network/state/custom_distance.nim
@@ -1,7 +1,22 @@
 import stint
 
 const MID* = u256(2).pow(u256(255))
+const MAX* = high(Uint256)
 
+# Custom distance function described in: https://notes.ethereum.org/h58LZcqqRRuarxx4etOnGQ#Storage-Layout
+# The implementation looks different than in spec, due to the fact that in practice
+# we are operating on unsigned 256bit integers instead of signed big ints.
+# Thanks to this we do not need to use:
+#  - modulo operations
+#  - abs operation
+# and the results are eqivalent to function described in spec.
+# 
+# The way it works is as follows. Let say we have integers modulo 8:
+# [0, 1, 2, 3, 4, 5, 6, 7]
+# and we want to calculate minimal distance between 0 and 5.
+# Raw difference is: 5 - 0 = 5, which is larger than mid point which is equal to 4.
+# From this we know that the shorter distance is the one wraping around 0, which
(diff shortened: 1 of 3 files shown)


fluffy/network/state/custom_distance.nim
```diff
@@ -0,0 +1,15 @@
+import stint
+
+const MID* = u256(2).pow(u256(255))
+
+proc distance*(node_id: UInt256, content_id: UInt256): UInt256 =
```
kdeme, line 5: Looks like this is indeed equivalent with the spec algo.
I do think we should put some nice comment that it is the equivalent of the one of the spec. Unless the spec updates it to this version.

fluffy/network/state/custom_distance.nim
```diff
@@ -0,0 +1,15 @@
...
+  let rawDiff = 
+    if node_id > content_id:
+      node_id - content_id
+    else:
+      content_id - node_id
+
+  if rawDiff > MID:
+    u256(0) - rawDiff
```
kdeme, line 13: I don't like this line that much though, I know it is fine but relying on warp-around here is not really that clear and not sure we should allow that in this type of code. Perhaps `high(UInt256) - rawDiff + 1` is better, or perhaps some other rewrite can be done. 

Anyway, I don't mind it that much but it might also simply not get accepted in a spec as it is also a pretty language specific thing.

KonradStaniec: My motivation here for going that way was:

1. I thought that way will be clearer as `-` signalises that we are going backwards on number axis.
2. The unsigned integer overflow has well defined behaviour (it is treated `modulo uint.max+ 1`) so it should be implemented similarly in most languages. But yea not sure how it goes in languages without types like python.

But if you think that `high(UInt256) - rawDiff + 1` is more readable then sure we can go with it. I will also leave a comment how to interpret this result.

kdeme: Yeah, not sure if my suggestion is any more clear. I guess I was more concerned on point 2.  But I think the now adjusted one with MAX is. Anyway, doesn't matter that much.

kdeme (approved):

PR #828, opened 2021-09-14
bump nim-eth module

KonradStaniec:
new version:
- fixes bug in discovery v5
- exposed node verification function which can be used in different contexts

### DIFF
--- a/fluffy/network/state/portal_protocol.nim
+++ b/fluffy/network/state/portal_protocol.nim
@@ -23,7 +23,7 @@ const
   Alpha = 3 ## Kademlia concurrency factor
   LookupRequestLimit = 3 ## Amount of distances requested in a single Findnode
   ## message for a lookup or query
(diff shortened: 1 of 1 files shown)


fluffy/network/state/portal_protocol.nim
```diff
@@ -23,6 +23,8 @@ const
   Alpha = 3 ## Kademlia concurrency factor
   LookupRequestLimit = 3 ## Amount of distances requested in a single Findnode
   ## message for a lookup or query
+  FindNodeResultLimit = 16 ## Maximum amount of ENRs in the total Nodes messages
```
kdeme, line 26: This should be 32 I believe as differently from discv5, the spec for Nodes and FoundContent messages allow for 32 `max_length`. Perhaps rename it also to `EnrsResultLimit` or something similar

kdeme (approved):

PR #829, opened 2021-09-14
Update wire protocol of state network spec

kdeme:
Update of wire protocol following changes in this spec PR: https://github.com/ethereum/stateless-ethereum-specs/pull/70

### DIFF
--- a/fluffy/network/state/messages.nim
+++ b/fluffy/network/state/messages.nim
@@ -19,7 +19,7 @@ export ssz_serialization, stint
 
 type
   ByteList* = List[byte, 2048]
-  Bytes2* = List[byte, 2]
+  Bytes2* = array[2, byte]
 
 
   MessageKind* = enum
--- a/fluffy/network/state/messages.nim
+++ b/fluffy/network/state/messages.nim
@@ -17,10 +17,15 @@ import
 
 export ssz_serialization, stint
 
+const
+  contentKeysLimit = 64
+
 type
   ByteList* = List[byte, 2048]
   Bytes2* = array[2, byte]
 
+  ContentKeysList* = List[ByteList, contentKeysLimit]
+  ContentKeysBitList* = BitList[contentKeysLimit]
 
   MessageKind* = enum
     unused = 0x00
@@ -58,11 +63,11 @@ type
     payload*: ByteList
 
   OfferMessage* = object
-    contentKeys*: List[ByteList, 64]
+    contentKeys*: ContentKeysList
 
   AcceptMessage* = object
     connectionId*: Bytes2
-    contentKeys*: BitList[64]
+    contentKeys*: ContentKeysBitList
 
   Message* = object
     case kind*: MessageKind
--- a/fluffy/network/state/messages.nim
+++ b/fluffy/network/state/messages.nim
@@ -19,6 +19,8 @@ export ssz_serialization, stint
 
 type
   ByteList* = List[byte, 2048]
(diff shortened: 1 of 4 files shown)


kdeme:
Keeping as draft for now as spec changes are not merged

fluffy/network/state/messages.nim
```diff
@@ -55,15 +57,12 @@ type
     enrs*: List[ByteList, 32]
     payload*: ByteList
 
-  AdvertiseMessage* = List[ByteList, 32] # No container, heh...
-
-  # This would be more consistent with the other messages
-  # AdvertiseMessage* = object
-  #   contentKeys*: List[ByteList, 32]
+  OfferMessage* = object
+    contentKeys*: List[ByteList, 64]
```
KonradStaniec, line 61: maybe it would be good to move `64` to some kinda `const` and use it in `Offer` and `Accept` , to avoid potential changing and forgetting about another ?

fluffy/network/state/portal_protocol.nim
```diff
@@ -240,6 +244,18 @@ proc findContent*(p: PortalProtocol, dst: Node, contentKey: ByteList):
...
+    Future[PortalResult[AcceptMessage]] {.async.} =
+  let offer = OfferMessage(contentKeys: contentKeys)
+
+  trace "Send message request", dstId = dst.id, kind = MessageKind.offer
+
+  return await reqResponse[OfferMessage, AcceptMessage](
+    p, dst, PortalProtocolId, offer)
+
+  # TODO: Actually have to parse the offer message and get the uTP connection
+  # id, and initiate a uTP stream to get the data out.
```
KonradStaniec, line 257: I would expand a comment a little bit here from `and initiate a uTP stream` to `and initiate a uTP stream with given connection id`, as in standard utp interface `connection-id` is never exposed outside and is always generated randomly for each connection, and in our use case we will need to use `connection-id`  generated out side of utp by receiver of our call, so we will need to adjust standard utp interface a little bit.

One thing that is not specified anywhere (or I kinda missed it), what if we receive `connection-id` for which we already have some open stream.  If that happens we will not be able to open new connection with the same `id`, unless old one is closed.

KonradStaniec (approved):

fluffy/network/state/portal_protocol.nim
kdeme: > I would expand a comment a little bit here from `and initiate a uTP stream` to `and initiate a uTP stream with given connection id`

Sure.



> as in standard utp interface `connection-id` is never exposed outside and is always generated randomly for each connection


So that means that in the current (bittorrent) usage, the connection id is not "pre-negotiated" and just accepted as the random value that is received in the first packet its header. And for response packets, connection id + 1 is taken.

I guess that makes sense as I think in this usage the connection is "longer lived" and multiple pieces are transferred on that same connection. It does make you think about the effectiveness in our use case.



> One thing that is not specified anywhere (or I kinda missed it), what if we receive `connection-id` for which we already have some open stream. If that happens we will not be able to open new connection with the same `id`, unless old one is closed.

Yup, I think all you can do (with current spec) is ignore that transfer and the other side should eventually time-out. One could still try to wait out until the current stream with same connection id finishes, but some time-out values should be added in the specification then I think.
Or you could send a packet with another connection id just to make the whole thing fail immediately.

kdeme:
I will go ahead and merge this, even though the spec changes are not merged yet. It most likely will go in this direction, and any additional updates can be tackled separately.

PR #831, opened 2021-09-16
Use custom distance function in state network

KonradStaniec:
Initialise  state network with custom distance calculator (one using the function defined in https://notes.ethereum.org/h58LZcqqRRuarxx4etOnGQ#Storage-Layout). Also add `logDist` and `idAtDistance` functions consistent with defined custom function

### DIFF
--- a/fluffy/network/state/custom_distance.nim
+++ b/fluffy/network/state/custom_distance.nim
@@ -53,7 +53,7 @@ func myLog2Distance(value: UInt256): uint16 =
 
 func atDistance*(id: UInt256, dist: uint16): UInt256 =
   # TODO With current distance function there are always two ids at given distance
-  # so we might as well did: id - u256(dist), maybe it is worth discussing if every client
+  # so we might as well do: id - u256(dist), maybe it is worth discussing if every client
   # should use the same id in this case.
   id + u256(2).pow(dist)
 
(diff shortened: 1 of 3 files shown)


fluffy/network/state/custom_distance.nim
```diff
@@ -30,3 +32,41 @@ proc distance*(node_id: UInt256, content_id: UInt256): UInt256 =
...
+  var comp = value
+  var ret = 0'u16
+  while (comp > 1):
+    comp = comp shr 1
+    ret = ret + 1
+  return ret
+
+func atDistance*(id: UInt256, dist: uint16): UInt256 =
+  # TODO With current distance function there are always two ids at given distance
+  # so we might as well did: id - u256(dist), maybe it is worth discussing if every client
```
kdeme, line 56:
```suggestion
  # so we might as well do: id - u256(dist), maybe it is worth discussing if every client
```

fluffy/tests/test_custom_distance.nim
```diff
@@ -28,3 +28,32 @@ suite "State network custom distance function":
...
+      logDistance(u256(8), u256(0)) == 3
+      logDistance(UInt256.zero, MID) == 255
+      logDistance(UInt256.zero, MID + UInt256.one) == 254
+  
+  test "Calculate id at log distance":
+    let logDistances = @[
+      0'u16, 1, 2, 3, 4, 5, 6, 7, 8
+    ]
+
+    # for each log distance, calulate node-id at given distance from node zero, and then
```
kdeme, line 48:
```suggestion
    # for each log distance, calculate node-id at given distance from node zero, and then
```

kdeme (approved):

Issue #835, opened 2021-09-20
Fix failure of tests to be reported as error


kdeme:
Something is wrong with the Fluffy tests, failure does not return as an error.

PR #836, opened 2021-09-21
Disable mixed usage of cliBuilder to fix incorrect test failure reporting

kdeme:
Some of the tests are build under `cliBuilder:` and some are not. This results in incorrect exit status. 

We have to either run all under cliBuilder or none, opted for none for now as I am not sure this functionality is being used (I don't use it) and it increases compile time (I have not assessed how much).

Resolves https://github.com/status-im/nimbus-eth1/issues/835

### DIFF
--- a/fluffy/tests/all_fluffy_tests.nim
+++ b/fluffy/tests/all_fluffy_tests.nim
@@ -5,8 +5,6 @@
 #  * MIT license ([LICENSE-MIT](LICENSE-MIT) or http://opensource.org/licenses/MIT)
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-import ../../test_macro
-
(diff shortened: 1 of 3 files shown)


kdeme:
Testing how long clean builds take on my machine shows that the time increase is not really substantial (talking few seconds here).

However, a more important issue is that when updating unittest2 (and testutils), the macro does not seem to work any more.
This is what we see as output then:
```
[Suite] Portal Protocol Message Encodings

[Suite] Portal Tests

[Suite] Content Network

[Suite] Discovery Rpc

[Suite] State network custom distance function

[Suite] Bridge client parser suite
```

So, I'll go for this PR (for now?) instead of https://github.com/status-im/nimbus-eth1/pull/838

PR #838, opened 2021-09-21
Make all tests support cliBuilder and enable it

kdeme:
Just to see time increase


kdeme:
close, see https://github.com/status-im/nimbus-eth1/pull/836

PR #839, opened 2021-09-21
Bump unittest2 and testutils

kdeme:
Draft as this will probably show up as empty tests like:

```
[Suite] parse bytecode

[Suite] gasMeter

[Suite] memory

[Suite] stack

[Suite] Genesis

[Suite] Custom Genesis

[Suite] Precompiles

...
```


kdeme:
Tests are not being run properly, as you can see from the time it takes also.

See: https://github.com/status-im/nimbus-eth1/pull/839/checks?check_run_id=3664076604#step:18:173

PR #858, opened 2021-10-06
Jordan/tx pool

mjfh:
Todo
   * verify nim-lang/Nim#8952 (see zah's comment)

This PR has been obsolated by the cleaned up version https://github.com/status-im/nimbus-eth1/pull/933


nimbus/docs.static/dochack.js
```diff
@@ -0,0 +1,1957 @@
+/* Generated by the Nim Compiler v1.5.1 */
```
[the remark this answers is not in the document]
zah: This seems to be a relevant issue.
https://github.com/nim-lang/Nim/issues/8952

As far as I understand, you are saying that using `nim doc` within our project generates documentation files that refer to this file, but it's missing. Sounds like the issue above needs to be updated/re-opened with that description.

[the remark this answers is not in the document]
arnetheduck: we generally don't include generated files like this - if it's needed, a `make doc` command can fetch it from `vendor` (and it should be in gitignore as appropriate)

nimbus/utils/tx_pool.nim
```diff
@@ -0,0 +1,713 @@
...
+  # Pack txs
+  if algoAutoTxsPacker in xp.pAlgoFlags:
+    if xp.pStagedItems:
+      # Incrementally pack txs by appropriate fetching by items from
+      # the `staged` bucket.
+      xp.bucketsUpdatePacked
+      xp.pStagedItems = false
+
+
+proc processJobs(xp: TxPoolRef): int
```
arnetheduck, line 403: why is there a job queue here? why not perform the operations eagerly?

nimbus/utils/tx_pool.nim
```diff
@@ -0,0 +1,713 @@
...
+proc ethBlock*(xp: TxPoolRef): EthBlock
+    {.inline,gcsafe,raises: [Defect,CatchableError].} =
+  ## Getter, retrieves the block made up by the txs from the `packed` bucket.
+  xp.ethBlockAssemble
+
+# core/tx_pool.go(474): func (pool SetGasPrice,*TxPool) Stats() (int, int) {
+# core/tx_pool.go(1728): func (t *txLookup) Count() int {
+# core/tx_pool.go(1737): func (t *txLookup) LocalCount() int {
+# core/tx_pool.go(1745): func (t *txLookup) RemoteCount() int {
+proc nItems*(xp: TxPoolRef): TxTabsItemsCount
```
arnetheduck, line 569: where does this naming come from? it's certainly not canonincal in nim in general, or in our other libraries

nimbus/utils/tx_pool/tx_job.nim
```diff
@@ -0,0 +1,271 @@
...
+    # hide complexity unless really needed
+    when JobWaitEnabled:
+      jobsAvail: AsyncEvent                ## Fired if there is a job available
+
+const
+  txJobIdMax* = ##\
+    ## Wraps around to `1` after last ID
+    999999.TxJobID
+
+{.push raises: [Defect].}
```
arnetheduck, line 104: this goes after the imports, generally

nimbus/utils/tx_pool/tx_job.nim
```diff
@@ -0,0 +1,271 @@
...
+      setHeadArgs*: tuple[
+        head:  Hash256]
+
+
+  TxJobPair* = object     ## Responding to a job queue query
+    id*: TxJobID          ## Job ID, queue database key
+    data*: TxJobDataRef   ## Data record
+
+
+  TxJobRef* = ref object ##\
```
arnetheduck, line 89: Same question: why do we need a "job queue"? What's the advantage, from a Nim perspective (that doesn't have a language-based concurrent scheduler like go does)

nimbus/utils/tx_pool.nim
mjfh: Guess this was discussed before: jobs may come in asynchroneosly

nimbus/utils/tx_pool.nim
mjfh: len() would not be approriate as there is the length of a sub-list and the length of the full list with cascaded sub-lists. I am happy tio use any other name.

nimbus/utils/tx_pool/tx_item.nim
```diff
@@ -0,0 +1,277 @@
...
+
+proc info*(item: TxItemRef): string {.inline.} =
+  ## Getter
+  item.info
+
+proc local*(item: TxItemRef): bool {.inline.} =
+  ## Getter
+  item.local
+
+proc status*(item: TxItemRef): TxItemStatus {.inline.} =
```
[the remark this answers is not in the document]
mjfh: true, at the moment i use that for read-only items while developing, could be removed later

nimbus/utils/tx_pool/tx_job.nim
mjfh: ok, will cange that

nimbus/utils/tx_pool/tx_job.nim
mjfh: same answer as given at the fist question

nimbus/utils/tx_pool/tx_item.nim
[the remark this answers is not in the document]
mjfh: sorry, wrong answer to the right question -- can remove inlines ok

nimbus/utils/tx_pool.nim
arnetheduck: yes - but where do these jobs come in asynchronously, and why is it useful that the _pool_ is asynchronous? This is a massive PR that follows none of our usual guidelines to work incrementally and anchor changes in issues and documented discussion before writing the code - it's very difficult to review, meaning that questions are to be expected.

arnetheduck: > Guess this was discussed before

where was this discussed? can you link to the discussion?

arnetheduck:
* What expectations of number of transactions in pool are there here? 
* Are there any benchmarks? 
* How do we control memory usage such that it doesn't grow unbounded?
* What happens during a reorg, when transactions in a block that gets orphaned no longer are part of the chain?

nimbus/utils/tx_pool.nim
zah: Since block creation is a potentially lengthy process, a queue of some sort is needed in order to accumulate new incoming transactions that arrived while the block is being executed (through the interruptible async-friendly EVM).

nimbus/docs.static/dochack.js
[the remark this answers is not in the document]
mjfh: As an exception to the rule, I did neither find dochack.js nor nimdoc.out.css on ./vendor/. As far as I remember, with a build from scratch on my local system, I had to build the indexes explicitely. I will go back to this later as stated in the Todo, above.

nimbus/utils/tx_pool.nim
arnetheduck: so - let's assume we create one block per month per validator, roughly - is it reasonable to pay the overhead of a queue then?  how does that affect the block creation strategy?

It's important to remember that the block creation of eth1, where blocks are continuously created for mining no longer is relevant to the design of a tx pool - instead, how and when are blocks created in an eth2-driven process, and how does that affect the design of a tx pool in general?

zah: I'm not sure what you imply here. I think it's definitely not reasonable to say that it's OK for the client to lose its network peers while creating a block because creating a block is a rare event. We would also need to be able to run robust simulations of the software where block creation will be a much more frequent event.

zah: The queue doesn't need to be async though, which may make the overall structure of the code simpler.

arnetheduck: > The queue doesn't need to be async though, 

well, this would be one point, yes - ie a queue is a cost, both computationally and conceptually, so you need some benefit to justify it - in this case, it looks like a copy of the go design that has a built-in scheduler, something we generally don't - since it's here, documenting the justification would make sense

mjfh:
This PR has been obsolated by the cleaned up version #933

PR #900, opened 2021-12-07
crash test scenario

mjfh:
details:
  Example code for inspecting nested block chain and accounts cache
  database transaction framework. There seems to be a pathological
  case where the system crashes after a rollback (as appeared in the
  tx-pool packer code.)

### DIFF
--- a/tests/test_txcrash.nim
+++ b/tests/test_txcrash.nim
@@ -1,0 +1,359 @@
+# Nimbus
+# Copyright (c) 2018-2019 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or
+#    http://www.apache.org/licenses/LICENSE-2.0)
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT) or
+#    http://opensource.org/licenses/MIT)
+# at your option. This file may not be copied, modified, or distributed except
+# according to those terms.
+
+import
+  std/[os, sequtils, strformat, strutils],
+  ../nimbus/[chain_config, config, constants, genesis],
+  ../nimbus/db/[accounts_cache, db_chain],
+  #not-needed# ../nimbus/forks,
+  ../nimbus/p2p/chain,
+  #not-needed# ../nimbus/p2p/[dao, executor],
+  ../nimbus/transaction,
+  ../nimbus/transaction/call_evm,
+  ../nimbus/vm_state,
+  ../nimbus/vm_types,
+  ./test_clique/undump,
+  eth/[common, keys, p2p, trie/db],
+  stint,
+  unittest2
+
+type
+  CaptureSpecs = tuple
+    network: NetworkID
+    file: string
+    numBlocks: int
+    numTxs: int
+
+const
+  baseDir = [".", "tests", ".." / "tests", $DirSep] # path containg repo
+  repoDir = ["replay", "status", "test_clique"]     # alternative repo paths
+
+  goerliCapture: CaptureSpecs = (
+    network: GoerliNet,
+    # file: "goerli68161.txt.gz",
+    file: "goerli51840.txt.gz",
+    numBlocks: 22000,  # unconditionally load blocks
+    numTxs:       30)  # txs following (not in block chain)
+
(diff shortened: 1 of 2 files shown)


tests/test_txcrash.nim
```diff
@@ -0,0 +1,284 @@
...
+    block:
+      let
+        accTx1 = vmState.stateDB.beginSavepoint
+        accTx2 = vmState.stateDB.beginSavepoint
+      vmState.stateDB.modBalance(eAddr)
+      vmState.stateDB.commit(accTx2)
+
+      if rollbackOk:
+        vmState.stateDB.rollback(accTx1)
+        dbTx.rollback()
```
jangko, line 124: `dbTx.rollback` here guarantee the `stateRoot` in `stateDB` points to something incomplete/missing nodes in the trie. 
the key point is when you call `dbTx.rollback`, the `stateRoot` generated in line 112 by `stateDB.persist` will become invalid. `stateDB.rollback` only operates on cached accounts, never rollback the `stateRoot` itself. no wonder `stateDB.persist` at line 146 crash.

in general, db.beginTransaction + db.commit + db.dispose/rollback should be used at outer block/scope of stateDB operations, never interleaved with stateDB's operations. 

if you need to call `db.commit/rollback/dispose` in the middle of `stateDB` operations, don't forget to also update it's `stateRoot`(don't ask me how, I've never do that myself :) ).

jangko: btw, `stateRoot` is equivalent to `rootHash` in stateDB/hexary trie.

mjfh: thanks for commenting/contributing -- actually, the test scenario is strictly nested (not interleaved) with doubly nested db.commit/rollback/dispose at the outmost frame

mjfh: > the key point is when you call `dbTx.rollback`, the `stateRoot` generated in line 112 by `stateDB.persist` will become invalid. `stateDB.rollback` only operates on cached accounts, never rollback the `stateRoot` itself. no wonder `stateDB.persist` at line 146 crash.

Could you explain this a bit more, please? How would I know that I rollback the `stateRoot` itself?

jangko: basic usage:
```nim
stateDB.new / init with stateRoot-A
   ...
stateDB.persist         # stateDB will have stateRoot-B
```

wrong usage:
```nim
stateDB.new / init with stateRoot-A
db.beginTx
   ...
stateDB.persist       # stateDB will have stateRoot-B
   ...
stateDB.rollback     # stateDB still have stateRoot-B
db.rollback             # everything between stateRoot-A and stateRoot-B will be purged

##-#-##

db.beginTx
   ....
stateDB.persist     # crash, because stateDB still hold stateRoot-B
db.commit
```

how to fix it:
```nim
stateDB.new / init with stateRoot-A
db.beginTx
   ...
stateDB.persist        # stateDB will have stateRoot-B
   ...
stateDB.rollback     # stateDB still have stateRoot-B
db.rollback             # everything between stateRoot-A and stateRoot-B will be purged

##-#-##
stateDB.recreate with stateRoot-A or a good stateRoot
or set the stateRoot of stateDB to a good stateRoot  _and_ purge internal cache.

db.beginTx
   ...
stateDB.persist     # will not crash, transition from good-stateRoot to new-stateRoot
db.commit
```

mjfh: Thanks really, I will check that out -- still this defies the meaning of a rollback() facility, if that makes sense. BTW, the crash occurs only in the third transaction frame:
```
   db.beginTx
   beginSavepoint, commit, persist
   beginSavepoint, rollback
   db.rollback

   db.beginTx
   beginSavepoint, commit, persist => crash
```
Also, if I look at the snapshot()/commit()/rollback() wrappers in the vm/computaion source, there is no provision for restoring the state root -- can this  be a serious show stopper?

jangko: https://github.com/status-im/nimbus-eth1/blob/10a3946194f3592a8162c504eb51ee95f7703272/nimbus/p2p/chain/persist_blocks.nim#L59

mjfh: I see, ok

mjfh: see re-named branch jordan/accounts-cache-scenario

PR #904, opened 2021-12-10
Jordan/accounts cache scenario

mjfh:
branch renamed from jordan/tx-crash-scenario, mutated to generaral cover of db/accounts tx nesting

### DIFF
--- a/tests/test_accounts_cache.nim
+++ b/tests/test_accounts_cache.nim
@@ -358,7 +358,7 @@ proc runner(noisy = true; capture = goerliCapture) =
 # Main function(s)
 # ------------------------------------------------------------------------------
 
-proc txCrashMain*(noisy = defined(debug)) =
+proc accountsCacheMain*(noisy = defined(debug)) =
   noisy.runner
 
 when isMainModule:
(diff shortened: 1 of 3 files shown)


tests/test_accounts_cache.nim
```diff
@@ -0,0 +1,373 @@
...
+      vmState.stateDB.rollback(accTx)
+
+    # The following statement will cause a crash at the next `persist()` call.
+    dbTx.rollback()
+
+  # In order to survive without an exception in the next `persist()` call, the
+  # following function could be added to db/accounts_cache.nim:
+  #
+  #   proc clobberRootHash*(ac: AccountsCache; root: KeccakHash; prune = true) =
+  #     ac.trie = initSecureHexaryTrie(ac.db, rootHash, prune)
```
jangko, line 185: I have no objection to this PR, but as a reminder, only reinit the `ac.trie` is not enough, you should consider the accounts in the cache too. if there is any accounts in the cache they must in sync with the new rootHash.

mjfh: good point -- will add it to the comment (lest we forget.)

PR #905, opened 2021-12-10
Arrow Glacier fork

jlokier:
Add the new [Arrow Glacier fork](https://eips.ethereum.org/EIPS/eip-4345).

Only the difficulty calculation is changed, but as a new fork it still affects a number of places in the code.

To the best of my knowledge the change is only scheduled on Mainnet.

In addition:

- The fork date comments in `chain_config.nim` have been checked against the real networks, set consistently in UTC instead of random timezones, and made neater.  Maybe we'll keep these when transferring config to a file someday.

- It's added to forkid hash tests (EIP-2124/EIP-2364), of course.

### DIFF
--- a/nimbus/chain_config.nim
+++ b/nimbus/chain_config.nim
@@ -38,6 +38,7 @@ type
     muirGlacierBlock   : Option[BlockNumber]
     berlinBlock        : Option[BlockNumber]
     londonBlock        : Option[BlockNumber]
+    arrowGlacierBlock  : Option[BlockNumber]
     clique             : CliqueOptions
 
   ChainConfig* = object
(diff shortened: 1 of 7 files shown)


jlokier:
> Forks usually come with updated test suites. Anything in the upstream repos we should bump?

The forkid tests have already been updated, straight out of Geth's own.
Aside from that, yes there are quite a few upstream tests. The latest fixtures test suite does cover this fork (unlike Muir Glacier fork, which does not appear in the suite).  

It's not quite as simple as bumping the repos; there's a little extra functionality needed on our side of the test harness, and the upstream fixtures repo has other, substantial changes as well, so I'm doing that upstream bump separately.

Still, I've checked Arrow Glacier is covered, and as it's an almost trivial fork compared with others (except the other Glacier fork) I'm unconcerned about anything subtle being missed, assuming it passes basic tests from upstream when those are merged.

jangko (approved):

PR #913, opened 2021-12-14
combine pragma pragma and `ignore` attribute to make conditional `evm…

jangko:
…` option in config

because nim-confutils not yet able to handle
`when` in object definition, we combine nim pragma pragma
and `ignore` attribute from nim-confutils
to achieve similar effect.


jangko:
closing in favor of #915

PR #916, opened 2021-12-16
Add state and history content key tests

kdeme:
Added tests using test vectors as currently provided in PR https://github.com/ethereum/portal-network-specs/pull/123

Currently not (yet) using the test vector results for history content keys as they are not according to specifications. (Union of containers vs raw containers with content type in it).


kdeme:
I'll merge this and code/test vectors can be adjusted if they do get changed to the Union version for the history content keys.

PR #919, opened 2021-12-23
History content keys

kdeme:
Specs were changed to use SSZ `Union`


kdeme:
I'll merge this as the `Union` change is in the spec, there might be more changes coming with regards to the Endianness in `toContentId` as this is currently not properly specified.

PR #922, opened 2022-01-07
Add uTP over discv5 in portal wire protocol

kdeme:
Still wip

### DIFF
--- a/fluffy/network/wire/portal_protocol.nim
+++ b/fluffy/network/wire/portal_protocol.nim
@@ -14,8 +14,10 @@ import
   std/[sequtils, sets, algorithm],
   stew/results, chronicles, chronos, nimcrypto/hash, bearssl,
   ssz_serialization,
-  eth/rlp, eth/p2p/discoveryv5/[protocol, node, enr, routing_table, random2, nodes_verification],
+  eth/rlp, eth/p2p/discoveryv5/[protocol, node, enr, routing_table, random2,
+    nodes_verification],
   ../../content_db,
+  ./portal_stream,
   ./messages
 
 export messages, routing_table
@@ -34,18 +36,6 @@ const
   initialLookups = 1 ## Amount of lookups done when populating the routing table
 
 type
-  ContentResultKind* = enum
-    ContentFound, ContentMissing, ContentKeyValidationFailure
-
-  ContentResult* = object
-    case kind*: ContentResultKind
-    of ContentFound:
-      content*: seq[byte]
-    of ContentMissing:
-      contentId*: Uint256
-    of ContentKeyValidationFailure:
-      error*: string
-
   ToContentIdHandler* =
     proc(contentKey: ByteList): Option[ContentId] {.raises: [Defect], gcsafe.}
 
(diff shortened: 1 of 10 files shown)


fluffy/network/wire/portal_stream.nim
```diff
@@ -0,0 +1,106 @@
...
+    # communicate that with the peer however. Or have some more async waiting
+    # until a spot becomes free, like with an AsyncQueue. Although the latter
+    # probably can not be used here directly. This system however does needs
+    # some agreement on timeout values of how long a uTP socket may be
+    # "listening" before it times out because of inactivity.
+    # Or, depending on the direction, it might also depend on the time out
+    # values of the discovery v5 talkresp message.
+    # TODO: Should the content key also be stored to be able to validate the
+    # received data?
+    contentRequests*: Table[uint16, ByteList]
```
KonradStaniec, line 37: shouldn't table keys be tuples `(ConnectionId, NodeId)`, to avoid clashes between different peers i.e situtation like

1. we receive `FindContent` from peer `A`. Content is to large so we put into to table, and respond with generated connection id
2. before we receive stream from `A` we receive `FindContent` from `B` which also is to big, so we generate connecitonId (which happens to be the same) and respond
3. Now `A` opens stream and receives content meant for `B`

fluffy/network/wire/portal_stream.nim
```diff
@@ -0,0 +1,106 @@
...
+      if stream.contentRequests.pop(client.connectionId, content):
+        let fut = client.writeAndClose(content.asSeq())
+        return fut
+
+      var contentKeys: ContentKeysList
+      if stream.contentOffers.pop(client.connectionId, contentKeys):
+        let fut = client.readAndClose()
+        return fut
+
+      # TODO: Is there a scenario where this can happen,
```
KonradStaniec, line 77: Imo it shouldn't happen as `AllowConnectionCallback` is fired strictly before `AcceptConnectionCallback` so if connection won't be known then this will never be fired

fluffy/network/wire/portal_protocol.nim
```diff
@@ -144,6 +136,12 @@ proc handleFindContent(p: PortalProtocol, fc: FindContentMessage): seq[byte] =
       else:
         var connectionId: Bytes2
         brHmacDrbgGenerate(p.baseProtocol.rng[], connectionId)
+        # Note: This connection id passed is the `receive_conn_id` from the peer
+        # that is supposed to connect to this node, and thus this node its
+        # `send_conn_id`.
+        # uTP protocol uses BE for all values in the header, incl. connection id
+        let id = uint16.fromBytesBE(connectionId)
```
KonradStaniec, line 143: I wonder, if it would be better to have proc like `proc allowContent(content): Bytes2` on `PortalStream` object, which would generate connectionId and register content for deliver. This way would avoind exposing this `uint16.fromBytesBE` conversions which is implementation detail how utp works.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -175,16 +174,20 @@ proc handleOffer(p: PortalProtocol, o: OfferMessage): seq[byte] =
...
+          contentKeysBitList.setBit(i)
+          discard contentKeys.add(contentKey)
     else:
       # Return empty response when content key validation fails
       return @[]
 
   var connectionId: Bytes2
   brHmacDrbgGenerate(p.baseProtocol.rng[], connectionId)
 
+  let id = uint16.fromBytesBE(connectionId)
```
KonradStaniec, line 186: Simmilar comment as with content handling

fluffy/network/wire/portal_stream.nim
kdeme: Yes, was planning on doing this still, I was thinking of using the `UtpSocketKey.

fluffy/network/wire/portal_protocol.nim
kdeme: Yup, I'm planning to move any access to these in separate calls in PortalStream yes. Same with the connect calls and such probably.

kdeme:
There is more work to do on this, but I'll merge this already so it can be further used.

PR #926, opened 2022-01-12
bump nim-eth and nim-ssz-serialization vendor modules

kdeme:
- nim-eth: uTP updates, ENR improvemens and extra discv5 metrics
- nim-ssz-serialization: require BitList pairs


kdeme:
Also updating the cache keys to be able to get CI back in order on Windows.

PR #927, opened 2022-01-13
move rocksdb support to eth1

arnetheduck:
only used here / causes unnecessary build deps

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -284,6 +284,13 @@ jobs:
           env CC=gcc make ${DEFAULT_MAKE_FLAGS}
           build/nimbus --help
           # CC, GOARCH, and CGO_ENABLED are needed to select correct compiler 32/64 bit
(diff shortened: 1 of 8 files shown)


jlokier:
Nice! This will simplify `nimbus-eth1` as well.

I am a little concerned that future changes in `kvstore.nim` done in `nim-eth` may break things in -eth1 unnoticed because it tends to be -eth2 authors making such changes without testing -eth1, much like the way changes in a published library API base class in OOP tend to break derived classes in applications.

But it's easily outweighed by the flexibility of not having to worry about breaking other users of `nim-eth` when we want to change users of rocksdb.  Also we'll dropping rocksdb in due course. If we still need a small general purpose kvstore after -eth1 has its state DB in place, sqlite is better for that than rocksdb.

jlokier:
However, could you hold off merging this until the Segmentation Faults in the matching PR in `nimbus-eth1` have been fixed or explained? Otherwise this will get in the way of other changes to `nim-eth` needed by -eth1. Thanks!

PR #938, opened 2022-01-21
test environment for studying crash of hexary trie

mjfh:
why:
  the persistent test case will crash unless in genesis.toBlock():
  + pruneTrie is set false, or
  + the directive "tdb.put(emptyRlpHash.data,emptyRlp)" is added right
    before the "for k, v in account.storage:" loop


stefantalpalaru:
See if you're running into this problem: https://github.com/status-im/nimbus-eth2/issues/3121

mjfh:
> See if you're running into this problem: [status-im/nimbus-eth2#3121](https://github.com/status-im/nimbus-eth2/issues/3121)

Thanks, but this is not the problem. For windows, there were stale file locks which prevent the deferred removal of the persistent database.

There are segfault problems on Linux/i386 which I cannot reproduce on my machine .. but I am on it :)

mjfh:
> > See if you're running into this problem: [status-im/nimbus-eth2#3121](https://github.com/status-im/nimbus-eth2/issues/3121)
> 
> Thanks, but this is not the problem. For windows, there were stale file locks which prevented the deferred removal of the persistent database.
> 
> There are segfault problems on Linux/i386 which I cannot reproduce on my machine .. but I am on it :)

But I realise that this might be a simillar issue on the Linix/i386. And yes, it haunts me too.

PR #939, opened 2022-01-23
setup nimbus tx pool for app wide usage

jangko:
currently the txpool is used at:
- sealer engine
- graphql ethapi
- json rpc ethapi
- ws rpc ethapi


mjfh (approved):
:)

PR #947, opened 2022-01-28
Improve utp test app

KonradStaniec:
- nicer signatures for api calls in utp testing app

### DIFF
--- a/fluffy/tools/utp_testing/utp_rpc_types.nim
+++ b/fluffy/tools/utp_testing/utp_rpc_types.nim
@@ -1,0 +1,50 @@
+# Copyright (c) 2022 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
+
+
+import 
+  std/hashes,
+  json_rpc/jsonmarshal,
+  stew/byteutils,
+  eth/p2p/discoveryv5/[enr, protocol],
+  eth/utp/[utp_discv5_protocol, utp_router]
+
+export jsonmarshal, enr
+
+type SKey* = object
+  id*: uint16
+  nodeId*: NodeId
+
+proc `%`*(value: SKey): JsonNode =
+  let hex = value.nodeId.toBytesBE().toHex()
+  let numId = value.id.toBytesBE().toHex()
+  let finalStr = hex & numId
+  newJString(finalStr)
+
+proc fromJson*(n: JsonNode, argName: string, result: var SKey)
+    {.raises: [Defect, ValueError].} =
+  n.kind.expect(JString, argName)
+  let str = n.getStr()
+  let strLen = len(str)
+  if (strLen >= 64):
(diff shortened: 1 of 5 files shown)


fluffy/tools/utp_testing/utp_test_client.nim
```diff
@@ -0,0 +1,17 @@
...
+import
+  std/os,
+  json_rpc/rpcclient,
+  ./utp_rpc_types,
+  ../../rpc/[rpc_types, rpc_discovery_api]
+
+export utp_rpc_types, rpc_types
+
+createRpcSigs(RpcClient, currentSourcePath.parentDir / "utp_test_rpc_calls.nim")
+createRpcSigs(RpcClient, currentSourcePath.parentDir.parentDir.parentDir / "rpc" / "rpc_calls" / "rpc_discovery_calls.nim")
```
kdeme, line 17: Using the `/../` proc from `std/os` instead of 3x parentDir is perhaps a bit nicer

fluffy/tools/utp_testing/utp_rpc_types.nim
```diff
@@ -0,0 +1,50 @@
...
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
+
+
+import 
+  std/hashes,
+  json_rpc/jsonmarshal,
+  stew/byteutils,
+  eth/p2p/discoveryv5/[enr, protocol],
```
kdeme, line 14: Why does `enr` need to be imported & exported here? It isn't used? (protocol I guess for `NodeId`?, although one can also import just `node` for that).

fluffy/tools/utp_testing/utp_test.nim
```diff
@@ -36,45 +38,68 @@ procSuite "Utp integration tests":
...
   # should be successful
-  proc pingTillSuccess(client: RpcHttpClient, enr: JsonNode): Future[void] {.async.}=
+  proc pingTillSuccess(client: RpcHttpClient, enr: Record): Future[void] {.async.}=
     var failed = true
     while failed:
-      let pingRes = (await client.call("ping", %[enr])).getBool()
-      if pingRes:
-        failed = false
+      let futRes = client.discv5_ping(enr)
+      yield futRes
```
kdeme, line 45: Why yield and not await?

fluffy/tools/utp_testing/utp_rpc_types.nim
KonradStaniec: yep it not needed (it was leftover from earlier configuration), fixed

fluffy/tools/utp_testing/utp_test.nim
KonradStaniec: tbh in that case I find `if futRes.completed():` more readable and succinct than enclosing it in `try .. except` block.

kdeme: True, but I think there was some reason for using await and not yield and I vaguely remember it having to do with cancellations

KonradStaniec: hmm ok thats good to know. 

But this is unfortunate then as doc in `asyncloop.nim` mentions that:
```

## The most reliable way to handle exceptions is to use ``yield`` on a future
## then check the future's ``failed`` property. For example:
```

kdeme: Yeah, I might be wrong.

PR #949, opened 2022-01-31
Remove stacktrace and linetrace in Windows i386 builds 

kdeme:


stefantalpalaru:
Have you tried this?

https://github.com/status-im/nimbus-eth2/blob/bef13b6cce244510ba989d3ce2a6176426592604/config.nims#L68

kdeme:
> Have you tried this?
> 

Nope, I will, but would be good to add regardless.

Would be good in general to have a look at the diff with the nimbus-eth2 config.nims and see what is missing.

PR #952, opened 2022-02-02
Add utp test to ci

KonradStaniec:


.github/workflows/fluffy.yml
```diff
@@ -8,6 +8,72 @@ on:
...
+        run: |
+          docker build -t test-utp --no-cache --build-arg BRANCH_NAME=${{ github.ref_name }} fluffy/tools/utp_testing/docker
+      
+      - name: run test app with simulator
+        run: |
+          SCENARIO="drop-rate --delay=15ms --bandwidth=10Mbps --queue=25 --rate_to_client=10 --rate_to_server=10" docker-compose -f fluffy/tools/utp_testing/docker/docker-compose.yml up -d
+      
+      - name: wait 5 seconds for containers to start
+        run: |
+          sleep 5
```
kdeme, line 61: Isn't there a docker command to check if the containers are started?

Or is the issue also about the applications in the container that need to be at a certain stage in startup (I'd guess not as you already have the `repeatTillSuccess` call)?

kdeme (approved):

.github/workflows/fluffy.yml
KonradStaniec: tbh It is mainly for improving debugging experience if something goes wrong as next step is `docker ps -a` which shows if containers are running or if they crashed with what exit code. 

Without this sleep `docker ps -a` sometimes showed that containers are running, but they would fail 1s later.

Issue #943, opened 2022-01-25
Flaky test: Portal State - Random node lookup from each node.


jangko:
```text
[Suite] Portal testnet tests
  [OK] Discv5 - RoutingTableInfo at start
  [OK] Discv5 - Random node lookup from each node
  [OK] Portal State - RoutingTableInfo at start
{"code":-32000,"message":"portal_state_lookupEnr raised an exception","data":"Record not found in DHT lookup."}
    /home/runner/work/nimbus-eth1/nimbus-eth1/fluffy/scripts/test_portal_testnet.nim(114, 16): Check failed: enr == nodeInfo.nodeENR
    enr was (0)
    nodeInfo.nodeENR was (1, id: "v4", ip: 127.0.0.10, secp256k1: 0x0365DD989C9B9B74DA568EE572585BE6E9151E7C18F515CEE37E7749204A873BB6, udp: 9009)
{"code":-32000,"message":"portal_state_lookupEnr raised an exception","data":"Record not found in DHT lookup."}
    /home/runner/work/nimbus-eth1/nimbus-eth1/fluffy/scripts/test_portal_testnet.nim(114, 16): Check failed: enr == nodeInfo.nodeENR
    enr was (0)
    nodeInfo.nodeENR was (1, id: "v4", ip: 127.0.0.8, secp256k1: 0x026DB3F86661F6F5A626EFBD70CE245A9332E2B0F4E467B6ECC3A4EC9FAB38A023, udp: 9007)
{"code":-32000,"message":"portal_state_lookupEnr raised an exception","data":"Record not found in DHT lookup."}
    /home/runner/work/nimbus-eth1/nimbus-eth1/fluffy/scripts/test_portal_testnet.nim(114, 16): Check failed: enr == nodeInfo.nodeENR
    enr was (0)
    nodeInfo.nodeENR was (1, id: "v4", ip: 127.0.0.10, secp256k1: 0x0365DD989C9B9B74DA568EE572585BE6E9151E7C18F515CEE37E7749204A873BB6, udp: 9009)
  [FAILED] Portal State - Random node lookup from each node
  [OK] Portal History - RoutingTableInfo at start
  [OK] Portal History - Random node lookup from each node
Error: Process completed with exit code 1.
```

It fails in different configuration at different time:
linux-i386:  https://github.com/status-im/nimbus-eth1/runs/4932202504?check_suite_focus=true
[...]

PR #953, opened 2022-02-02
Improve the tests of the local testnet

kdeme:
The local testnet test was rather flaky and would occasionally
fail. It has been made more robust by adding the ENRs directly
to the routing table instead of doing some random lookups.

Additionally, the amount of nodes were increased (=64), ip limits
configuration was added, and the bits-per-hop value was set to 1
in order to make the lookups more likely to hit the network
instead of only the local routing table.

Failure is obviously still possible to happen when sufficient
packets get lost. If this turns out to be the case with the current
amount of nodes, we might have to revise the testing strategy here.


kdeme:
Failures in this PR are due to issues with the custom distance function for the state network.
Skipped that test and will merge for now, lets see if it resolves https://github.com/status-im/nimbus-eth1/issues/943

PR #958, opened 2022-02-08
WIP: Super compact account+storage state history

jlokier:
This is work in progress on the replacement account+storage state history database.

Files have been prepared that fit the Mainnet state history history, full archive of all blocks, into < 200 GB, and the other types of storage are expected to fit into < 1 TB total (block bodies, Merkle hashes etc).  (It may be much smaller, I don't want to give an overly-speculative number until we have figures for everything together.)

The state history database is queried by this code.

- There is an option `--db-compare` to specify a file, and when it's used all block processing compares the initial state used by the EVM against the state read from the `--db-compare` file.  This is to validate that the file contains all the correct states and the query algorithm is correct.

- There's a new subcommand `blockExec START END` (`block-exec` was tried but it gets truncated at `-`).  When this is used, blocks are processed from the local database in the range `START` to `END` inclusive, without any networking.  The blocks and states are read from the RocksDB local database, and if `--db-compare` is used, the states in that file are compared with those fetched from the hexary trie in RocksDB.

This version of the code is not intended for merge, but it is close to the code that will be merged, most likely as soon as a Mainnet state history is run successfully using it.


arnetheduck:
superseded by aristo and future designs

PR #975, opened 2022-02-18
Add block bodies to the propagation and lookups

kdeme:
- Read and propagate block bodies and receipts next to the headers
- Add block bodies content (via lookups) to the eth_getBlockByHash
call
- Test the above in test_portal_testnet


kdeme:
I had misinterpreted the original json block data. It is actually the block that is dumped rlp encoded: `[header, txs, uncles]` and not the block header + block body + receipts: `[header, [txs, uncles], receipts]`. So some adjustment is needed and receipts can not be tested yet. I'll look into getting our data dump in the `[header, [txs, uncles], receipts]` format.

PR #986, opened 2022-03-03
Bump nim-eth

KonradStaniec:
Bump nim-eth. Main changes are multiple improvements in uTP.


jangko:
it breaks fluffy tests

KonradStaniec:
yep, there was regression in uTP implementation. Fixed in https://github.com/status-im/nim-eth/pull/483, with more test + logging.

PR #989, opened 2022-03-09
Update terminal total difficulty handling

mjfh:
why:
  Testing against a re[play unit test for Devnet4 made it necessary to
  adjust the TTD handling. Without updated, importing fails at block
  #5646 which is the parent of the terminal PoW block.

also:
  needed to temporarily silence the test_merge.nim unit tests.


nimbus/db/db_chain.nim
```diff
@@ -59,11 +59,17 @@ proc `$`*(db: BaseChainDB): string =
   result = "BaseChainDB"
 
 proc ttd*(db: BaseChainDB): DifficultyInt =
-  if db.config.terminalTotalDifficulty.isSome:
```
zah, line 62: Goerli is an example of a PoA engine that will have a TTD in the future (it will be upgraded to a merge-based network).

Would the check here make sense for it?

tests/replay/pp.nim
```diff
@@ -12,73 +12,21 @@
 ## ----------------------------------------------------
```
zah, line 12: Is the binary file above necessary for something? If it's needed, it might be better to push it to a submodule where we can more easily rewrite the history and truncate the size of the repo.

nimbus/db/db_chain.nim
```diff
@@ -395,6 +401,30 @@ proc persistHeaderToDb*(self: BaseChainDB; header: BlockHeader): seq[BlockHeader
     self.totalDifficulty = score
     result = self.setAsCanonicalChainHead(headerHash)
 
+proc isBlockAfterTtd*(db: BaseChainDB; blockHeader: BlockHeader): bool =
+  ## Check whether the argument block is the *terminal PoW block* and the
+  ## parent is the canonical head.
+
+  # Easy cases: check whether applicable or TTD reached, already.
+  if not db.config.poaEngine or
```
zah, line 409: The `not` operator has a stronger precedence over the `or` operator, so the code here will be parsed like this:

```nim
if (not db.config.poaEngine) or
   (db.config.terminalTotalDifficulty.isNone) or
   (db.config.terminalTotalDifficulty.get <= db.totalDifficulty):
  return false
```

Since `not poaEngine` will be true for most networks, `isBlockAfterTtd` will always return `false` in such networks. I guess this was not the intention of the code.

nimbus/db/db_chain.nim
mjfh: Thanks, so this clause will just go away (3 instances, i guess)

nimbus/db/db_chain.nim
mjfh: Oops, right  -- was copied and re-written from an _and_ condition (but should go away, anyway as discussed above)

tests/replay/pp.nim
mjfh: Sorry I was too quick and did not see the `devnetfour5664.txt.gz` mentioned above. Yes, this can go anywhere as long as it is not used in the standard unit tests.

tests/customgenesis/devnet5.json
```diff
@@ -0,0 +1,867 @@
...
+    "byzantiumBlock": 0,
+    "constantinopleBlock": 0,
+    "petersburgBlock": 0,
+    "istanbulBlock": 0,
+    "berlinBlock": 0,
+    "londonBlock": 0,
+    "mergeForkBlock": 1000,
+    "terminalTotalDifficulty": 500000000000
+  },
+  "genesis" : { /* added for nimbus */
```
jangko, line 17: I'm a bit surprised to know nim-json-serialization accepts comments.

jangko:
LGTM, I will the `test_merge` later.

tests/customgenesis/devnet5.json
mjfh: Json? should allow comments

mjfh: Ahh, ok.

Just for the records, these comments work on Nethermind json as well :)

jangko:
probably the new opcode `prevRandao` not executed properly.

mjfh:
superseded by https://github.com/status-im/nimbus-eth1/pull/992

Jiangzy520:
Hello, sir. My team is testing an eth2 0 requires a lot of goerli. If you have a lot of goerli, I can choose to buy it,

PR #990, opened 2022-03-09
Jordan/experimental merge support2

mjfh:
This is a cleaned up version of https://github.com/status-im/nimbus-eth1/pull/979

Syncing _Nimbus_ against _Devnet5_ with Nimbus can be done with:
  ```
    nimbus \
        --custom-network:./tests/customgenesis/devnet5.json \
        --bootstrap-file:../nimbus-eth1-blobs/custom-network/devnet5-enode.txt \
        ...
```
where _nimbus-eth1-blobs_ is cloned from [here](https://github.com/status-im/nimbus-eth1-blobs). Alternatively, _Devnet5_  configuration can be retrieved from [here](https://github.com/eth-clients/merge-testnets/tree/main/merge-devnet-5). Note that [genesis.json](https://github.com/eth-clients/merge-testnets/blob/main/merge-devnet-5/genesis.json) needs to be converted for Nimbus to work with.

### CODE REVIEW

File: nimbus/sync/protocol_eth66.nim [hunk 1]
```diff
@@ -0,0 +1,247 @@
...
+
+func toHex*(x: KeccakHash): string = x.data.toHex
+macro tracePacket*(msg: static[string], args: varargs[untyped]) =
+  quote do:
+    if tracePackets:
+      trace `msg`, `args`
+
+p2pProtocol eth(version = ethVersion,
+                peerState = PeerState,
+                useRequestIds = true):
```
kdeme, line 57: From doing a quick diff, it seems `useRequestIds` is the only functional change compared to eth/65.
You've also stated that there are barely eth/65 nodes around; mostly some Nethermind nodes (which are also doing chunking?).

Considering the above I'd suggest to just use eth/66 and drop the compile time option. I'd also suggest to put that change in a separate PR so that it can be tested and merged separately. Smaller PRs = easier review, especially as it turns out there aren't that many changes in this eth66 file for example.

mjfh: Yes it is the same, nicely enough.

I have not tested _eth/66_ extensively on  _Mainnet_, or _Goerly_. And on the merge test networks there is a problem with Geth (Although _Nethermind_ does work fine.)  I feel we should be able to back paddle to _eth/65_ auto-magically, at least for a while?

mjfh:
Superseded by several PRs extracted from here

PR #993, opened 2022-03-15
Activate wire protocol eth/66

mjfh:
and:
 Disentangle protocol_eth66.nim from import sections

why:
  Importing the protocol_eth66 module is not necessary. There is
  no need to know too many details of the underlying wire protocol. All
  that is needed will be exported by blockchain_sync.nim.


nimbus/p2p/blockchain_sync.nim
```diff
@@ -89,13 +112,15 @@ proc availableWorkItem(ctx: SyncContext): int =
...
-  var numBlocks = (ctx.endBlockNumber - nextRequestedBlock).toInt
+  var numBlocks = (ctx.endBlockNumber - nextRequestedBlock).truncate(int)
   if numBlocks > maxHeadersFetch:
     numBlocks = maxHeadersFetch
   ctx.workQueue[result] = WantedBlocks(startIndex: nextRequestedBlock, numBlocks: numBlocks.uint, state: Initial)
 
-proc persistWorkItem(ctx: SyncContext, wi: var WantedBlocks): ValidationResult =
-  result = ctx.chain.persistBlocks(wi.headers, wi.bodies)
+proc persistWorkItem(ctx: SyncContext, wi: var WantedBlocks): ValidationResult
+    {.gcsafe, raises: [Defect,CatchableError].} =
```
kdeme, line 121: nitpick: space between Exceptions/Defect raises list

nimbus/p2p/blockchain_sync.nim
```diff
@@ -9,8 +9,16 @@ import
...
-  eth/[p2p, p2p/private/p2p_types, p2p/rlpx, p2p/peer_pool],
-  ../sync/protocol_eth65
+  eth/[p2p, p2p/private/p2p_types, p2p/rlpx, p2p/peer_pool]
+
+# Link to actual wire protocol stacs. This should be the only place where the
+# protocol specs file is linked in. Subsequent module imports should rely on
+# the `ethVersion` constant (set in *protocol_ethXX*) if protocool details are
+# needed to be known.
+import ../sync/protocol_eth66
+export         protocol_eth66
```
kdeme, line 19: nitpick:
```suggestion
export protocol_eth66
```

nimbus/p2p/blockchain_sync.nim
```diff
@@ -89,13 +112,15 @@ proc availableWorkItem(ctx: SyncContext): int =
...
+  var numBlocks = (ctx.endBlockNumber - nextRequestedBlock).truncate(int)
   if numBlocks > maxHeadersFetch:
     numBlocks = maxHeadersFetch
   ctx.workQueue[result] = WantedBlocks(startIndex: nextRequestedBlock, numBlocks: numBlocks.uint, state: Initial)
 
-proc persistWorkItem(ctx: SyncContext, wi: var WantedBlocks): ValidationResult =
-  result = ctx.chain.persistBlocks(wi.headers, wi.bodies)
+proc persistWorkItem(ctx: SyncContext, wi: var WantedBlocks): ValidationResult
+    {.gcsafe, raises: [Defect,CatchableError].} =
+  catchException("persistBlocks"):
```
kdeme, line 122: Are you using this because there is still a "raw" `Exception` indicated as being possibly raised?

nimbus/p2p/blockchain_sync.nim
mjfh: ok, will fix that :)

nimbus/sync/protocol_eth66.nim
```diff
@@ -155,7 +158,7 @@ p2pProtocol eth(version = ethVersion,
     proc blockHeaders(p: Peer, headers: openArray[BlockHeader])
 
   requestResponse:
-    # User message 0x05: GetBlockBodies.
+    # User message 0x05: GetBlockBodhasheses.
```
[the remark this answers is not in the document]
mjfh: yep

nimbus/p2p/blockchain_sync.nim
```diff
@@ -44,6 +55,18 @@ type
...
 
+template catchException(info: string; code: untyped) =
+  try:
+    code
+  except CatchableError as e:
+    raise (ref CatchableError)(msg: e.msg)
+  except Defect as e:
+    raise (ref Defect)(msg: e.msg)
+  except:
+    let e = getCurrentException()
```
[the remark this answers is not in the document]
mjfh: ok

nimbus/p2p/blockchain_sync.nim
mjfh: indeed:

...
[..]/vm_compile_info.nim(23, 10) Warning: *** Compiling with nimvm enabled [User]
[..]/blockchain_sync.nim(120, 22) template/generic instantiation from here
[..]/status/nimbus-eth1/nimbus/p2p/blockchain_sync.nim(122, 21) Error: can raise an unlisted exception: Exception
...

kdeme: OK, it would be good to figure out where that comes from and whether it can really raise that Exception, but that is not in the scope of this PR which is about activating eth/66, so it can be handled later. 
You did bring it kind of in scope of this PR by adding `{.push raises:[Defect].}` and its related changes, so perhaps try to avoid that next time.

nimbus/p2p/blockchain_sync.nim
kdeme: Opinions will differ on whether this improves readability or not.

However, adding spaces to align with certain words on other lines should be avoided. Next time the path changes you have to re-align also the next line.

This is related to the Aligned indents mentioned at https://status-im.github.io/nim-style-guide/formatting.style.html .
And just the general NEP1: https://nim-lang.org/docs/nep1.html#introduction-spacing-and-whitespace-conventions

It should be noted that mostly our older code will be full of these style mistakes, and should not be taken as example.

nimbus/p2p/blockchain_sync.nim
```diff
@@ -9,8 +9,16 @@ import
   std/[sets, options, random, hashes, sequtils],
   chronos, chronicles,
   eth/common/eth_types,
-  eth/[p2p, p2p/private/p2p_types, p2p/rlpx, p2p/peer_pool],
-  ../sync/protocol_eth65
+  eth/[p2p, p2p/private/p2p_types, p2p/rlpx, p2p/peer_pool]
+
+# Link to actual wire protocol stacs. This should be the only place where the
```
kdeme, line 14: I also don't fully understand this comment.
It is indeed good practise to export protocol_eth66 when some of its public definitions are also used in the public API of the module where you are exporting.

However from what I understand it is used here also to replace imports from `protocol_eth66` to `blockchain_sync`, while nothing specifically from the `blockchain_sync` code is used? Or am I wrong on this?

mjfh: There were lots of paces where protocol_eth65 was explicitly imported. Also, at some places a hard coded `65` protocol version constant was used. Importing by a neutral module name as `blockchain_sync` digging to the file system tree and fixing individual logic can be avoided.

Any suggestion foe a better comment I can replace with?

kdeme: I'm not against having a neutral module name, just that in the current way it is done it imports additional unused code in some tests.

You could instead have a `protocol_eth` file that imports & exports `protocol_eth66`. Then next time a version changes you only need to change it there. And you could still add compile time flags for static version selection in that file (like you were doing before in the other PR), if you'd want/need that.

(Else, if you want to support just 1 version and drop the rest, you can also just rename `protocol_eth66` to `protocol_eth`. And then just adjust code in that file when newer versions come.)

mjfh: There a lots of things one can do but I do not see that this overly pollutes the name space - it is just nitpicking :) IMHO.

PR #1005, opened 2022-03-22
Handle EIP-3675 (TheMerge) switch block

mjfh:
why:
  If configured/activated, this block is advertised as FORK_NEXT in the
  status(0x01) message of the RPLx wire protocol.

### DIFF
--- a/nimbus/chain_config.nim
+++ b/nimbus/chain_config.nim
@@ -39,6 +39,12 @@ type
     berlinBlock        : Option[BlockNumber]
     londonBlock        : Option[BlockNumber]
     arrowGlacierBlock  : Option[BlockNumber]
+
+    mergeForkBlock     : Option[BlockNumber] ##\
+      ## EIP-3675 (TheMerge) switch block: "For the purposes of the EIP-2124\
+      ## fork identifier, nodes implementing this EIP *MUST* set the\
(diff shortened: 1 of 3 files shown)


nimbus/chain_config.nim
```diff
@@ -63,6 +69,7 @@ type
     berlinBlock*        : BlockNumber
     londonBlock*        : BlockNumber
     arrowGlacierBlock*  : BlockNumber
+    mergeForkBlock*     : Option[BlockNumber] # EIP-3675 (TheMerge) switch block
```
jangko, line 72: Do we really need `Option[BlockNumber]` here? I think without using `Option`, forkId calculation still correct.

mjfh: The difference to the go ref implemewntation is that some block numbers are optional (formally all) and the ones defined must be monotonous. Only the last two are consecutive options.

Maybe there is another way though ...

jangko: > What happens if arrowGlacierBlock is BlockNumber.high and mergeForkBlock is 1000 (as in the current test situation)?

set both to the same block number. so far we don't have problem calculating forkId without `Option`.

see `tests/test_forkid.nim`

PR #1019, opened 2022-03-29
Support for local accounts

mjfh:
Sneak preview

### DIFF
--- a/nimbus/utils/tx_pool.nim
+++ b/nimbus/utils/tx_pool.nim
@@ -10,10 +10,6 @@
 
 ## TODO:
 ## =====
-## * Support `local` accounts the txs of which would be prioritised. This is
-##   currently unsupported. For now, all txs are considered from `remote`
-##   accounts.
-##
 ## * No uncles are handled by this pool
 ##
 ## * Impose a size limit to the bucket database. Which items would be removed?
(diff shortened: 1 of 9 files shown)


jangko:
Can we add something like `proc addRemote(xp: TxPoolRef; tx: Transaction)` and `proc addLocal(xp: TxPoolRef; tx: Transaction)`? it will be much more easier to use.

mjfh:
> Can we add something like `proc addRemote(xp: TxPoolRef; tx: Transaction)` and `proc addLocal(xp: TxPoolRef; tx: Transaction)`? it will be much more easier to use.

There is no problem if this would add the TX and tag the account implicitly.

To avoid unwanted side effects,  I would do it with a prototype like `proc addRemote(xp: TxPoolRef; tx: Transaction; force = false): bool` where the _force_ flag always untags the account, with `force = false` it adds the TX only if the underlying account is already untagged (aka _remote_.)

Another possibility would be to tag a sequence of TXs (chained by consecutive _nonces_) but this is more effort to design cleanly ...

nimbus/utils/tx_pool.nim
```diff
@@ -845,30 +849,80 @@ proc disposeItems*(xp: TxPoolRef; item: TxItemRef;
...
+  ## Tag argument `account` local which means that the transactions from this
+  ## account -- together with all other local accounts -- will be considered
+  ## first for packing.
+  xp.txDB.setLocal(account)
+
+proc resLocal*(xp: TxPoolRef; account: EthAddress) =
+  ## Untag argument `account` as local which means that the transactions from
+  ## this account -- together with all other untagged accounts -- will be
+  ## considered for packing after the locally tagged accounts.
+  xp.txDB.setLocal(account)
```
jangko, line 865: resLocal?

jangko (approved):

Issue #870, opened 2021-10-19
Fluffy ContentDB needs a query furthest by distance function


kdeme:
[ContentDB](https://github.com/status-im/nimbus-eth1/blob/master/fluffy/content_db.nim#L33) needs a call that queries furthest by distance: Retrieval of the content from the database which is furthest from a provided node_id.

I believe this is mostly to be used to call from the local nodes own `node_id`, in order to be able to purge data furthest away when a node would reach its storage limit.

PR #1026, opened 2022-03-31
Add api to get n furthest elements from db

KonradStaniec:
Partially solves: https://github.com/status-im/nimbus-eth1/issues/870

caveats:

Currently works by querying for all elements in database and doing all necessary 
work on program level. This is mainly due to two facts:
- sqlite does not have build xor function, also it does not handle bitwise
operations on blobs as expected
- our nim wrapper for sqlite does not support create_function api of sqlite
so we cannot create custom function comparing blobs at sql level. If that 
would be possible we may be able to all this work by one sql query

Also for now using hardcoded xor function (as this is the one used for history network). Before customizing distance function it would be good to decide how do we want to deal with different content dbs for each network i.e answet question on top of the file:
```
# 1. More kvstores are added per network, and thus depending on the network a
# different kvstore needs to be selected.
# 2. Or more kvstores are added per network and per content type, and thus
# content key fields are required to access the data.
# 3. Or databases are created per network (and kvstores pre content type) and
# thus depending on the network the right db needs to be selected
```
At this point I am slightly leaning in direction of having different db (i.e different sqlite file) for each network, then each network can manage its own db in the way that is most efficient. The problem would be of course handling json rpc queries which must combine data from different networks, but I am not sure that this will be big problem.

### DIFF
--- a/fluffy/content_db.nim
+++ b/fluffy/content_db.nim
@@ -8,7 +8,7 @@
 {.push raises: [Defect].}
 
 import
-  std/options,
+  std/[options, heapqueue],
   eth/db/kvstore,
   eth/db/kvstore_sqlite3,
   stint,
@@ -30,10 +30,24 @@ export kvstore_sqlite3
 # thus depending on the network the right db needs to be selected.
 
 type
+  RowInfo = tuple
+    contentId: array[32, byte]
+    payloadLength: int64
+
+  ObjInfo* = object
+    contentId*: array[32, byte]
+    payloadLength*: int64
+    distFrom*: UInt256
+
   ContentDB* = ref object
     kv: KvStoreRef
     sizeStmt: SqliteStmt[NoParams, int64]
     vacStmt: SqliteStmt[NoParams, void]
+    getAll: SqliteStmt[NoParams, RowInfo]
+
+# we want objects to be sorted from largest distance to closests
+proc `<`(a, b: ObjInfo): bool = 
+  return a.distFrom < b.distFrom
 
 template expectDb(x: auto): untyped =
   # There's no meaningful error handling implemented for a corrupt database or
@@ -47,7 +61,7 @@ proc new*(T: type ContentDB, path: string, inMemory = false): ContentDB =
         "working database (out of memory?)")
     else:
       SqStoreRef.init(path, "fluffy").expectDb()
-
+  
   let getSizeStmt = db.prepareStmt(
     "SELECT page_count * page_size as size FROM pragma_page_count(), pragma_page_size();",
     NoParams, int64).get()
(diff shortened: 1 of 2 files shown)


fluffy/content_db.nim
```diff
@@ -47,7 +61,7 @@ proc new*(T: type ContentDB, path: string, inMemory = false): ContentDB =
         "working database (out of memory?)")
     else:
       SqStoreRef.init(path, "fluffy").expectDb()
-
+
```
kdeme, line 64: whitespace

fluffy/content_db.nim
```diff
@@ -56,7 +70,59 @@ proc new*(T: type ContentDB, path: string, inMemory = false): ContentDB =
     "VACUUM;",
     NoParams, void).get()
 
-  ContentDB(kv: kvStore db.openKvStore().expectDb(), sizeStmt: getSizeStmt, vacStmt: vacStmt)
+  let kvStore = kvStore db.openKvStore().expectDb()
+
+  # this need to go after opening openKvStore, as it checks that table name kvstore
+  # already exists.
```
kdeme, line 76:
```suggestion
  # this need to go after `openKvStore`, as it checks that the table name kvstore
  # already exists.
```

fluffy/content_db.nim
```diff
@@ -56,7 +70,59 @@ proc new*(T: type ContentDB, path: string, inMemory = false): ContentDB =
...
+  let getKeysStmt = db.prepareStmt(
+    "SELECT key, length(value) FROM kvstore",
+    NoParams, RowInfo
+  ).get()
+
+  ContentDB(kv: kvStore, sizeStmt: getSizeStmt, vacStmt: vacStmt, getAll: getKeysStmt)
+
+proc getNFurtherstElements*(db: ContentDB, distanceFrom: UInt256, n: uint64): seq[ObjInfo] =
```
kdeme, line 84: Perhaps call `distanceFrom` instead `target` or `targetId`. This naming gets a bit confusing later when actually doing the distance calculation

kdeme, line 84: Typo:
```suggestion
proc getNFurthestElements*(db: ContentDB, distanceFrom: UInt256, n: uint64): seq[ObjInfo] =
```

fluffy/content_db.nim
```diff
@@ -56,7 +70,59 @@ proc new*(T: type ContentDB, path: string, inMemory = false): ContentDB =
...
+  ## Get at most n furthest elements from database in order from furtherst to closest.
```
kdeme, line 85:
```suggestion
  ## Get at most n furthest elements from database in order from furthest to closest.
```

fluffy/content_db.nim
```diff
@@ -56,7 +70,59 @@ proc new*(T: type ContentDB, path: string, inMemory = false): ContentDB =
...
+  if n == 0:
+    return newSeq[ObjInfo]()
+
+  var heap = initHeapQueue[ObjInfo]()
+
+  var ri: RowInfo
+  for e in db.getAll.exec(ri):
+    let contentId = UInt256.fromBytesBE(ri.contentId)
+    # TODO: Currently it asumes xor distance, but when we start testing networks with
+    # other distance functions this may not necessary be true expected
```
kdeme, line 108:
```suggestion
    # TODO: Currently it assumes xor distance, but when we start testing networks with
    # other distance functions this needs to be adjusted to the custom distance function
```

fluffy/tests/test_content_db.nim
```diff
@@ -20,7 +22,20 @@ proc genByteSeq(length: int): seq[byte] =
     inc i
   return resultSeq
 
+proc generateNRanomU256(rng: var BrHmacDrbgContext, n: int): seq[UInt256] =
```
kdeme, line 25:
```suggestion
proc generateNRandomU256(rng: var BrHmacDrbgContext, n: int): seq[UInt256] =
```

kdeme (approved):
I think this will do for now, especially considering it is an action that should not occur frequently.

I did have the idea before of trying to use the length of the "common prefix" as indicator of how far an id is from another (you do loose granularity of course for the LSBs). If that works, such approach would even be better if we had some sorted database. Anyhow, food for thought.

One of the reasons why I'm thinking of the above, is that when you want to offer content, you might want to load a group of content items that are close to each other / close to a specific id.

KonradStaniec:
> I think this will do for now, especially considering it is an action that should not occur frequently.
> 
> I did have the idea before of trying to use the length of the "common prefix" as indicator of how far an id is from another (you do loose granularity of course for the LSBs). If that works, such approach would even be better if we had some sorted database. Anyhow, food for thought.
> 
> One of the reasons why I'm thinking of the above, is that when you want to offer content, you might want to load a group of content items that are close to each other / close to a specific id.

hmm if we would like to keep database sorted then we should probably use rocksdb instead of sqlite. In rocks db one can define custom comparator function for keys, and then whole db is sorted in that order. With such approach, finding keys furthest from some id would as simple as creating db iterator [iter](https://github.com/facebook/rocksdb/wiki/Iterator#introduction) and setting it up on the end.

I am not sure if sqlite got api do define table ordering other that adding `ORDER BY` clause to select statement, for which we would need to define some custom compassion functions to have results ordered by length of common prefix

PR #1035, opened 2022-04-04
snappy revamp

arnetheduck:
not entirely sure why nimbus-eth1 has a copy of snappy but here's a new
version: https://github.com/status-im/nim-snappy/pull/10


jangko:
why nimbus has a copy of snappy: https://eips.ethereum.org/EIPS/eip-706

arnetheduck:
@jangko but is it actually used? the changes are kind of significant but things seem to compile fine anyway

kdeme:
It doesn't look like this is ever defined: https://github.com/status-im/nim-eth/blob/e62fdfe6f1c8cbef48768a4d12fb2929518649d8/eth/p2p/private/p2p_types.nim#L15

Probably hasn't been tested with it enabled since ages.

jangko:
that's right. DEVp2p snappy compression was implemented before EIP-706 finalized. and since then forgotten. but we can revive it

PR #1036, opened 2022-04-05
Comment out unnecessary action in unit test

mjfh:


tests/test_txpool2.nim
```diff
@@ -289,10 +289,12 @@ proc runTxHeadDelta*(noisy = true) =
           # Commit to block chain
           check chain.persistBlocks([blk.header], [body]).isOk
 
-          # PoS block canonical head must be explicitly set using setHead. The
-          # function `persistHeaderToDb()` used in `persistBlocks()` does not
-          # reliably do so due to scoring.
-          chainDB.setHead(blk.header)
+          # I not for other reason, setting head is irrelevant for this test
+          #
+          # # PoS block canonical head must be explicitly set using setHead.
```
kdeme, line 294: It's fine to just remove the piece of code if it is not needed.

If for some reason you need to get it back, that's what git is for.

PR #1037, opened 2022-04-05
Inital poke implementation

KonradStaniec:
Initial implementation of POKE mechanism i.e offering content to nodes which we discovered do not have it during content lookup.

One caveat for now:
- Currently, due to how code is structured, we may only offer content which is in our db i.e it is in our range. It means we cannot offer content to nodes in case when content is in their range and not in our range. There is one one `todo` in offer procedure which would fix it i.e passing content directly of `offer` procedure, instead of using db as communication channel. It would also cause problem when we add deleting stuff from db, as offer-accept is asynchronous  so the content could be deleted before be send it via uTP.

### DIFF
--- a/fluffy/network/history/history_network.nim
+++ b/fluffy/network/history/history_network.nim
@@ -130,11 +130,21 @@ proc getBlockHeader*(
 
   let headerContent = maybeHeaderContent.unsafeGet()
 
-  let maybeHeader = validateHeaderBytes(headerContent, hash)
+  let maybeHeader = validateHeaderBytes(headerContent.content, hash)
 
   # content is in range and valid, put into db
   if maybeHeader.isSome() and h.portalProtocol.inRange(contentId):
-     h.contentDB.put(contentId, headerContent)
+    # TODO this bit is quite troubling, currently we may trigger offer/accept
+    # only when content is in our db and we save content only when is in our range
+    # which means we cannot propagate content which is not in our range, but maybe
+    # in range of other nodes. 
+    h.contentDB.put(contentId, headerContent.content)
+    # content is valid and in db we may propagate it through the network
+    h.portalProtocol.triggerPoke(
+      headerContent.nodesInterestedInContent,
+      keyEncoded, 
+      contentId
+    )
 
   return maybeHeader
 
@@ -164,7 +174,7 @@ proc getBlock*(
 
   let bodyContent = maybeBodyContent.unsafeGet()
 
-  let maybeBody = validateBodyBytes(bodyContent, header.txRoot, header.ommersHash)
+  let maybeBody = validateBodyBytes(bodyContent.content, header.txRoot, header.ommersHash)
 
   if maybeBody.isNone():
     return none(Block)
@@ -173,7 +183,17 @@ proc getBlock*(
 
   # content is in range and valid, put into db
   if h.portalProtocol.inRange(contentId):
-     h.contentDB.put(contentId, bodyContent)
+    # TODO this bit is quite troubling, currently we may trigger offer/accept
+    # only when content is in our db and we save content only when is in our range
+    # which means we cannot propagate content which is not in our range, but maybe
+    # in range of other nodes. 
+    h.contentDB.put(contentId, bodyContent.content)
+    # content is valid and in db we may propagate it through the network
+    h.portalProtocol.triggerPoke(
+      bodyContent.nodesInterestedInContent,
+      keyEncoded, 
+      contentId
+    )
 
   return some[Block]((header, blockBody))
 
--- a/fluffy/network/history/history_network.nim
+++ b/fluffy/network/history/history_network.nim
@@ -139,7 +139,7 @@ proc getBlockHeader*(
     # which means we cannot propagate content which is not in our range, but maybe
     # in range of other nodes. 
     h.contentDB.put(contentId, headerContent.content)
-    # content is valid and in db we may propagate it through the network
+    # content is valid and in the db, it may be propagated it through the network
     h.portalProtocol.triggerPoke(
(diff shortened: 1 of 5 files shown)


fluffy/network/history/history_network.nim
```diff
@@ -173,7 +183,17 @@ proc getBlock*(
 
   # content is in range and valid, put into db
   if h.portalProtocol.inRange(contentId):
-     h.contentDB.put(contentId, bodyContent)
+    # TODO this bit is quite troubling, currently we may trigger offer/accept
```
kdeme, line 186: I was planning on putting the poke in the `contentLookup` proc just before it returns the content (with an `asyncSpawn`). I assume that you had the same in mind, but solved it like this because of the `offer` call using the database, hence the "quit troubling` wording.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -758,11 +781,36 @@ proc lookup*(p: PortalProtocol, target: NodeId): Future[seq[Node]] {.async.} =
...
+    contentKey: ByteList,
+    contentId: ContentId) =
+  ## Trigers asynchronous offer-accept interaction to provided nodes.
+  ## Provided content should be in range of provided nodes
+  ## Provided content should be in database
+  ## TODO Related to todo in `proc offer` it maybe better to pass content to
+  ## offer directly to avoid potential problems when content is not really in database
+  ## this will be especially important when we introduce deleting content
+  ## from database
+  let keys = List[ByteList, contentKeysLimit].init(@[contentKey])
```
kdeme, line 796: You should be able to do `ContentKeysList.init()` directly iirc

fluffy/network/wire/portal_protocol.nim
```diff
@@ -823,20 +880,20 @@ proc contentLookup*(p: PortalProtocol, target: ByteList, targetId: UInt256):
...
-
         portal_lookup_content_requests.observe(requestAmount)
-        return some(content.content)
+        return some(ContentLookupResult.init(content.content, nodesWithoutContent))
     else:
       # TODO: Should we do something with the node that failed responding our
       # query?
       discard
-
+
```
kdeme, line 894: whitespace

fluffy/network/history/history_network.nim
```diff
@@ -130,11 +130,21 @@ proc getBlockHeader*(
...
 
   # content is in range and valid, put into db
   if maybeHeader.isSome() and h.portalProtocol.inRange(contentId):
-     h.contentDB.put(contentId, headerContent)
+    # TODO this bit is quite troubling, currently we may trigger offer/accept
+    # only when content is in our db and we save content only when is in our range
+    # which means we cannot propagate content which is not in our range, but maybe
+    # in range of other nodes. 
+    h.contentDB.put(contentId, headerContent.content)
+    # content is valid and in db we may propagate it through the network
```
kdeme, line 142:
```suggestion
    # content is valid and in the db, it may be propagated it through the network
```

fluffy/network/wire/portal_protocol.nim
```diff
@@ -144,12 +144,28 @@ type
     Content
 
   FoundContent* = object
+    dst*: Node
```
kdeme, line 147:
```suggestion
    src*: Node
```
it is kind of the `src` node with respect to the found content

fluffy/network/wire/portal_protocol.nim
```diff
@@ -168,9 +184,16 @@ func localNode*(p: PortalProtocol): Node = p.baseProtocol.localNode
 func neighbours*(p: PortalProtocol, id: NodeId, seenOnly = false): seq[Node] =
   p.routingTable.neighbours(id = id, seenOnly = seenOnly)
 
+proc inRangeImpl(
```
kdeme, line 187: Can just be called `inRange` as it will be a directly useful proc also, and it shouldn't give naming symbol issues considering it has different parameter list.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -758,11 +781,36 @@ proc lookup*(p: PortalProtocol, target: NodeId): Future[seq[Node]] {.async.} =
   p.lastLookup = now(chronos.Moment)
   return closestNodes
 
+proc triggerPoke*(
+    p: PortalProtocol,
+    nodes: seq[Node], 
+    contentKey: ByteList,
+    contentId: ContentId) =
+  ## Trigers asynchronous offer-accept interaction to provided nodes.
```
kdeme, line 789:
```suggestion
  ## Triggers asynchronous offer-accept interaction to provided nodes.
```

fluffy/network/wire/portal_protocol.nim
```diff
@@ -811,6 +861,13 @@ proc contentLookup*(p: PortalProtocol, target: ByteList, targetId: UInt256):
 
       case content.kind
       of Nodes:
+        let maybeRadius = p.radiusCache.get(content.dst.id)
+        if maybeRadius.isSome() and p.inRangeImpl(content.dst.id, maybeRadius.unsafeGet(), targetId):
+          # we only return nodes which may be interested in content
+          # also we do not need to chec for duplicates in nodesWithoutContent
+          # as we never make requests two times to the same node
```
kdeme, line 868:
```suggestion
          # Only return nodes which may be interested in content.
          # No need to checc for duplicates in nodesWithoutContent
          # as requests are never made two times to the same node.
```

fluffy/network/history/history_network.nim
KonradStaniec: yup, I have also initially planning to add it in contentLookup, but having it here is the only way to make sure that:
a) Content is really in database, before triggering poke
b) Content is really validated before spreading it around

PR #1039, opened 2022-04-05
Enable JWT authentication for websockets

mjfh:


nimbus/nimbus.nim
```diff
@@ -11,19 +11,26 @@ import
...
-  os, strutils, net, options,
-  stew/shims/net as stewNet,
-  eth/keys, db/[storage_types, db_chain, select_backend],
-  eth/common as eth_common, eth/p2p as eth_p2p,
-  chronos, json_rpc/rpcserver, chronicles,
-  eth/p2p/rlpx_protocols/les_protocol,
-  ./sync/protocol_ethxx,
-  ./p2p/blockchain_sync, eth/net/nat, eth/p2p/peer_pool,
+  std/[os, strutils, net, options],
+  "."/[conf_utils, config, constants, context, genesis, sealer, utils, version],
```
kdeme, line 15: The order of imports we typically do is: stdlib, external imports, local imports. See https://status-im.github.io/nim-style-guide/language.import.html

nimbus/rpc/jwt_auth.nim
```diff
@@ -0,0 +1,317 @@
...
+# at your option. This file may not be copied, modified, or distributed except
+# according to those terms.
+#
+# Ackn:
+#   nimbus-eth2/beacon_chain/spec/engine_authentication.nim
+#   go-ethereum/node/jwt_handler.go
+
+import
+  std/[base64, json, options, os, strutils, times],
+  ../config,
```
kdeme, line 17: idem above

nimbus/rpc/jwt_auth.nim
```diff
@@ -0,0 +1,317 @@
...
+  jwtSecretFile* = ##\
+    ## A copy on the secret key in the `dataDir` directory
+    "jwt.hex"
+
+  jwtMinSecretLen* = ##\
+    ## Number of bytes needed with the shared key
+    32
+
+type
+  #[
```
kdeme, line 42: Github doesn't seem to parse the multi-line comments properly, perhaps better not to use it to make it easier for reviewing.

Unused code shouldn't be committed anyhow tbh.

nimbus/rpc/jwt_auth.nim
```diff
@@ -0,0 +1,317 @@
...
+
+              let rc = auth[7..^1].strip.verifyTokenHS256(key)
+              if rc.isOk:
+                return ok()
+
+              debug "Could not authenticate",
+                error = rc.error
+
+              case rc.error:
+              of jwtTokenValidationError, jwtMethodUnsupported:
```
jangko, line 286: iirc we need to reply with Http401 error code "Unauthorized" if token validation failed? 
how CL/beacon client react if we don't reply with Http401?

mjfh: Good point. But see the example `clientVerify()` on line 40 of https://github.com/status-im/nim-websock/blob/main/tests/testhooks.nim. The error return code is used for logging.

tests/test_jwt_auth.nim
```diff
@@ -0,0 +1,293 @@
...
+      let
+        config = @[dataDirCmdOpt,jwtStrippedCmdOpt].makeConfig
+        secret = fakeKey.fakeGenSecret.jwtSharedSecret(config)
+        lines = config.jwtSecret.get.string.readLines(1)
+
+      check secret.isOk
+      check 0 < lines.len
+
+      let
+        hexKey = secret.value.mapIt(it.toHex(2)).join
```
mratsim, line 168: nimcrypto.utils has a constant-time hex conversion: https://github.com/cheatfate/nimcrypto/blob/a5742a9a214ac33f91615f3862c7b099aec43b00/nimcrypto/utils.nim#L150=

nimbus/rpc/jwt_auth.nim
```diff
@@ -0,0 +1,317 @@
...
+      # client using authentication. This keeps it lower-risk initially.
+      warn "Could not write JWT secret to data directory",
+        jwtSecretPath
+    return ok(newSecret)
+
+  try:
+    let lines = config.jwtSecret.get.string.readLines(1)
+    if lines.len == 0:
+      return err(jwtKeyEmptyFile)
+    let secret = utils.fromHex(lines[0])
```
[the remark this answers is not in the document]
mratsim: Also we might want a new proc with the following signature (or using `Result`):

```Nim
proc fromHex(dst: var openarray[byte], src: string): int =
  ## Parse `src` from hex-string to byte.
  ## Stores the result in `dst`
  ## Returns the number of bytes written
  ## In case of an error returns -1
```

This avoids having secrets lingering in seq/Nim GC when we can avoid it.
And it's more efficient than the hex->seq[byte]->toArray construct here.

cc @cheatfate

tests/test_jwt_auth.nim
mjfh: good :)

nimbus/rpc/jwt_auth.nim
```diff
@@ -0,0 +1,317 @@
...
+    return ok()
+
+  debug "Could not authenticate",
+    error = rc.error
+
+  case rc.error:
+  of jwtTokenValidationError, jwtMethodUnsupported:
+    return err("Unauthorized")
+  else:
+    return err("Malformed Token")
```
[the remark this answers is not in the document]
mjfh: oops -- cruft

nimbus/rpc/jwt_auth.nim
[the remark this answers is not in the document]
mjfh: > There should be a comment that says that nimcrypto.utils.fromHex is constant-time to avoid people refactoring that away.

Why would it not not be O(1) -- just curious?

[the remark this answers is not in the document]
mratsim: constant-time as in cryptographic constant-time, so resistant to all side-channel attacks including timing attacks and cache attacks: https://github.com/mratsim/constantine/wiki/Constant-time-arithmetics

This is different from O(1).

For example, digit conversion from nimcrypto uses arithmetic: 
https://github.com/cheatfate/nimcrypto/blob/a5742a9a214ac33f91615f3862c7b099aec43b00/nimcrypto/utils.nim#L54-L58=

```Nim
proc hexDigit(x: int, lowercase: bool = false): char =
  var off = uint32(0x41 - 0x3A)
  if lowercase:
    off += 0x20
  char(0x30'u32 + uint32(x) + (off and not((uint32(x) - 10) shr 8)))
```

While stdlib toHex uses an array: https://github.com/nim-lang/Nim/blob/7f6e800cafc7b73625893fb5280eb8b51a15b252/lib/pure/strutils.nim#L910-L919=
```Nim
func toHexImpl(x: BiggestUInt, len: Positive, handleNegative: bool): string =
  const
    HexChars = "0123456789ABCDEF"
  var n = x
  result = newString(len)
  for j in countdown(len-1, 0):
    result[j] = HexChars[int(n and 0xF)]
    n = n shr 4
    # handle negative overflow
    if n == 0 and handleNegative: n = not(BiggestUInt 0)
```

In particular `HexChars[int(n and 0xF)]` is vulnerable to cache side-channels.

Using an array is possible if-and-only-if the lookup function scans through all the array and select the value with bit manipulation like in Constantine:
- https://github.com/mratsim/constantine/blob/bde4f97b5679629feb36d2a801e612e6f5c0fd0c/constantine/math/io/io_bigints.nim#L382-L398=
- https://github.com/mratsim/constantine/blob/742cecce0898c71564f9709003996604f7af34d9/constantine/platforms/constant_time/ct_routines.nim#L200-L208=
```Nim
func nativeEndianToHex(bytes: openarray[byte], order: static[Endianness]): string =
  ## Convert a byte-array to its hex representation
  ## Output is in lowercase and not prefixed.
  ## This assumes that input is in platform native endianness
  const hexChars = "0123456789abcdef"
  result = newString(2 + 2 * bytes.len)
  result[0] = '0'
  result[1] = 'x'
  for i in 0 ..< bytes.len:
    when order == system.cpuEndian:
      let bi = bytes[i]
      result[2 + 2*i] = hexChars.secretLookup(SecretWord bi shr 4 and 0xF)
      result[2 + 2*i+1] = hexChars.secretLookup(SecretWord bi and 0xF)
    else:
      let bmi = bytes[bytes.high - i]
      result[2 + 2*i] = hexChars.secretLookup(SecretWord bmi shr 4 and 0xF)
      result[2 + 2*i+1] = hexChars.secretLookup(SecretWord bmi and 0xF)


func secretLookup*[T; S: Ct](table: openArray[T], index: S): T =
  ## Return table[index]
  ## This is constant-time, whatever the `index`, its value is not leaked
  ## This is also protected against cache-timing attack by always scanning the whole table
  var val: S
  for i in 0 ..< table.len:
    let selector = S(i) == index
    selector.ccopy(val, S table[i])
  return T(val)
```

A potential way to exploit the stdlib `toHex` is reenacting Colin Percival's HyperThreading attack by trying to evict the conversion table from the CPU cache. (Though with the table size, risk is low). See presentation "Full AES key extraction in 65 milliseconds using cache attacks" https://www.iacr.org/conferences/crypto2005/r/6.ppt

nimbus/rpc/jwt_auth.nim
```diff
@@ -0,0 +1,317 @@
...
+
+  JwtAuthAsyHandler* = ##\
+    ## Asynchroneous JSW authenticator prototype. This is the definition
+    ## appicable for the `verify` entry of a `ws.Hook`.
+    proc(req: HttpTable): Future[Result[void,string]]
+      {.closure, gcsafe, raises: [Defect].}
+
+  JwtSharedKey* = ##\
+    ## Convenience type, needed quite often
+    array[jwtMinSecretLen,byte]
```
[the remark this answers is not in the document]
mjfh: ok

nimbus/rpc/jwt_auth.nim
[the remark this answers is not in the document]
mjfh: Ahh ok, the crucial item was that it not a constant time _bound_ (my assumption) but guards against side channel attacks -- similar jargon for different things but it holds almost everywhere (finite number of exceptions, only) :)

PR #1042, opened 2022-04-06
Increase packet size of utp over discv5

KonradStaniec:


fluffy/fluffy.nim
```diff
@@ -111,8 +111,13 @@ proc run(config: PortalConf) {.raises: [CatchableError, Defect].} =
 
   # One instance of UtpDiscv5Protocol is shared over all the PortalStreams.
   let
+    # Overhead of talkReq message when sending 1150 bytes data packet, is 107bytes
+    # also each uTP packet have 20bytes header.
+    # Total max packet size will be: 1150 + 20 + 107 = 1277bytes, which
```
kdeme, line 116: Could use some const with a comment that explains where 107 comes from in more detail like is done for the `talkRespOverhead` const in Portal wire protocol. Its pretty much the same except for the addition of the protocol id rlp encoded string

kdeme (approved):

PR #1046, opened 2022-04-07
No priority for RPC transactions

mjfh:
Currently, there is no meaningful way to have a single local transaction. Only an account can be local.

This will be changed and local transactions will be introduced.


mjfh:
prriority was ok (commit still missing, will do in a new PR)

mjfh:
superseded by https://github.com/status-im/nimbus-eth1/pull/1047

PR #1053, opened 2022-04-11
Bump chronos and json-rpc and avoid http conn re-use for blockwalk

kdeme:
Currently re-using http connections for the json-rpc requests
seems to slow down the requests considerably. Avoid this by
forcing a close after each requests in the blockwalk tool.


arnetheduck:
you'll need one more chronos bump I think to catch: https://github.com/status-im/nim-chronos/pull/267

PR #1065, opened 2022-04-14
Add put method which preserves max size

KonradStaniec:
Add put method to content db which preserves total size and return information about if database was pruned. 
If db was pruned there are two important information which may be necessary for radius adjustments:
- what is current furthest element in db (so caller will never drop radius below this value)
- what fraction of content was deleted
In general we do not want to perform pruning to often thats why we like to delete even 25% of db at one go.

Radius adjustments will be done in next pr.

Also it is a bit not natural that we have content db in history network and in portal protocol, maybe some refactor is needed there.

Current open questions are:
- what is reasonable size for portal node database or if we want to allow nodes without storage ?
- is there possibility for some attacker to craft content and content keys which will be correct and close to our id so it would never be delted and pollute our db ?

### DIFF
--- a/fluffy/conf.nim
+++ b/fluffy/conf.nim
@@ -33,6 +33,9 @@ const
   defaultAdminListenAddressDesc = $defaultAdminListenAddress
   defaultDataDirDesc = defaultDataDir()
   defaultClientConfigDesc = $(defaultClientConfig.httpUri)
+  # 100mb seems a bit smallish we may consider increasing defaults after some
+  # network measurements
+  defaultStorageSize* = uint32(1000 * 1000 * 100)
 
 type
   PortalCmd* = enum
@@ -183,6 +186,14 @@ type
       desc: "Kademlia's b variable, increase for less hops per lookup"
       defaultValue: DefaultBitsPerHop
       name: "bits-per-hop" .}: int
+
+    # TODO maybe it is worth defining minimal storage size and throw error if
+    # value provided is smaller than minimum
+    storageSize* {.
+      desc: "Maximum amount (in bytes) of content which will be stored " &
+            "in local database."
+      defaultValue: defaultStorageSize
+      name: "storage-size" .}: uint32
 
     case cmd* {.
       command
--- a/fluffy/content_db.nim
+++ b/fluffy/content_db.nim
@@ -173,7 +173,7 @@ proc get*(db: ContentDB, key: openArray[byte]): Option[seq[byte]] =
 
   return res
 
-proc put*(db: ContentDB, key, value: openArray[byte]) =
+proc put(db: ContentDB, key, value: openArray[byte]) =
   db.kv.put(key, value).expectDb()
 
 proc contains*(db: ContentDB, key: openArray[byte]): bool =
@@ -194,7 +194,7 @@ proc get*(db: ContentDB, key: ContentId): Option[seq[byte]] =
   # TODO: Here it is unfortunate that ContentId is a uint256 instead of Digest256.
   db.get(key.toByteArrayBE())
 
-proc put*(db: ContentDB, key: ContentId, value: openArray[byte]) =
+proc put(db: ContentDB, key: ContentId, value: openArray[byte]) =
   db.put(key.toByteArrayBE(), value)
 
 proc contains*(db: ContentDB, key: ContentId): bool =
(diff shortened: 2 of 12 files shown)


fluffy/content_db.nim
```diff
@@ -187,3 +202,44 @@ proc contains*(db: ContentDB, key: ContentId): bool =
...
+  key: ContentId, 
+  value: openArray[byte],
+  target: UInt256): PutResult = 
+  db.put(key, value)
+  let dbSize = db.size()
+  
+  if dbSize < int64(db.maxSize):
+    return PutResult(kind: ContentStored)
+  else:
+    # TODO maybe caller should decide how many elements should be deleted and what
```
kdeme, line 239: How do you see that work? I was rather thinking that the db would arrange this all (perhaps with some setting configurable at startup). In the current setup I don't really see any other information than the `furthestNonDeletedElement` needed to be returned?

I was actually thinking of perhaps moving the `Radius` into the `ContentDB` object, which would require you not to return anything then. It would however link the `Radius` for all portal wire protocols (which I think is fine however) in the current setup of 1 shared ContentDB. 
Not sure though, perhaps some wrapper calls around the db function to only allow data in range to be stored is also sufficient.

fluffy/content_db.nim
```diff
@@ -187,3 +202,44 @@ proc contains*(db: ContentDB, key: ContentId): bool =
...
+    # fraction of content should be left ?
+    let (furthestNonDeletedElement, deletedFraction) = db.deleteNelemsNoMoreThan(target, 100000, 0.25)
```
kdeme, line 241: You're putting here the large amount of 100k items to be sure you are getting enough of them to reach the 0.25 drop?
Sounds like something that will highly dependable on size of the items and size of the database.

Perhaps you can just go without this number and do the size drop calculation directly in the logic currently present in `getNFurthestElements` ?

fluffy/conf.nim
```diff
@@ -33,6 +33,9 @@ const
   defaultAdminListenAddressDesc = $defaultAdminListenAddress
   defaultDataDirDesc = defaultDataDir()
   defaultClientConfigDesc = $(defaultClientConfig.httpUri)
+  # 100mb seems a bit smallish we may consider increasing defaults after some
+  # network measurements
+  defaultStorageSize* = uint32(1000 * 1000 * 100)
```
kdeme, line 38: For reference: Current network with first 25K blocks and full radius is ~21MB in database size. So this is fine for now, but will obviously increase in the future (depends to what kind of radius we want to map the default)

fluffy/network/history/history_network.nim
```diff
@@ -197,7 +202,11 @@ proc getBlock*(
 
   # content is in range and valid, put into db
   if h.portalProtocol.inRange(contentId):
-    h.contentDB.put(contentId, bodyContent.content)
+    # TODO handle radius adjustments
+    discard h.contentDB.putAndPrune(
```
kdeme, line 206: I think we can just call it `put` and make it the only public proc to put data into the db.
Unless we go with a system where some task would be lazily be pruning the data.

KonradStaniec: Yea I was thinking about it, but was not sure if this is direction we want to go, but if you are thinking the same will do 👍

fluffy/content_db.nim
KonradStaniec: To be honest it is rather another way around i.e I want to delete as most elements as possible (I have chosen 100k as 100k `ObjInfo` objects will take around 7mb of ram memory which is imo acceptable), but I do not want to delete more  that 25% of database.

fluffy/content_db.nim
KonradStaniec: > How do you see that work? I was rather thinking that the db would arrange this all (perhaps with some setting configurable at startup). In the current setup I don't really see any other information than the furthestNonDeletedElement needed to be returned?

Yea you are probably right, I will change todo to maybe add some configuration.

> I was actually thinking of perhaps moving the Radius into the ContentDB object, which would require you not to return anything then. It would however link the Radius for all portal wire protocols (which I think is fine however) in the current setup of 1 shared ContentDB.
> Not sure though, perhaps some wrapper calls around the db function to only allow data in range to be stored is also sufficient.

I am not sure here, tbh I think I would tie radius to db only when we would decide to go with having db per portal network, as it seems as reasonable use case to have different radius on different network and different storage capacity.

PR #1073, opened 2022-05-06
Snap sync preview

mjfh:
Inherited and rebased from https://github.com/status-im/nimbus-eth1/commit/6a17e5b7ad126a6d612047cb0b46ce7beb02a20c

Current status
+ unit tests ok
+ can sync against mainnet shadow 3 with standard sync
+ hive/nodocker tests ok for consesus and graphql

Snap itself is not tested, yet

PR #1076, opened 2022-05-06
Squashed snap-sync-preview patch

mjfh:
This is a squashed version of https://github.com/status-im/nimbus-eth1/pull/1073 which was inherited and rebased from https://github.com/status-im/nimbus-eth1/commit/6a17e5b7ad126a6d612047cb0b46ce7beb02a20c. Collected patch set comments are available as _nimbus/sync/ChangeLog.md_  sorted in chronological order, oldest first.

Current status
+   unit tests ok
+  can sync against mainnet shadow 3 with standard sync
+  hive/nodocker tests ok for consesus and graphql

Snap itself is not tested, yet

### DIFF
--- a/nimbus/chain_config.nim
+++ b/nimbus/chain_config.nim
@@ -8,7 +8,7 @@
 # those terms.
 
 import
-  std/[tables, strutils,sequtils, options, times],
+  std/[tables, strutils, options, times],
   eth/[common, rlp, p2p], stint, stew/[byteutils],
   nimcrypto/hash,
   json_serialization, chronicles,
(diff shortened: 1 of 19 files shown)


jlokier:
I'll just mention that I put `peer=sp` at the end deliberately, because I found the logs significantly less useful with it at the start while reading them at high speed and in real time.

(Actually all the log parameters I added in code from this era and later are ordered following a consistent design pattern for readability, that's why block numbers are before hashes, for example.)

With peer=sp at the start, the most useful _varying_ fields (starting with block number) to understand how things are progressing tend to be in slightly different column positions from one line to the next, and wrapped or truncated on the screen.

PR #1093, opened 2022-05-17
Add pruning metrics

KonradStaniec:
Adds new pruning metrics.
Addition:
- counter of pruning events which happens during node lifetime
- gauge of quantity of elements deleted in each pruning event
- new grafana panel showing, those statistics in history network. Putting all networks in one graph made it harder to discern what is happening

### DIFF
--- a/fluffy/network/wire/portal_protocol.nim
+++ b/fluffy/network/wire/portal_protocol.nim
@@ -79,6 +79,14 @@ declareHistogram portal_content_enrs_packed,
   "Portal wire protocol amount of enrs packed in a content message",
   labels = ["protocol_id"], buckets = enrsBuckets
 
+declareCounter portal_pruning_counter,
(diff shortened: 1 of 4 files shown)


fluffy/network/wire/portal_protocol.nim
```diff
@@ -79,6 +79,14 @@ declareHistogram portal_content_enrs_packed,
   "Portal wire protocol amount of enrs packed in a content message",
   labels = ["protocol_id"], buckets = enrsBuckets
 
+declareCounter portal_pruning_counter,
+  "Number of pruning event which happened during node lifetime",
+  labels = ["protocol_id"]
+
+declareGauge portal_pruning_deleted_elements,
```
kdeme, line 86: Perhaps Histogram would also have worked well here?

Issue #1075, opened 2022-05-06
uTP sockets leak sometimes on cancellation


kdeme:
Sometimes on a cancellation of the PortalStream `connectTo` the underlying uTP socket leaks. It does not occur always however.

Cancellation is currently only done in the `contentLookup` code (https://github.com/status-im/nimbus-eth1/blob/master/fluffy/network/wire/portal_protocol.nim#L939) on a `findContent` call.

The leak appears to happen sometimes when at this point at moment of cancel: https://github.com/status-im/nimbus-eth1/blob/master/fluffy/network/wire/portal_protocol.nim#L585

The PortalStream `connectTo` does a `UtpDiscv5Protocol` `connectTo`. There is however a cancelCallback which should clean this up: https://github.com/status-im/nim-eth/blob/01684a2130051357bf0e5bf0c9357fcf195f15d9/eth/utp/utp_router.nim#L250
Sometimes it does clean it up, sometimes it doesn't get triggered.

This is very easy to reproduce with Fluffy:
Start Fluffy & run the blockwalk tool with some hash that is available on the network so that `contentLookup`s & cancels are done.

It is also directly visibly in the `utp_established_connections` metric.

PR #1107, opened 2022-05-30
Fix utp connection leak on cancel

KonradStaniec:
Fixes https://github.com/status-im/nimbus-eth1/issues/1075 by bumping nim-eth and adding fixes to portal_protocol.

Summary: The main culprit of this leak is that cancellation and `await` do not play with each other very well when allocating some resource inside the `async` procs.

Consider simplified code:
```
import
  chronos

type SomeResourceWhichNeedToBeReleased = object

proc freeResource(res: SomeResourceWhichNeedToBeReleased): void = discard

proc use(res: SomeResourceWhichNeedToBeReleased): void= discard

proc allocateResource: Future[SomeResourceWhichNeedToBeReleased] = discard

proc someProcWhichUsesResource: Future[void] {.async.} =
  let resource = await allocateResource()
  try:
    resource.use()
  finally:
    resource.freeResource()

proc someProcWhichStartAndCancel() {.async.} =
  let fut = someProcWhichUsesResource()

  fut.cancel()

```

What was happening in current case was:
1. When calling `fut.cancel()`, the path with the [child](https://github.com/status-im/nim-chronos/blob/master/chronos/asyncfutures2.nim#L287) is taken where child future is `allocateResource` future (and the parent future is `someProcWhichUsesResource`)
2. But child future have already finished but not yet given control to the corresponding yield statement, so child cancel returns [false](https://github.com/status-im/nim-chronos/blob/master/chronos/asyncfutures2.nim#L284). So before cancel return, only the flag `mustCancel` on cancelled future is set to [true](https://github.com/status-im/nim-chronos/blob/master/chronos/asyncfutures2.nim#L295)
3. In the next eventLoop step control is given to the yield statement in `await allocateResource()` [yield](https://github.com/status-im/nim-chronos/blob/master/chronos/asyncmacro2.nim#L274), as the child future already finished (`allocateResource` future is child of `someProcWhichUsesResource` future ). As this child future already finished and not cancelled, resource was already allocated so to clean it up, call to `freeResource` is needed.
4. But as `mustCancel`  was set to true in point `2` for parent future, `await` actually throws and cancellation [exception](https://github.com/status-im/nim-chronos/blob/master/chronos/asyncmacro2.nim#L287) 
5. So it leads to the situation that there is no way to release resources allocated by `SomeResourceWhichNeedToBeReleased`

### DIFF
--- a/fluffy/network/wire/portal_protocol.nim
+++ b/fluffy/network/wire/portal_protocol.nim
@@ -643,24 +643,35 @@ proc findContent*(p: PortalProtocol, dst: Node, contentKey: ByteList):
         # validation is required, using a length prefix here might be beneficial for
         # this.
         let readFut = socket.read()
+
+        readFut.cancelCallback = proc(udate: pointer) {.gcsafe.} =
+          debug "Socket read cancelled",
+            socketKey = socket.socketKey
+          # In case this `findContent` gets cancelled while reading the data,
+          # send a FIN and clean up the socket.
+          socket.close()
+
         if await readFut.withTimeout(p.stream.readTimeout):
           let content = readFut.read
-          return ok(FoundContent(src: dst, kind: Content, content: content))
-        else :
-          return err("Reading data from socket timed out, content request failed")
-      finally:
-        if socket.atEof():
           # socket received remote FIN and drained whole buffer, it can be
           # safely destroyed without notifing remote
           debug "Socket read fully",
             socketKey = socket.socketKey
           socket.destroy()
-        else:
-          # socket read interrupted, either cancelled or timeout on reading
-          # data, inform remote by sending FIN and cleaning up resources
-          debug "Socket read interrupted",
+          return ok(FoundContent(src: dst, kind: Content, content: content))
+        else :
+          debug "Socket read time-out",
             socketKey = socket.socketKey
           socket.close()
+          return err("Reading data from socket timed out, content request failed")
+      except CancelledError as exc:
+        # even though we already installed cancelCallback on readFut, it is worth
+        # catching CancelledError in case that withTimeout throws CancelledError 
+        # but readFut have already finished.
+        debug "Socket read cancelled",
+          socketKey = socket.socketKey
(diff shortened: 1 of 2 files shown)


fluffy/network/wire/portal_protocol.nim
```diff
@@ -614,37 +614,53 @@ proc findContent*(p: PortalProtocol, dst: Node, contentKey: ByteList):
...
+        # A FIN does not necessarily mean that the data read is complete. Further
+        # validation is required, using a length prefix here might be beneficial for
+        # this.
+        let readFut = socket.read()
+        if await readFut.withTimeout(p.stream.readTimeout):
+          let content = readFut.read
+          return ok(FoundContent(src: dst, kind: Content, content: content))
+        else :
+          return err("Reading data from socket timed out, content request failed")
+      finally:
```
kdeme, line 651: We have been avoiding the use of finally in async procs due to bugs on finally + closure iterators (and also issues when raising in finally)
I don't know the exact state of this now but considering following still opens bugs I'd stay away from it:
https://github.com/nim-lang/Nim/issues/15501
https://github.com/status-im/nim-chronos/issues/132
https://github.com/nim-lang/Nim/issues/11887
(You can find more if you search for finally / defer )

KonradStaniec: that nice to know 👍 rewritten it to catching cancelled exception and closing socket there

kdeme:
> What was happening in current case was:

I haven't dived deep into this but would you say than that setting the `mustCancel` flag to true in the case that the child cancel failed is a bug?

KonradStaniec:
tbh at first glance it smells a little bit like bug, as it the child future already finished then we should definitely give back control to parent future instead of just throwing (and possibly leaving some dangling resources).

That saying, to be sure I also would need to analyse how futures interplay with cancellation a bit more to be sure, especially as `mustCancel` was introduced to fix some kinda race: https://github.com/status-im/nim-chronos/pull/107

PR #1113, opened 2022-06-02
Use queries with custom xor function

KonradStaniec:
Removes custom logic to sort elements, by using custom xor function on sqlite level and sql query to sort results by xor distance.

Next step is to add nice queries to get content in range of node radius like:
`"SELECT key WHERE xorDistance(?, key) <= radius LIMIT limit"`

### DIFF
--- a/fluffy/content_db.nim
+++ b/fluffy/content_db.nim
@@ -29,16 +29,11 @@ export kvstore_sqlite3
 # 3. Or databases are created per network (and kvstores pre content type) and
 # thus depending on the network the right db needs to be selected.
 
-const
-  # Maximal number of ObjInfo objects held in memory per database scan. 100k
-  # objects should result in memory usage of around 7mb which should be
-  # appropriate for even low resource devices
-  maxObjPerScan = 100000
-
 type
   RowInfo = tuple
     contentId: array[32, byte]
     payloadLength: int64
+    distance: array[32, byte]
 
   ObjInfo* = object
     contentId*: array[32, byte]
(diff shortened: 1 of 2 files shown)


fluffy/content_db.nim
```diff
@@ -233,41 +206,35 @@ proc contains*(db: ContentDB, key: ContentId): bool =
...
-    return (UInt256.zero, 0'i64, 0'i64, 0'i64)
+proc deleteContentFraction(
+  db: ContentDB,
+  target: UInt256,
+  fraction: float64): (UInt256, int64, int64, int64) =
 
-  let lastIdx = len(furthestElements) - 1
+  doAssert(
+    fraction > 0 and fraction < 1, 
+    "Deleted fraction shohould be > 0 and < 1"
```
kdeme, line 216:
```suggestion
    fraction > 0 and fraction < 1,
    "Deleted fraction should be > 0 and < 1"
```

fluffy/content_db.nim
```diff
@@ -233,41 +206,35 @@ proc contains*(db: ContentDB, key: ContentId): bool =
...
+  var numOfDeletedElements: int64 = 0
 
-    if bytesDeleted + elem.payloadLength < bytesToDelete:
-      db.del(elem.contentId)
-      bytesDeleted = bytesDeleted + elem.payloadLength
+  var ri: RowInfo
+  var bytesDeleted: int64 = 0
+  let targetBytes = target.toByteArrayBE()
+  for e in db.getAllOrderedByDistanceStmt.exec(targetBytes, ri):
+    if bytesDeleted + ri.payloadLength < bytesToDelete:
```
kdeme, line 227: It doesn't matter that much in our current setting, as an arbitrary fraction is chosen anyhow, but it would be more correct if you still allow the last deletion to reach the amount of bytesToDelete (or slightly more).

kdeme (approved):

fluffy/content_db.nim
KonradStaniec: I think it is matter of preference i.e current approach interprets `fraction` as at most data to delete, where with allowing this last deletion it would  be at least data to delete. I slightly prefer `at most` interpretation as it allows to discard case when the whole database could be deleted when someone would provide fraction of 0.99 (or something similar)

PR #1115, opened 2022-06-06
Change test data and propagation format

KonradStaniec:
Changes format of seed data from ` [header, txs, uncles]` to more natural `header, [txs, uncles], receipts]`, also add receipts to the mix.

@kdeme do we need to reseed this data to our fleet right now ? Or we may do it after we have means to propagate multiuple items at once ?


kdeme:
The reason the format parsing is the way it is, is because these are the way the files that we got currently are formatted. Until I got some files in the right format, we shouldn't change this. (Sure I can reformat all files, but still don't have the receipts atm).



> @kdeme do we need to reseed this data to our fleet right now ? Or we may do it after we have means to propagate multiuple items at once ?

This doesn't change anything to the actual format they are send over the Portal wire. That is why the re-serialization step is in there in the first place.

kdeme (changes requested):
As mentioned in call, lets await merging this till we have the tooling in here also to create the files.

Might also have to adjust the serialization to SSZ but that can be another PR as the protocol will have to follow this first / too.

KonradStaniec:
Yep as discussed https://github.com/status-im/nimbus-eth1/pull/1115/commits/0398a410dce8284833864bacd34168095d2409c8 adds the tool to download data from local node instance through json-rpc.

One limitation for now is that it is using premix to do the whole downloading thing, which is hardcoded to `127.0.0.1:8545` host. It is not big deal (as this is default geth host), but we may improve it at some point.

> Might also have to adjust the serialization to SSZ but that can be another PR as the protocol will have to follow this first / too.

I agree, adjusting tooling and protocol level to SSZ should be done in separate pr.

fluffy/tools/eth_data_exporter.nim
```diff
@@ -0,0 +1,153 @@
...
+      desc: "default name of the file with history data"
+      defaultValue: defaultFileName
+      name: "filename" .}: string
+
+type DataRecord = object
+  rlp: string
+  number: uint64
+
+proc writeBlock(writer: var JsonWriter, blck: Block) {.raises: [IOError, Defect].} = 
+  let enc = rlp.encodeList(blck.header, blck.body, blck.receipts)
```
kdeme, line 80: Perhaps group to 1 let

fluffy/tools/eth_data_exporter.nim
```diff
@@ -0,0 +1,153 @@
...
+    return requestBlock(num, flags = {DownloadReceipts})
+  except CatchableError as e:
+    fatal "Error while requesting Block", error = e.msg
+    quit 1
+
+proc createAndOpenFile(config: ExporterConf): OutputStreamHandle =
+  # Creates directory and file specified in config, if file already exists 
+  # program is aborted with info to user, to avoid losing data
+
+  try:
```
kdeme, line 100: Not super important considering it is a tool, but it would be nicer / cleaner if stew/io2 calls are used here, as already done in Fluffy on other locations.

KonradStaniec: I have changed it were possible although `fileOutput` still can throw as underneath it uses std lib open file.

kdeme: Can put the try / except clause only over that call now then, or not?
And remove std/os import?

KonradStaniec: good point about try/except, although `std/os` is still needed for path concatenation i.e `/`

kdeme (approved):

PR #1116, opened 2022-06-07
Add nimbus_addPeer rpc call

hmel:
Adds a nimbus_addPeer rpc call which adds an enode to the list of trusted peers.

### DIFF
--- a/nimbus/rpc/common.nim
+++ b/nimbus/rpc/common.nim
@@ -9,8 +9,8 @@
 
 import
   std/[strutils, tables],
-  nimcrypto, eth/common as eth_common, stint, json_rpc/server,
-  eth/p2p, eth/p2p/enode,
+  nimcrypto, eth/common as eth_common, stint, json_rpc/server, json_rpc/errors,
+  eth/p2p, eth/p2p/enode, eth/p2p/peer_pool,
   ../config, ./hexstrings
 
 type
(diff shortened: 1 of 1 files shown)


nimbus/rpc/common.nim
```diff
@@ -57,3 +57,10 @@ proc setupCommonRpc*(node: EthereumNode, conf: NimbusConf, server: RpcServer) =
         listener: $enode.address.tcpPort
       )
     )
+
+  server.rpc("nimbus_addPeer") do(enode: string) -> bool:
+    var res = ENode.fromString(enode)
+    if res.isOk:
+      asyncCheck node.peerPool.connectToNode(newNode(res.get()))
```
kdeme, line 64: `asyncCheck` is deprecated, use `asyncSpawn` instead. Also be aware that when using these, the proc called must not raise, which I believe is fine here as connectToNode is mostly doing an rlpx connect.

nimbus/rpc/common.nim
```diff
@@ -57,3 +57,10 @@ proc setupCommonRpc*(node: EthereumNode, conf: NimbusConf, server: RpcServer) =
...
+      return true
```
kdeme, line 65: Not sure what you would prefer as behaviour here, but `connectToNode` might fail. If you need to know this, you will have to `await` in this call to get that result.

hmel: This is just a request to add a node. We don't want to block the caller in an rpc call.

nimbus/rpc/common.nim
hmel: Updated

kdeme (approved):

PR #1118, opened 2022-06-09
Add getting receipts from history network

KonradStaniec:
Add getting and validating receipts from history network.

For now it does not get used in any endpoint as there are no endpoints which can be served only by history network and require receipts, although maybe we can update `eth_getLogs` endpoint to support the case of getting logs for block with provided `blockHash`. It is pretty recent addition  and https://eth.wiki/json-rpc/API marks is `future` and nimbus itself does not support it yet (from what I see nimbus in general does not support `get_Logs`). 

There is quite a bit duplication of code with logic to get `BlockBody` but I would wait with refactor until at least we have getting receipts used in some json_rpc endpoint.

### DIFF
--- a/fluffy/network/history/history_network.nim
+++ b/fluffy/network/history/history_network.nim
@@ -57,46 +57,74 @@ func getEncodedKeyForContent(
 
   return encodeKey(contentKey)
 
+proc getContentFromBytes(bytes: openArray[byte], T: type): Result[T, string] =
+  var rlp = rlpFromBytes(bytes)
+  try:
+    let content = rlp.read(T)
+    ok[T](content)
+  except RlpError as e:
+    err(e.msg)
+
 proc validateHeaderBytes*(
     bytes: openArray[byte], hash: BlockHash): Option[BlockHeader] =
-  try:
-    var rlp = rlpFromBytes(bytes)
-
-    let blockHeader = rlp.read(BlockHeader)
-
-    if not (blockHeader.blockHash() == hash):
-      # TODO: Header with different hash than expected, maybe we should punish
-      # peer which sent us this ?
-      return none(BlockHeader)
-
-    return some(blockHeader)
-
-  except MalformedRlpError, UnsupportedRlpError, RlpTypeMismatch:
-    # TODO add some logging about failed decoding
+  
+  let headerResult = getContentFromBytes(bytes, BlockHeader)
+  
+  if headerResult.isErr():
+    error "Failed to decode header ", msg = headerResult.error()
     return none(BlockHeader)
+  
+  let header = headerResult.unsafeGet()
 
-proc validateBodyBytes*(
-    bytes: openArray[byte], txRoot: KeccakHash, ommersHash: KeccakHash):
-    Option[BlockBody] =
+  if not (header.blockHash() == hash):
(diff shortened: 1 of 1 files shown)


fluffy/network/history/history_network.nim
```diff
@@ -199,6 +199,73 @@ proc getBlock*(
...
+
+    if receiptRoot != calculatedReceiptsRoot:
+      # we got receipts which do not match
+      # header. For now just ignore it, but maybe we should penalize peer
+      # sending us such data?
+      return none(seq[Receipt])
+
+    return some(receipts)
+
+  except RlpError, MalformedRlpError, UnsupportedRlpError, RlpTypeMismatch:
```
kdeme, line 218: RlpError is enough, it contains the other 3 errors iirc.
Also keep the try/except as small as possible.

fluffy/network/history/history_network.nim
```diff
@@ -199,6 +199,73 @@ proc getBlock*(
...
+
+  if maybeReceiptsFromDb.isSome():
+    info "Fetched receipts from database", hash
+    return maybeReceiptsFromDb.unsafeGet()
+
+  let maybeReceiptsContent = await h.portalProtocol.contentLookup(keyEncoded, contentId)
+
+  if maybeReceiptsContent.isNone():
+    warn "Failed fetching receipts from the network", hash
+    return newSeq[Receipt]()
```
kdeme, line 244: Can't discern a failed lookup from a header without receipts if this gets returned.

fluffy/network/history/history_network.nim
```diff
@@ -199,6 +199,73 @@ proc getBlock*(
...
+      return none(seq[Receipt])
+
+    return some(receipts)
+
+  except RlpError, MalformedRlpError, UnsupportedRlpError, RlpTypeMismatch:
+    # TODO add some logging about failed decoding
+    return none(seq[Receipt])
+
+proc getReceipts*(
+  h: HistoryNetwork,
```
kdeme, line 223: nit: indentation

fluffy/network/history/history_network.nim
```diff
@@ -199,6 +199,73 @@ proc getBlock*(
...
+  if maybeReceiptsContent.isNone():
+    warn "Failed fetching receipts from the network", hash
+    return newSeq[Receipt]()
+
+  let receiptsContent = maybeReceiptsContent.unsafeGet()
+
+  let maybeReceipts = validateReceiptsBytes(receiptsContent.content, header.receiptRoot)
+
+  if maybeReceipts.isNone():
+    return newSeq[Receipt]()
```
kdeme, line 251: idem above

fluffy/network/history/history_network.nim
```diff
@@ -199,6 +199,73 @@ proc getBlock*(
...
+  let receipts = maybeReceipts.unsafeGet()
+
+  # receips are valid, propagate it to interested peers
+  h.portalProtocol.triggerPoke(
+    receiptsContent.nodesInterestedInContent,
+    keyEncoded,
+    receiptsContent.content
+  )
+
+  # content is in range and valid, put into db
```
kdeme, line 264: This comment is no longer needed / valid

PR #1126, opened 2022-06-15
Add seed db file format

KonradStaniec:
Add possibility to export history data to sqlite with range query support, this should enable efficient bulk offers in fluffy.
Next step is how to use this data format in fluffy.

### DIFF
--- a/fluffy/tools/eth_data_exporter.nim
+++ b/fluffy/tools/eth_data_exporter.nim
@@ -6,8 +6,9 @@
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
 # Tool to download chain history data from local node, and save it to the json
-# file.
-# Data of each block is rlp encoded list of: 
+# file or sqlite database.
+# In case of json:
+# Data of each block is rlp encoded list of:
 # [blockHeader, [block_transactions, block_uncles], block_receipts]
 # Json file has following format:
 # {
@@ -18,7 +19,11 @@
 #   ...,
 #   ...,
 # }
-#
+# In case of sqlite:
+# Data is saved in format friendly to history network i.e one tablse with 3
+# columns: contentid, contentkey, content.
+# Such format enables queries to quickly find content in range of some node
+# which makes it possible to offer content to nodes in bulk.
 #
 
 {.push raises: [Defect].}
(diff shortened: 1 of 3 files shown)


fluffy/tools/eth_data_exporter.nim
```diff
@@ -18,7 +19,11 @@
 #   ...,
 #   ...,
 # }
-#
+# In case of sqlite:
+# Data is saved in format friendly to history network i.e one tablse with 3
```
kdeme, line 23:
```suggestion
# Data is saved in a format friendly to history network i.e one table with 3
```

fluffy/tools/eth_data_exporter.nim
```diff
@@ -70,14 +80,32 @@ type
     filename* {.
       desc: "default name of the file with history data"
       defaultValue: defaultFileName
+      defaultValueDesc: $defaultFileName
       name: "filename" .}: string
+    mode* {.
+      desc: "default mode of data export"
```
kdeme, line 86:
```suggestion
      desc: "Mode of data export"
```

Although I would rather call it more explicitly storage mode or storage type or format as mentioned above

fluffy/tools/eth_data_exporter.nim
```diff
@@ -141,6 +175,37 @@ proc run(config: ExporterConf) =
       fatal "Error occoured while closing file", error = e.msg
       quit 1
 
+proc writeToDb(config: ExporterConf) =
+  let db = SeedDb.new(distinctBase(config.dataDir), config.filename)
+  defer:
+    db.close()
+
+  for i in config.initialBlock..config.endBlock:
+    let blck = downloadBlock(i)
```
kdeme, line 184: nit: 1 let

fluffy/tools/seed_db.nim
```diff
@@ -0,0 +1,137 @@
+# Nimbus
+# Copyright (c) 2021-2022 Status Research & Development GmbH
```
kdeme, line 2:
```suggestion
# Copyright (c) 2022 Status Research & Development GmbH
```

fluffy/tools/seed_db.nim
```diff
@@ -0,0 +1,137 @@
+# Nimbus
```
kdeme, line 1: I think this file should not go in tools but in a more general folder (perhaps just top folder for now) considering it will also be used in Fluffy to read the data.

KonradStaniec: tbh I have started with this file in top folder then moved it here as I was not sure what is perfect place for it, so yep I will move it to the top :+1:

PR #1135, opened 2022-06-19
bump serialization family(serialization, json, toml)

jangko:


jangko:
@kdeme or @KonradStaniec , do you have any idea what might be the cause of failing test?

kdeme:
@jangko Nothing related with your PR. I've seen it fail before. I need to figure out some better way to wait or just add more wait time for i386.

PR #1145, opened 2022-06-30
Store proved snap accounts

mjfh:


kdeme (commented):
Why the nim-ssz-serialization module bump to an older version. This breaks things for Fluffy, quite clearly visible in CI too.

PR #1164, opened 2022-07-14
Additional seed endpoints

KonradStaniec:
Adds additional seed endpoints which uses seed db which enables easy range queries. This enables offering multiple content items which are in remote nodes range.

Ultimately I have chosen json rpc approach instead of additional seed mode, as with seed mode it is not obvious it there should be separate seed db per network or one seed db for all networks. We can revisit this approach when we will have more networks to work on.

Also add test in testnet test, which uses new endpoints to seed additional block to the network. (also it is a bit not ideal as it requires adding binary sqlite3 file to repo)

Also added re-tries in testnet test instead of sleep, as those seem to make this test less flakey. (although it still does not solve this problem entirely)

It also fixes two things in history network:
- When doing retries there was not `continue` in loop in case of failed lookup which made it possible to call `unsafeGet` on empty option. (In commit d23651aaa67ce7c33f601af7374c5906372842e5, changed to early return as I saw this was the intention)
- When requesting body for block without transactions and uncles,  just return empty body without any network and db calls. It was necessary as we do not store empty stuff in seed db.


fluffy/scripts/test_portal_testnet.nim
```diff
@@ -227,3 +287,65 @@ procSuite "Portal testnet tests":
...
+  asyncTest "Portal History - Propagate content from seed db":
+    let clients = await connectToRpcServers(config)
+
+    var nodeInfos: seq[NodeInfo]
+    for client in clients:
+      let nodeInfo = await client.portal_history_nodeInfo()
+      await client.close()
+      nodeInfos.add(nodeInfo)
+
+    const dbFile = "./fluffy/tests/blocks/"
```
kdeme, line 300: it's a bit inconsistent with the usage of json files we do in the other tests, and also less adjustable / readable.

fluffy/populate_db.nim
```diff
@@ -221,3 +221,26 @@ proc historyPropagateBlock*(
...
     return err(blockDataTable.error)
+
+proc historyGetHashesInRange*(
+  db: SeedDb,
+  nodeId: UInt256,
+  radius: UInt256,
+  max: int64): seq[BlockHash] =
+  var hashes: seq[BlockHash]
+  # specifing `max` as int.high means all content in range will be returned
+  let contentsInRange = db.getContentInRange(nodeId, radius, max, int64(0))
```
kdeme, line 232: Why not use the `getContentInRange` without the offset parameter?

fluffy/rpc/rpc_portal_debug_api.nim
```diff
@@ -54,3 +54,65 @@ proc installPortalDebugApiHandlers*(
...
+
+  rpcServer.rpc("portal_" & network & "_offerContentInNodeRange") do(
+      dbPath: string,
+      dbName: string,
+      nodeId: NodeId,
+      max: uint32,
+      starting: uint32) -> bool:
+      ## Offers `max` closest elements starting from `starting` index to peer
+      ## with given `nodeId`.
+      ## Maxium value of `max` is 64 , as this is limit for single offer.
```
kdeme, line 86:
```suggestion
      ## Maximum value of `max` is 64 , as this is limit for a single offer.
```

fluffy/rpc/rpc_portal_debug_api.nim
```diff
@@ -54,3 +54,65 @@ proc installPortalDebugApiHandlers*(
...
+      ## `starting` argument is needed as seed_db is read only, so if there is
```
kdeme, line 87: So this way will require more tooling around it to actually inject data into the network.

This is different from the current propagate calls and their usage.

fluffy/scripts/test_portal_testnet.nim
```diff
@@ -227,3 +287,65 @@ procSuite "Portal testnet tests":
...
+      await clients[0].close()
+
+    let db = SeedDb.new(path = dbFile, name = dbName)
+
+    # using UInt256.high as radius means we will get all hashes which are stored
+    # in seed db. This is necessary as we do not have canoncial indicies network
+    # and need to now upfront what blockhashes to request.
+    # setting max as 100 as this is enough to get all things from db (there
+    # are 81 pieces of content in seed db)
+    let hashes = historyGetHashesInRange(db, UInt256.zero(), UInt256.high, 100)
```
kdeme, line 321: For getting all the hashes, the `for hash in blockData.get().blockHashes():` done higher up sounds like a cleaner option, but of course it requires the json file.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -1329,3 +1329,37 @@ proc resolve*(p: PortalProtocol, id: NodeId): Future[Option[Node]] {.async.} =
...
+
+  if r.isSome():
+    return some((node, r.unsafeGet()))
+
+  let pongResult = await p.ping(node)
+
+  if pongResult.isOk():
+    # If pong is successful, radius of the node should definitly be in local
+    # radius cache
+    return some((node, p.radiusCache.get(id).unsafeGet()))
```
kdeme, line 1363: I don't like the idea of doing an unsafeGet here without check.
If somehow the code gets altered on the ping/pong side (potentially gets buggy or whatever), or some way the push the data fast enough out the cache (unlikely), this makes the node crash here.

fluffy/seed_db.nim
```diff
@@ -125,13 +126,23 @@ proc getContentInRange*(
     db: SeedDb,
     nodeId: UInt256,
     nodeRadius: UInt256,
-    max: int64): seq[ContentDataDist] =
+    max: int64,
+    offset: int64): seq[ContentDataDist] =
+  ## Return `max` amout of content in `nodeId` range, starting from `offset` postion
```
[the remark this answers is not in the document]
KonradStaniec: Not really it means it will return `max` closest items to nodeid, except first `offset` ones. This enables iteration over db in batches of closest items.

fluffy/rpc/rpc_portal_debug_api.nim
```diff
@@ -54,3 +54,65 @@ proc installPortalDebugApiHandlers*(
...
+
+    db.close()
+
+    for contentData in contentInRange:
+      let cid = UInt256.fromBytesBE(contentData.contentId)
+      p.storeContent(cid, contentData.content)
+
+    return true
+
+  rpcServer.rpc("portal_" & network & "_offerContentInNodeRange") do(
```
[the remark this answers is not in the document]
KonradStaniec: So tbh I was a bit torn here i.e it would be easy to replicate `portal_history_propagate` on top of the seed db with something like getting closest nodes to local node with their radiuses, then iterating over db with batches of 64 and starting concurrent offers to them, but I am wondering if core fluffy node is really place for such logic. It is not really composable with other calls, it is also not really deterministic, and is also pretty opinionated about the way how to spread content.

In contrast call which offer some content to particular nodeId is a bit more composable with other calls, and all the  optimisation can be done at client level. It is possible to define flow like on some client level:
- call `portal_history__findNodes` which will return sent of nodes
- call `portal_history_offerContentInNodeRange` to this nodes. (can be called concurrently if needed)

Also such call does not impose any spreading strategy, and something similar to `portal_history_propagate` can probably be implemented on top of it.

Ultimately at this point In time, `portal_history_propagate` (If you think is desirable I can add it in next pr.) may serve us better as does not require additional tooling and makes spreading a bit easier, but at some point in time, imo it will be better to have set of composable endpoints  which will enable building specific functionalists on top of them.

fluffy/network/wire/portal_protocol.nim
KonradStaniec: Sure, I was thinking of refactoring `ping` to return its radius but it would require a bit to much code moving for my taste and it would be better done in separate pr.

I will add check here, and leave some `todo`

fluffy/seed_db.nim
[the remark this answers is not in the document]
kdeme: Right, so it still goes over all items and skips `offset` amount?

fluffy/rpc/rpc_portal_debug_api.nim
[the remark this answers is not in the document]
KonradStaniec: As disussed in call, this will be merged as it is as those calls maybe become useful. After that I will create with `portal_history_propagate` which will give us faster propagations in testnet quicker.

fluffy/seed_db.nim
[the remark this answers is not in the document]
KonradStaniec: it depends how sqlite query is implemented as offset is passed to query, but in general I think this is true.

PR #1176, opened 2022-07-28
Fix testnet tests

KonradStaniec:
Improves stability of fluffy testnet tests.

Main culprit of failures was to agressive `defaultContentReadTimeout`, which could sometimes lead to failure of all offers with larger content, which in turn lead to given content not being available on testnet

On 2 core vm in which I was testing changes  (and is pretty similar to the ones used in CI) even standard `talkreq` call would fail with timeouts when there are 64 fluffy nodes running, even though response timeout is set to 4 second on discv5 level. So 2 second for reading all content is definitely to small especially as:
- control over incoming connection is given to the app almost just after receiving utp SYN packet, which means it is at least 1 round trip before receiving any data from sender as sender of data first needs to receive SYN-ACK, and only after that he can send first data packet
- Initial utp window size fits only 2 data packets, so if content is larger than 3 packets, then sender need to receive at least one ack before sending 3rd packet.

In general 15s is maybe a bit on larger side, so maybe its worth making this configurable and increase it only for this testnet tests.

Another finds:
- 64 fluffy nodes were sometimes consuming 3.5GB ram on the machine which seems pretty large. On 32 bit systems, I would not be suprised if some kind OOM would happen during this tests. It maybe worth more investigation.
- During whole test run all nodes usually make around 18k offers, from which only 2k results in success. Most of offers ends up with no accepted content, I would say around 15k. Rest ends up with different kinds of timeouts. 
- In our testnet a lot of nodes receive invalid offered content as they are unable at the beginning to retrieve header for giver body or receipit.

### DIFF
--- a/fluffy/network/history/history_network.nim
+++ b/fluffy/network/history/history_network.nim
@@ -502,10 +502,6 @@ proc processContentLoop(n: HistoryNetwork) {.async.} =
       let (contentKeys, contentItems) =
         await n.portalProtocol.stream.contentQueue.popFirst()
 
-      let lenReceived = len(contentItems)
-
-      debug "Receied inoming offered content", contentKeys, len = lenReceived
-
       # content passed here can have less items then contentKeys, but not more.
       for i, contentItem in contentItems:
         let contentKey = contentKeys[i]
--- a/fluffy/network/history/history_network.nim
+++ b/fluffy/network/history/history_network.nim
@@ -280,21 +280,21 @@ proc getBlockHeader*(
 
   let headerFromDb = n.getContentFromDb(BlockHeader, contentId)
   if headerFromDb.isSome():
-    info "Fetched block header from database", hash
+    info "Fetched block header from database", hash, contentKey = keyEncoded
     return headerFromDb
 
   for i in 0..<requestRetries:
     let headerContentLookup =
       await n.portalProtocol.contentLookup(keyEncoded, contentId)
     if headerContentLookup.isNone():
-      warn "Failed fetching block header from the network", hash
+      warn "Failed fetching block header from the network", hash, contentKey = keyEncoded
       return none(BlockHeader)
 
     let headerContent = headerContentLookup.unsafeGet()
 
     let res = validateBlockHeaderBytes(headerContent.content, hash)
     if res.isOk():
-      info "Fetched block header from the network", hash
+      info "Fetched block header from the network", hash, contentKey = keyEncoded
       # Content is valid we can propagate it to interested peers
       n.portalProtocol.triggerPoke(
         headerContent.nodesInterestedInContent,
@@ -306,7 +306,7 @@ proc getBlockHeader*(
 
       return some(res.get())
     else:
-      warn "Validation of block header failed", err = res.error, hash
+      warn "Validation of block header failed", err = res.error, hash, contentKey = keyEncoded
 
   # Headers were requested `requestRetries` times and all failed on validation
   return none(BlockHeader)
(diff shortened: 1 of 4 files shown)


fluffy/network/history/history_network.nim
```diff
@@ -353,15 +353,18 @@ proc getBlockBody*(
...
     else:
-      warn "Validation of block body failed", err = res.error, hash
+      warn "Validation of block body failed", err = res.error, hash, contentKey = keyEncoded
 
   return none(BlockBody)
 
 proc getBlock*(
     n: HistoryNetwork, chainId: uint16, hash: BlockHash):
     Future[Option[Block]] {.async.} =
+  debug "Trying to retrieve block with hash", hash
```
kdeme, line 363: Not sure if that is really needed, perhaps on trace rather then

fluffy/network/wire/portal_protocol.nim
```diff
@@ -703,6 +708,9 @@ proc offer(p: PortalProtocol, o: OfferRequest):
   ## by the cleanup process before it will be transferred, so this way does not
   ## guarantee content transfer.
   let contentKeys = getContentKeys(o)
+
+  debug "Offering content", contentKeys = contentKeys
```
kdeme, line 712: Again, don't think this should be a debug message, will be quite spammed

fluffy/network/wire/portal_protocol.nim
```diff
@@ -726,6 +734,7 @@ proc offer(p: PortalProtocol, o: OfferRequest):
     let acceptedKeysAmount = m.contentKeys.countOnes()
     portal_content_keys_accepted.observe(acceptedKeysAmount.int64)
     if acceptedKeysAmount == 0:
+      debug "No content acceppted", contentKeys = contentKeys
```
kdeme, line 737: idem ^

fluffy/network/wire/portal_protocol.nim
```diff
@@ -790,14 +799,17 @@ proc offer(p: PortalProtocol, o: OfferRequest):
 
             let dataWritten = await socket.write(output.getOutput)
             if dataWritten.isErr:
-              debug "Error writing requested data", error = dataWritten.error
+              debug "Error writing requested data", error = dataWritten.error, contentKeys = contentKeys
               # No point in trying to continue writing data
               socket.close()
               return err("Error writing requested data")
 
+    debug "Content successfully offered", contentKeys = contentKeys
```
kdeme, line 807: idem ^

fluffy/network/wire/portal_stream.nim
```diff
@@ -102,6 +102,9 @@ proc addContentOffer*(
 
   # uTP protocol uses BE for all values in the header, incl. connection id.
   let id = uint16.fromBytesBE(connectionId)
+
+  debug "Register new incoming offer", contentKeys
```
kdeme, line 106: idem ^

fluffy/scripts/test_portal_testnet.nim
```diff
@@ -81,12 +84,13 @@ proc withRetries[A](
...
+proc retryUntil[A](
   f: FutureCallback[A],
   c: CheckCallback[A],
-  checkFailMessage: string): Future[A] =
-  # some reasonable limits, which will cause waits as: 1, 2, 4, 8, 16 seconds
-  return withRetries(f, c, 5, seconds(1), checkFailMessage)
+  checkFailMessage: string,
+  nodeIdx: int): Future[A] =
+  # some reasonable limits, which will cause waits as: 1, 2, 4, 8, 16, 32 seconds
+  return withRetries(f, c, 6, seconds(1), checkFailMessage, nodeIdx)
```
kdeme, line 93: Did you increase the rettries here even more because there are still failures with 5?

kdeme (commented):
I think the timeout of 15s is fine. I actually set this even higher on my accumulator branch because epoch accumulators are ~0.5MB SSZ encoded, and I would see almost everything time-out.

PR #1178, opened 2022-07-31
Add verification of block data by use of accumulator

kdeme:
- Add getter for epoch accumulator
- Add initial network getter for master accumulator
- Add specific storage for master accumulator
- Add canonical chain verification of history content: default
disabled still
- Add options to start Fluffy with a locally build accumulator
or build the accumulator of history data provided locally


kdeme:
This PR futher adds accumulator functionality, and the possibility to verify content as part of canonical chain.
More refactor is required in several areas, but that will have to go in a second PR and I would like to get this in before it bit rots.

More importantly, it can't be default enabled now as it will make the local testnet tests very slow (it does work however). Secondly it needs a 8K (master accumulator) and 1.1MB (epoch accumulator) file, so I was thinking of dropping that in a secondary repository for test data.

fluffy/tools/eth_data_exporter.nim
```diff
@@ -139,60 +210,82 @@ proc downloadBlock(i: uint64, client: RpcClient): Block =
...
       fh.close()
     except IOError as e:
-      fatal "Error occoured while closing file", error = e.msg
+      fatal "Error occured while closing file", error = e.msg
       quit 1
 
-proc writeToDb(config: ExporterConf, client: RpcClient) =
-  let db = SeedDb.new(distinctBase(config.dataDir), config.filename)
+proc writeBlocksToDb(config: ExporterConf, client: RpcClient) =
+  let db = SeedDb.new(distinctBase(config.dataDir), defaultBlockFileName)
```
KonradStaniec, line 288: shouldn't this use 'config.filename' ? (otherwise it will always write to `eth-block-data.json.sqlite3` if I am not mistaken)

fluffy/tools/eth_data_exporter.nim
```diff
@@ -139,60 +210,82 @@ proc downloadBlock(i: uint64, client: RpcClient): Block =
...
+  let fh = createAndOpenFile(string config.dataDir, string config.fileName)
+
+  try:
+    var writer = JsonWriter[DefaultFlavor].init(fh.s, pretty = true)
+    writer.beginRecord()
+    for i in config.initialBlock..config.endBlock:
+      let blck = client.downloadHeader(i)
+      writer.writeHeaderRecord(blck)
+      if (i mod 8192) == 0 and i != config.initialBlock:
+        info "Downloaded 8192 new block headers", currentHeader = i
```
KonradStaniec, line 251: this log message is not really correct as if we start from block lets say 12k, it will be shown after around 4k headers

fluffy/network/history/history_network.nim
```diff
@@ -532,6 +747,24 @@ proc start*(n: HistoryNetwork) =
...
   n.processContentLoop = processContentLoop(n)
 
+proc initMasterAccumulator*(
+    n: HistoryNetwork,
+    accumulator: Option[Accumulator]) {.async.} =
+  if accumulator.isSome():
+    n.contentDB.putAccumulator(SSZ.encode(accumulator.get()))
+    info "Successfully retrieved master accumulator from local data"
+  else:
+    while true:
```
KonradStaniec, line 757: Just for my undestanding, this is inifinite loop so that fluffy won't be able to start without some kind of master accumulator ?

fluffy/network/history/accumulator_db.nim
```diff
@@ -0,0 +1,200 @@
...
+  ssz_serialization, ssz_serialization/[proofs, merkleization],
+  ../../common/common_types,
+  ../../populate_db,
+  "."/[history_content, accumulator]
+
+type
+  # TODO:
+  # This database should probably dissappear and become part of ContentDB and
+  # SeedDB. Some of the ContentDB work has been done regarding this.
+  AccumulatorDB* = ref object
```
KonradStaniec, line 23: Does this accumulator db is still needed (with introduction of permanent db) ? From what I can find it is only used in some tests ?

KonradStaniec (approved):
No blockers here from my side.

fluffy/tools/eth_data_exporter.nim
kdeme: Woops, that's a leftover, I'll adjust

fluffy/network/history/accumulator_db.nim
kdeme: Yeah, it is not really needed for normal operation, I think I can remove all of that and make the tests work without. Part of the refactor I was mentioning.

fluffy/network/history/history_network.nim
kdeme: Yes, I'm not fully sure yet if I should make it blocking, probably not, but when the canonical chain verification would be turned on, and no accumulator can be found, it cannot accept any data through offers. It could still request data though so...

fluffy/tools/eth_data_exporter.nim
```diff
@@ -222,39 +315,110 @@ proc writeToDb(config: ExporterConf, client: RpcClient) =
...
+  case config.cmd
+  of ExporterCmd.exportBlockData:
+    try:
+      exportBlocks(config, client)
+    finally:
+      waitFor client.close()
+
+  of ExporterCmd.exportAccumulatorData:
+    var headers: seq[BlockHeader]
+    for i in config.initialBlock..config.endBlock:
```
kdeme, line 406: Need not to allow still starting from an initial block for the accumulator build

PR #1182, opened 2022-08-03
Add getting block by number through accumulator

KonradStaniec:
Adds getting block by block number thorugh accumulator. 
It does not update testnet test yet as those would require passing large accumulators around. To test functionality new test suite is added to test history network interface. (at least for getting block by number)

### DIFF
--- a/fluffy/network/history/accumulator.nim
+++ b/fluffy/network/history/accumulator.nim
@@ -37,6 +37,26 @@ type
     historicalEpochs*: List[Bytes32, maxHistoricalEpochs]
     currentEpoch*: EpochAccumulator
 
+  BlockHashResultType* = enum
+    BHash, HEpoch, UnknownBlockNumber
+
+  BlockHashResult* = object
+    case kind*: BlockHashResultType
+    of BHash:
+      blockHash*: BlockHash
+    of HEpoch:
+      epochHash*: Bytes32
+      epochIndex*: uint64
+      blockRelativeIndex*: uint64
+    of UnknownBlockNumber:
+      discard
+
+proc newEmptyAccumulator*(): Accumulator =
+  return Accumulator(
+    historicalEpochs:  List[Bytes32, maxHistoricalEpochs].init(@[]),
+    currentEpoch: List[HeaderRecord, epochSize].init(@[])
+  )
+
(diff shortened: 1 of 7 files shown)


fluffy/network/history/accumulator.nim
```diff
@@ -37,6 +37,26 @@ type
...
+    of BHash:
+      blockHash*: BlockHash
+    of HEpoch:
+      epochHash*: Bytes32
+      epochIndex*: uint64
+      blockRelativeIndex*: uint64
+    of UnknownBlockNumber:
+      discard
+
+proc newEmptyAccumulator*(): Accumulator =
```
kdeme, line 54: Is this needed? If so, or in any case, it should use the style we usually follow

fluffy/network/history/accumulator.nim
```diff
@@ -140,3 +170,21 @@ proc verifyHeader*(
         err("Proof verification failed")
     else:
       err("Need proof to verify header")
+
+func getHeaderHashForBlockNumber*(a: Accumulator, bn: UInt256): BlockHashResult=
```
kdeme, line 174: It seems to me that this code is made a bit more complex with the BlockHashResult and then the cases in the history network code, than as when all of this code would live in the history network code directly. 
It also doesn't really belong in accumulator code in my opinion.

fluffy/network/history/accumulator.nim
KonradStaniec: it is not strictly necessary tbh, it is just a little helper to explicitly get empty accumulator. As for style you meand this should be `Init` instead of `new` as this object, not ref ? (if yes then , yep you are right here 👍 )

fluffy/network/history/accumulator.nim
KonradStaniec: I think I disagree on this as all of those calucluations are really implementation details `Accumulator` (at some point we should probably make its fields private and expose only needed stuff outside)

Also having separate type (`BlockHashResult`) and function in accumulator make it explicit at type system level how many possibilities responses are there when asking for hash given some block number. It forces caller to make decision based on this 3 possible outcomes. Also it makes it testable without whole ceremony of running history network.

I agree it is a bit more mouthful, but imo it is good tradeoff.

PR #1188, opened 2022-08-12
Bump nim eth

KonradStaniec:
Bumps nim eth to allow passing user data to uTP. Also fixes small bug in data exporter

### DIFF
--- a/fluffy/fluffy.nim
+++ b/fluffy/fluffy.nim
@@ -118,6 +118,8 @@ proc run(config: PortalConf) {.raises: [CatchableError, Defect].} =
       registerIncomingSocketCallback(@[
         stateNetwork.portalProtocol.stream,
         historyNetwork.portalProtocol.stream]),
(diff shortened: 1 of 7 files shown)


KonradStaniec:
Also bumps ubuntu in github actions to avoid upcoming brownouts - https://github.com/actions/runner-images/issues/6002

fluffy/fluffy.nim
```diff
@@ -118,6 +118,8 @@ proc run(config: PortalConf) {.raises: [CatchableError, Defect].} =
       registerIncomingSocketCallback(@[
         stateNetwork.portalProtocol.stream,
         historyNetwork.portalProtocol.stream]),
+      # for now we do not use user data in callbacks
+      nil,
```
kdeme, line 122: Could the default not been set to nil here, so that this is not needed when you don't have user data (and remains signature compatible)?

PR #1191, opened 2022-08-16
Refactor portal stream

KonradStaniec:
Clean up a bit portal stream interaction with UtpDiscv5Protocol transaport. Improvements:

- encapsulate whole stream logic in `portal_stream` by introducing `StreamManager`.
- do not leak `portal_stream` constants to other parts of the app.
- use `getUserData` on utp router simplify callbacks passed to UtpDiscv5Protocol (they no longer need to capture opened streams`)
- move `contentQueue` to network, that way each network is responsible for each own content queue, where stream is producer of content and network is consumer, this allows different limits content limits per network.

### DIFF
--- a/fluffy/network/state/state_network.nim
+++ b/fluffy/network/state/state_network.nim
@@ -23,10 +23,8 @@ const
 type StateNetwork* = ref object
   portalProtocol*: PortalProtocol
   contentDB*: ContentDB
+  contentQueue*: AsyncQueue[(ContentKeysList, seq[seq[byte]])]
   processContentLoop: Future[void]
-
-func setStreamTransport*(n: StateNetwork, transport: UtpDiscv5Protocol) =
-  setTransport(n.portalProtocol.stream, transport)
 
 func toContentIdHandler(contentKey: ByteList): Option[ContentId] =
   toContentId(contentKey)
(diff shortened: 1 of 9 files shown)


fluffy/network/wire/portal_stream.nim
```diff
@@ -67,9 +67,13 @@ type
     connectionTimeout: Duration
     contentReadTimeout*: Duration
     rng: ref HmacDrbgContext
-    udata: pointer
     contentQueue*: AsyncQueue[(ContentKeysList, seq[seq[byte]])]
```
kdeme, line 70: I think you forgot the remove the contentQueue here

fluffy/network/state/state_network.nim
```diff
@@ -77,15 +75,25 @@ proc new*(
...
 
-  return StateNetwork(portalProtocol: portalProtocol, contentDB: contentDB)
+  return StateNetwork(
+    portalProtocol: portalProtocol,
+    contentDB: contentDB,
+    contentQueue: cq
+  )
 
 proc processContentLoop(n: StateNetwork) {.async.} =
   try:
```
kdeme, line 99: The stream popped from should be adjusted here too

fluffy/network/wire/portal_stream.nim
KonradStaniec: hmm this queue is shared between each network and it stream, where stream only pushes items on it, and each network will read from it, so the stream and each network should have reference to it. (or am I missing something ?)

kdeme: Right, my bad. Didn't realize the put happened still in stream code.

PR #1203, opened 2022-08-25
Switch to the new style of importing vendor packages

zah:
In preparation of our migration to the new Nimble-based setup
and [nim-workspace][1], we switch to a new model where the vendor
packages are no longer imported through a locally generated Nimble
dir, but rather through an auto-generated `nimbus-build-system.paths`
file that features regular `--path:` statements.

This file will be imported only within the nimbus-build-system
environment in order to avoid any unwanted interference in working
copies based on the new Nimble setup.

[1]: https://github.com/status-im/nim-workspace

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,7 @@ nimcache
 # Nimble user files
 nimble.develop
 nimble.paths
+
+# nimbus-build-system files
+nimbus-build-system.paths
+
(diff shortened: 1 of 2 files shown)


arnetheduck:
shouldn't this be written by `make update` (like the nimble links were)?

PR #1213, opened 2022-09-05
Add mpt proof verification to proxy

KonradStaniec:
- Add mpt proof verification to the proxy
- Add proof verification test


lc_proxy/validate_proof.nim
```diff
@@ -0,0 +1,54 @@
...
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
+
+import
+  std/[sequtils, typetraits],
+  stint,
+  eth/common/eth_types as etypes,
```
kdeme, line 13: is this giving conflict with `web3/ethtypes`? It shouldn't I think? (albeit rather confusing)

lc_proxy/rpc/rpc_eth_lc_api.nim
```diff
@@ -59,12 +60,25 @@ proc installEthApiHandlers*(lcProxy: LightClientRpcProxy) =
...
-    # TODO this could be realised by eth_getProof as it return also balance
-    # of the account
-    let b = await lcProxy.client.eth_getBalance(address, blockId(blockNumber))
-
-    return encodeQuantity(b)
+    let proof = await lcProxy.client.eth_getProof(address, @[], blockId(blockNumber))
+
+    let proofValid = isAccountProofValid(
+      executionPayload.stateRoot,
+      proof.address,
```
kdeme, line 73: Perhaps we can just pass the proof? Or does it contain much unneeded extra?

KonradStaniec: It contains whole  section about storage slots proofs which I would like to keep separate form proving just the account. 

I was thinking about repackaging it into some separate object, but ultimately decided that it is too much ceremony from time being.

lc_proxy/validate_proof.nim
arnetheduck: they are the same: `_` is ignored as are case differences except for first letter

lc_proxy/rpc/rpc_eth_lc_api.nim
kdeme: Ok, fair enough

lc_proxy/validate_proof.nim
kdeme: Right, of course. For a moment there I thought only snake and camel case are interchangeable, but that makes no sense as Nim also doesn't care about the case.

PR #1216, opened 2022-09-08
Implement getStorageAt endpoint

KonradStaniec:
Implements getStorageAt endpoint in proxy.

Two things discovered during testing/using;
- for some reason web sockets client `nim-websock` does not work with alchemy api. There is strange error with regards to tls - `Error: unhandled exception: (RecvFatalAlert) Fatal alert (40) received from the peer (code: 296) [TLSStreamProtocolError]`. When using `news` it works.
- Out proof validation code does not handle validating proof of no existence. For now I have left `todo` to add it nim-eth.


lc_proxy/validate_proof.nim
```diff
@@ -40,13 +55,94 @@ proc isAccountProofValid*(
...
+  if maybeAccount.isSome():
+    let account = maybeAccount.unsafeGet()
+
+    if account.storageRoot == etypes.EMPTY_ROOT_HASH:
+      # valid account with empty storage, in that case getStorageAt
+      # return 0 value
+      return ok(u256(0))
+
+    if len(proof.storageProof) != 1:
+      return err("no storage proof for reqested slot")
```
kdeme, line 126:
```suggestion
      return err("no storage proof for requested slot")
```

kdeme: It is slightly strange to see the `!= 1` check with then a `no storage proof` error message.
I assume you check here for exactly 1 because in the eth_getProof needs to return exactly 1 proof as you only request for 1?

lc_proxy/rpc/rpc_eth_lc_api.nim
```diff
@@ -33,6 +38,19 @@ type LightClientRpcProxy* = ref object
...
+template checkPreconditions(payload: Opt[ExecutionPayloadV1], quantityTag: string) =
+  if payload.isNone():
+    raise newException(ValueError, "Syncing")
+
+  if quantityTag != "latest":
+    # TODO for now we support only latest block, as its semanticly most streight
+    # forward i.e it is last received and valid ExecutionPayloadV1.
+    # Ultimatly we could keep track of n last valid payload and support number
+    # queries for this set of blocks
+    # `Pending` coud be mapped to some optimisc header with block fetched on demand
```
kdeme, line 51:
```suggestion
    # TODO: for now we support only latest block, as its semantically most straight
    # forward, i.e it is last received and a valid ExecutionPayloadV1.
    # Ultimately we could keep track of n last valid payloads and support number
    # queries for this set of blocks.
    # `Pending` could be mapped to some optimistic header with the block
    # fetched on demand.
```

lc_proxy/validate_proof.nim
```diff
@@ -8,26 +8,41 @@
...
 func toMDigest(arg: FixedBytes[32]): MDigest[256] =
   MDigest[256](data: distinctBase(arg))
 
-proc isAccountProofValid*(
+proc isValidProof(
+    branch: seq[seq[byte]],
+    rootHash: KeccakHash,
+    key, value: seq[byte]): bool =
+  try:
+    # TODO Investigate if this handle proof of non-existence.
```
kdeme, line 30:
```suggestion
    # TODO: Investigate if this handles proof of non-existence.
```

lc_proxy/validate_proof.nim
KonradStaniec: yep. Also `proc getStorageData` takes `requestedSlot` as param, therefore it expected at least one storage proof. One could imagine more genreral api where `getStorageData` takes `requestedSlots: seq[UIn256]`as param and we validate that there is at least this many proofs in response, but I would not generalize it until it is needed.

kdeme: Yeah, was considering that too while writing my comment, but it is indeed not needed atm.

PR #1223, opened 2022-09-13
Add improved mpt trie proof verification

KonradStaniec:
- bumps `nim-eth`
- uses new prof verification api


kdeme (commented):
I think the code could perhaps better live in nim-eth then? It would be useful to probably more than just the lc proxy.

Also, is the original code still useful (namely `isValidBranch`)?
It would be good to have this cleared out and cleaned up before it lingers and we forget about it.

KonradStaniec:
Sure I can move it to `nim-eth` then. I considered it, but thought that we don't validate proofs from untrusted source anywhere else so that why I started in proxy. I will leave this pr as pr adapting rpc to use this new api.

Imo trie `isValidBranch` maybe useful to validate consistency of proofs generated by trie `getBranch`. I would just not use it to validate proofs which come from the network, as it has some assertions in it and its parsing of mpt nodes is a bit to lax.

PR #1224, opened 2022-09-14
Add missing nimbus-eth2 modules

KonradStaniec:
Adds missing eth2 modules required to build lc-proxy


kdeme (approved):
lgtm, but as mentioned already, I'm confused as why this used to work without.

KonradStaniec:
yup I am also a bit confused, I assumed that transitive dependencies are pulled by default, but from my discussions with Zah and you it seems it shouldn't have worked form the start.

PR #1226, opened 2022-09-15
Add ci and initial readme

KonradStaniec:
Add CI workflow for light client proxy


KonradStaniec:
Remove `windows i386` platform from matrix as it seems `nimbus-eth2` also does not use it (https://github.com/status-im/nimbus-eth2/blob/unstable/.github/workflows/ci.yml) and the build was failing there due to some missing dependecies problem  which are satisfied on any other platform.

PR #1229, opened 2022-09-16
Beacon sync skeleton

jangko:


nimbus/sync/protocol/wire_protocol.md
```diff
@@ -0,0 +1,550 @@
+# Ethereum Wire Protocol (ETH)
```
kdeme, line 1: Why this file here? We can just link to https://github.com/ethereum/devp2p/blob/master/caps/eth.md ?

PR #1232, opened 2022-09-20
Add eth_chainId endpoint to light client proxy

KonradStaniec:
- Add `eth_chainId` endpoint as it is needed by metamask
- Adds validation that chain id configured by light client is the same one used by data provider

### DIFF
--- a/lc_proxy/rpc/rpc_eth_lc_api.nim
+++ b/lc_proxy/rpc/rpc_eth_lc_api.nim
@@ -39,7 +39,7 @@ type LightClientRpcProxy* = ref object
   client*: RpcClient
   server*: RpcServer
   executionPayload*: Opt[ExecutionPayloadV1]
-  chainId: Option[Quantity]
+  chainId: Quantity
 
 template checkPreconditions(payload: Opt[ExecutionPayloadV1], quantityTag: string) =
   if payload.isNone():
@@ -58,13 +58,7 @@ proc installEthApiHandlers*(lcProxy: LightClientRpcProxy) =
   template payload(): Opt[ExecutionPayloadV1] = lcProxy.executionPayload
 
   lcProxy.server.rpc("eth_chainId") do() -> HexQuantityStr:
-    if lcProxy.chainId.isSome():
-      return encodeQuantity(lcProxy.chainId.get())
-    else:
-      # we got some unknown chainId, probably someone using proxy on some private
-      # unknown network. Retrieve it form data provider.
-      let chainId = await lcProxy.client.eth_chainId()
-      return encodeQuantity(chainId)
+    return encodeQuantity(lcProxy.chainId)
 
   lcProxy.server.rpc("eth_blockNumber") do() -> HexQuantityStr:
     ## Returns the number of most recent block.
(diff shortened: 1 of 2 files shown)


lc_proxy/rpc/rpc_eth_lc_api.nim
```diff
@@ -112,3 +123,55 @@ proc installEthApiHandlers*(lcProxy: LightClientRpcProxy) =
...
+    client: RpcClient,
+    eth1Network: Option[Eth1Network]): T =
+
+  return LightClientRpcProxy(
+    client: client,
+    server: server,
+    chainId: networkToChainId(eth1Network)
+  )
+
+proc start*(p: LightClientRpcProxy): Future[void] {.async.} =
```
kdeme, line 154: I think it would be more immediately clear what is being done here if:
- the name of the proc would be something like `verifyChaindId` or `checkChainId`
- The `isNone()` is pulled outside of the proc.

And additionally by not having a `ChainId` as an `Option`, but rather have it always filled in, and do the check if it is one of the known chain ids where it is needed.
I think right now the code is somewhat more work to understand than it should be.

KonradStaniec: So I am 👍 on name changing, but I am wonder if changing `ChainId` to be one of the 5 known is not overly restrictive.

It is basically as in network metadata docs:
```
# If the eth1Network is specified, the Eth1Monitor will perform some
# additional checks to ensure we are connecting to a web3 provider
# serving data for the same network. The value can be set to `None`
# for custom networks and testing purposes.
eth1Network*: Option[Eth1Network]
```

It easily for me to imagine that someone would like to test proxy running it with some custom local network against custom geth provider.

KonradStaniec: at the second thought, I am also fan of using `option.get`  as near as possible to `option.isNone` so If we leave chainId as `Option[ChainId]` then imo it is better to have this check in `verifyChaindId` than somewhere outside.

kdeme: > at the second thought, I am also fan of using `option.get` as near as possible to `option.isNone` 

I guess I wasn't clear on that, but I meant to also take the `get` out of it and just pass in `ChaindId`

> It easily for me to imagine that someone would like to test proxy running it with some custom local network against custom geth provider.

I probably should have expressed myself better, but I'm not advocating for this not to be possible.
What I meant instead was: As `ChainId` is practically a number, that every network has, making it an `Option` is conceptually confusing. If it would not be an `Option`, and instead of where now an `isSome` or `isNone` is called, there would be an `isKnownId` check or similar (basically the `networkToChainId` ...) it would be more clear what is going on here imo. Anyway, not that important, just a suggestion.

arnetheduck: fwiw, we've mostly stopped using `std/options` and are using `results.Opt` instead which gives access to `?` and `valueOr`, both of which make for a lot smoother code

kdeme: > fwiw, we've mostly stopped using `std/options` and are using `results.Opt` instead which gives access to `?` and `valueOr`, both of which make for a lot smoother code

Yeah, that's great. I've hit already some cases where the mixed usage of options and result is bugging me, and was planning on moving all our options to `result.Opt`

PR #1235, opened 2022-09-21
Fix cors preflight request

KonradStaniec:
Few fixes to cors handler in Nimbus.

1. When allowing all origins i.e `allowedOrigins: seq[Uri]` is empty, do not check if origin is in the list, just allow it.
2. When getting additional headers like `Access-Control-Request-Method` , they should be taken from request header list not response, as  initial `let resp = req.getResponse()` creates new response with empty header list, therefore those headers will never be present
3. Add additional check to preflight check, as preflight check should have `Access-Control-Request-Method` header specified
4. Response to preflight check with response in 200 range.

Making all those changes make is possible for browser based metamask extension to connect with local proxy node.

Side note: `httpCors` proc looks like middleware which should be part of some library, and should take seq of allowed headers and allowed methods as parameters from the caller.
Side note1: We could configure additional header in response  - `Access-Control-Max-Age' which would make browser cache the prefilght request result for given time. This would avoid double http request with each call. But is should probably be done in some seprate pr.

### DIFF
--- a/nimbus/rpc/cors.nim
+++ b/nimbus/rpc/cors.nim
@@ -45,7 +45,8 @@ proc httpCors*(allowedOrigins: seq[Uri]): HttpAuthHook =
     # this section shared by all http method
     let origin = parseUri(origins[0])
     let resp = req.getResponse()
-    if not allowedOrigins.containsOrigin(origin):
+
+    if not everyOriginAllowed and not allowedOrigins.containsOrigin(origin):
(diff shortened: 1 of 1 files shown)


jangko (approved):

jangko:
Ideally, CORS handling should be part of chronos-http. Then all http derivative services(graphql, json-rpc, rest, ws, etc) can benefit from it instead each of them have their own version of CORS handler.

PR #1239, opened 2022-09-23
Add headers with proof content type and use it for verification

kdeme:
- Add BlockHeaderWithProof content type & content
- Use BlockHeaderWithProof content to verify if chain data is part of the canonical chain
- Adjust parser & seeder code to be able to seed these headers with proof
- Adjust eth_data_exporter to be able to generate these headers with proof

There is currently quite some ugliness & clean-up needed for which a big part is due to:
1. Supporting both BlockHeader and BlockHeaderWithProof on the network
2. Supporting or not supporting proofs for headers which fall in the current epoch accumulator that doesn't get finished pre-merge


fluffy/network/history/history_network.nim
```diff
@@ -295,6 +291,71 @@ const requestRetries = 4
...
+
+  return verifyHeader(accumulator, header, proof)
+
+proc getVerifiedBlockHeader*(
+    n: HistoryNetwork, chainId: uint16, hash: BlockHash):
+    Future[Option[BlockHeader]] {.async.} =
+  let (keyEncoded, contentId) =
+    getEncodedKeyForContent(blockHeaderWithProof, chainId, hash)
+
+  let headerFromDb = n.getContentFromDb(BlockHeader, contentId)
```
KonradStaniec, line 312: I must say this is a bit misleading, that we are retrieving here a `BlockHeader ` and later save `BlockHeaderWithProof` . I needed to checkoout `getContentFromDb` implementation to see that inside it expects `BlockHeaderWithProof`. Either some comment or maybe separate method would be good

fluffy/data/history_data_seeding.nim
```diff
@@ -67,18 +73,25 @@ proc propagateAccumulatorData*(
     Future[Result[void, string]] {.async.} =
   ## Propagate all epoch accumulators created when building the accumulator
   ## from the block headers.
-  ## dataFile holds block data
-  let epochAccumulators = buildAccumulatorData(dataFile)
-  if epochAccumulators.isErr():
-    return err(epochAccumulators.error)
+  ## dataFile a set of consecutive headers.
+  let res = buildAccumulatorData(dataFile)
+  if res.isErr():
```
KonradStaniec, line 78: this could probably be `?` as error types are the same.

KonradStaniec (approved):
I do not see any blocking issues :+1:

fluffy/network/history/history_network.nim
kdeme: Yeah, I had the same thoughts on this, but then went with it anyway apparently :face_in_clouds: 

I'll see how to improve this, either here or in a follow PR

fluffy/data/history_data_seeding.nim
kdeme: I wish, but `?` doesn't work for `async` (Unless I missed this being fixed).

kdeme:
close for https://github.com/status-im/nimbus-eth1/pull/1281

PR #1241, opened 2022-09-26
Add simple block cache. Handle numeric params

KonradStaniec:
Cache several recently received execution blocks. 

This is mainly needed as metamask sometimes uses pattern of interaction which looks like:
1. bn <- `eth_blockNumber`
2. `eth_getBalance(address, 'bn')`

Without caching, proxy would be unable to handle this case as execution block could change between call.

For now hardcoded cache size to `64` block which with recent block times and sizes should cache blocks for around 13min, and take 6.5mb in memory. Ultimately cache size could be configurable.

Adding some kind of caching is also needed to serve `eth_getBlockByNumber` with blocks which are verified by light client. (currently this call is forwarded to data provider without any verification)

### DIFF
--- a/lc_proxy/block_cache.nim
+++ b/lc_proxy/block_cache.nim
@@ -13,10 +13,9 @@ import
   stew/[results, keyed_queue]
 
 
-## Simple block cache implementation which main purpose is to cache execution
-## payloads received through block gossip (and validated by light clicent).
-## Payloads are stored in order of arrival. When cache is full full oldest
-## payloads are deleted.
+## payloads received through block gossip (and validated by light client).
+## Payloads are stored in order of arrival. When cache is full the oldest
+## payload is deleted first.
 type BlockCache* = ref object
   max: int
   blocks: KeyedQueue[BlockHash, ExecutionPayloadV1]
--- a/lc_proxy/block_cache.nim
+++ b/lc_proxy/block_cache.nim
@@ -1,0 +1,63 @@
+# light client proxy
+# Copyright (c) 2022 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
+
+import
+  std/tables,
+  web3/ethtypes,
+  stew/[results, keyed_queue]
+
+
+## Simple block cache implementation which main purpose is to cache execution
+## payloads received through block gossip (and validated by light clicent).
+## Payloads are stored in order of arrival. When cache is full full oldest
+## payloads are deleted.
+type BlockCache* = ref object
+  max: int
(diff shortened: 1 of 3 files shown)


lc_proxy/rpc/rpc_eth_lc_api.nim
```diff
@@ -35,26 +37,77 @@ template encodeHexData(value: UInt256): HexDataStr =
...
+  var payload: ExecutionPayloadV1
+
+  case tag.kind
+  of LatestBlock:
+    # this will always be ok as we always validate that cache is not empty
+    payload = proxy.blockCache.latest.get
+  of BlockNumber:
+    let payLoadResult = proxy.blockCache.getByNumber(tag.blockNumber)
+    if payLoadResult.isErr():
+      raise newException(ValueError, "Unknown block with number " & $tag.blockNumber)
```
kdeme, line 106: I think error message can be improved here, now it seems like an "invalid" number, while it might just be "Block not stored in block cache"?

lc_proxy/block_cache.nim
```diff
@@ -0,0 +1,63 @@
...
+import
+  std/tables,
+  web3/ethtypes,
+  stew/[results, keyed_queue]
+
+
+## Simple block cache implementation which main purpose is to cache execution
+## payloads received through block gossip (and validated by light clicent).
+## Payloads are stored in order of arrival. When cache is full full oldest
+## payloads are deleted.
```
kdeme, line 19:
```suggestion
## payloads received through block gossip (and validated by light client).
## Payloads are stored in order of arrival. When cache is full the oldest
## payload is deleted first.
```

lc_proxy/block_cache.nim
```diff
@@ -0,0 +1,63 @@
...
+
+import
+  std/tables,
+  web3/ethtypes,
+  stew/[results, keyed_queue]
+
+
+## Simple block cache implementation which main purpose is to cache execution
+## payloads received through block gossip (and validated by light clicent).
+## Payloads are stored in order of arrival. When cache is full full oldest
```
kdeme, line 18: Not just stored in order of arrival but also in order of block number? Although I think no check is done on that, would that be needed?

KonradStaniec: not really, as we are following optimistic head (through optimistic processor and optimisitic updates) there are forks possible so it is possible that we receive blocks `1 -> 2 -> 3` and then `2' -> 3'`.

Thats actually why I decided to store blocks in order of arrival and when getting block by number look for them in from newest to oldest. This approach favours newest fork.

There probably still some edge cases where we would serve stale block from the cache, but I would like to avoid implementing branch resolving logic at proxy level. (at least for now)

kdeme: > Thats actually why I decided to store blocks in order of arrival and when getting block by number look for them in from newest to oldest. This approach favours newest fork.

Right, I missed the part of `getByNumber` where you iterate over the blocks.
So that means than that "stored in order of arrival" is only important really for the `latest` block?

KonradStaniec: not sure I understand question correctly, but yea with this approach `latest` actually means `latest received and validated block`. In most cases it will also be a block with highest block number, but not always i.e when fork happen.

PR #1250, opened 2022-09-29
Working on making vm2 async. (Not done yet.)

AdamSpitz:
This code is not even close to being ready for use.

"make test" should still work, though.

What's in here:

  - There's a duplicated async version of the vm2 inner loop. (Search for UGLY-duplicatedForAsync.) I'm not happy with how much duplication there is, and maybe there's some clever macro I could write to make code that works for both sync and async, but doing macros-on-macros strikes me as a little bit too much magic for now, especially since I'm currently struggling with compiler bugs.

  - The asyncness has been propagated upward, to the point where I can at least run it through some tests. (I cut the propagation along other paths, though. Search for FIXME-eventuallyPropagateAsyncFurtherUpward.)

  - I've got a basic test sub-suite that just runs a few of the EVM-specific tests. (Run "make test-evm".) I'm wondering whether was an easier way for me to create a sub-suite of tests. (See FIXME-howDoIMakeASubSuite.)

  - In the implementation of each opcode, we can now (optionally) specify an asynchronous handler. (Search for asyncHandlers.)

  - And the thing that's broken because I encountered that Nim compiler bug is in op_handlers.nim; there are a few lines of code in there that you can uncomment to see the bug. What's *supposed* to happen is that if asyncHandlers is present, its "run" proc gets put into runAsynchronously in the vmOpHandlersRec structure. But for some reason that I don't understand, that's not working. Anyway, the idea is that when the async version of the EVM is running an opcode that has an asynchronous handler, it should run that instead of the synchronous handler. That should allow us to implement asynchronous handlers for the few opcodes that need them, while still allowing us to have a fully-synchronous version of the EVM (which might be useful for performance comparisons, until/unless we become satisfied that we don't need it).

### DIFF
--- a/nimbus.nimble
+++ b/nimbus.nimble
@@ -66,6 +66,10 @@ task test, "Run tests":
 task test_rocksdb, "Run rocksdb tests":
   test "tests/db", "test_kvstore_rocksdb", "-d:chronicles_log_level=ERROR -d:unittest2DisableParamFiltering"
 
+# FIXME-howDoIMakeASubSuite
+task test_evm, "Run EVM tests":
+  test "tests", "evm_tests", "-d:chronicles_log_level=ERROR -d:unittest2DisableParamFiltering"
+
 task fluffy, "Build fluffy":
   buildBinary "fluffy", "fluffy/", "-d:chronicles_log_level=TRACE -d:chronosStrictException -d:PREFER_BLST_SHA256=false"
 
--- a/nimbus/vm2/interpreter/op_dispatcher.nim
+++ b/nimbus/vm2/interpreter/op_dispatcher.nim
@@ -26,6 +26,10 @@ import
   chronicles,
   macros
 
+# Need to exclude ServerCommand because it contains something
+# called Stop that interferes with the EVM operation named Stop.
+import chronos except ServerCommand
+
 export
   Fork, Op,
   oph_defs,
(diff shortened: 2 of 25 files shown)


nimbus/vm2/interpreter/op_dispatcher.nim
```diff
@@ -42,23 +46,56 @@ template handleStopDirective(k: var Vm2Ctx) =
...
+# UGLY-duplicatedForAsync
+template asyncHandleStopDirective(k: var Vm2Ctx) =
+  #trace "op: Stop"
+  if not k.cpt.code.atEnd() and k.cpt.tracingEnabled:
+    # we only trace `REAL STOP` and ignore `FAKE STOP`
+    k.cpt.opIndex = k.cpt.traceOpCodeStarted(Stop)
+    k.cpt.traceOpCodeEnded(Stop, k.cpt.opIndex)
+
+# UGLY-duplicatedForAsync
+template asyncHandleFixedGasCostsDirective(fork: Fork; op: Op; k: var Vm2Ctx) =
```
zah, line 77: One way to write code that works in roughly the same way in async and non-async context is to rely on higher-order templates. You can pass an `awaiter` template that will await in the async case and do nothing under normal compilation.

Here is one example of such template from the Faststreams library:

https://github.com/status-im/nim-faststreams/blob/master/faststreams/inputs.nim#L522-L552

As you can see, this template is called either with `fsAwait` or `noAwait` which are defined here:

https://github.com/status-im/nim-faststreams/blob/c653d05f277dca0f374732c5b9b80f2368faea33/faststreams/async_backend.nim#L70-L71

https://github.com/status-im/nim-faststreams/blob/c653d05f277dca0f374732c5b9b80f2368faea33/faststreams/async_backend.nim#L29-L30

nimbus/vm2/interpreter/op_handlers/oph_defs.nim
```diff
@@ -27,29 +29,48 @@ type
                               ## back via argument descriptor ``k``
     proc(k: var Vm2Ctx) {.gcsafe.}
 
+  # AARDVARK - does the gcsafe pragma make any sense here? I don't
+  # really understand how gcsafe interacts with async. --Adam
+  Vm2AsyncOpFn* =             ## here's the async version
+    proc(k: var Vm2Ctx): Future[void] {.gcsafe.}
```
zah, line 35: Yes, see this:

https://status-im.github.io/nim-style-guide/language.proctypes.html

nimbus/vm2/types.nim
```diff
@@ -81,7 +82,7 @@ type
     instr*:                 Op
     opIndex*:               int
     parent*, child*:        Computation
-    continuation*:          proc() {.gcsafe.}
+    continuation*:          proc(): Future[void] {.gcsafe.}
```
zah, line 85: I was not expecting this change and also the change introducing the async version of each opcode.

Instead, some of the opcodes will become async in practice by introducing an interruption point in their body through the `chainTo` template. This creates the continuation that will be executed once the asynchronously awaited data becomes available. 

The question is then how do we know when the data becomes available? To answer this, `chainTo` can be modified to optionally take an async operation that results in a `Future[void]` object which will be stored in the `Computation` object (e.g. in a field such as `pendingAsyncOperation`). Under async execution, the main loop will `await` this future before calling the `continuation` closure.

How do we know what kind of async operation to start? We can configure the `Computation` with a factory object for the various type of async operations that the EVM uses (e.g. `getStorageSlot`, `setStorageSlot`, etc). 

To make this nearly free under sync execution, you would pass a factory object that implement all the operations synchronously and you'll have to make sure that the main loop is proceeding without triggering any async operations in Chronos. At the moment, I'm not sure whether `waitFor` over an already completed `Future` will be enough for this. The alternative would be that the `Computation` object can enriched be with even more states such as having a synchronous closure as an alternative to the `pendingAsyncOperation`. The factory object will then be responsible for configuring the `Computation` in the right way when the "async" operation is started.

nimbus.nimble
```diff
@@ -66,6 +66,10 @@ task test, "Run tests":
 task test_rocksdb, "Run rocksdb tests":
   test "tests/db", "test_kvstore_rocksdb", "-d:chronicles_log_level=ERROR -d:unittest2DisableParamFiltering"
 
+# FIXME-howDoIMakeASubSuite
+task test_evm, "Run EVM tests":
```
zah, line 70: This looks like a perfectly good way to do this.

nimbus/vm2/types.nim
arnetheduck: `waitFor` can only be called at the top of the stack - nested calls are not supported, so we can have no designs that require calling it

zah:
Superseded by https://github.com/status-im/nimbus-eth1/pull/1260

PR #1255, opened 2022-10-07
Bump nimbuseth2

KonradStaniec:
-bump nimbus-eth2 and required transitive dependencies


kdeme (approved):
Just to add context: Removal of CI tests in Windows i386 is due to the fact that most/not all dependencies are no longer CI tested on Windows i386.
This means that vendor module bumps can cause failing CI (which is probably what happened here with the segfault on nim-chronos bump

PR #1259, opened 2022-10-10
Make portal_historyOffer spec compliant

KonradStaniec:
One of the endpoint `portal_historyOffer` got respecified (https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/portal-network-specs/assembled-spec/jsonrpc/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false).

Now it takes contentKey and content, gossip this content to interested peers, and returns the number of peers to which content was gossiped. It nicely maps to top our `neighborhoodGossip` proc.

It is worth mentioning that this edpoint is used as endpoint to inject data via bridge node (https://eth-portal.readthedocs.io/en/latest/bridge.html#bridge-node-intro)

### DIFF
--- a/fluffy/network/history/history_network.nim
+++ b/fluffy/network/history/history_network.nim
@@ -768,6 +768,12 @@ proc validateContent(
 
   return true
 
+proc gossipDiscardPeers(
+    p: PortalProtocol,
+    contentKeys: ContentKeysList,
+    content: seq[seq[byte]]): Future[void] {.async.} =
+  discard await p.neighborhoodGossip(contentKeys, content)
+
 proc processContentLoop(n: HistoryNetwork) {.async.} =
   try:
     while true:
@@ -779,7 +785,7 @@ proc processContentLoop(n: HistoryNetwork) {.async.} =
       # TODO: Differentiate between failures due to invalid data and failures
       # due to missing network data for validation.
       if await n.validateContent(contentKeys, contentItems):
-        asyncSpawn n.portalProtocol.neighborhoodGossip(contentKeys, contentItems)
+        asyncSpawn n.portalProtocol.gossipDiscardPeers(contentKeys, contentItems)
 
   except CancelledError:
     trace "processContentLoop canceled"
--- a/fluffy/network/history/history_network.nim
+++ b/fluffy/network/history/history_network.nim
@@ -768,7 +768,7 @@ proc validateContent(
 
   return true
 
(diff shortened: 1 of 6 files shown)


fluffy/network/history/history_network.nim
```diff
@@ -768,6 +768,12 @@ proc validateContent(
 
   return true
 
+proc gossipDiscardPeers(
```
kdeme, line 771: Ugh, too bad we need this, a bit ugly :(
Perhaps give it the exact name of the other call + discard to be more clear why:

```suggestion
proc neighborhoodGossipDiscard(
```

or 


```suggestion
proc neighborhoodGossipDiscardPeers(
```

fluffy/network/wire/portal_protocol.nim
```diff
@@ -1118,10 +1118,15 @@ proc getNClosestNodesWithRadius*(
...
 
 proc neighborhoodGossip*(
-    p: PortalProtocol, contentKeys: ContentKeysList, content: seq[seq[byte]])
-    {.async.} =
+    p: PortalProtocol,
+    contentKeys: ContentKeysList,
+    content: seq[seq[byte]]): Future[int] {.async.} =
+  ## Returns number of peers to which content was gossiped
+
+  var numberOfGossipedNodes = 0
```
kdeme, line 1126: Can avoid this var high up in the proc. Just return 0 on the failure ones and either have `let` variables in the if/else or create a var right above the if/else. Those versions would be more readable and less error prone imo.

kdeme:
I realize only now that how `portal_historyOffer` got "reused" there makes the rest of the json rpc spec rather inconsistent. Anyways, better to follow spec for now.

fluffy/network/history/history_network.nim
KonradStaniec: I also do not like, but `asyncSpawn` requires `Future[void]` : <

fluffy/network/wire/portal_protocol.nim
KonradStaniec: I started with versions using let variables, but decided that it will be more readable this way. Anyhow, can go back to let version 👍

PR #1260, opened 2022-10-13
Added basic async capabilities for vm2.

AdamSpitz:
This is a whole new Git branch, not the same one as [last time](https://github.com/status-im/nimbus-eth1/pull/1250) - there wasn't much worth salvaging. Main differences:

I didn't do the "each opcode has to specify an async handler" junk that I put in last time. Instead, in oph_memory.nim you can see sloadOp calling asyncChainTo and passing in an async operation. That async operation is then run by the execCallOrCreate (or asyncExecCallOrCreate) code in interpreter_dispatch.nim.

In the test code, the (previously existing) macro called "assembler" now allows you to add a section called "initialStorage", specifying fake data to be used by the EVM computation run by that test. (In the long run we'll obviously want to write tests that for-real use the JSON-RPC API to asynchronously fetch data; for now, this was just an expedient way to write a basic unit test that exercises the async-EVM code pathway.)

There's also a new macro called "concurrentAssemblers" that allows you to write a test that runs multiple assemblers concurrently (and then waits for them all to finish). There's one example test using this, in test_op_memory_lazy.nim, though you can't actually see it doing so unless you uncomment some echo statements in async_operations.nim (in which case you can see the two concurrently running EVM computations each printing out what they're doing, and you'll see that they interleave).

A question: is it possible to make EVMC work asynchronously? (For now, this code compiles and "make test" passes even if ENABLE_EVMC is turned on, but it doesn't actually work asynchronously, it just falls back on doing the usual synchronous EVMC thing. See FIXME-asyncAndEvmc.)


nimbus/transaction/call_evm.nim
```diff
@@ -177,8 +178,10 @@ proc rpcEstimateGas*(cd: RpcCallData, header: BlockHeader, chainDB: BaseChainDB,
 
   hi
 
-proc txCallEvm*(tx: Transaction, sender: EthAddress, vmState: BaseVMState, fork: Fork): GasInt =
-  var call = CallParams(
+proc callParamsForTx(tx: Transaction, sender: EthAddress, vmState: BaseVMState, fork: Fork): CallParams =
+  # Is there a nice idiom for this kind of thing? Should I
```
jangko, line 182: object construction syntax works for both regular object and ref object, and no need to worry about `new result` or `new(SomeRefType)`.
If you still love assignment syntax, you can just ignore the fields initialization when creating object.

nimbus/vm2/computation.nim
```diff
@@ -137,6 +138,12 @@ template getStorage*(c: Computation, slot: UInt256): UInt256 =
   else:
     c.vmState.readOnlyStateDB.getStorage(c.msg.contractAddress, slot)
 
+template hasStorage*(c: Computation, slot: UInt256): bool =
+  when evmc_enabled:
+    c.host.hasStorage(c.msg.contractAddress, slot)
```
jangko, line 143: do evmc spec allow host to have this kind of 'hasStorage' operation?

I think it only has `getStorage`.

https://github.com/status-im/nim-evmc/blob/144b13e0ccef229bc9118311e1735d0a23ac4dd5/evmc/evmc.nim#L388

nimbus/vm2/evmc_api.nim
```diff
@@ -110,6 +110,12 @@ proc getStorage*(ctx: HostContext, address: EthAddress, key: UInt256): UInt256 =
   var key = toEvmc(key)
   UInt256.fromEvmc ctx.host.get_storage(ctx.context, address, key.addr)
 
+proc hasStorage*(ctx: HostContext, address: EthAddress, key: UInt256): bool =
```
jangko, line 113: https://github.com/status-im/nim-evmc/blob/144b13e0ccef229bc9118311e1735d0a23ac4dd5/evmc/evmc.nim#L388

nimbus/vm2/async_operations.nim
```diff
@@ -0,0 +1,88 @@
...
+  if c.hasStorage(slot): return  # already have it, no need to fetch
+
+  # FIXME-onDemandStorageNotImplementedYet
+  # (I sketched in this code, but haven't actually tried running it yet.)
+  echo("Attempting to for-real fetch slot " & $(slot))
+  # ethAddressStr("0xfff33a3bd36abdbd412707b8e310d6011454a7ae")
+  # check hexDataStr(0.u256).string == res.string
+  let ethAddress = c.msg.contractAddress
+  let address: Address = Address(ethAddress)
+  let quantity: int = slot.truncate(int)  # this is probably wrong; what's the right way to convert this?
```
jangko, line 35: currently, truncate is the correct way to convert from UInt256 to smaller ints. But beware of int size, it could be 32 bit or 64 bit,  make sure it doesn't overflow.

nimbus/vm2/interpreter/op_dispatcher.nim
```diff
@@ -98,7 +98,7 @@ proc toCaseStmt(forkArg, opArg, k: NimNode): NimNode =
     # Wrap innner case/switch into outer case/switch
     let branchStmt = block:
       case op
-      of Create, Create2, Call, CallCode, DelegateCall, StaticCall:
+      of Create, Create2, Call, CallCode, DelegateCall, StaticCall, Sload:
```
jangko, line 101: why Sload? hmm, I don't understand this.

@mjfh, can you please take a look at this?

nimbus/transaction/call_common.nim
```diff
@@ -35,6 +36,11 @@ type
     noAccessList*: bool                 # Don't initialise EIP-2929 access list.
     noGasCharge*:  bool                 # Don't charge sender account for gas.
     noRefund*:     bool                 # Don't apply gas refund/burn rule.
+    # FIXME-whereDoesTheFactoryObjectBelong
```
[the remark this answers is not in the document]
AdamSpitz: Sounds good!

(Just to make sure I'm clear on what you're suggesting: when you say VMState, you mean BaseVMState, right? I might be missing it, but I don't see anything called VMState.)

nimbus/vm2/interpreter/op_dispatcher.nim
AdamSpitz: The idea is that sloadOp calls the asyncChainTo template (newly-created, analogous to the already-existing chainTo) to set a continuation and pendingAsyncOperation in the Computation. The interpreter_dispatch.nim code needs to notice that the continuation is there and break out of the opcode-interpreting loop so that it can run the pendingAsyncOperation and then run the continuation before continuing on with the rest of the opcodes. Is there a better way for me to make that happen?

nimbus/transaction/call_common.nim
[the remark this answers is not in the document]
jangko: right, I mean BaseVMState

nimbus/vm2/computation.nim
AdamSpitz: Yes, you're right. And even in the non-EVMC code, it feels very ugly for me to have added this new operation.

The problem I'm trying to solve is: if we want to do on-demand data fetching (start with an empty DB and then fetch the data on the fly, as needed), how can we tell whether a zero means "haven't fetched yet, so we need to send off a network request for this storage slot" or "already fetched it, and its value is zero"?

Maybe I could keep a separate Set that remembers which slots have already been requested and have come back with a zero? (Or something more space-efficient, like a Bloom filter? Except that's not quite right.)

nimbus/vm2/interpreter/op_dispatcher.nim
jangko: does it also interrupt non async evm? I understand that create and call family need to break off from continuation in non async evm. But if the Sload also do the same, can it resume execution cleanly?

maybe come comment also helpful to reduce confusion.

AdamSpitz: I believe it should resume cleanly. ("make test" passes, anyway.) I'm unsure about how big the performance hit is, though - that's something I need to test.

nimbus/vm2/interpreter_dispatch.nim
```diff
@@ -243,15 +255,73 @@ else:
...
         if c.continuation.isNil:
           c.afterExec()
           break
-        (before, c.child, c, c.parent) = (true, nil.Computation, c.child, c)
+        if not c.pendingAsyncOperation.isNil:
+          before = false
+          shouldPrepareTracer = false
+          let p = c.pendingAsyncOperation
+          c.pendingAsyncOperation = nil
+          # FIXME-areAsyncOperationsNecessaryDuringSynchronousExecution
```
zah, line 267: I personally imagine a different solution, where the synchronous version of the code installs already completed Futures in the `pendingAsyncOperation` field. The code can then check if the Future is already completed before calling `waitFor` (thus, the `waitFor` acts a fallback that's never expected to be executed).

nimbus/vm2/interpreter_dispatch.nim
```diff
@@ -222,17 +225,30 @@ when vm_use_recursion:
...
+      # a child (i.e. call or create) or a pendingAsyncOperation.
+      if not c.pendingAsyncOperation.isNil:
+        let p = c.pendingAsyncOperation
+        c.pendingAsyncOperation = nil
+        # This should never actually be called, because in synchronous
+        # mode every async operation should be an already-resolved
+        # Future. But we can check, just in case. (Or should we just
+        # make this an assertion?)
+        if not p.finished():
+          waitFor(p)
```
arnetheduck, line 238: it is essential that `waitFor` is not called in a nested context - since all of nimbus-eth1 is running in an "outer" `waitFor` (or similar event loop), this call is a no-no and must be rearchitected to use await instead

arnetheduck: the same goes for all the other `waitFor` calls here - in tests, use `asyncTest` instead of `waitFor`

zah: This cannot be re-architected, because this is the synchronous version of the EVM. We also have an asynchronous version which is using `await`. The key thing to understand here is that the `waitFor` will never be executed under certain configurations of the EVM which never start truly async operations. It will be executed only when you have plugged an "asynchronous back-end" (e.g. you use a storage engine that uses an external data provider to satisfy all queries), but you use synchronous EVM driving process. Nimbus itself will not feature such usage pattern, but it may appear in tests.

zah: In other words, the only viable alternative to having a `waitFor` here is to assert on these unusual configurations, but this is only loss of functionality.

zah: Please note that not having a synchronous mode of execution at all is also not acceptable, because this would prevent plugging Nimbus in fuzzing engines that expect synchronous EVM state transitions.

nimbus/vm2/async_operations.nim
```diff
@@ -0,0 +1,98 @@
...
+      # check hexDataStr(0.u256).string == res.string
+      let ethAddress = c.msg.contractAddress
+      let address: Address = Address(ethAddress)
+      let quantity: int = slot.truncate(int)  # this is probably wrong; what's the right way to convert this?
+      let blockId: BlockIdentifier = blockId(c.vmState.parent.blockNumber.truncate(uint64)) # ditto
+      let res = await client.eth_getStorageAt(address, quantity, blockId)
+      echo("Fetched slot " & $(slot) & ", result is " & $(res))
+      let v = res  # will res be the actual value, or do I need to convert or something?
+      c.vmState.mutateStateDB:
+        db.setStorage(c.msg.contractAddress, slot, UInt256.fromBytesBE(v))
```
zah, line 54: This should be done with care. There are many side-effects of writing something to storage (e.g. this would trigger state root re-computations).

In general, since we want to support truly stateless operation, the writing to storage should be optional. In a gradually syncing client (i.e. one using beam sync), we'll probably need new a storage paradigm that is aware that only portions of the state is present (and authenticated against a particular root hash / at a particular block height). The client should avoid computing the state root hash with such a partial database.

nimbus/vm2/async_operations.nim
```diff
@@ -0,0 +1,98 @@
...
+      echo("Attempting to for-real fetch slot " & $(slot))
+      # ethAddressStr("0xfff33a3bd36abdbd412707b8e310d6011454a7ae")
+      # check hexDataStr(0.u256).string == res.string
+      let ethAddress = c.msg.contractAddress
+      let address: Address = Address(ethAddress)
+      let quantity: int = slot.truncate(int)  # this is probably wrong; what's the right way to convert this?
+      let blockId: BlockIdentifier = blockId(c.vmState.parent.blockNumber.truncate(uint64)) # ditto
+      let res = await client.eth_getStorageAt(address, quantity, blockId)
+      echo("Fetched slot " & $(slot) & ", result is " & $(res))
+      let v = res  # will res be the actual value, or do I need to convert or something?
```
zah, line 52: You probably need to store it at a particular index in the EVM stack? See the current non-async version of `getStorage` and think about the implications under speculative execution.

nimbus/vm2/async_operations.nim
```diff
@@ -0,0 +1,98 @@
...
+
+# Used in synchronous mode.
+# FIXME: is there some way to make this a singleton,
+# so that we don't bother creating a new one each time,
+# without ruining the gcsafeness?
+proc noLazyDataSource*(): LazyDataSource =
+  LazyDataSource(
+    ifNecessaryGetStorage: (proc(c: Computation, slot: UInt256): Future[void] {.async.} =
+      # FIXME-areAsyncOperationsNecessaryDuringSynchronousExecution 
+      discard
```
zah, line 31: We've discussed that this might eagerly execute the current `getStorage` implementation and it can return a resolved Future.

nimbus/vm2/interpreter_dispatch.nim
```diff
@@ -45,12 +46,14 @@ const
...
-proc selectVM(c: Computation, fork: Fork) {.gcsafe.} =
+proc selectVM(c: Computation, fork: Fork, shouldPrepareTracer: bool) {.gcsafe.} =
   ## Op code execution handler main loop.
   var desc: Vm2Ctx
   desc.cpt = c
 
-  if c.tracingEnabled:
+  # It's important not to re-prepare the tracer after
+  # an async operation, only after a call/create.
+  if c.tracingEnabled and shouldPrepareTracer:
```
zah, line 56: Why is `shouldPrepareTracer` necessary as a separate parameter? How does it differ from the existing `c.tracingEnabled`?

nimbus/vm2/interpreter_dispatch.nim
arnetheduck: > viable alternative 

then `doAssert` - nested `waitFor` leads to undefined and difficult-to-debug context and order-dependent behavior (it's likely an assert _inside_ chronos will be added when there's a nested call like this) - the problems happen because when doing `await`, the future is either scheduled on the queue _or_ executed in the same call stack frame - when the latter happens, livelocks, deadlocks and unexpected state changes happen.. we could support nested waitFor, but then we'd have to ensure that _every_ await leads to a queued / delayed operation so that everything happens on a "pure" stack without nesting, which would incur a significant performance penalty for all operations that _could_ be run directly.

arnetheduck: > Please note that not having a synchronous mode of execution at all is also not acceptable,

whatever else happens, using `waitFor` inside library code is _not_ acceptable either (the only place to use it is in `main` effectively) - so either two separate versions of the code need to be written, or some other solution needs to be found - trying to cram both async and sync versions into the same code often results in false sharing - ie the semantics are different but you try to make it appear as if they were the same.

This is a fallacy of templates like `fsAsync` which disregard the significant semantic differences leading to either undefined behavior (in the async case) or suboptimal and more complex code (in the sync case) - usually what's needed is factoring the code in a different way which splits out the sync parts (so that no _significant_ logic is duplicated), then have different "wrappers" that glue the pieces together - we disabled everything `async` in faststreams for this precise reason: it's too complex to try to please both worlds at the same time, in general.

zah: > doAssert

If Chronos wants to outlaw any nested `waitFor` calls, the right solution would be indeed to add an assert there. This would automatically catch inappropriate usages of the code here (i.e. the fail-fast quality will be preserved).

On the other hand, if we place an assert here, we only lose the ability to write short test programs that drive the EVM in synchronous ways with all possible storage back-ends.

> faststreams

Err, the async faststreams are still there and they have been working without issues for months in the Nim language server. Since there is no `fsAsync` template I'm not sure whether you are referring to the `fsMultiSync` macro or to the `fsAwait` template which is just an alias to the normal `await`, so the rest of your message is difficult to comprehend for me (I can't think of any part of the FastStreams APIs that can lead to undefined behavior and I would argue that the library offers the most optimal approach to async programming when compared to the available alternatives).

arnetheduck: > if Chronos wants to outlaw

it's already outlawed, and doesn't actually work (ie it's undefined behavior that leads to leaks and Defect's) - adding an assert in chronos is not possible without adding runtime state ("am I in a nested waitFor / poll loop"), so it hasn't been done yet - presumably because doing so would require an exception-safe mechanism to detect whether we're in a nested call which is .. somewhat expensive (ie it affects "correct" code also that doesn't do nested waitFor). Regardless, the rule is simple: don't call `waitFor` (or any of its brethren, ie `poll`), in libraries, at all.

> async faststreams are still there and they have been working without issues

this is not the case in nimbus-eth2 where we disable them via compile-time flags (the faststream default is not to expose async), for stability reasons - arguably nimbus-eth2 is a more complex use case where stability is more important and compile-time bugs are hit more frequently - it was impossible to work with many libraries when the async fs backend was enabled - perhaps the situation has improved since then. Notably though, the std/asyncdispatch backend in faststreams is still disabled in the faststreams unit tests because they don't work / compile.

More than that, sync and async code has different lifetime / borrowing requirements and guarantees - ie `ref` and `var` have completely different semantics and performance profiles - trying to target both in a single codebase is .. difficult - the only way around that is leaving memory safety and concurrent modification constraints behind (via pointers and / or `ref`) as some of the faststreams libraries do - this leads to other problems, as can be seen in the faststreams issue tracker were most "advanced" uses of faststreams (ie anything that is not a simple linear read/write of a stream) actually don't work (runways, ranges, readInto, peeking etc).

zah: Outlawing nested `waitFor` in Chronos will require a single `bool` variable and it will be extremely cheap. You set the variable to true just after a new async event/completion is triggered and you set it back to false just before entering the OS "wait for next event" mechanism. You'll never notice this any profiler. The claim that nested `waitFor` is fundamentally unsupportable is also wrong. The async run-time in your favorite programming language (Rust) is built on the equivalent of using nested `waitFor` loops.

Nevertheless, since the decision here is not so important (it affects only unusual configurations of the EVM), we can agree to use `doAssert` just so the progress of the more important features can continue.

Many of the claims about FastStreams are also wrong, but I'm not interested in continuing this off-topic discussion.

arnetheduck: > extremely cheap

the exception safety requirement means there's a setjmp involved as well, not just the bool.

> here is not so important

it's important we don't bake code flows that will create problems downstream.

> rust

rust indeed has a different/inverted model for continuing futures - quite unlike that of most other async frameworks, and quite unlike that of chronos - it's of course not impossible to rewrite chronos to work the same way (with some quite significant changes), but _currently_, chrosos simply does not support it.

zah: I don't think you need to add new `setjmp` calls to implement the check in Chronos. The ones that are needed are already there.

nimbus/vm2/interpreter_dispatch.nim
AdamSpitz: tracingEnabled is: "Do we want to do tracing at all?" It's checked in many places, not just here.

shouldPrepareTracer is: "In this particular call to selectVM, do we really want to re-initialize the tracer?" (When I was implementing this code, I got stuck for quite a while on a bug where a slot was "disappearing" after SLOAD was executed, and it turned out to be because this line of code was resetting the tracer.) I agree that the variable name shouldPrepareTracer doesn't make that super-clear, though.

Maybe there's a cleaner solution? (e.g. I considered pulling the call to prepareTracer upward into executeOpcodes, or even further upward into execCallOrCreate. But that seems even more awkward.) I don't think we want to just set tracingEnabled to false, though - we don't want to disable tracing, we just don't want to reset the tracer after every async operation.

PR #1280, opened 2022-10-26
Add initial implementation of portal powered light client

KonradStaniec:
Implements light client which underneath uses light client portal network. For now much of stuff is copied from libp2p version, in time we can probably extract some common abstractions.

Some of portal light client networks procs are still not implemented, but interfaces should be roughly as defined here.


kdeme (commented):
I didn't have a detailed look yet but I'm a bit reluctant on adding this duplication of code for something we don't really need yet.

On the other hand it is probably not such a good plan to request/make changes to make the beacon light client manager more generic in nimbus-eth2 (making it more complex), considering we are not sure the Portal version will survive.

But did you have a look at what could perhaps be separated and what not? For example anything non network related. As I'm assuming the only real difference is the network backend, being the Portal `LightClientNetwork` versus the `Eth2Node` object?

The code itself is rather contained in its own light client files, so perhaps not too awful to deal with later on.

fluffy/network/beacon_light_client/portal_light_client_manager.nim
```diff
@@ -0,0 +1,340 @@
...
+  n.getLightClientFinalityUpdate()
+
+# https://github.com/ethereum/consensus-specs/blob/v1.2.0/specs/altair/light-client/p2p-interface.md#getlightclientoptimisticupdate
+proc doRequest(
+    e: typedesc[OptimisticUpdate],
+    n: LightClientNetwork
+): Future[NetRes[altair.LightClientOptimisticUpdate]] =
+  n.getLightClientOptimisticUpdate()
+
+template valueVerifier[E](
```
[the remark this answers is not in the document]
KonradStaniec: yep, but as you mention it is not exprted from orginal manager

KonradStaniec:
It is as you mention i.e main differentiatior at this point is usage of the Portal LightClientNetwork versus the Eth2Node.

I looked of what can be exported or re-used and there would certainly be some functions and types which can be shared, but at this point I consider this whole endeavour highly experimental. Thats why I would like for it to be separated as possible from libp2p light client as much as possible, even at cost of some code duplication. From what is worth, most important part, the one which proccesses light client objects i.e `light_client_processor` is shared so it is not that bad.

In future, I think it would be possible to encapuslate also gossip in LightClientManager, and then LightClient object can take manager as argument, so that libp2p and portal versions would just differ only in manager implementation. Which makes sense, as this module is responsible for networking and portal and libp2p are different networking stacks.

My current plan is roughly:
- implement other content types
- then try to run libp2p light client on mainnet (which small modification that it will also store objects in db) and try to make portal light client to follow this libp2p client.

fluffy/network/beacon_light_client/portal_light_client.nim
```diff
@@ -0,0 +1,160 @@
+# beacon_chain
```
kdeme, line 1: Name of `portal_light_client.nim`. I would change it to `beacon_light_client`. Everything in this folder is Portal.

fluffy/network/beacon_light_client/portal_light_client_manager.nim
```diff
@@ -0,0 +1,340 @@
+# beacon_chain
```
kdeme, line 1: Same regarding naming here.

kdeme:
> My current plan is roughly:

I think how to do the validation part of offered content is also quite important.

PR #1281, opened 2022-10-26
Add headers with proof content type and use it for verification

kdeme:
- Add BlockHeaderWithProof content type & content
- Use BlockHeaderWithProof content to verify if chain data is part of the canonical chain
- Adjust parser & seeder code to be able to seed these headers with proof
- Adjust eth_data_exporter to be able to export custom header ranges for which to build proofs (mostly for testing)

There is currently quite some ugliness & clean-up needed for which a big part is due tos upporting both BlockHeader and BlockHeaderWithProof on the network.

This PR supersedes https://github.com/status-im/nimbus-eth1/pull/1239. A lot of the changes from #1239 were already added in other smaller PRs, just not the headers with proofs part.
Still in draft because:

- [x] Most importantly, I'll look into making changes to the `BlockHeaderWithProof` type, to have an SSZ `Union` with a possibly `None` for the proofs. And make the proofs an SSZ vector so it can default fail at deserialization when the depth is not correct (with a List this would be custom addition)
- [x] There is quite some duplication on the building of proofs (for testing, for data generation, etc.). Might want to align some of that.
- [x] Not fully sure yet if the local testnet tests should use headers + epoch accumulator as input data, and build headers with proof on the fly, as it is done now. We could just go with headers with proof (this was done in the previous PR) immediately as input data. That way the proof building code would only be located in the tooling. However, you'd probably end up with duplication on storing these (even though temporary). Storing headers + epoch accumulators + also headers with proofs, which is why I didn't do it that way in this PR.
- [ ] Also, not sure if for tests we should use directly ssz files and/or e2s files, or rather stick with json + hex encoded data. I'm currently leaning towards the first.
- [ ] 2 local testnets tests will fail and are skipped now. One of the already was skipped. Another one is new. The test already fails at content offering. I believe the tests are flawed in the "validate the offered content" flow where we need the header to be verified now. It is basically based on luck right now (do the headers come first or not?) that it passes. Would need to rewrite them so that they work in 2 steps: headers offered, then the rest.


kdeme:
Sticking with building the proofs on the fly, with `EpochAccumulatorCached` it doesn't take that long and else I need to store more data locally for injecting those headers with proofs into the network.

Also sticking with ssz/e2s files for the headers for testing, but undecided on that one really.

The faulty local testnet test fails is for another PR

PR #1290, opened 2022-11-07
Make portal network generic in db store and db get

KonradStaniec:
Currently portal network in fixed on using `ContentDb` which may not always be a right choice.

For example for light client network will have totally different characteristics than other networks:
- every node stores everything
- pruning based on on deleting elements outside of weak subjectivity period.

To enable such use cases portal network will be now parametrised by two callbacks :

```
  DbGetHandler* =
    proc(contentKey: ByteList): Result[seq[byte], DbError] {.raises: [Defect], gcsafe.}

  PortalStoreHandler* =
    proc(
      contentKey: ByteList,
      contentId: ContentId,
      content: seq[byte]) {.raises: [Defect], gcsafe.}

```

Then each network will be responsible for creating and setting this callbacks.

Currently all networks use `ContentDb` so default `ContentDb` impls are provided in seprate file.


fluffy/network/wire/portal_protocol.nim
```diff
@@ -313,33 +320,35 @@ proc handleFindContent(
     maxPayloadSize = maxDiscv5PacketSize - talkRespOverhead - contentOverhead
     enrOverhead = 4 # per added ENR, 4 bytes offset overhead
 
-  let contentIdOpt = p.toContentId(fc.contentKey)
```
kdeme, line 316: I think we should keep the content key validation (deserialization) outside of the db code.
That keeps those different types of errors nicely separated and that way you don't need the variant result.

After all, failure on deserialization is a *real* failure. While not finding something in a database is just regular usage. We shouldn't put this under the same umbrella.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -350,16 +359,20 @@ proc handleOffer(p: PortalProtocol, o: OfferMessage, srcId: NodeId): seq[byte] =
...
+      let error = contentResult.error
+      case error.kind
+      of NoKeyInDb:
+        if p.inRange(error.contentId):
           contentKeysBitList.setBit(i)
           discard contentKeys.add(contentKey)
-    else:
-      # Return empty response when content key validation fails
-      return @[]
+      of InvalidContentKey:
```
kdeme, line 373: Same story as above

fluffy/network/wire/portal_protocol.nim
```diff
@@ -1186,68 +1193,13 @@ proc neighborhoodGossip*(
...
-        )
-    of Static:
-      # If the config is set statically, radius is not adjusted, and is kept
-      # constant thorugh node life time, also database max size is disabled
-      # so we will effectivly store fraction of the network
-      p.contentDB.put(key, content)
+    contentKey: ByteList,
+    contentId: ContentId,
+    content: seq[byte]) =
+  if p.portalStore != nil:
```
kdeme, line 1201: `doAssert` instead? Or do we have some case for no storing?

fluffy/network/wire/portal_protocol.nim
```diff
@@ -128,12 +120,27 @@ const
...
 
   DbGetHandler* =
-    proc(contentDB: ContentDB, contentId: ContentId):
-      Option[seq[byte]] {.raises: [Defect], gcsafe.}
+    proc(contentKey: ByteList): Result[seq[byte], DbError] {.raises: [Defect], gcsafe.}
+
+  DbStoreHandler* =
+    proc(
+      contentKey: ByteList,
+      contentId: ContentId,
```
kdeme, line 142: Have both these arguments is not great as one is derived from the other, but I understand you'd want `contentKey` to be more flexible for future handlers for other networks.

I think having the `contentKey` in its deserialized form is probably fine? content id derivation can happen inside the call if needed. Shouldn't result in double derivations from what I see in the code that calls it.
Might result in encoding twice however (depends on toContentId implementation, but it is the case for history network).

However, you do the encoding twice also now as you don't reuse the variable ;)

fluffy/network/wire/portal_protocol.nim
```diff
@@ -158,9 +165,9 @@ type
     protocolId*: PortalProtocolId
     routingTable*: RoutingTable
     baseProtocol*: protocol.Protocol
-    contentDB*: ContentDB
     toContentId*: ToContentIdHandler
     dbGet*: DbGetHandler
+    portalStore*: DbStoreHandler
```
kdeme, line 170: Naming between the getter and setter should be similar.

eg. `dbGet`, `dbPut`. or `getContent` and `storeContent`. etc.

fluffy/network/history/history_network.nim
```diff
@@ -15,6 +15,7 @@ import
   ../../content_db,
   ../../../nimbus/constants,
   ../wire/[portal_protocol, portal_stream, portal_protocol_config],
+  ../content_db_callbacks,
```
kdeme, line 18: From your explanation, these are and should remain directly related to the `ContentDB`. In that case I think they should be exported from there.

fluffy/data/history_data_seeding.nim
```diff
@@ -47,14 +48,17 @@ proc propagateEpochAccumulator*(
...
     info "Gossiping epoch accumulator", rootHash
 
     p.storeContent(
-      history_content.toContentId(key), encodedAccumulator)
+      encKey,
+      history_content.toContentId(encKey),
+      encodedAccumulator
+    )
     discard await p.neighborhoodGossip(
       ContentKeysList(@[encode(key)]), @[encodedAccumulator])
```
kdeme, line 63: This is for example the double encoding I'm talking about in my other comment

fluffy/network/wire/portal_protocol.nim
KonradStaniec: From my point of view having both:
```
contentKey: ByteList,
contentId: ContentId
```

in function signature effectively means that the caller is responsible for  deriving correct contentId from contentKey and function should not double check it.

I was thinking about creating new object like:

```
ValidContentId = object
  contentKey: ByteList,
  contentId: ContentId
```

to make it more clear more api wise, but ultimately decided that it is a bit too much ceremony at this point

fluffy/network/wire/portal_protocol.nim
KonradStaniec: Yea the option was to either have `DbGetHandler` as 
``` 
DbGetHandler* =
    proc(contentKey: ByteList): Result[seq[byte], DbError] {.raises: [Defect], gcsafe.}
```

and make the handler do the validation of contentKey (and then it requires this varint to handle additional case)

or 

```
DbGetHandler* =
    proc(contentKey: ByteList, contentId: ContentId): results.Opt[seq[byte]] {.raises: [Defect], gcsafe.}
```

So make the caller responsible for deriving contentId.

I ultimately went with first one as I saw the chance to get rid of `ToContentIdHandler` from `portal_protocol` , as it is currently only used in `neighborhoodGossip` and could be easily delted from there, by requirig passing contentIds by the called.


But on the second thought, the second version is nicely symmetrical with `DbStoreHandler` and both of them look almost like pretty standard db interface.

fluffy/network/history/history_network.nim
KonradStaniec: yea I wanted to leave `ContentDB` without dependency on `PortalProtocol`, but given the fact that whole pruning thing is designed around radius it is either way implicit dependency.

fluffy/network/wire/portal_protocol.nim
kdeme: Well, from the db point of view, the `contentId` is the db key currently, and the `contentKey` (which is solely the network key, not db key) is not used directly for db access.

If we have in mind to use content key and not content id as db key in some future different content db, then this would needed to be passed as parameter. But I'm not sure this is the case.

If it is the case, and like I said above, I think it would be better to separate the deserialization errors from the dbGet code.

KonradStaniec: sure, I will go with second callback version and leave `ToContentIdHandler` as separate validator of content keys.

I have in mind using content keys to store light client updates with key being light client update period. Then getting ranges it pretty efficient operation, also deleting oldest is pretty efficient then.

kdeme: fyi, you can still pass the just the `ContentKey` then. I don't think it will result in multiple content id derivations (as that should be avoided as it does hashing).

KonradStaniec: hmm but then I will need to handle case when `toContentId` fails, so return this var int.

One way around it would be to make portal protocol a bit more generic i.e have callback

```
toContentKey[A]* = proc(contentKey: ByteList): Result[A, string] {.raises: [Defect], gcsafe.}
```

then all db handlers could take `contentKey: A`  as parameters. Then each network would be that parametrize this type param.

Although I am not sure we need such sophistication right now.

kdeme: > hmm but then I will need to handle case when `toContentId` fails, so return this var int.

No, I'm talking about `ContentKey`. Not `ByteList`. So the deserialisation is done. Content Id deriviation on a proper `ContentKey` object should never fail.

KonradStaniec: ok, so then you need to make it generic on portal network level, as different network have different `ContentKey`. I would say this is good next step.

PR #1301, opened 2022-11-16
Initial implementation of getting range of lc updates

KonradStaniec:
Add initial separate db for portal light client and initial implementation of get range of updates.

There are still some rough edges which needs to be solved on spec level (like content keys for optimistic and final updates) but this pr completes at least be bones implementations.


fluffy/network/beacon_light_client/light_client_content.nim
```diff
@@ -41,12 +41,21 @@ type
...
 
-  #TODO Following types will need revision and improvements
   LightClientUpdateKey* = object
+    startPeriod*: uint64
+    count*: uint64
 
+  #TODO Following types will need revision and improvements
+  # - Cannot user empty objects as SSZ spec does not support containers without
+  # fields
+  # - Cannot use any other type which serializes to empty bytearray (like List[byte, 0])
```
kdeme, line 51: You should be able to do this by using `discard` in the actual `ContentKey` case enumeration. 

We used to do that for the `MasterAccumulator` key before it got removed (https://github.com/status-im/nimbus-eth1/blob/d6673468345da112c672aa0714eeb37c84007f51/fluffy/network/history/history_content.nim#L50).

And we are still doing this for the Headers with proof: https://github.com/status-im/nimbus-eth1/blob/master/fluffy/network/history/history_content.nim#L132

It results in a `None` which is allowed as first type in the enum. I mentioned this also in the portal spec PR.

It did just make me realize that SSZ spec states that the first value may be the None. Which is not the case here. Also, it doesn't allow for two `None`s I guess. Although it is a bit ambiguously defined.

kdeme: By the way, is there some value to having a hash in there? Basically to verify if what a node is offering is something new to what we have already?

fluffy/network/beacon_light_client/light_client_db.nim
```diff
@@ -0,0 +1,192 @@
+# Nimbus
+# Copyright (c) 2021-2022 Status Research & Development GmbH
```
kdeme, line 2:
```suggestion
# Copyright (c) 2022 Status Research & Development GmbH
```

fluffy/network/beacon_light_client/light_client_db.nim
```diff
@@ -0,0 +1,192 @@
...
+      # TODO: as this should not fail, maybe it is better to raiseAssert ?
+      if contentKeyResult.isNone():
+        return Opt.none(seq[byte])
+
+      let ck = contentKeyResult.get()
+
+      if ck.contentType == lightClientUpdate:
+        let
+          # TODO: add validation that startPeriod is not from the furure,
+          # this requires db to be aware to current beacon time
```
kdeme, line 138: Sounds more like this validation would belong outside of the database handling

```suggestion
          # TODO: add validation that startPeriod is not from the future,
          # this requires db to be aware off the current beacon time
```

fluffy/network/beacon_light_client/light_client_db.nim
```diff
@@ -0,0 +1,192 @@
...
+# Copyright (c) 2021-2022 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
+
+import
+  std/[options, heapqueue],
```
kdeme, line 11: heapqueue?

fluffy/network/beacon_light_client/light_client_content.nim
KonradStaniec: hmm maybe not hash, as hash does not tell us anything if the update is useful to our local node i.e if it advances our store in any way. 

Knowledge of the slots would be useful.
 
`OptimisticUpdate` have inside `attested_header` which esentialy validaties optimistic slot so we could have, so for this update we could have `light_client_optimistic_update_key   = Container(optimisticSlot: uint64)`

`FinalityUpdate` have inside `attested_header` and `finalized_header` so we could have some thing like `light_client_finality_update_key  = Container(optimisticSlot: uint64, finalSlot: uint64)`

With those we could know before hand if node is offering to us something which improves our local situation

KonradStaniec: Yup ssz only allow for one `None` , and if you have first selector as `None` then the second type option must be not None -  `Must have at least 2 type options if the first is None`

Our ssz impl follows this pretty good becouse it does not even compile two `None` option i.e
```
import
  ssz_serialization,
  ssz_serialization/codec,
  ssz_serialization/types


type
  ContentType* = enum
    lightClientFinalityUpdate = 0x00
    lightClientOptimisticUpdate = 0x01

  ContentKey* = object
    case contentType*: ContentType
    of lightClientFinalityUpdate:
      discard
    of lightClientOptimisticUpdate:
      discard


let key = ContentKey(contentType: lightClientOptimisticUpdate)
let encoded = SSZ.encode(key)
let decoded = SSZ.decode(encoded, ContentKey)

```

fails to compile with
```
instantiation of `unsupported` from here
/Users/konradstaniec/Work/status/nimbus-eth1/vendor/nim-stew/stew/shims/macros.nim(245, 18) Error: node is not a symbol
```

fluffy/network/beacon_light_client/light_client_db.nim
KonradStaniec: yea possibly although at this point I am not sure where will be the best place to do it. Thats why this little reminder.

fluffy/network/beacon_light_client/light_client_db.nim
KonradStaniec: hmm leftover for older version of pruning in contentdb, can be deleted here and in contentdb

fluffy/network/beacon_light_client/light_client_content.nim
kdeme: > With those we could know before hand if node is offering to us something which improves our local situation

Yeah, it sounds like a good way to avoid useless data transfers. Lets go with that?



> Our ssz impl follows this pretty good becouse it does not even compile two `None` option i.e
I don't think I actually added that. It probably fails because there is not a "real" value. Adding a third case with some actual value (not discard) probably works. Anyways, not so important, needs to be improved on ssz library side.

KonradStaniec: > Yeah, it sounds like a good way to avoid useless data transfers. Lets go with that?

Sure. It can probably mean something a bit  different depending wheter it is used in FINDCONTENT of OFFERCONTENT context.

So for example in find content context `Container(optimisticSlot: uint64)` optimistic slot can be requesting side current best optimistic slot, then responder may respond only if they have better update.

And in offering context it is just offered update attested header slot. Although it will mean nodes will need to cache locally few latest updates to avoid offering content which they do not have, but this is probably reasonable tradeoff.

kdeme: > And in offering context it is just offered update attested header slot. Although it will mean nodes will need to cache locally few latest updates to avoid offering content which they do not have, but this is probably reasonable tradeoff.

You mean, because they might have received even a newer one? I think in that scenario it is fine to send the even newer one? As in `optimisticSlot` would be treated as the minimum latest slot.

PR #1310, opened 2022-11-21
Add eth_getCode to lc proxy

KonradStaniec:
Add get_code json rpc api to lc proxy

### DIFF
--- a/lc_proxy/rpc/rpc_eth_lc_api.nim
+++ b/lc_proxy/rpc/rpc_eth_lc_api.nim
@@ -13,6 +13,7 @@ import
   stew/[byteutils, results],
   chronicles,
   json_rpc/[rpcproxy, rpcserver, rpcclient],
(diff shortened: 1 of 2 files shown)


lc_proxy/rpc/rpc_eth_lc_api.nim
```diff
@@ -192,6 +196,42 @@ proc installEthApiHandlers*(lcProxy: LightClientRpcProxy) =
     else:
       raise newException(ValueError, accountResult.error)
 
+  lcProxy.proxy.rpc("eth_getCode") do(address: Address, quantityTag: string) -> HexDataStr:
+    let
+      executionPayload = lcProxy.getPayloadByTag(quantityTag)
```
kdeme, line 201: What I don't understand about this is that this is using the block cache underneath right? So it would only work for very recent blocks that it cached since start-up?

PR #1311, opened 2022-11-22
Stateless mode (not working yet)

AdamSpitz:
NOT WORKING YET: New "stateless" mode.

The basic idea is that just *before* we're about to need some account or slot or block header, we prefetch it asynchronously. (e.g. Accounts and slots are fetched by using eth_getProof, then putting the fetched nodes into the database.)

You can start in this mode by running:

    nimbus --sync-mode=stateless --stateless-data-source-url=https://mainnet.infura.io/whatever

(At which point nimbus-eth1 will wait for an eth2 client to feed it blocks to run via newPayload in the Engine API.)

Alternatively, you can run the block with a particular hash by running:

    nimbus statelesslyRun --stateless-data-source-url=https://mainnet.infura.io/whatever --stateless-block-hash=1234ABCetc.

(This will simply fetch the block header corresponding to that hash from the data source, then run it, without bothering to wait for an eth2 client.)

I say "not working yet" because the state roots are still coming out wrong.

Also, block processing times are still much too slow to be practical: on the order of five minutes per block. We'll need to do precalculated witnesses or something like that in order to make this viable.

One more note: I've removed the concurrentAssemblers test that I implemented when I did the basic async EVM stuff; it was useful as a temporary way to exercise the async code, but the test was based on a very hacky idea and it was more trouble than it was worth.


nimbus/db/accounts_cache.nim
```diff
@@ -63,17 +69,36 @@ const
...
   result.db = db
-  result.trie = initSecureHexaryTrie(db, root, pruneTrie)
+  result.trie = initAccountsTrie(db, root, pruneTrie)
   result.witnessCache = initTable[EthAddress, WitnessData]()
+  result.shouldCheckForMissingNodes = shouldCheckForMissingNodes
   discard result.beginSavepoint
 
-proc init*(x: typedesc[AccountsCache], db: TrieDatabaseRef, pruneTrie: bool = true): AccountsCache =
-  init(x, db, emptyRlpHash, pruneTrie)
+proc statelessInit*(x: typedesc[AccountsCache], db: TrieDatabaseRef,
```
zah, line 90: What's the difference between `init` and `statelessInit` here?

nimbus/p2p/executor/process_transaction.nim
```diff
@@ -106,6 +112,18 @@ proc processTransactionImpl(
...
+
+proc processTransactionImpl(
+    vmState: BaseVMState; ## Parent accounts environment for transaction
+    tx:      Transaction; ## Transaction to validate
+    sender:  EthAddress;  ## tx.getSender or tx.ecRecover
+    header:  BlockHeader; ## Header for the block containing the current tx
+    fork:    Fork): Result[GasInt,void]
+    # wildcard exception, wrapped below
+    {.gcsafe, raises: [Exception].} =
+  return waitFor(asyncProcessTransactionImpl(vmState, tx, sender, header, fork))
```
zah, line 125: That's OK as long this is still work-in-progress, but you'll have to restore the proper sync version before merging this PR.

nimbus/vm2/async/data_sources/on_demand.nim
```diff
@@ -0,0 +1,30 @@
+import
```
zah, line 1: This should probably be called `data_sources/json_rpc` or `data_sources/web3`. Fluffy will add another on-demand data source which will probably be called `portal`.

nimbus/vm2/async/data_fetching.nim
```diff
@@ -0,0 +1,225 @@
...
+  ok[void]()
+
+proc storeBlockHeader(chainDB: BaseChainDB, header: BlockHeader) =
+  chainDB.persistHeaderToDbWithoutSetHeadOrScore(header)
+
+proc assertThatWeHaveStoredBlockHeader(chainDB: BaseChainDB, blockNumber: BlockNumber, header: BlockHeader) =
+  let h = chainDB.getBlockHash(blockNumber)
+  doAssert(h == header.blockHash, "stored the block header for block " & $(blockNumber))
+
+template raiseExceptionIfError[E](whatAreWeVerifying: untyped, r: Result[void, E]) =
```
zah, line 137: This should probably be a generic proc to avoid the multiple evaluation of the `r` parameter.

nimbus/vm2/async/rpc_api.nim
```diff
@@ -0,0 +1,157 @@
...
+from web3 import Web3, BlockObject, FixedBytes, newWeb3, fromJson, fromHex
+from ../../premix/downloader import request
+from ../../premix/parser import prefixHex, parseBlockHeader, parseReceipt, parseTransaction
+
+# Trying to do things the new web3 way:
+# from web3 import eth_getProof, Address, StorageProof, blockId, `%`
+# from ../../lc_proxy/validate_proof import getAccountFromProof
+
+
+var durationSpentDoingFetches*: times.Duration
```
zah, line 28: Consider using the `metrics` library for this. This would allow visualizing the data in Grafana once we deploy long-running instances of Nimbus-eth1 on our "fleet" servers. As an example, here are some metrics from our Nimbus eth2 fleet:

https://metrics.status.im/d/pgeNfj2Wz23/nimbus-fleet-testnets?orgId=1&refresh=5m

Take a look at the "Block & Attestation Delay" panels at the bottom of the screen which are currently using histograms:

```nim
# at the module top-level scope
declareHistogram data_fetching_duration,
  "Data fetch duration", buckets = [0.25, 0.5, 1, 2, 4, 8, Inf]

# inside functions
data_fetching_duration.observe(durationInSecondsAsFloat) # Usually based on `Moment.now()`
```

nimbus/vm2/async/rpc_api.nim
```diff
@@ -0,0 +1,157 @@
...
+
+
+proc makeAnRpcClient*(web3Url: string): Future[RpcClient] {.async.} =
+  let myWeb3: Web3 = waitFor(newWeb3(web3Url))
+  return myWeb3.provider
+
+
+proc fetchBlockHeaderWithHash*(rpcClient: RpcClient, h: Hash256): Future[BlockHeader] {.async.} =
+  let t0 = now()
+  let r = request("eth_getBlockByHash", %[%h.prefixHex, %false], some(rpcClient))
```
zah, line 43: `request` uses `waitFor` internally, which is not quite appropriate. Furthermore, all HTTP requests may hang forever unless they are guarded with a timeout. Study the usages of `awaitWithTimeout` from the nimbus-eth2 codebase.

nimbus/vm2/interpreter/op_handlers/oph_blockdata.nim
```diff
@@ -30,9 +31,11 @@ when not defined(evmc_enabled):
 const
   blockhashOp: Vm2OpFn = proc (k: var Vm2Ctx) =
     ## 0x40, Get the hash of one of the 256 most recent complete blocks.
-    let (blockNumber) = k.cpt.stack.popInt(1)
-    k.cpt.stack.push:
-      k.cpt.getBlockHash(blockNumber)
+    let cpt = k.cpt  # so it can safely be captured by the asyncChainTo closure below
```
zah, line 34: Isn't this quite expensive to copy?
Can you explain the problem being solved in more detail? Perhaps we need to get back to the drawing board.

AdamSpitz: I'm a bit confused about what you're asking.

If you're asking why I'm saying `let cpt = k.cpt`, the problem I'm solving is that if I instead refer directly to `k.cpt` inside the asyncChainTo closure, I get error messages like this:

    Error: 'k' is of type <var Vm2Ctx> which cannot be captured as it would violate memory safety, declared here: /home/adam/Projects/nimbus-eth1/nimbus/vm2/interpreter/op_handlers/oph_blockdata.nim(32, 32)

By doing what I did, I avoid the need for the closure to capture k. I don't actually need k (the Vm2Ctx) inside the asyncChainTo closure; I only need cpt (the Computation). (And this turns out to be true in all of the places where I've used this pattern.)

I expected this to be cheap; the Computation type is defined as a `ref object`, so I assumed that all that was being copied into the closure was a single reference. (But I didn't actually look at the C code to verify that.) Am I wrong?

PR #1312, opened 2022-11-22
Handle blocks from cache

KonradStaniec:
Until now, blocks provided by `eth_getBlockByNumber` and `eth_getBlockByHash` was served from untrusted data source.

With this pr, the blocks will be served from proxy internal cache of blocks validated by beacon light client. This means that only most recent blocks will be handled (which is line with proxy intended use cases)

### DIFF
--- a/lc_proxy/block_cache.nim
+++ b/lc_proxy/block_cache.nim
@@ -60,3 +60,8 @@ proc getByNumber*(
       break
 
   return payloadResult
+
+proc getByHash*(
+    self: BlockCache,
+    hash: BlockHash): Opt[ExecutionPayloadV1] =
+  return self.blocks.eq(hash)
--- a/lc_proxy/block_cache.nim
+++ b/lc_proxy/block_cache.nim
@@ -61,7 +61,7 @@ proc getByNumber*(
 
   return payloadResult
 
-proc getByHash*(
+proc getPayloadByHash*(
     self: BlockCache,
     hash: BlockHash): Opt[ExecutionPayloadV1] =
   return self.blocks.eq(hash)
--- a/lc_proxy/rpc/rpc_eth_lc_api.nim
+++ b/lc_proxy/rpc/rpc_eth_lc_api.nim
@@ -113,7 +113,7 @@ proc getPayloadByTagOrThrow(
   let tagResult = getPayloadByTag(proxy, quantityTag)
 
   if tagResult.isErr:
-     raise newException(ValueError, "No block for given tag " & quantityTag)
+    raise newException(ValueError, "No block stored for given tag " & quantityTag)
 
   return tagResult.get()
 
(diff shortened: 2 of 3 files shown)


lc_proxy/block_cache.nim
```diff
@@ -60,3 +60,8 @@ proc getByNumber*(
       break
 
   return payloadResult
+
+proc getByHash*(
```
kdeme, line 64:
```suggestion
proc getBlockByHash*(
```
or 

```suggestion
proc getPayloadByHash*(
```

lc_proxy/rpc/rpc_eth_lc_api.nim
```diff
@@ -98,25 +99,25 @@ proc getPayloadByTag(
...
+    proxy: LightClientRpcProxy,
+    quantityTag: string): ExecutionPayloadV1 {.raises: [ValueError, Defect].} =
 
-  return payload
+  let tagResult = getPayloadByTag(proxy, quantityTag)
 
-proc installEthApiHandlers*(lcProxy: LightClientRpcProxy) =
-  template payload(): Opt[ExecutionPayloadV1] = lcProxy.executionPayload
+  if tagResult.isErr:
+     raise newException(ValueError, "No block for given tag " & quantityTag)
```
kdeme, line 116:
```suggestion
     raise newException(ValueError, "No block stored for given tag " & quantityTag)
```

To state that it is not necessarily an error on their end.

lc_proxy/rpc/rpc_eth_lc_api.nim
```diff
@@ -239,9 +240,23 @@ proc installEthApiHandlers*(lcProxy: LightClientRpcProxy) =
...
+  lcProxy.proxy.rpc("eth_getBlockByNumber") do(quantityTag: string, fullTransactions: bool) -> Option[BlockObject]:
+    let executionPayload = lcProxy.getPayloadByTag(quantityTag)
+
+    if executionPayload.isErr:
+      return none(BlockObject)
+
+    return some(asBlockObject(executionPayload.get()))
+
+  lcProxy.proxy.rpc("eth_getBlockByHash") do(blockHash: BlockHash, fullTransactions: bool) -> Option[BlockObject]:
+    let executionPayload = lcProxy.blockCache.getByHash(blockHash)
```
kdeme, line 254:
```suggestion
    let executionPayload = lcProxy.blockCache.getPayloadByHash(blockHash)
```
Lets use payload so it is consistent with `getPayloadByTag`

lc_proxy/rpc/rpc_utils.nim
```diff
@@ -0,0 +1,86 @@
...
+    timestamp     : fromUnix payload.timestamp.unsafeQuantityToInt64,
+    extraData     : bytes payload.extraData,
+    mixDigest     : payload.prevRandao.asEthHash,
+    nonce         : default(etypes.BlockNonce),
+    fee           : some payload.baseFeePerGas
+  )
+  return uint64(len(rlp.encode(bh)))
+
+proc asBlockObject*(p: ExecutionPayloadV1): BlockObject =
+  # TODO currently we always calculate txHashes as BlockObject does not have
```
kdeme, line 55: In Nimbus and Fluffy we use the `BlockObject` of the local RPC code. There we do have this implemented (although twice due to yet other differences).

Having it both in nimbus-eth1 and web3 is really a pain, especially with all the slight differences. But ok, not something to fix here I guess. It keeps lingering on however.

kdeme (approved):
Looks fine. 

I wasn't so sure at first about actually replacing these calls to use the cache, but I don't know the details on the use cases for them in wallets, especially on older blocks. Are there any actually? So I think it is fine. 

I was basically thinking that you probably get all the data on a request to reconstruct the header and body and verify them when the block was requested by hash? But by number would still be an issue and require a local mapping that is simply too large.

PR #1376, opened 2022-12-15
Add experimental beacon chain block proofs for Portal network

kdeme:
PoC to show on how to possibly verify on the Portal history network if a BlockHeader (and thus body/receipts) is part of canonical chain, post merge. Pre-merge, the hardcoded portal accumulator is already used.

Currently completely unspecified and also not possible to use as there is no access to historical_roots possible over the network.


kdeme:
Going to merge this even though it is experimental and not currently used in the Fluffy code.
Merging because it is well isolated from other code yet still has tests to keep it from bit rot.

PR #1386, opened 2022-12-16
Add light client bridge

KonradStaniec:
Add light client bridge which is a node which follows libp2p lc updates and at the same time it is participant in portal light client network. Updates are saved in portal light client db format.

### DIFF
--- a/fluffy/network/beacon_light_client/beacon_light_client_bridge.nim
+++ b/fluffy/network/beacon_light_client/beacon_light_client_bridge.nim
@@ -1,0 +1,230 @@
+# Nimbus
+# Copyright (c) 2022 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
+
+import
+  std/[os, strutils],
+  chronicles, chronicles/chronos_tools, chronos,
+  eth/keys,
+  beacon_chain/eth1/eth1_monitor,
+  beacon_chain/gossip_processing/[optimistic_processor, light_client_processor],
+  beacon_chain/spec/beaconstate,
+  beacon_chain/spec/datatypes/[phase0, altair, bellatrix],
+  beacon_chain/[light_client, nimbus_binary_common, version],
+  "."/[
+    light_client_db,
+    light_client_network,
+    light_client_content,
+    beacon_light_client_bridge_conf
+  ],
+  ".."/wire/[portal_stream, portal_protocol_config, portal_protocol]
+
+# TODO Find what can throw exception
+proc run() {.raises: [Exception, Defect].} =
+  {.pop.}
+  var config = makeBannerAndConfig(
+    "Beacon light client bridge " & fullVersionStr, BridgeConf)
+  {.push raises: [Defect].}
+
+  # Required as both Eth2Node and LightClient requires correct config type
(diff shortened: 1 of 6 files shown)


fluffy/network/beacon_light_client/beacon_light_client_bridge.nim
```diff
@@ -0,0 +1,230 @@
+# Nimbus
```
kdeme, line 1: I don't think we should put this application under `fluffy/network/beacon_light_client`. I consider these folders purely for code more related to the core specification.

So I'd suggest to get either a bridge folder straight under `fluffy` or in `fluffy/tools`

kdeme (commented):
I understand that this data is getting saved directly into the Portal ContentDB (`portalProtocol.storeContent()`).

How is the gossip over the network supposed to happen?

I kind of assumed that there would be a full fluffy node running, which gets these updates from the consensus light client, and immediately gossips them into the network. But this is not the case.

Will you launch the gossip somehow separately?

kdeme:
> I kind of assumed that there would be a full fluffy node running, which gets these updates from the consensus light client, and immediately gossips them into the network. But this is not the case.

fyi, for this reason I was actually thinking of having it either as:
- Integrated in Fluffy itself, with special run option as consensus light client bridge
- Separate tool that makes use of the portal JSON RPC API to gossip & store data.

The latter sounds cleaner in terms of separation and would in theory allow for other clients to use the bridge also (when they get consensus light client network implemented).
The first has less moving parts and as a dev it is slight easier to just run one tool.

KonradStaniec:
> Integrated in Fluffy itself, with special run option as consensus light client bridge

I really not a fan of bringing whole eth2 consensus world directly into fluffy client. I would keep those separated and have fluffy to use only portal powered light client.

> How is the gossip over the network supposed to happen?
> 
> I kind of assumed that there would be a full fluffy node running, which gets these updates from the consensus light client, and immediately gossips them into the network. But this is not the case.
> 
> Will you launch the gossip somehow separately?

I am not sure why full fluffy node with all other networks would be necessary here. As it currently stands bridge is member of portal light client network, so other nodes can find it in it s routing table and also can find its content.

As far as gossip, it now just the question of when and how to `neighborhoodGossip`, as I see it:
- gossiping bootstraps is quite unknown, we could do it, but receiving side would need to explicitly have such bootstrap 
- continuous batch of `LightClientUpdate` can be offered periodically to neighbouring peers.
- finality and optimistic update, can be probably offered after each received update.

> Separate tool that makes use of the portal JSON RPC API to gossip & store data.

Not sure why we would need JSON RPC API to gossip & store data. In my mind bridge, was this pretty small piece of software, which part of two worlds - libp2p and portal. It downloads and validates objects from libp2p world, and makes them available in portal light client network.

fluffy/network/beacon_light_client/beacon_light_client_bridge.nim
KonradStaniec: ah I got it there initially, but wanted to make `beacon_light_client` module more self contained. Sure, I can bring it back up there :+1:

kdeme:
Replying here for reference:

> I really not a fan of bringing whole eth2 consensus world directly into fluffy client. I would keep those separated and have fluffy to use only portal powered light client.

Which is why we could go for a separated tool/bridge that interacts with a fluffy node through the standardized Portal JSON-RPC to store & gossip data.


> I am not sure why full fluffy node with all other networks would be necessary here. As it currently stands bridge is member of portal light client network, so other nodes can find it in it s routing table and also can find its content.

I wasn't fully aware of that intent here as only `storeContent` is being done. Meaning no gossip would every be executed.
Hence my response of where the gossip would be done.

> Not sure why we would need JSON RPC API to gossip & store data. In my mind bridge, was this pretty small piece of software, which part of two worlds - libp2p and portal. It downloads and validates objects from libp2p world, and makes them available in portal light client network.

And it could still be a small piece of software. By only doing the libp2p part and sending data to a Fluffy client over JSON-RPC. And it would be consistent with for example an eth1 bridge. And compatible with other clients.

PR #1396, opened 2022-12-23
Add light client mode to fluffy

KonradStaniec:
Adds light client mode to fluffy.


kdeme (commented):
It is unclear to me what the current use for this light client mode in fluffy is.

Is it of a temporary intent until we can get/verify the data through the Portal beacon light client network?

PR #1401, opened 2022-12-28
introduce runtime EIPs management

jangko:
the EVMC interface become a barrier when talking about runtime EIP configuration. It only works on the EVM host side but currently there is no way to tell the EVM about which EIP(s) should be switch on or off.

related issue ethereum/evmc#453


jangko:
It is nearly impossible to add this feature while the EVM still using compile time construct. It will only works for few EIP. closing.

PR #1409, opened 2023-01-09
Some work on EIP-4895 (withdrawals).

AdamSpitz:
I don't want to actually merge this into the master branch yet. Don't trust it.

Block execution checks the withdrawals in the body against the withdrawalsRoot in the header, then adds the withdrawal amounts to the appropriate account balances.

I think I've implemented the V2 pieces of the Engine API.

I still need to implement timestamp-based (rather than block-number-based) forking. And I bet I'm missing some other details, too.


nimbus/rpc/engine_api.nim
```diff
@@ -41,322 +41,402 @@ proc invalidFCU(com: CommonRef, header: EthBlockHeader): ForkchoiceUpdatedRespon
...
 
-  # https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#engine_newpayloadv1
-  # cannot use `params` as param name. see https:#github.com/status-im/nim-json-rpc/issues/128
-  server.rpc("engine_newPayloadV1") do(payload: ExecutionPayloadV1) -> PayloadStatusV1:
-    trace "Engine API request received",
-      meth = "newPayloadV1", number = $(distinctBase payload.blockNumber), hash = payload.blockHash
+# FIXME-Adam: How do I do this? IIUC, Shanghai (and later) forks
+# happen based on timestamp rather than block number.
+#
+# The stuff in hardforks.nim and chain_config.nim seems relevant.
```
jangko, line 54: yes, the hardfork transition should be handled in one place, in `common/common.nim` and `hardforks.nim`. the td vs ttd comparison in engine api should be fixed too, of course in another PR.

AdamSpitz:
I ended up recreating these changes from scratch because the bits rotted. That's merged in now, so I'm closing this PR.

PR #1413, opened 2023-01-10
implement EOF suite: EIP-3540, 3670, 4200, 4750, and 5450

jangko:
I set the schedule to Cancun. Only need to modify 4 lines of code to move it back to Shanghai if needed.
It pass deployment test in `test_eof.nim`. I try to test it against Shanghai and EOF test vectors from ethereum/tests, but the test vectors itself have bugs.

Do not merge it until we have confirmation about which hard fork to include these EIPs and when it pass test from ethereum/tests and the test vectors have been fixed.


jangko:
If EIP-1153(Transient Storage) accepted, then add TLOAD, and TSTORE to stack table.

tersec:
Replaced by https://github.com/status-im/nimbus-eth1/pull/3229

PR #1414, opened 2023-01-10
Update eth/6? messages Get/PooledTransactions

mjfh:


nimbus/sync/handlers.nim
```diff
@@ -385,15 +387,22 @@ method getReceipts*(ctx: EthWireRef, hashes: openArray[Hash256]): seq[seq[Receip
...
+  var someActive = false
   for txHash in hashes:
     let res = txPool.getItem(txHash)
     if res.isOk:
-      result.add res.value.tx
+      someActive = true
+      result.add some(res.value.tx)
     else:
-      result.add Transaction()
+      result.add none(Transaction)
```
jangko, line 399: after I re-read the spec, turn out the response of GetPooledTransactions should be like this: :)

```text
The transactions must be in same order as in the request, but it is OK to skip
transactions which are not available. This way, if the response size limit is reached,
requesters will know which hashes to request again (everything starting from the last
returned transaction) and which to assume unavailable (all gaps before the last returned
transaction).
```

It is OK to skip transactions which are not available.

I think empty RLP encoding will not work.

mjfh: Ahh good point ok, that makes it easier. I mis-read slightly.

But then the last paragraph at  `PooledTransactions`: "A peer may respond with an empty list iff none of the hashes match transactions in its pool." does not give any additional information. It just states the obvious.

mjfh:
superseded by #1415

PR #1415, opened 2023-01-10
Update eth/6? messages Get/PooledTransactions

mjfh:
supersedes #1414


nimbus/sync/handlers.nim
```diff
@@ -392,7 +392,6 @@ method getPooledTxs*(ctx: EthWireRef, hashes: openArray[Hash256]): seq[Transacti
     if res.isOk:
       result.add res.value.tx
     else:
-      result.add Transaction()
```
jangko, line 395: no objections

Issue #1424, opened 2023-01-11
Timestamp-based forking for Shanghai (and later)


AdamSpitz:
Activation of the Shanghai fork will be determined by a timestamp, not a block number.

e.g. [EIP-4895](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4895.md) says, "Beginning with the execution timestamp FORK_TIMESTAMP, execution clients MUST introduce the following extensions to payload validation and processing..."

jangko had some comments [here](https://github.com/status-im/nimbus-eth1/pull/1409) on how to go about implementing this.

  - We probably need both shanghaiBlock and shanghaiTime, because hive and ethereum/tests are still using a block number.
  - The relevant files are common/common.nim and hardforks.nim.

AdamSpitz:
Note also that geth has put in a hard-coded value called timestampThreshold, used to decide whether a forkid.next field is a block number or a timestamp.

https://github.com/ethereum/go-ethereum/blob/master/core/forkid/forkid.go#L48

jangko:
in EIP-4895, geth still keep both `shanghaiBlock` and `shanghaiTime`. but in EIP-4844, as can be seen in this [commit](https://github.com/mdehoog/go-ethereum/commit/b2d6f07276f6cb3cf9e306333eab56fc54324f9c), they remove `shanghaiBlock` and use only timestamp.
Sure this will complicated fork transition and forkid calculation.

jangko:
fixed by #1465. closing

Issue #1410, opened 2023-01-10
remove forkOverride from test runner and processTransaction


jangko:
The early general state test is using fork override because at that time there is no proper hard fork management.
It was handy at that time, but now it becomes unnecessarily complicated.

Multiple source of hardfork only cause confusion and maintenance cost.

places where to look at when addressing this issue:
- test_generalstate_json.nim
- evmstate tool
- BaseVMState.fork usage, maybe in evm_call
- processTransaction

PR #1434, opened 2023-01-16
Time based forking

AdamSpitz:
https://github.com/status-im/nimbus-eth1/issues/1424

Replaced the ForkToBlockNumber array with a (more complicated) object called ForkTransitionTable that knows the difference between block-based forks, time-based forks, and the special cases like MergeFork and Shanghai (which IIUC needs to be doable via either block number or time, at least temporarily).

Created a type called ForkDeterminationInfo to represent the information needed to determine which hard fork a particular block is on. In the past all we needed was a block number; now we need to pass in both a block number and timestamp (and if mergeForkBlock isn't specified, maybe TD as well), so I created a type for it (though I don't love the name).

Several things are still unresolved. (See various places that say FIXME-Adam.) But I think this is progress.

### DIFF
--- a/nimbus/common/hardforks.nim
+++ b/nimbus/common/hardforks.nim
@@ -8,7 +8,7 @@
 # those terms.
 
 import
-  std/options,
+  std/[options, times],
   eth/common,
   json_serialization,
   ../utils/utils,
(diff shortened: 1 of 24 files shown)


nimbus/common/hardforks.nim
```diff
@@ -47,12 +47,119 @@ type
...
+# FIXME: Is this called anywhere?
+func forkDeterminationInfoIncludingTd*(n: BlockNumber, t: EthTime, td: DifficultyInt): ForkDeterminationInfo =
+  ForkDeterminationInfo(blockNumber: n, time: some(t), td: some(td))
+
+func adjustForNextBlock*(t: EthTime): EthTime =
+  # FIXME-Adam: what's the right thing to do here?
+  # How do we calculate "the timestamp for the block
+  # after this one"?
+  #
+  # If this makes no sense, what should the callers
```
jangko, line 116: see my comment below for `evm/state.nim`

nimbus/common/hardforks.nim
```diff
@@ -254,56 +401,104 @@ func calculateForkIds*(c: ChainConfig,
...
+    result.toFork = forkFalse
+
+func time_blockToFork(t: Option[EthTime]): BlockToFork =
+  if t.isSome:
+    result.data = t.get.toUnix.u256
+    result.toFork = timeMaybe
+  else:
+    result.toFork = forkFalse
+
+# FIXME: is BlockToForks even necessary anymore? I feel like
```
jangko, line 476: If you think it's enough to keep only `ForkTransitionTable`, then please do it.

nimbus/common/hardforks.nim
```diff
@@ -47,12 +47,119 @@ type
...
+    td: f.td
+  )
+
+# This function is awkward because there are various different ways now of
+# doing a hard-fork transition (block number, ttd, time, block number *or*
+# time). We used to have a simple array called forkToBlock that mapped each
+# HardFork to a BlockNumber; now we have this ForkTransitionTable, which
+# contains a couple of arrays and also special cases for MergeBlock and
+# Shanghai.
+func isGTETransitionThreshold*(map: ForkTransitionTable, forkDeterminer: ForkDeterminationInfo, fork: HardFork): bool =
```
jangko, line 133: Please add test for `isGTETransitionThreshold`.

nimbus/core/executor/process_transaction.nim
```diff
@@ -129,7 +129,10 @@ proc processTransaction*(
     {.gcsafe, raises: [Defect,CatchableError].} =
   ## Variant of `processTransaction()` with `*fork* derived
   ## from the `vmState` argument.
-  let fork = vmState.com.toEVMFork(header.blockNumber)
+  ##
+  ## FIXME-Adam: Hmm, I'm getting the block number and timestamp from
```
jangko, line 133: I have a plan to remove this fork override in near future, see #1410.

PR #1442, opened 2023-01-23
attempt to make ci green

jangko:


kdeme:
I don't think we can make it green for fluffy and verified proxy as it uses nimbus-eth2 and to bump that one we need to move nimbus-eth1 also to Nim 1.6.

So I think I'll revert the bumps and first try to move nimbus-eth1 to Nim 1.6

jangko:
close in favor of #1448

PR #1445, opened 2023-01-23
Bump Nim to 1.6 and resolve the related issues

kdeme:
- Still has 3 tests that are commented out. Appears to be an issue with the ./tests/macro_assembler.nim. To be looked into
- I'm aware that we can get rid of the `raises[Defect]` to `raises[]`, but not doing this in this PR yet.


nimbus/db/accounts_cache.nim
```diff
@@ -135,13 +135,20 @@ proc getAccount(ac: AccountsCache, address: EthAddress, shouldCreate = true): Re
...
-      account: rlp.decode(recordFound, Account),
-      flags: {IsAlive}
-      )
+    try:
+      result = RefAccount(
+        account: rlp.decode(recordFound, Account),
+        flags: {IsAlive}
+        )
+    except RlpError:
+      raiseAssert("No RlpError should occur on decoding account from trie")
```
kdeme, line 151: idem^

kdeme:
Will go ahead and merge this, the macro_assembler will be fixed later so that those 3 tests can be run again

PR #1462, opened 2023-02-10
Silence some compiler gossip -- part 6, evm

mjfh:


nimbus/evm/computation.nim
```diff
@@ -22,6 +22,8 @@ import
 export
   common
 
+{.push raises: [].}
```
kdeme, line 25: nitpick: we put or `push raises` before the imports normally.

kdeme (approved):

nimbus/evm/computation.nim
mjfh: ok :)

.. still solving some issues with the windows system

PR #1465, opened 2023-02-14
Time based forking

AdamSpitz:
This problem is driving me a little bit crazy.

I recreated my changes from scratch, but for some reason I'm still causing "make fluffy-test" to fail because of that side-effect thing. I don't understand why.


kdeme (commented):
Debugging this together with @AdamSpitz

Original error occurs only with changes on this branch and not in master:
```
/home/runner/work/nimbus-eth1/nimbus-eth1/fluffy/network/history/history_network.nim(202, 22) template/generic instantiation of `error` from here
/home/runner/work/nimbus-eth1/nimbus-eth1/vendor/nim-stew/stew/results.nim(824, 24) template/generic instantiation of `raiseResultDefect` from here
/home/runner/work/nimbus-eth1/nimbus-eth1/vendor/nim-stew/stew/results.nim(372, 6) Error: 'raiseResultDefect' can have side effects
> /home/runner/work/nimbus-eth1/nimbus-eth1/vendor/nim-stew/stew/results.nim(374, 63) Hint: 'raiseResultDefect' calls `.sideEffect` '$'
>> /home/runner/work/nimbus-eth1/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/dollars.nim(89, 6) Hint: '$' called by 'raiseResultDefect'
>>> /home/runner/work/nimbus-eth1/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/dollars.nim(111, 15) Hint: '$' calls `.sideEffect` 'addQuoted'
>>>> /home/runner/work/nimbus-eth1/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system.nim(2887, 6) Hint: 'addQuoted' called by '$'
```

There is basically the `raiseResultDefect` call done from the `error` call in Results, which are `func`s and thus don't allow side effects. If the `$` call for the specific Value type in the Result has side effects, this fails.

The `$` call used appears to be the Nim system generic one for objects. The error however is specific on the location where it calls the `$` for a subfield. 

The value in `history_network.nim` that causes the issue is `BlockHeader`. Adding a custom `$` for it in history_network works around the compiler error. 
After further debugging, it turns out adding a custom `$` which just prints the `timestamp` field does the following:
- Compiles as is.
- Fails compilation when `std/times` is imported.

Importing `std/times` on master also generates the same error. So the issue might be that suddenly `times` import leaks further  through some modules. There has been an an additional import of times in this Branch.
Not 100% sure that this is the field in `BlockHeader` causing it, but likely. To be further investigated...

nimbus/common/hardforks.nim
```diff
@@ -8,7 +8,7 @@
 # those terms.
 
 import
-  std/options,
+  std/[options, times],
```
kdeme, line 11: Perhaps this one...

kdeme:
> After further debugging, it turns out adding a custom `$` which just prints the `timestamp` field does the following:
> 
>     * Compiles as is.
> 
>     * Fails compilation when `std/times` is imported in `history_network.nim`

Come to think of it, it might not be the `timestamp` that gives the issue, as it wouldn't compile as is on this branch either then normally...

nimbus/common/hardforks.nim
kdeme: Or not..

```
import std/times except `$`
```
instead does not work

kdeme:
> 
> Come to think of it, it might not be the `timestamp` that gives the issue, as it wouldn't compile as is on this branch either then normally... Although it does look like the more likely one as its `$` does have side effects.

To be certain, I made following change, which does compile:
```
diff --git a/lib/pure/times.nim b/lib/pure/times.nim
index 113f73d2a..60f7cbbd3 100644
--- a/lib/pure/times.nim
+++ b/lib/pure/times.nim
@@ -2174,7 +2174,8 @@ proc `$`*(time: Time): string {.tags: [], raises: [], benign.} =
     let dt = dateTime(1970, mJan, 01, 00, 00, 00, 00, local())
     let tm = dt.toTime()
     doAssert $tm == "1970-01-01T00:00:00" & format(dt, "zzz")
-  $time.local
+  {.noSideEffect.}:
+    $time.local
 
 #
 # TimeInterval

```


So it definitely seems to be this `Time` `$` leaking. Where/how and why now, I don't know.

kdeme:
So yeah, I cannot explain why, but the most "fun" workaround is adding:

```
proc `$`*(x: BlockHeader): string =
  $x
```

We can go with that for now as I cannot figure out in reasonable time frame how to avoid hitting the `Time` its `$`, sigh.

kdeme:
@AdamSpitz It's a slightly differrent, yet similar, error now on a different location / test.

PR #1479, opened 2023-02-23
@

mjfh:
Test suite for gc crash

Compile run as:
```
./gabbleblotchit.mk nobrun # crashing binary w/default gc
```
or
```
./gabbleblotchit.mk brun  # not crashing on boehm gc 
```

Little help page:
```
./gabbleblotchit.mk
```


mjfh:
maybe solved by #2205

PR #1482, opened 2023-02-28
More work on withdrawals

AdamSpitz:
I don't think this is done yet, but I believe the tests should pass, and I want to get this merged in before the bits rot.


AdamSpitz:
Okay, looks like CI finally passes. I'm gonna merge. The Shanghai stuff isn't *done*, but this is at least some progress.

PR #1488, opened 2023-03-07
Bump Nim compiler (mitigate gc problem)

mjfh:


zah:
The bump to Nim 1.6.12 has been merged in alternative PR. Nim 1.6.12 includes the necessary fix for the GC issue.

PR #1497, opened 2023-03-14
nimbus-verified-proxy integration with status-go

siphiuel:
This is depended on by https://github.com/siphiuel/lc-proxy-wrapper.

Custom `nim-confutils` dependency is due to adding a fix suggested in https://github.com/status-im/nim-confutils/issues/31
.
Tracking issue: https://github.com/status-im/status-go/issues/3282

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,7 @@ TOOLS_CSV := $(subst $(SPACE),$(COMMA),$(TOOLS))
 	nimbus \
 	fluffy \
 	nimbus_verified_proxy \
+	libverifproxy \
 	test \
 	test-reproducibility \
 	clean \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -272,9 +273,52 @@ evmstate_test: | build deps evmstate
...
+	$(ENV_SCRIPT) nim c --noLinking:on -d:"libp2p_pki_schemes=secp256k1" --header:verifproxy.h --noMain:on --nimcache:nimcache/libverifproxy -o:build/$@ $(NIM_PARAMS) nimbus_verified_proxy/nimbus_verified_proxy.nim
+	echo -e $(BUILD_END_MSG) "build/$@"
+
+# Shared library for verified proxy
+OS = $(shell $(CC) -dumpmachine)
+ifneq (, $(findstring darwin, $(OS)))
+  SHAREDLIBEXT = dylib
+else
+ifneq (, $(findstring mingw, $(OS))$(findstring cygwin, $(OS))$(findstring msys, $(OS)))
+  SHAREDLIBEXT = lib
```
arnetheduck, line 287: `.lib` / `.a` are static libraries, not shared - ie shared libraries use `.dll` / `.so` - which is intended here?

siphiuel: Yes, thanks for spotting, will fix. This is a partial carryover from earlier implementation that relied only on dynamic libs.

PR #1499, opened 2023-03-15
Bump Nim to 1.6.12

zah:


zah:
@jangko, the macOS CI failure here looks like a potentially temporary hiccup (the CI runner failed to download some of the per-requisite software), but the Windows issue in the test suite may be a genuine problem. Would you be able to take a look at it?

kdeme:
I re-ran the all failed jobs because of the macOS hiccup, and it appears that also the Windows failures passed this time. 

Perhaps it was a failure not related to these changes. Did you ever see this before @jangko ? 
You can still access the previous failed run here: https://github.com/status-im/nimbus-eth1/actions/runs/4427335721/jobs/7764780057

PR #1509, opened 2023-03-17
Bumped the tests. Withdrawals pass but EOF still fails.

AdamSpitz:


AdamSpitz:
@jangko, can you take a look at this?

I've bumped tests/fixtures/eth_tests to v12, and now the withdrawals-related tests pass, but there are also a bunch of EOF-related tests that fail. I'm not sure how to disable those tests, or whether it'd be easier to just merge in your EOF-implementation code to make those tests pass.

jangko:
I think the `tests/fixtures/eth_tests` should be bumped one commit ahead to [9a90f73](https://github.com/ethereum/tests/commits/9a90f730bd87c11ee377a336dcb7eef3aa9b0195).
In that commit, EIPs not included in Shanghai is moved into separate folder.

AdamSpitz:
Wow, you're right. Off by one commit. Neat!

PR #1513, opened 2023-03-21
Add clientInfo to ENR for easier testnet debugging

kdeme:
As mentioned in the code, this is a temporary measure, as we probably do not want this information that easily searchable in the DHT for security reasons.


kdeme:
fixes https://github.com/status-im/nimbus-eth1/issues/1755

PR #1527, opened 2023-04-03
Use closeWait when closing early on a readContentOffer

kdeme:


kdeme:
Unnecessary change. Socket will anyhow wait for FIN-ACK before it gets destroyed.

PR #1554, opened 2023-04-21
Allow dataRadius as optional parameter in Portal ping JSON-RPC

kdeme:
However, ignore it currently, as it might be removed anyhow.

Originally wanted to merge this, and then remove it again or not depending on https://github.com/ethereum/portal-network-specs/pull/202.

However, options appear to be broken in nim-json-rpc, making them not so optional. So this will remain draft.


kdeme:
Closing as no longer applicable: specs got changed

PR #1572, opened 2023-05-03
nimbus-verified-proxy integration with status-go

vitvly:
This is depended on by https://github.com/siphiuel/lc-proxy-wrapper.

Custom `nim-confutils` dependency is due to adding a fix suggested in https://github.com/status-im/nim-confutils/issues/31
.
Tracking issue: https://github.com/status-im/status-go/issues/3282

Supersedes https://github.com/status-im/nimbus-eth1/pull/1497


nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -40,14 +40,27 @@ func getConfiguredChainId(networkMetadata: Eth2NetworkMetadata): Quantity =
   else:
     return networkMetadata.cfg.DEPOSIT_CHAIN_ID.Quantity
 
-proc run(config: VerifiedProxyConf) {.raises: [CatchableError].} =
+type OnHeaderCallback* = proc (s: cstring) {.cdecl.}
```
arnetheduck, line 43:
```suggestion
type OnHeaderCallback* = proc (s: cstring) {.cdecl, raises: [].}
```

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -163,6 +176,14 @@ proc run(config: VerifiedProxyConf) {.raises: [CatchableError].} =
...
         info "New LC finalized header",
           finalized_header = shortLog(forkyHeader)
+        if finalizedHeaderCallback != nil:
+            notice "### Invoking finalizedHeaderCallback"
+            {.gcsafe.}:
+              try:
+                finalizedHeaderCallback(Json.encode(finalizedHeader))
+              except Exception as e:
+                notice "finalizedHeaderCallback exception"
+
```
arnetheduck, line 186:
```suggestion
            notice "### Invoking finalizedHeaderCallback"
            finalizedHeaderCallback(Json.encode(finalizedHeader))
```

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -171,6 +192,14 @@ proc run(config: VerifiedProxyConf) {.raises: [CatchableError].} =
         info "New LC optimistic header",
           optimistic_header = shortLog(forkyHeader)
         optimisticProcessor.setOptimisticHeader(forkyHeader.beacon)
+      if optimisticHeaderCallback != nil:
```
arnetheduck, line 195: ditto

nimbus_verified_proxy/nimbus_verified_proxy.nim
siphiuel: @arnetheduck thank you for review!
The `Json.encode()` invocation throws `SerializationError`. If not wrapped with `try-except` the following compiler error is shown:
```
Error: writeValue writer`gensym1293, optimisticHeader can raise an unlisted exception: SerializationError
```

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -40,14 +40,27 @@ func getConfiguredChainId(networkMetadata: Eth2NetworkMetadata): Quantity =
...
+proc setOptimisticHeaderCallback*(cb: OnHeaderCallback) {.exportc, dynlib.} =
+  optimisticHeaderCallback = cb
+  echo "optimistic header callback set"
+
+proc setFinalizedHeaderCallback*(cb: OnHeaderCallback) {.exportc, dynlib.} =
+  finalizedHeaderCallback = cb
+  echo "finalized header callback set"
+
+
+proc run(config: VerifiedProxyConf) {.raises: [CatchableError, Exception].} =
```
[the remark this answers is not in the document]
siphiuel: With this change, i get the following compiler error on that line:
```
Error: Exception can raise an unlisted exception: Exception

File: nimbus_verified_proxy/nimbus_verified_proxy.nim (hunk 2 above)
Comment (arnetheduck) [reply nimbus_verified_proxy/nimbus_verified_proxy.nim:186]: right, then the right approach is to specifically catch that (`try: ... except SerializationError...`).

File: nimbus_verified_proxy/nimbus_verified_proxy.nim (hunk 4 above)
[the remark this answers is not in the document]
Comment (arnetheduck) [reply nimbus_verified_proxy/nimbus_verified_proxy.nim:56]: For this exception, you need to track down where the `Exception` is coming from - usually, it's the `callback` that I noted above should be annotated, else the compiler will assume the most broad exception raising.

You can figure it out by putting try/except around specific lines until it goes away: the proper fix is what I suggest above for the callbacks, to annotate them with `raises: [], gcsafe`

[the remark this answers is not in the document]
Comment (siphiuel) [reply nimbus_verified_proxy/nimbus_verified_proxy.nim]: Thanks, figured it out - it was `getCLIParams()` that threw the exception.

File: nimbus_verified_proxy/nimbus_verified_proxy.nim (hunk 2 above)
Comment (siphiuel) [reply nimbus_verified_proxy/nimbus_verified_proxy.nim]: Done

File: nimbus_verified_proxy/nimbus_verified_proxy.nim (hunk 3 above)
Comment (siphiuel) [reply nimbus_verified_proxy/nimbus_verified_proxy.nim:199]: Done.

**Comment by siphiuel** [comment]
@arnetheduck can you please have a look at the fixes? Hopefully this looks better now.

**Comment by siphiuel** [comment]
@arnetheduck @zah can you please review the fixes made by @Ivansete-status ?

File: nimbus_verified_proxy/nimbus_verified_proxy.nim [hunk 5]
```diff
@@ -249,10 +283,42 @@ proc run(config: VerifiedProxyConf) {.raises: [CatchableError].} =
   while true:
     poll()
 
-when isMainModule:
+proc quit*() {.exportc, dynlib.} = 
+  echo "Quitting"
+
+proc NimMain() {.importc.}
+proc startProxyViaJson*(configJson: cstring) {.exportc, dynlib.} =
```
arnetheduck, line 290: So, we need to think about this API in particular - when called from `go`, this would block the go thread until finished.

As an alternative to this model, I would propose the following API:

```nim
type Context = object
  thread: Thread[ptr Context]
  config: cstring
  stop: bool
  onFinalizedHeader: proc...

proc runContext(ctx: ptr Context) {.thread.} =
  let node = newLightClientNode(ctx.config)

  while not ctx[].stop and node.running:
    let timeout = sleepAsync(100.millis)
    waitFor timeout

  # do cleanup
  node.stop()

proc startLightClientProxy(config: cstring, onFinalizedHeader: ...): ptr Context =
  let ctx = createShared(Context, 1)
  ctx.config = cast[cstring](allocShared0(len(config) + 1))
  ctx.onFinalizedHeader = ...
  copyMem(ctx.config, config, len(config))

  createThread(ctx.thread, runContext, ctx)

proc stopLightClientProxy(ctx: ptr Context) =
  ctx[].stop = false
  ctx.thread.joinThreads()
  deallocShared(ctx.config)
  ...
  freeShared(ctx)
```

In the above setup, the light client runs in its own thread and all "shared" memory is allocated inside the `Context` object which is freed when `go` wants to stop the proxy.

Instead of calling `run(config)` which calls `runForever`, we use a custom polling loop that wakes up every 100ms to check if we should shut down - later, we can replace this with a more sophisticated method but this is sufficient for now.

So, to rehash the key ideas:
* `go` calls a "start" function that takes as parameter the configuration and any callbacks that are needed
* the "start" function allocates a context which is given back to `go` - `go` doesn't interact with this context _directly_ but it's used in the rest of the API
  * the "start" function does not block
* the node operates as usual in its own thread - the only difference is the polling for the `stop` boolean that happens every now and then
* notice in particular how the `node` instance is allocated _inside_ the thread
* when it's time to shut down, go calls the "stop" function giving the context - it blocks until the node has stopped

Later, the `Context` object will also hold a queue for interacting with the node from `go` - this queue will be used to make sure that there are no unsanctioned memory interactions between `go` and `nim` - it looks like so:

```nim
proc sendRequest(ctx: ptr Context, data: ptr byte, len: cint) =
  let copy = allocShared(len)
  copyMem(copy, data, len)
  ctx.queue.add((copy, len))
```

If a return value is needed, this is easiest to do via a callback:


```nim
proc sendRequest(ctx: ptr Context, data: ptr byte, len: cint, user: pointer, callback: proc(user: pointer, response: ptr byte, len: cint) {.cdecl.}) =
  let copy = allocShared(len)
  copyMem(copy, data, len)
  ctx.queue.add((copy, len, callback))
```

when the result of the call is ready, the callback is simply called making it easy to integrate with the async behavior of go

arnetheduck: Confirming that cgo can be called from non-go threads: https://github.com/golang/go/commit/6c976393aea607e67f4d31e3a2ae7b3c0dc15ade

arnetheduck: https://github.com/status-im/nim-style-guide/pull/15 contains more examples - will be adding a go example soon as well

arnetheduck:
let's move  the code needed to compile the proxy as a library to a separate file like so: 

* https://github.com/waku-org/nwaku/blob/master/library
* https://github.com/status-im/nimbus-eth2/tree/unstable/beacon_chain/libnimbus_lc

Similar to the above two libraries, the C header file should be written manually to expose the functions - it's probably easier to create a static library instead of a dll/so and link everything statically.

We can also create the PR for confutils and get that merged ASAP so that it is not a blocker.

arnetheduck:
> Tried rebasing, but this causes issues

did you upgrade your copy of the dependencies (nim-stew etc) after rebasing?

vitvly:
i think i did - i made sure that all submodule deps are identical to the ones in `master`, except for `nim-confutils`. Then did `make mrproper`, `make update`, `make libverifproxy`. Must be something that i've messed up, i guess.

vitvly:
The draft PR for rebased version here: https://github.com/status-im/nimbus-eth1/pull/1971

vitvly:
Closing in favor of https://github.com/status-im/nimbus-eth1/pull/1971

PR #1585, opened 2023-05-18
Add shanghai timestamp to MainNet, Goerli, and Sepolia

jangko:
Looks like just adding shanghai timestamp is not enough, txpool should also be fixed.
Txpool should generate withdrawal root for post Shanghai blockheader too.
Where withdrawal info should come from? directly from engine API? or from Casper consensus mechanism?


zah:
The `withdrawals` are supplied by the Consensus client in the [PayloadAttributesV2](https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#payloadattributesv2) structure which is supplied in the [forkchoiceUpdated](https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#engine_forkchoiceupdatedv2) call.

PR #1647, opened 2023-07-18
Remove slots from content keys according to current specs

kdeme:
Might or might not go on with this change, need to create some local testnet first.

Branch is there to be able to easily create test vectors according to spec


kdeme:
closing in favour of https://github.com/status-im/nimbus-eth1/pull/1810 and https://github.com/status-im/nimbus-eth1/pull/1811

Issue #1640, opened 2023-07-12
cli option to run fluffy without POKE functionality


kdeme:
In order to properly test gossip mechanisms (e.g. in Portal Hive), we need the option to turn off the POKE functionality as it influences how data moves around the network.

kdeme:
Resolved in https://github.com/status-im/nimbus-eth1/pull/1711

PR #1711, opened 2023-08-24
disable POKE for fluffy tests by default (#1640)

mynameisdaniil:


kdeme (commented):
While this works, I was thinking of rather making it into a runtime option.

This will especially be useful for portal-hive testing (the main reason we need it, although in our own testing we could also have tests that require POKE to be turned on or off).
The reason for this is that portal-hive uses our Fluffy docker image, available on docker hub, and else we need to build a second version for this, which I wouldn't recommend.

So instead I think it could be a runtime option, probably a hidden cli option, which could be added to the portal config, and a new `disablePoke` field in the `PortalProtocol`. wdyt?

Makefile
```diff
@@ -241,7 +241,7 @@ all_fluffy_portal_spec_tests: | build deps
 
 all_fluffy_tests: | build deps
 	echo -e $(BUILD_MSG) "build/$@" && \
-	$(ENV_SCRIPT) nim c -r $(NIM_PARAMS) -d:chronicles_log_level=ERROR -d:nimbus_db_backend=sqlite -d:mergeBlockNumber:38130 -o:build/$@ "fluffy/tests/$@.nim"
+	$(ENV_SCRIPT) nim c -r $(NIM_PARAMS) -d:chronicles_log_level=ERROR -d:nimbus_db_backend=sqlite -d:mergeBlockNumber:38130 -d:disablePoke=true -o:build/$@ "fluffy/tests/$@.nim"
```
kdeme, line 244: I'm not sure if we should actually disable this here. But I guess it does no harm.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -933,6 +935,12 @@ proc triggerPoke*(
     nodes: seq[Node],
     contentKey: ByteList,
     content: seq[byte]) =
+  ## See: https://github.com/status-im/nimbus-eth1/issues/1640                                                                          
+  ## In order to properly test gossip mechanisms (e.g. in Portal Hive),                                                                 
+  ## we need the option to turn off the POKE functionality as it influenceshow                                                          
+  ## how data moves around the network.
```
kdeme, line 941: I don't think we need to link the issue here. Issues are typically linked in code when a workaround is needed because of that issue, and then removed again when the issue is resolved and the work around removed.


```suggestion                                                                        
  ## In order to properly test gossip mechanisms (e.g. in Portal Hive),                                                                 
  ## we need the option to turn off the POKE functionality as it influences                                                          
  ## how data moves around the network.                                                                                                 
```

mynameisdaniil:
@kdeme pls take a look again. More like this?

fluffy/conf.nim
```diff
@@ -46,6 +46,8 @@ const
   defaultTableIpLimitDesc* = $DefaultTableIpLimit
   defaultBucketIpLimitDesc* = $DefaultBucketIpLimit
 
+  defaultDisablePoke* = false
```
kdeme, line 49: You can move this constant to the portal_protocol_config file as it is portal protocol specific.

fluffy/conf.nim
```diff
@@ -226,6 +228,12 @@ type
       defaultValue: none(TrustedDigest)
       name: "trusted-block-root" .}: Option[TrustedDigest]
 
+    disablePoke* {.
+      desc: "Disable POKE functionality for gossip mechanisms test"
```
kdeme, line 232: I'd suggest to add `hidden` to this parameter here, as it is a flag that should not be used in normal operations.

There are some other options in the list which use `hidden` also.

kdeme, line 232:
```suggestion
      desc: "Disable POKE functionality for gossip mechanisms testing"
```

mynameisdaniil:
@kdeme I've addressed the issues you've mentioned. Please take one more look.

kdeme (approved):

Issue #1227, opened 2022-09-16
NH gossip: Avoid offering to the node that we received the data from


kdeme:
Fluffy should keep track which node offered its received content and not gossip the received content again to this node.

kdeme:
Resolved in https://github.com/status-im/nimbus-eth1/pull/1719

PR #1719, opened 2023-08-29
don't gossip data back to the node we rcvd it from [#1227]

mynameisdaniil:
Not sure how to go about testing it yet.

### DIFF
--- a/fluffy/eth_data/history_data_seeding.nim
+++ b/fluffy/eth_data/history_data_seeding.nim
@@ -60,7 +60,7 @@ proc propagateEpochAccumulator*(
       encodedAccumulator
     )
     discard await p.neighborhoodGossip(
-      ContentKeysList(@[encKey]), @[encodedAccumulator])
+      Opt.none(NodeId), ContentKeysList(@[encKey]), @[encodedAccumulator])
 
     return ok()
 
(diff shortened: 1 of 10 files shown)


fluffy/eth_data/history_data_seeding.nim
```diff
@@ -60,7 +60,7 @@ proc propagateEpochAccumulator*(
       encodedAccumulator
     )
     discard await p.neighborhoodGossip(
-      ContentKeysList(@[encKey]), @[encodedAccumulator])
+      ContentKeysList(nodeId, @[encKey]), @[encodedAccumulator])
```
mynameisdaniil, line 63: I'm not sure what `nodeId` should it be in this case. Should it be `nodeId` of the current node? What would be a better way to get it in this part of the code?

kdeme:
Looks good. Testing this is not so convenient indeed. It probably would require using some individual functions of the portal protocol and not the full implementation. Another way could be perhaps by accessing the metrics, but again, not so convenient / conventional

fluffy/network/wire/portal_protocol.nim
```diff
@@ -1128,10 +1128,11 @@ proc getNClosestNodesWithRadius*(
 
 proc neighborhoodGossip*(
     p: PortalProtocol,
+    srcNodeId: Opt[NodeId],
     contentKeys: ContentKeysList,
     content: seq[seq[byte]]): Future[int] {.async.} =
   ## Returns number of peers to which content was gossiped
-
+
```
kdeme, line 1135: Remove whitespace

fluffy/eth_data/history_data_seeding.nim
```diff
@@ -10,6 +10,7 @@
 import
   std/[strformat, os],
   stew/results, chronos, chronicles,
+  eth/p2p/discoveryv5/protocol,
```
kdeme, line 13: What is this import needed for? Is it for `NodeId`?

If so, I think it is better to add `node` module to the export list of `portal_protocol.nim` as it gets added as parameter in a public API call.

PR #1729, opened 2023-09-05
Bump submodules and related fixes

kdeme:
Bumps for nim-ssz-serialization + fixes and then anything of module that needs a bump because of this including nimbus-eth2 and nimbus-build-system (Nim 1.6.14).

The latter does make nimbus build fail due to https://github.com/nim-lang/Nim/issues/22190

### DIFF
--- a/fluffy/common/common_types.nim
+++ b/fluffy/common/common_types.nim
@@ -34,11 +34,11 @@ func decodeRlp*(input: openArray[byte], T: type): Result[T, string] =
 func decodeSsz*(input: openArray[byte], T: type): Result[T, string] =
   try:
     ok(SSZ.decode(input, T))
-  except SszError as e:
+  except SerializationError as e:
     err(e.msg)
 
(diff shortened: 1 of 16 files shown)


nimbus/evm/interpreter/op_handlers.nim
```diff
@@ -51,6 +51,7 @@ proc mkOpTable(selected: EVMFork): array[Op,Vm2OpExec] {.compileTime.} =
   # Collect selected <fork> entries
   for (subList,subName) in allHandlersList:
     for w in subList:
+      # This fails, sigh: https://github.com/nim-lang/Nim/issues/22190
```
jangko, line 54: as a temporary workaround, maybe we can use `let` to replace `const`

narimiran:
> The latter does make nimbus build fail due to https://github.com/nim-lang/Nim/issues/22190

The offending commit in `version-1-6` branch is now reverted. Please test with `HEAD` of that branch to see if this solved the problem.

PR #1756, opened 2023-09-18
Correct the shortcut for ommershash verification post Shanghai

kdeme:
There was a faulty length check being done in the assumption that the uncles field was already the list. However it is still the RLP encoded data at that point, hence we check directly for the RLP encoded value for empty list.


kdeme:
Merging this without CI as the CI issues are unrelated and this changes has been tested well locally + in portal-hive.
Need this change on the fluffy fleet.

PR #1759, opened 2023-09-18
Local docker dev improvements

mynameisdaniil:
brings build time down to ~250s on my machine.
There's still room for improvement + README needs an update to address this changes

### DIFF
--- a/fluffy/README.md
+++ b/fluffy/README.md
@@ -157,6 +157,21 @@ The code follows the
 [Status Nim Style Guide](https://status-im.github.io/nim-style-guide/).
 
 
+## Build local dev container for portal-hive
+
+To develop code against portal-hive tests you will need:
+
+1) Clone and build portal-hive ([#1](https://github.com/ethereum/portal-hive))
+
+2) Modify `Dockerfile` for fluffy in `portal-hive/clients/fluffy/Dockerfile` ([#2](https://github.com/ethereum/portal-hive/blob/main/docs/overview.md#running-a-client-built-from-source))
+
+3) Build local dev container using following command: ```docker build --tag fluffy-dev --file ./fluffy/tools/docker/Dockerfile.portalhive .``` You may need to change fluffy-dev to the tag you using in portal-hive client dockerfile.
+
+4) Run the tests
+
+Also keep in mind that `./vendors` is dockerignored and cached. If you have to make local changes to one of the dependencies in that directory you'll have to remove `vendors/` from `./fluffy/tools/docker/Dockerfile.portalhive.dockerignore`.
+
+
 ## Metrics and their visualisation
 
 To enable metrics run Fluffy with the `--metrics` flag:
--- a/fluffy/tools/docker/Dockerfile.portalhive
+++ b/fluffy/tools/docker/Dockerfile.portalhive
@@ -1,0 +1,26 @@
+FROM debian:stable-slim as build
+
+ENV DEBIAN_FRONTEND=noninteractive TZ="Etc/UTC"
+ENV NPROC=2
+
+RUN apt update \
+  && apt install make bash build-essential git -y \
+  && apt-get clean \
+  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+COPY ../../ /fluffy
+
+WORKDIR /fluffy
+
+RUN --mount=type=cache,target=/fluffy/vendor \
+    make deps -j${NPROC}
+
+RUN --mount=type=cache,target=/fluffy/vendor <<EOF
+  set -e
+  make fluffy -j${NPROC}
+  cp /fluffy/build/fluffy /usr/local/bin/fluffy
+EOF
+
+FROM debian:stable-slim as app
+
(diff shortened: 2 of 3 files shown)


docker/debug/Dockerfile.fluffy
```diff
@@ -0,0 +1,18 @@
...
+    --mount=type=cache,target=./build \
+    make deps -j2
+
+RUN --mount=type=cache,target=./vendor \
+    --mount=type=cache,target=./build <<EOF
+  set -e
+  make fluffy -j2
+EOF
+
+RUN cp ./build/fluffy /usr/local/bin/fluffy
```
kdeme, line 18: We could still have a new image here where we copy the binary to, like in the release Dockerfile.  It would make the end image much smaller, but the total size used wouldn't change I think as that would remain taken up in some docker build cache. Still, if wanted one could clean-out just that cache I think? wdyt?

docker/debug/Dockerfile.fluffy
```diff
@@ -0,0 +1,18 @@
+from debian:stable-slim as build
+
+RUN apt update && apt install make bash build-essential git libnatpmp-dev librocksdb-dev -y
+
+# RUN git clone --depth=1 https://github.com/status-im/nimbus-eth1.github
+COPY ../../ ./
+
+RUN --mount=type=cache,target=./vendor \
+    --mount=type=cache,target=./build \
+    make deps -j2
```
kdeme, line 10: Can make the cores used an `ENV`

docker/debug/Dockerfile.fluffy
```diff
@@ -0,0 +1,18 @@
+from debian:stable-slim as build
+
+RUN apt update && apt install make bash build-essential git libnatpmp-dev librocksdb-dev -y
+
+# RUN git clone --depth=1 https://github.com/status-im/nimbus-eth1.github
+COPY ../../ ./
```
kdeme, line 6: Probably the best way to save on some space (and copy time) here is being more selective on what to copy?

kdeme (commented):
The copy-from-local-repo way is definitaly more handy for development, such as in portal-hive. Not only speedwise but also in avoiding commiting to some branch.

We do try to keep everything fluffy inside the fluffy folder. I think this should go into `./fluffy/tools/docker/debug/`.
We can also move the docker folder to `./fluffy/docker/` instead, and have a release folder for our release Dockerfile.

README.md
```diff
@@ -394,6 +394,13 @@ graph legend.
 To see a single metric, click on its name in the legend. Click it again to go back
 to the combined view. To edit a panel, click on its title and select "Edit".
 
+### Build local development cotainer for portal-hive
```
kdeme, line 397:
```suggestion
### Build local development container for portal-hive
```

docker/debug/Dockerfile.fluffy
kdeme: Actually, it also helps for total size considering stale images (older builds) will be much smaller.

.dockerignore
```diff
@@ -0,0 +1 @@
+vendor/
```
kdeme, line 1: could we add `**/.git` here?

docker/debug/Dockerfile.fluffy
```diff
@@ -0,0 +1,18 @@
...
+
+RUN --mount=type=cache,target=./vendor \
```
kdeme, line 8: Could you explain how this works together with the `.dockerignore` for `vendor`.
Does it mean that vendor folder will never be copied from local repo, but instead populated in docker image here by this command, and cached separately by docker?

Have you tested that in several scenarios of code change that might be done during development, e.g.:
- Adjust a fluffy nim file: this will work as it gets copied and has nothing to do with vendor, right?
- Adjust a git submodule revision: This will work as it will recognize the new hash and make deps will update the revision, correct?
- Make a local change (no commit) to a submodule. I assume this will not work, as that change is vendor local and does not get copied? While such thing would be handy, I guess it is fine if it works by having a local branch of the submodule with the change commited. As long as it is explained/commented.
- ...

docker/debug/Dockerfile.fluffy
```diff
@@ -0,0 +1,18 @@
+from debian:stable-slim as build
+
+RUN apt update && apt install make bash build-essential git libnatpmp-dev librocksdb-dev -y
```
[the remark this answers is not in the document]
mynameisdaniil: yeah, I've started by trying to build it on alpine to save space

docker/debug/Dockerfile.fluffy
mynameisdaniil: true that

docker/debug/Dockerfile.fluffy
mynameisdaniil: yep, exactly

docker/debug/Dockerfile.fluffy
mynameisdaniil: You got it absolutely right. The answers are yes, yes and yes. For the latter case one can just remove `vendors/` from `.dockerignore`

kdeme (approved):

PR #1763, opened 2023-09-21
Load and use the stored ENR to allow incr of seqNum on updates

kdeme:
Currently the stored ENR is not reloaded in restarts. This means that the ENR sequence number always remains at 1. For better propagation of updated ENRs the sequence number should be incremented on ENR updates.


kdeme:
I will get back to this later, needs some nim-eth changes.

PR #1765, opened 2023-09-21
Update exception annotations for upcoming core db extension

mjfh:


fluffy/network/history/history_network.nim
```diff
@@ -183,7 +187,7 @@ func encode*(receipts: seq[Receipt]): seq[byte] =
 # TODO: Failures on validation and perhaps deserialisation should be punished
 # for if/when peer scoring/banning is added.
 
-proc calcRootHash(items: Transactions | PortalReceipts | Withdrawals): Hash256 =
+proc calcRootHash(items: Transactions | PortalReceipts | Withdrawals): Hash256 {.catchRaise.} =
```
kdeme, line 190: This shouldn't raise, where does that come from now? And I don't want to introduce this here and have it all added up in all the calls below that shouldn't raise.

I already disliked the fact that `CatchableError` was introduced lower (but before it was `RlpError`, so not a huge diff), which we have to `raiseAssert` because of that.

fluffy/network/history/history_network.nim
```diff
@@ -25,6 +25,10 @@ logScope:
 
 export accumulator
 
+# Annotation helpers
+{.pragma:    noRaise, gcsafe, raises: [].}
+{.pragma: catchRaise, gcsafe, raises: [CatchableError].}
```
kdeme, line 30: I consider these 2 type of `raises` helpers rather anti-helpers.

The first one is already implied with the `{.push raises: [].}` that we now tend to do at every file.
The second one is more like an "easy way out" with dealing with your specific exceptions.

kdeme (changes requested):

fluffy/network/history/history_network.nim
mjfh: The exceptions will come from the `CoreDB` database. I decided that it is not a `Defect` if the is an adjustment problem with the multi MPT descriptors. The current database interface does not support `Result[Blob,SomeError]` type. I will migrate to that but do not want to change too much at the moment.

I will prettify your fluffy code.

kdeme: The prettify commit is not enough. Those calls simply should not raise, hence the use of `Result`.
And the fact that they raise `CatchableError` makes it even worse. As it would require an except clause for `CatchableError` where they are called, potentially missing new raises when new calls are introduced in that code.

Is it `newCoreDbRef(LegacyDbMemory).mptPrune` that raises this? I'd say a more specific error, that in our scenario cannot occur here would be more acceptable, and then it can be catched and `raiseAssert`ed.

Or perhaps we need to go back to using here the `HexaryTrie` with MemoryDb like before. It wouldn't be bad to have a "safe" hexary trie version. All that is being done here is build one from transactions/receipts/withdrawals to be able to know the roots.

mjfh: Just note that `HexaryTrie` is not available unless you want to get locked in there.

Using exceptions are just a cheap way to proceed (with the same excuse for the `RLP` module not updated.)

But I can do better here and provide both, `Result[]` return codes and `Defect` asserts for a legacy API which would not imply any annotation changes. Errors will be opaque objects can then be decoded by a backend specific method. I will give it a try.

BTW: What is the feature of the `SafeHexaryTrie`? I have not found anything "safe". Only that key arguments are pre-hashed -- or did I miss something?

So this PR can be mothballed and eventually disposed.

kdeme: > Just note that `HexaryTrie` is not available unless you want to get locked in there.

Not sure what you mean with this?



> BTW: What is the feature of the `SafeHexaryTrie`? I have not found anything "safe". Only that key arguments are pre-hashed -- or did I miss something?

Well, for the use case here, which is just calculating the tree root, I would think it could be safe, as in, it does not raise exceptions.

PR #1776, opened 2023-09-25
Core db update providing additional results code interface

mjfh:
This PR provides two APIs for functions:
* The legacy API where **no** function returns a `CatchableError`
*  A new API where functions return `Result[]` codes

This is different for iterators which might throw exceptions derived from  `CoreDbApiError` (which in turn is derived from `CatchableError`.)

To be checked whether feasible:
 > The legacy API function will defect on any exception which means to defect on an `RlpError` for the `HexaryTrie`.


jangko:
LGTM

nimbus/db/aristo/aristo_fetch.nim
```diff
@@ -32,6 +32,13 @@ proc fetchPayloadImpl(
     return err((vid, rc.error[1]))
   ok rc.value.legs[^1].wp.vtx.lData
 
+proc fetchPayloadImpl(
+    db: AristoDbRef;
+    root: VertexID;
+    path: openArray[byte];
+      ): Result[PayloadRef,(VertexID,AristoError)] =
```
kdeme, line 39: nitpick, but kinda all over the place: Space after commas.

nimbus/db/aristo/aristo_init/rocks_db/rdb_put.nim
```diff
@@ -49,13 +49,13 @@ proc getFileSize(fileName: string): int64 =
     defer: f.close
     try:
       result = f.getFileSize
-    except:
+    except CatchableError:
       discard
```
kdeme, line 53: Is it okay to just return default int64 here?

nimbus/db/aristo/aristo_transcode.nim
```diff
@@ -17,8 +17,11 @@ import
   stew/endians2,
   "."/[aristo_constants, aristo_desc, aristo_get]
 
+# Annotation helper
+{.pragma:  noRaise, gcsafe, raises: [].}
```
kdeme, line 21:
```suggestion
{.pragma: noRaise, gcsafe, raises: [].}
```

nimbus/db/core_db/legacy_db.nim
```diff
@@ -0,0 +1,340 @@
...
+
+type
+  LegacyApiRlpError* = object of CoreDbApiError
+    ## For re-routing exceptions in iterator closure
+
+  LegacyDbRef* = ref object of CoreDbRef
+    tdb: TrieDatabaseRef # copy of descriptor reference captured with closures
+
+  HexaryTrieRef = ref object
+    trie: HexaryTrie     # nedded for decriptor capturing with closures
```
kdeme, line 28:
```suggestion
    trie: HexaryTrie     # needed for descriptor capturing with closures
```

nimbus/db/core_db/legacy_db.nim
```diff
@@ -0,0 +1,340 @@
...
+
+proc init*(
+    db: LegacyDbRef;
+    dbType: CoreDbType;
+    tdb: TrieDatabaseRef;
+      ): CoreDbRef
+      {.gcsafe.}
+
+# ------------------------------------------------------------------------------
+# Private helpers, exceprion management
```
kdeme, line 58:
```suggestion
# Private helpers, exception management
```

nimbus/db/core_db/base.nim
```diff
@@ -390,233 +178,343 @@ proc newCoreDbCaptRef*(db: CoreDbRef; methods: CoreDbCaptFns): CoreDbCaptRef =
...
+    cld.distinctBase.parent()
+
+  # ----------------
+
+  func kvt*(db: CoreDbRef): CoreDbKvtRef =
+    ## Legacy pseudo constructor, see `toKvt()` for production constructor
+    db.toKvt.CoreDbKvtRef
+
+  proc get*(kvt: CoreDbKvtRef; key: openArray[byte]): Blob =
+    kvt.distinctBase.get(key).expect "kvt/get()"
```
kdeme, line 382: Does this mean it would `Defect` if it fails getting the value?

kdeme (commented):
Glanced over it (mostly on context for Fluffy) as it is a bit much to give a proper review.

Mostly wondering about the `expects`.

nimbus/db/aristo/aristo_fetch.nim
mjfh: it is nitpick :)

nimbus/db/core_db/base.nim
mjfh: it won't fail, just defect

nimbus/db/aristo/aristo_init/rocks_db/rdb_put.nim
mjfh: yes, as returned from `getFileSize()`

nimbus/db/core_db/base.nim
kdeme: That's the strangest API ever. 
You request a value from a database and if it doesn't have it the whole program panics?

mjfh: It defects on error, otherwise returns value, see `select_backend.nim` for an example (or look up `results.nim`.). 

The equivalent for `CatchableError` would be `tryGet()` (or `tryValue()` which is the same.) BTW, this had a problem lately in a complex scenario which caused me a lot of headaches, commented on in `core_db/legacy_rocksdb.nim`. It was too messy to distil an issue -- took to much time already to get there.

kdeme: So after looking up myself, it seems the answer to my question is: It returns empty Blob when value is not in the db. And `Defect` on any DB related issue.

PR #1779, opened 2023-09-27
short-circuit pings if they happen too often

mynameisdaniil:


fluffy/conf.nim
```diff
@@ -233,6 +233,19 @@ type
       defaultValueDesc: $defaultDisablePoke
       name: "disable-poke" .}: bool
 
+    timeoutExpiration* {.
```
kdeme, line 236: These settings are very internal to the protocol that I don't think they need to become cli options.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -1255,6 +1290,7 @@ proc revalidateNode*(p: PortalProtocol, n: Node) {.async.} =
 proc revalidateLoop(p: PortalProtocol) {.async.} =
   ## Loop which revalidates the nodes in the routing table by sending the ping
   ## message.
+  # NOTE This is actually the place to control timings
```
kdeme, line 1293: That's right. It shouldn't be done inside the `ping` call itself. As an external user should still be able to call ping and be certain that the message was send. That way you also don't have to create an error result there.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -566,6 +592,15 @@ proc recordsFromBytes*(rawRecords: List[ByteList, 32]): PortalResult[seq[Record]
 
 proc ping*(p: PortalProtocol, dst: Node):
     Future[PortalResult[PongMessage]] {.async.} =
+
+  let nowMilliSeconds = uint64(epochTime() * 1000) # milliseconds
```
kdeme, line 596: We use chronos `now` functionality and work with that: https://github.com/status-im/nim-chronos/blob/master/chronos/timer.nim#L436

See usage in the rest of the code, e.g. https://github.com/status-im/nimbus-eth1/blob/master/fluffy/network/wire/portal_protocol.nim#L1283

fluffy/network/wire/portal_protocol.nim
```diff
@@ -194,6 +196,23 @@ type
     # content is in their range
     nodesInterestedInContent*: seq[Node]
 
+proc gcOldTimestamps(p: PortalProtocol) {.async.} =
```
kdeme, line 199: I understand what you are trying to do here, and it can be a solution. But I think a simpler way is to just delete an entry from the `table` when we replace a node.

The revalidation (pinging at interval) of nodes is done only on nodes that are in the routing_table. So this table is inherently linked to the nodes in the routing table. That means it is fine to keep an entry in there as long as the Node for that entry is also still in the routing table. So I think the more natural way would be to remove an entry when we also remove it from the routing table.

Replacement happens here: https://github.com/status-im/nimbus-eth1/blob/master/fluffy/network/wire/portal_protocol.nim#L523
You can make a PortalProtocol version of that call that wraps the routing table one.

Eventually, this could/should probably become part of the routing_table code itself (possibly still as a HashMap or as an extra field in the `Node` object or somewhere else). But that could be a second step as it is also need to figure out how that would interact with discv5.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -194,6 +196,23 @@ type
...
+  # oldest ones
+  while true:
+    let nowMilliSeconds = uint64(epochTime() * 1000)
+    for nodeId, timeout in p.pingTimings:
+      # NOTE not sure how to go about this cause uint64 - uint64 still uint64
+      # so in case where clocks moved backwards (think NTP correction) there might be a case
+      # where this works incorrectly
+      if nowMilliSeconds - timeout > p.timeoutExpiration:
+        p.pingTimings.del(nodeId)
+    # TODO this value should probably be adaptive, relative to the size
```
kdeme, line 212: I think you can skip the adaptive relative to size of routing table for now.

fluffy/network/wire/portal_protocol_config.nim
```diff
@@ -33,6 +33,8 @@ const
   defaultRadiusConfig* = RadiusConfig(kind: Dynamic)
   defaultRadiusConfigDesc* = $defaultRadiusConfig.kind
   defaultDisablePoke* = false
+  defaultTimeoutExpiration* = 60 * 60 * 24 * 1000 # One day
+  defaultTimeoutInterval = 1 * 1000 # 1 sec
```
kdeme, line 37: I think we are rather aiming here for something like `30.seconds` to `1.minutes`, perhaps even longer.

tests/fixtures/eth_tests
kdeme, line 1: Accidental bump of eth_tests?

kdeme (commented):
I think the general idea is good. Location of where to apply it should be changed and potentially how to delete entries can be made easier imo.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -568,6 +572,9 @@ proc ping*(p: PortalProtocol, dst: Node):
     Future[PortalResult[PongMessage]] {.async.} =
   let pongResponse = await p.pingImpl(dst)
 
+  # Update last time we pinged this node
+  p.pingTimings[dst.id] = now(chronos.Moment)
```
kdeme, line 576: Perhaps only call this if the pong response was actually ok.
If it wasn't it would get deleted anyhow.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -1252,15 +1259,28 @@ proc revalidateNode*(p: PortalProtocol, n: Node) {.async.} =
         if nodes.len > 0: # Normally a node should only return 1 record actually
           discard p.routingTable.addNode(nodes[0])
 
+proc getNodeForRevalidation(p: PortalProtocol): Opt[Node] =
+  var ret = Opt.none(Node)
+
+  for i in 0..numberOfNodeRetries:
```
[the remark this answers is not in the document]
mynameisdaniil: Perhaps it comes down to what behaviour we want: lazy or eager. I figured that eager tactics would be better here, since it's relatively cheap to check a few or all nodes to find something we haven't pinged yet. But I'm totally fine with making this lazy.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -1252,15 +1259,28 @@ proc revalidateNode*(p: PortalProtocol, n: Node) {.async.} =
...
+proc getNodeForRevalidation(p: PortalProtocol): Opt[Node] =
+  let node = p.routingTable.nodeToRevalidate()
+  if node.isNil:
+    return Opt.none(Node)
+
+  let now = now(chronos.Moment)
+  let timestamp = p.pingTimings.getOrDefault(node.id, now)
+
+  if (timestamp + timeoutInterval) <= now:
+    return Opt.some(node)
```
kdeme, line 1271: Nothing wrong here, but you can omit the return statements in this if/else and use "expression" returns. See style guide: https://status-im.github.io/nim-style-guide/language.result.html

fluffy/network/wire/portal_protocol.nim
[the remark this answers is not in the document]
kdeme: Yes, I prefer not to be "eager" here, as it will only occur frequently when the routing table is small, and we don't need to overly run the revalidate in that case.

kdeme (approved):

PR #1799, opened 2023-10-03
Use admin prefix for nodeInfo and addPeer JSON-RPCs

kdeme:


jangko:
This will breaks hive test.
Need to fix hive's enode.sh too.

https://github.com/ethereum/hive/blob/b97408ce87bd0856e054fcdc8a54cf327716a8b3/clients/nimbus-el/enode.sh#L12

kdeme:
Thanks for pointing that out!

PR #1813, opened 2023-10-09
implement portal_historyTraceRecursiveFindContent

mynameisdaniil:
I'm still working out some kinks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 
 # Nimble packages
 /vendor/.nimble
-/vendor/
 
 # ntags/ctags output
 /tags
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 
 # Nimble packages
 /vendor/.nimble
+/vendor/
 
 # ntags/ctags output
 /tags
--- a/fluffy/network/wire/portal_protocol.nim
+++ b/fluffy/network/wire/portal_protocol.nim
@@ -12,12 +12,13 @@
 
 import
   std/[sequtils, sets, algorithm, tables],
-  stew/[results, byteutils, leb128, endians2], chronicles, chronos,
+  stew/[results, byteutils, leb128, endians2, shims/net], chronicles, chronos,
   nimcrypto/hash, bearssl, ssz_serialization, metrics, faststreams,
   eth/rlp, eth/p2p/discoveryv5/[protocol, node, enr, routing_table, random2,
     nodes_verification, lru],
   ../../seed_db,
   "."/[portal_stream, portal_protocol_config],
+  ../state/state_distance,
   ./messages
 
 export messages, routing_table, protocol
@@ -191,6 +192,26 @@ type
   ContentLookupResult* = object
     content*: seq[byte]
     utpTransfer*: bool
+    # List of nodes which do not have requested content, and for which
+    # content is in their range
+    nodesInterestedInContent*: seq[Node]
+
+  TraceResponse* = object
+    nodeId*: NodeId
+    ts*: Moment
+    respondedWith*: seq[NodeId]
+    enr*: Record
+    ip*: ValidIpAddress
+    port*: Port
+    distance*: UInt256
+    distance_log2*: uint16
+
+  TraceContentLookupResult* = object
+    origin*: NodeId
+    receivedFrom*: NodeId
+    content*: seq[byte]
+    ts*: Moment
+    responses*: seq[TraceResponse]
     # List of nodes which do not have requested content, and for which
     # content is in their range
     nodesInterestedInContent*: seq[Node]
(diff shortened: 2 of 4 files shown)


fluffy/rpc/rpc_types.nim
```diff
@@ -8,12 +8,46 @@
...
 import
+  std/tables,
   json_rpc/jsonmarshal,
+  chronos/timer,
   stew/results,
   eth/p2p/discoveryv5/[routing_table, enr, node]
 
 export jsonmarshal, routing_table, enr, node
 
+type
```
mynameisdaniil, line 19: On one hand I'm not really sure about all those additional types. On the other hand I feel like it isn't a good idea to construct final types/object in portal_protocol.
Also, as you may have noticed, most of the fields in this request are snake_cased. I feel like we should make them camelCased and also send a patch to glados (and trin) to fix that.

fluffy/rpc/rpc_portal_api.nim
```diff
@@ -173,6 +173,53 @@ proc installPortalApiHandlers*(
...
+
+    var responses = initTable[string, TraceItem]()
+    var metadata = initTable[string, NodeMetadata]()
+    for i in contentResult.responses:
+      var ids = newSeq[NodeId]()
+
+      for n in i.respondedWith:
+        ids.add(n)
+
+      metadata[fmt("0x{i.nodeId}")] = NodeMetadata(
```
mynameisdaniil, line 195: I'm pretty sure there must be a method to do this but I was unable to find it (nimsuggest is basically useless at this point, and just using search brings a lot of files, I might have missed something)

fluffy/network/wire/portal_protocol.nim
```diff
@@ -1064,6 +1085,144 @@ proc contentLookup*(p: PortalProtocol, target: ByteList, targetId: UInt256):
...
+
+        responses.add(TraceResponse(
+          nodeId: content.src.id,
+          ts: ts,
+          respondedWith: respondedWith,
+          enr: content.src.record,
+          ip: address.ip,
+          port: address.port,
+          distance: distance,
+          distance_log2: stateLogDistance(content.src.id, targetId)
```
kdeme, line 1184: Nothing network specific should be added here.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -12,12 +12,13 @@
...
 import
   std/[sequtils, sets, algorithm, tables],
-  stew/[results, byteutils, leb128, endians2], chronicles, chronos,
+  stew/[results, byteutils, leb128, endians2, shims/net], chronicles, chronos,
   nimcrypto/hash, bearssl, ssz_serialization, metrics, faststreams,
   eth/rlp, eth/p2p/discoveryv5/[protocol, node, enr, routing_table, random2,
     nodes_verification, lru],
   ../../seed_db,
   "."/[portal_stream, portal_protocol_config],
+  ../state/state_distance,
```
kdeme, line 21: nothing network specific should be imported here. Portal protocol should remain agnostic of network customizations.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -195,6 +196,26 @@ type
...
 
+  TraceResponse* = object
+    nodeId*: NodeId
+    ts*: Moment
+    respondedWith*: seq[NodeId]
+    enr*: Record
+    ip*: ValidIpAddress
+    port*: Port
+    distance*: UInt256
+    distance_log2*: uint16
```
kdeme, line 207: The logarithmic distance here is taken from the distance, I assume? Really shouldn't pass both then.

Are they both required in the json reply?

fluffy/rpc/rpc_types.nim
```diff
@@ -8,12 +8,46 @@
...
+  TraceItem* = object
+    timestamp_millis*: int64
+    responded_with*: seq[NodeId]
```
kdeme, line 22: These don't need to be the ENRs?

fluffy/rpc/rpc_types.nim
```diff
@@ -8,12 +8,46 @@
...
+
+type
+  NodeMetadata* = object
+    enr*: Record
+    ip*: string
+    port*: string
```
kdeme, line 28: ip and port is already available in the ENR normally?

Also a bit confused, it seems that things diverted a bit from the original description in the Trin PR?

fluffy/rpc/rpc_portal_api.nim
```diff
@@ -173,6 +173,53 @@ proc installPortalApiHandlers*(
...
+
+    return TraceNodeInfo(
+        content: contentResult.content.to0xHex(),
+        trace: TraceContainer(
+          origin: contentResult.origin,
+          target_id: contentKey,
+          received_content_from_node: contentResult.receivedFrom,
+          started_at: StartedAt(
+            secs_since_epoch: chronos.epochSeconds(contentResult.ts),
+            nanos_since_epoch: chronos.epochNanoSeconds(contentResult.ts)
```
kdeme, line 216: Why have both?

fluffy/rpc/rpc_types.nim
kdeme: Yes, we typically have separate types for the JSON-RPC, and I think that's good. 

Now, here you could actually go with one type as it is only a returned type and it's not actually used for anything internally. But you'd have to do the conversions by adding  `%` and  `fromJson` calls to keep that conversion code out of the portal wire protocol code (we want to avoid cluttering the code there with JSON related stuff).

I don't mind either way (current way or with those overloads), but it does seem that the current type can be made a bit more condense, see my comments/questions.

fluffy/network/wire/portal_protocol.nim
mynameisdaniil: Yep, they were. But I've fixed it.

fluffy/rpc/rpc_types.nim
mynameisdaniil: If you mean this PR (https://github.com/ethereum/trin/pull/482) it has almost nothing to do with how things actually worked, except for the general idea of responses and metadata. Also, there was noticeable de-sync between what data trin sent and what data glados actually utilized.
As for ip and port, I've fixed decoding in glados and now ENR is enough.

fluffy/rpc/rpc_portal_api.nim
mynameisdaniil: It turns out this data is ignored

fluffy/rpc/rpc_portal_api.nim
```diff
@@ -24,6 +24,12 @@ type
     content: string
     utpTransfer: bool
 
+  TraceNodeInfo* = object
```
kdeme, line 27: To be consistent with above type, perhaps this is better called `TraceContentInfo` ?

```suggestion
  TraceContentInfo* = object
```

fluffy/rpc/rpc_types.nim
```diff
@@ -10,7 +10,8 @@
 import
   json_rpc/jsonmarshal,
   stew/results,
-  eth/p2p/discoveryv5/[routing_table, enr, node]
+  eth/p2p/discoveryv5/[routing_table, enr, node],
+  ../network/wire/portal_protocol
```
kdeme, line 14: This import is no longer needed I assume?

fluffy/network/wire/portal_protocol.nim
```diff
@@ -1064,6 +1085,182 @@ proc contentLookup*(p: PortalProtocol, target: ByteList, targetId: UInt256):
...
+      error "Resulting query should have been in the pending queries"
+
+    let contentResult = query.read
+
+    if contentResult.isOk():
+      let content = contentResult.get()
+
+      case content.kind
+      of Nodes:
+        let duration = chronos.milliseconds(now(chronos.Moment) - ts)
```
kdeme, line 1169: This is the duration of receival since the start of the lookup. Are you sure this is not supposed to be seen at the duration of the actual individual request?

fluffy/network/wire/portal_protocol.nim
```diff
@@ -1064,6 +1085,182 @@ proc contentLookup*(p: PortalProtocol, target: ByteList, targetId: UInt256):
...
+          respondedWith: respondedWith,
+        )
+
+        metadata["0x" & $content.src.id] = NodeMetadata(
+          enr: content.src.record,
+          distance: distance,
+        )
+
+      of Content:
+        let duration = chronos.milliseconds(now(chronos.Moment) - ts)
```
kdeme, line 1216: Same remark as above.

kdeme (commented):
lgtm, just that remark against the actual meaning of the `durationMs`

fluffy/rpc/rpc_types.nim
mynameisdaniil: enr and node yes, portal_protocol - yes, since we're just returning objects from protocol

fluffy/network/wire/portal_protocol.nim
mynameisdaniil: I've looked up trin source code when I was implementing this.
https://github.com/ethereum/trin/blob/master/ethportal-api/src/types/query_trace.rs#L54
and

fluffy/rpc/rpc_types.nim
mynameisdaniil: Sorry, I was looking at another file.

fluffy/rpc/rpc_portal_api.nim
```diff
@@ -173,6 +179,23 @@ proc installPortalApiHandlers*(
         utpTransfer: contentResult.utpTransfer
       )
 
+  rpcServer.rpc("portal_historyTraceRecursiveFindContent") do(
```
kdeme, line 182: I hadn't noticed before, but the method name should used the `network` parameter here.

.gitignore
```diff
@@ -5,6 +5,7 @@
 
 # Nimble packages
 /vendor/.nimble
+/vendor/
```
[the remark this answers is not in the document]
mynameisdaniil: apparently not all subprojects in `/vendor` contain .nimble and I had a hard time with tests or something like that. The state that I've forked from was hard pushed probably and the tests caused merge problems on rebase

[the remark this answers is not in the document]
kdeme: I'm not sure how that helps. In case vendor modules are updated and you do a rebase, you correct them to the right version by running `make update`. Or you do it manually, for example in case you had the module also updated and you have a conflict.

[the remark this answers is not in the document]
mynameisdaniil: in case of a rebase, you're right it didn't help, but it helped along the way to ignore this change and not add it to commit

PR #1842, opened 2023-10-23
First go at the Fluffy book using mkdocs as for nimbus.guide

kdeme:
Still need to move bit of documentation from the main fluffy readme.md and then update that readme with the link to the fluffy book. Will do that in second PR with potentially some more changes.

### DIFF
--- a/fluffy/docs/the_fluffy_book/docs/index.md
+++ b/fluffy/docs/the_fluffy_book/docs/index.md
@@ -3,6 +3,12 @@
 Fluffy is the Nimbus client implementation of the
 [Portal network specifications](https://github.com/ethereum/portal-network-specs/blob/master/portal-network.md).
 
+The Portal Network aims to deliver a reliable, sync-free, and decentralized
+access to the Ethereum blockchain. The network can be used by a light client to
+get access to Ethereum data and as such become a drop-in replacement for full
+nodes by providing that data through the existing
+[Ethereum JSON RPC Execution API](https://github.com/ethereum/execution-apis).
+
 This book describes how to build, run and monitor the Fluffy client, and how to
 use and test its currently implemented functionality.
 
--- a/fluffy/docs/the_fluffy_book/docs/index.md
+++ b/fluffy/docs/the_fluffy_book/docs/index.md
@@ -3,14 +3,19 @@
 Fluffy is the Nimbus client implementation of the
 [Portal network specifications](https://github.com/ethereum/portal-network-specs/blob/master/portal-network.md).
 
-The Portal Network is a project still in research phase.
-This client is thus still highly experimental.
-
-Some of the sub-networks are already operational and can be tested in a local
-setting or in the public testnet.
-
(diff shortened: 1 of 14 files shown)


zah:
Where is the Fluffy guide going to be published? There is an ongoing effort to standardize the publishing platform across all Logos products and the current solution is based on Docusaurus:
https://github.com/status-im/infra-misc/issues/143

arnetheduck:
> Docusaurus

I think this doesn't really matter for this PR - ie the docusaurus migration is something that marketing needs to figure out and when/if they do, we'll cross that bridge (or rather, they will: they recommended mkdocs initially and might change away from docusaurus as well before the dust has settled). Ultimately, the renderer doesn't matter so much as the content and structure, which remains the same.

fluffy/docs/the_fluffy_book/docs/index.md
```diff
@@ -0,0 +1,31 @@
+# The Nimbus Fluffy Guide
+
+Fluffy is the Nimbus client implementation of the
+[Portal network specifications](https://github.com/ethereum/portal-network-specs/blob/master/portal-network.md).
+
+The Portal Network is a project still in research phase.
```
arnetheduck, line 6: we can move these warnings somewhere further down and focus on how to get up and running first, I think - maybe we can include a "readiness map" of some sort (ie which json-rpc reuqests are functional and to what degree) as well.

kdeme:
> Where is the Fluffy guide going to be published? There is an ongoing effort to standardize the publishing platform across all Logos products and the current solution is based on Docusaurus:
> [status-im/infra-misc#143](https://github.com/status-im/infra-misc/issues/143)

So my plan is to currently host it on GitHub pages of nimbus-eth1 and perhaps we can get either a new domain or use for example nimbus.guide/fluffy to link to it.

I agree with @arnetheduck and I'm not going to wait for any standardisation to happen before actually creating/deploying the book.
When that standardisation within Status happens, both the nimbus-eth2 book and fluffy book can be moved by the person/team that handles this.

fluffy/docs/the_fluffy_book/docs/index.md
kdeme: I've reorganised the index page a little. I'll add more information under the development status later on (or perhaps create a new page to where I link).
For now the goal is/was to move most existing documentation to the book and improve it somewhat to have better access for users/devs.

fluffy/docs/the_fluffy_book/docs/index.md
```diff
@@ -0,0 +1,31 @@
+# The Nimbus Fluffy Guide
+
+Fluffy is the Nimbus client implementation of the
+[Portal network specifications](https://github.com/ethereum/portal-network-specs/blob/master/portal-network.md).
```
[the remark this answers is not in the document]
kdeme: Here's an attempt https://github.com/status-im/nimbus-eth1/pull/1842/commits/715efe4f1412b777ea3af608b29076e6f776d620

PR #1933, opened 2023-12-11
Remove contentSize and contentCount from history network log

kdeme:
These are quite expensive database operations, especially the contentSize one, and are not that necessary to log periodically considering dbSize is already logged and is the more important one. ContentSize is mostly useful to see the database overhead, which we could also add through some json rpc debug call if needed.

### DIFF
--- a/fluffy/network/history/history_network.nim
+++ b/fluffy/network/history/history_network.nim
@@ -779,8 +779,6 @@ proc statusLogLoop(n: HistoryNetwork) {.async.} =
       info "History network status",
         radius = radiusPercentage.toString(10) & "%",
         dbSize = $(n.contentDB.size() div 1000) & "kb",
-        contentSize = $(n.contentDB.contentSize() div 1000) & "kb",
-        contentCount = n.contentDB.contentCount(),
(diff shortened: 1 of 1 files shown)


kdeme:
We can see here the cost of it on a 25GB database on our fleet machine, which makes sense as it has to iterate over all content:
```
All timings are in ms and are cpu time.
     Average,       StdDev,          Min,          Max,      Samples,         Test 
       0.157,        0.185,        0.091,        1.926,          100, DB put for content
       0.055,        0.013,        0.034,        0.086,          100, DB get for content
       0.004,        0.001,        0.003,        0.009,          100, DB contains for content
       0.102,        0.214,        0.069,        2.219,          100, DB delete for content
       0.023,        0.002,        0.020,        0.034,          100, Get DB size
       0.011,        0.001,        0.010,        0.022,          100, Get DB used size
     483.296,       11.425,      473.106,      525.281,          100, Get DB content size
      31.085,        0.885,       30.194,       33.985,          100, Get DB content count
     134.721,        3.188,      131.172,      143.192,          100, Get DB largest distance
```

And the effect on disk utilisation on the machines (these run 32 nodes which each do this once a minute).
image

PR #1947, opened 2023-12-18
Initial implementation of account and storage proof verification for the portal state network.

web3-developer:


web3-developer:
I'm still working on the unit tests locally so I haven't committed these yet.

fluffy/network/state/experimental/state_proof_verification.nim
```diff
@@ -0,0 +1,79 @@
...
+type
+  MptProof = seq[seq[byte]]
+  AccountProof* = distinct MptProof
+  StorageProof* = distinct MptProof
+
+
+proc verifyAccount*(
+    trustedStateRoot: KeccakHash,
+    address: EthAddress,
+    account: Account,
```
web3-developer, line 28: The EthAddress and Account types are from eth/common. @kdeme Do you think this is okay or do we want to use the SSZ types (The types used to encode/decode SSZ into objects)?

kdeme: I think that's fine for now. The Portal specification does not define currently an encoding even (it is just the RLP encoded data), so we can see which direction we go with here in the future.

fluffy/network/state/experimental/state_proof_verification.nim
```diff
@@ -0,0 +1,79 @@
...
+    proof: AccountProof): Result[void, string] =
+  assert trustedStateRoot.data.len() > 0
```
kdeme, line 30: We typically prefer `doAssert` over `assert`, so that the check does not get turned off due to compile options.

Hitting the asserts here would mean misuse of the API and/or not having checks build in properly somewhere up in the chain, in which case we want to know this asap.

kdeme: Actually, I just realized that the code in https://github.com/status-im/nim-eth/blob/master/eth/trie/hexary_proof_verification.nim#L187 is fairly recent as it was added for our nimbus-verified-proxy. The code itself already does seem to check (most?) of the inputs. 

It is being used in our nimbus-verified-proxy here: https://github.com/status-im/nimbus-eth1/blob/master/nimbus_verified_proxy/validate_proof.nim#L45

In that use case, the content is also returned as it needs to be used. Sorry for realizing this code only now, I had forgotten about it :/

fluffy/network/state/experimental/state_proof_verification.nim
```diff
@@ -0,0 +1,79 @@
+# Nimbus
+# Copyright (c) 2022-2023 Status Research & Development GmbH
```
kdeme, line 2:
```suggestion
# Copyright (c) 2023 Status Research & Development GmbH
```

fluffy/network/state/experimental/state_proof_verification.nim
```diff
@@ -0,0 +1,79 @@
...
+
+proc verifyContractStorageSlot*(
+    trustedStorageRoot: KeccakHash,
+    slotKey: UInt256,
+    slotValue: UInt256,
+    proof: StorageProof): Result[void, string] =
+  assert trustedStorageRoot.data.len() > 0
+  assert proof.MptProof.len() > 0
+
+  let key = toSeq(keccakHash(toBytesBE(slotKey)).data)
```
kdeme, line 57: Not of too much concern for this PR now, but we could probably avoid the sequence by using `openArray` down the verify API, TBI.

fluffy/network/state/experimental/state_proof_verification.nim
web3-developer: Sure, I'll leave it as is for now then. Once the specs have been finalized I'll update if needed.

fluffy/network/state/experimental/state_proof_verification.nim
web3-developer: Thanks, will update.

fluffy/network/state/experimental/state_proof_verification.nim
web3-developer: Thanks for the early feedback on this one. Yes it seems that this library is using `seq[byte]` instead of `openArray[byte]` which isn't ideal

fluffy/network/state/experimental/state_proof_verification.nim
kdeme: Yes, but the checks down the API throw not asserts but instead return a `Result` type as error. That's the opposite order that one expect (in case of usage of asserts, I would expect them on the lowest layer, where the API assumes verified input).

Additionally, having looked at it better now I also noticed that two of `len` checks are actually on an arrays. These will always pass.

kdeme: By the way, the code in https://github.com/status-im/nim-eth/blob/master/eth/trie/hexary_proof_verification.nim#L187 and by addition its usage in https://github.com/status-im/nimbus-eth1/blob/master/nimbus_verified_proxy/validate_proof.nim#L45 was developed for a proof-of-concept (nimbus_verified_proxy), and while it works, I don't think it has ever been thoroughly tested.

I can only find two tests for the happy cases: https://github.com/status-im/nim-eth/blob/master/tests/trie/test_hexary_proof.nim

web3-developer: > Yes, but the checks down the API throw not asserts but instead return a `Result` type as error. That's the opposite order that one would expect (in case of usage of asserts, I would expect them on the lowest layer, where the API assumes verified input).
> 
> Additionally, having looked at it better now I also noticed that two of `len` checks are actually on an arrays. These will always pass.

Oh thanks, I didn't notice that. I'll remove the checks on the arrays. I've also removed the checks on the proofs because they are getting checked in verifyMptProof already anyway.

web3-developer: > By the way, the code in https://github.com/status-im/nim-eth/blob/master/eth/trie/hexary_proof_verification.nim#L187 and by addition its usage in https://github.com/status-im/nimbus-eth1/blob/master/nimbus_verified_proxy/validate_proof.nim#L45 was developed for a proof-of-concept (nimbus_verified_proxy), and while it works, I don't think it has ever been thoroughly tested.
> 
> I can only find two tests for the happy cases: https://github.com/status-im/nim-eth/blob/master/tests/trie/test_hexary_proof.nim

Good to know. I'll make sure I cover it with lots of tests in that case.

PR #1948, opened 2023-12-19
Initial impl of state network bridge 1902

mynameisdaniil:


mynameisdaniil:
I've just noticed that in the link that you've sent me there are types that I re-created in my PR. Let me fix that and use correct types.

fluffy/network/state/state_content.nim
```diff
@@ -53,6 +55,17 @@ type
     address*: Address
     codeHash*: CodeHash
 
+  WitnessNode* = ByteList
+  AccountTrieProof* = List[WitnessNode, 32]
+
+  AccountState* = object
```
kdeme, line 61: The current specification puts the account node's value as RLP encoded bytes in the proof. And I think also the state root is put in there, although the spec does not explicitly state that.

I notice that you keep it separate and make one `AccountState` object containing all Account data and proof + state root, and then SSZ encode this instead. What's the plan here? Do we plan on suggesting this change, and for which reason?

If we were to do a change like this, I think I would keep the account data in a separate object and/or keep it still RLP encoded.

fluffy/network/state/state_content.nim
```diff
@@ -53,6 +55,17 @@ type
...
+    nonce*: uint64
+    balance*: UInt256
+    storageHash*: StorageHash # is it NodeHash?
```
kdeme, line 64: `storageRoot`

fluffy/network/state/state_network.nim
```diff
@@ -59,8 +60,45 @@ proc getContent*(n: StateNetwork, key: ContentKey):
   # domain types.
   return Opt.some(contentResult.content)
 
-proc validateContent(content: openArray[byte], contentKey: ByteList): bool =
-  true
+proc validateContent(
+  n: StateNetwork,
+  contentKey: ByteList,
+  conetentValue: seq[byte]): Future[bool] {.async.} =
```
kdeme, line 66:
```suggestion
  contentValue: seq[byte]): Future[bool] {.async.} =
```

kdeme, line 66:
```suggestion
    n: StateNetwork,
    contentKey: ByteList,
    conetentValue: seq[byte]): Future[bool] {.async.} =
```

fluffy/network/state/state_network.nim
```diff
@@ -59,8 +60,45 @@ proc getContent*(n: StateNetwork, key: ContentKey):
...
+      return true
+    of contractStorageTrieProof:
+      return true
+    of contractBytecode:
+      return true
+
+proc validateContent(
+  n: StateNetwork,
+  contentKeys: ContentKeysList,
+  contentValues: seq[seq[byte]]): Future[bool] {.async.} =
```
kdeme, line 88:
```suggestion
    n: StateNetwork,
    contentKeys: ContentKeysList,
    contentValues: seq[seq[byte]]): Future[bool] {.async.} =
```

fluffy/tests/state_data/proofs.full.block.0.json
```diff
@@ -0,0 +1,23 @@
+{
```
kdeme, line 1: These new testdata files should go in portal-spec-tests submodule.

Any old json files are legacy and eventually need to be moved there too.

fluffy/tests/test_state_network.nim
```diff
@@ -32,6 +34,76 @@ proc genesisToTrie(filePath: string): CoreDbMptRef =
...
+  asyncTest "Decode and use proofs":
+    let file = "./fluffy/tests/state_data/proofs.full.block.0.json"
+    var content: JsonNode
+    try:
+      content = io.readFile(file).parseJson()
+    except Exception as e:
+      echo "Skipping file ", file, " because of error"
+      quit(1)
+
+    let
```
kdeme, line 47: A network test is fine, but as first step, we definitely need to do this encoding/decoding testing without network, and that code would belong in test_state_content.

fluffy/tests/test_state_network.nim
```diff
@@ -32,6 +34,76 @@ proc genesisToTrie(filePath: string): CoreDbMptRef =
 
 procSuite "State Content Network":
   let rng = newRng()
+
+  asyncTest "Decode and use proofs":
+    let file = "./fluffy/tests/state_data/proofs.full.block.0.json"
+    var content: JsonNode
+    try:
+      content = io.readFile(file).parseJson()
+    except Exception as e:
```
kdeme, line 43: Don't use raw `Exception`.

You might also consider using `stew/io2` which uses `Result` based error handling.

fluffy/tests/test_state_network.nim
kdeme: Also, not completely against they way you are doing it now, but for testing I personally find the code cleaner and easier adaptable when using `json_serialization`.  E.g. https://github.com/status-im/nimbus-eth1/blob/master/fluffy/eth_data/history_data_json_store.nim#L190

fluffy/network/state/state_content.nim
mynameisdaniil: We don't really need a stateRoot here, since it contained in the key.
Everything else - agreed. Fixed that.

fluffy/network/state/state_content.nim
```diff
@@ -117,7 +121,7 @@ func toContentId*(contentKey: ContentKey): ContentId =
       h.update(key.address)
       h.update(key.codeHash.data)
 
-func toContentId*(contentKey: ByteList): results.Opt[ContentId] =
+proc toContentId*(contentKey: ByteList): results.Opt[ContentId] =
```
kdeme, line 124: can stay `func`?

fluffy/network/state/state_network.nim
```diff
@@ -25,7 +28,7 @@ type StateNetwork* = ref object
   contentQueue*: AsyncQueue[(Opt[NodeId], ContentKeysList, seq[seq[byte]])]
   processContentLoop: Future[void]
 
-func toContentIdHandler(contentKey: ByteList): results.Opt[ContentId] =
+proc toContentIdHandler(contentKey: ByteList): results.Opt[ContentId] =
```
kdeme, line 31: Can stay `func`?

fluffy/network/state/state_network.nim
```diff
@@ -59,8 +62,58 @@ proc getContent*(n: StateNetwork, key: ContentKey):
...
-proc validateContent(content: openArray[byte], contentKey: ByteList): bool =
-  true
+proc validateContent(
+    n: StateNetwork,
+    contentKey: ByteList,
+    contentValue: seq[byte]): Future[bool] {.async.} =
+  let key = contentKey.decode().valueOr:
+    return false
+
+  case key.contentType:
```
kdeme, line 72: Just a warning that I think when you will rebase you will need to add the `unused` case :/.

fluffy/network/state/state_network.nim
```diff
@@ -59,8 +62,58 @@ proc getContent*(n: StateNetwork, key: ContentKey):
...
+        warn "Received invalid account trie proof", error
+        return false
+      var proof: seq[seq[byte]]
+      for p in decodedProof:
+        proof.add(p.toSeq())
+      let
+        trieKey = keccakHash(key.accountTrieProofKey.address).data.toSeq()
+        value = proof[^1].decode(seq[seq[byte]])[^1]
+        stateRoot = MDigest[256](data: key.accountTrieProofKey.stateRoot)
+        result = verifyMptProof(proof, stateRoot, trieKey, value)
```
kdeme, line 88: Using `result` as variable name is something I prefer not to do. I guess it is not so risky here as the return type is different but I think in some cases it could lead to some accidental bugs in refactors or so.

fluffy/network/state/state_network.nim
```diff
@@ -59,8 +62,58 @@ proc getContent*(n: StateNetwork, key: ContentKey):
...
+  let key = contentKey.decode().valueOr:
+    return false
+
+  case key.contentType:
+    of accountTrieNode:
+      return true
+    of contractStorageTrieNode:
+      return true
+    of accountTrieProof:
+      var decodedProof = decodeSsz(contentValue, AccountTrieProof).valueOr:
```
kdeme, line 78:
```suggestion
      let decodedProof = decodeSsz(contentValue, AccountTrieProof).valueOr:
```

fluffy/network/state/state_network.nim
```diff
@@ -59,8 +62,58 @@ proc getContent*(n: StateNetwork, key: ContentKey):
...
+        warn "Received invalid account trie proof", error
+        return false
+      var proof: seq[seq[byte]]
+      for p in decodedProof:
+        proof.add(p.toSeq())
```
kdeme, line 83: Considering an SSZ List is a distinct `seq` it is better to just use `distinctBase` here. We also have a convenient `asSeq` template that you can use for Lists: https://github.com/status-im/nim-ssz-serialization/blob/f87c99be04f3051fe2d9b4facf5e6e71251e8555/ssz_serialization/types.nim#L154C34-L154C46

fluffy/network/state/state_network.nim
kdeme, line 88: We should be using our `verifyAccount` API here, but it looks like it currently doesn't fully map well on the received data objects.

Not for this PR perhaps but should get adjusted on both sides eventually to be seamless in usage.

fluffy/network/state/state_network.nim
```diff
@@ -59,8 +62,58 @@ proc getContent*(n: StateNetwork, key: ContentKey):
...
+        stateRoot = MDigest[256](data: key.accountTrieProofKey.stateRoot)
+        result = verifyMptProof(proof, stateRoot, trieKey, value)
+      case result.kind:
+        of ValidProof:
+          return true
+        else:
+          warn "Received invalid account trie proof"
+          return false
+    of contractStorageTrieProof:
+      return true
```
kdeme, line 96: You could remove all the return statements also and go expression based.

fluffy/tests/test_state_network.nim
```diff
@@ -32,6 +40,123 @@ proc genesisToTrie(filePath: string): CoreDbMptRef =
...
+      addressBytes = hexToByteArray[20](address)
+      stateRootBytes = hexToByteArray[sizeof(state_content.AccountTrieProofKey.stateRoot)](stateRoot)
+      key = AccountTrieProofKey(address: addressBytes, stateRoot: stateRootBytes)
+
+    let encodedKey = SSZ.encode(key)
+    check encodedKey.to0xHex() == hexKey
+    let decodedKey = decodeSsz(encodedKey, AccountTrieProofKey)
+    check decodedKey.isOk()
+
+  asyncTest "Encode/decode accountTrieProof":
```
kdeme, line 60:
```suggestion
  test "Encode/decode accountTrieProof":
```

fluffy/tests/test_state_network.nim
```diff
@@ -32,6 +40,123 @@ proc genesisToTrie(filePath: string): CoreDbMptRef =
...
+      discard accountTrieProof.add(witnessNode)
+
+    let
+      encodedProof = SSZ.encode(accountTrieProof)
+      decodedProof = decodeSsz(encodedProof, AccountTrieProof).get()
+
+    check decodedProof == accountTrieProof
+
+
+  asyncTest "Test account state proof":
```
kdeme, line 85:
```suggestion
  test "Test account state proof":
```

fluffy/tests/test_state_network.nim
```diff
@@ -32,6 +40,123 @@ proc genesisToTrie(filePath: string): CoreDbMptRef =
 
 procSuite "State Content Network":
   let rng = newRng()
+
+  test "Encode/decode accountTrieProofKey":
```
kdeme, line 44: I think these encoding/decoding tests should not go into the test_state_network.nim. Instead in the content encoding/decoding one.

fluffy/tests/test_state_network.nim
```diff
@@ -32,6 +40,123 @@ proc genesisToTrie(filePath: string): CoreDbMptRef =
...
+        Json.decode(content, state_bridge.JsonProofVector)
+      except SerializationError:
+        quit(1)
+    let
+      proof = decoded.proofs[0].proof.map(hexToSeqByte)
+      stateRoot = MDigest[256].fromHex(decoded.state_root)
+      address = hexToByteArray[20](decoded.proofs[0].address)
+      key = keccakHash(address).data.toSeq()
+      value = proof[^1].decode(seq[seq[byte]])[^1]
+      proofResult = verifyMptProof(proof, stateRoot, key, value)
```
kdeme, line 101: This proof test should probably be added in the proof verification tests?

It is sort of a mess to have these same/similar tests but with different APIs.

This needs to become all one (see above) way of usage.

fluffy/tools/state_bridge/state_bridge.nim
```diff
@@ -0,0 +1,113 @@
...
+# Portal Network <-> Portal Client (e.g. fluffy) <--JSON-RPC--> bridge <--REST--> consensus client (e.g. Nimbus-eth2)
+#
+# The Consensus client must support serving the Beacon LC data.
+#
+# Bootstraps and updates can be backfilled, however how to do this for multiple
+# bootstraps is still unsolved.
+#
+# Updates, optimistic updates and finality updates are injected as they become
+# available.
+#
```
kdeme, line 22: Not applicable?

fluffy/tools/state_bridge/state_bridge.nim
```diff
@@ -0,0 +1,113 @@
...
+  state*: JsonAccount
+  proof*: seq[string]
+
+type JsonProofVector* = object
+  `block`*: int
+  block_hash*: string
+  state_root*: string
+  proofs*: seq[JsonProof]
+
+proc run(config: BeaconBridgeConf) {.raises: [CatchableError].} =
```
kdeme, line 57: Naming not adjusted?

fluffy/tools/state_bridge/state_bridge.nim
```diff
@@ -0,0 +1,113 @@
...
+  setupLogging(config.logLevel, config.logStdout)
+
+  notice "Launching Fluffy fake state bridge",
+    cmdParams = commandLineParams()
+
+  let portalRpcClient = newRpcHttpClient()
+
+  proc backfill(rpcAddress: string, rpcPort: Port) {.async raises: [OSError].} =
```
kdeme, line 65: Naming? Not really a backfill?

fluffy/tools/state_bridge/state_bridge_conf.nim
```diff
@@ -0,0 +1,69 @@
...
+      name: "rpc-address" .}: string
+
+    rpcPort* {.
+      desc: "Listening port of the Portal JSON-RPC server"
+      defaultValue: 8545
+      name: "rpc-port" .}: Port
+
+    dataDir* {.
+      desc: "Data directory to lookup state data"
+      defaultValue: "./vendor/portal-spec-tests/tests/mainnet/state/"
```
kdeme, line 50: I wouldn't set this as default

fluffy/tools/state_bridge/state_bridge_conf.nim
```diff
@@ -0,0 +1,69 @@
...
+  confutils, confutils/std/net,
+  nimcrypto/hash,
+  ../../logging
+
+export net
+
+type
+  TrustedDigest* = MDigest[32 * 8]
+
+  BeaconBridgeCmd* = enum
```
kdeme, line 20: Naming?

fluffy/network/state/state_network.nim
mynameisdaniil: The problem is that I also need to convert each item on the list from `ByteList` to `seq[byte]`. I re-did it with `asSeq map toSeq` but not sure if this is the right way either.

fluffy/tests/test_state_network.nim
mynameisdaniil: Agreed. Yet I think this should be done as a separate PR.

fluffy/tools/state_bridge/state_bridge.nim
mynameisdaniil: It kinda is. Anyway I can't think of a better name for that method.

fluffy/tools/state_bridge/state_bridge_conf.nim
mynameisdaniil: For the time being I can't think of a better default value.

mynameisdaniil:
Re-based to current master

fluffy/tools/state_bridge/state_bridge_conf.nim
```diff
@@ -0,0 +1,69 @@
...
+
+import
+  confutils, confutils/std/net,
+  nimcrypto/hash,
+  ../../logging
+
+export net
+
+type
+  TrustedDigest* = MDigest[32 * 8]
```
kdeme, line 18: Unused type?

fluffy/tools/state_bridge/state_bridge_conf.nim
kdeme: Perhaps use it without a default?  Could add instead docs how data can be provided.

fluffy/tools/state_bridge/state_bridge_conf.nim
```diff
@@ -0,0 +1,69 @@
...
+      desc: "Listening port of the Portal JSON-RPC server"
+      defaultValue: 8545
+      name: "rpc-port" .}: Port
+
+    dataDir* {.
+      desc: "Data directory to lookup state data"
+      defaultValue: "./vendor/portal-spec-tests/tests/mainnet/state/"
+      name: "data-dir".}: string
+
+    blockAmount* {.
```
kdeme, line 53: Not used?

fluffy/tools/state_bridge/state_bridge_conf.nim
```diff
@@ -0,0 +1,69 @@
...
+      defaultValue: noCommand .}: StateBridgeCmd
+    of noCommand:
+      discard
+
+func parseCmdArg*(T: type TrustedDigest, input: string): T
+                 {.raises: [ValueError].} =
+  TrustedDigest.fromHex(input)
+
+func completeCmdArg*(T: type TrustedDigest, input: string): seq[string] =
+  return @[]
```
kdeme, line 69: Not needed?

fluffy/tests/state_network_tests/test_state_network.nim
```diff
@@ -6,16 +6,24 @@
...
+  eth/keys,
+  eth/common/[eth_types, eth_hash],
   eth/p2p/discoveryv5/protocol as discv5_protocol, eth/p2p/discoveryv5/routing_table,
-  ../../nimbus/[config, db/core_db, db/state_db],
-  ../../nimbus/common/[chain_config, genesis],
-  ../network/wire/[portal_protocol, portal_stream],
-  ../network/state/[state_content, state_network],
-  ../database/content_db,
-  ./test_helpers
+  ../../tools/state_bridge/state_bridge,
```
kdeme, line 17: I don't like it that much that it need to import state_bridge just for the json parsing. Would be nicer if that json parsing is somewhere common for both tests & bridge.

fluffy/tools/state_bridge/state_bridge.nim
```diff
@@ -0,0 +1,99 @@
...
+    echo "Backfilling...", config.rpcAddress, ":", config.rpcPort
+    await portalRpcClient.connect(config.rpcAddress, Port(config.rpcPort), false)
+    let files = collect(for f in walkDir(config.dataDir): f.path)
+    for file in files:
+      let
+        content = readAllFile(file).valueOr:
+          echo "Skipping file ", file, " because of error \n", error
```
kdeme, line 58: use chronicles logs instead of echo (it's a mix of both right now)

fluffy/tests/state_network_tests/test_state_content.nim
```diff
@@ -0,0 +1,61 @@
...
+
+    let
+      addressBytes = hexToByteArray[20](address)
+      stateRootBytes = hexToByteArray[sizeof(state_content.AccountTrieProofKey.stateRoot)](stateRoot)
+      key = AccountTrieProofKey(address: addressBytes, stateRoot: stateRootBytes)
+
+    let encodedKey = SSZ.encode(key)
+    check encodedKey.to0xHex() == hexKey
+    let decodedKey = decodeSsz(encodedKey, AccountTrieProofKey)
+    check decodedKey.isOk()
```
kdeme, line 36: We already have a set of content key tests using the portal spec repo test vectors: https://github.com/status-im/nimbus-eth1/blob/master/fluffy/tests/portal_spec_tests/mainnet/test_state_content.nim

I would keep this file just for the content item encodings.

fluffy/tests/state_network_tests/test_state_content.nim
```diff
@@ -0,0 +1,61 @@
+# Fluffy
+# Copyright (c) 2021-2023 Status Research & Development GmbH
```
kdeme, line 2:
```suggestion
# Copyright (c) 2023 Status Research & Development GmbH
```

kdeme:
The testutp failure seems to be unrelated to this PR and due to a docker image update of quic-network-simulator

PR #1951, opened 2023-12-20
Portal State Proof Verification with Tests

web3-developer:
This PR only covers implementation of the state proof verification with tests. It doesn't yet cover the generation of proofs based on touched account state after executing a block. This part can be covered in a separate PR.

Note that the portal network specs are preliminary and still in a state of flux. The proof format used in this change follows the common format for a single Merkle Patrica Trie proof which is basically an list of RLP serialized MPT nodes, starting from the root node and ending with the leaf node if it exists. The code supports successful verification of proofs showing the existence of nodes as well as missing leafs. 

The tests read in a list of genesis files and then for each, generate account and storage states using the eth/trie library (in memory), then for each account and storage slot generates a proof then checks the proof.


fluffy/network/state/experimental/state_proof_generation.nim
```diff
@@ -0,0 +1,30 @@
+# Nimbus
+# Copyright (c) 2023 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [].}
+
```
web3-developer, line 9: This is just a temporary solution for generating the single MPT proofs. We likely will use something different later when looking at the touched account state.

fluffy/network/state/experimental/state_proof_verification.nim
```diff
@@ -0,0 +1,67 @@
...
+  
+  let key = toSeq(keccakHash(address).data)
+  let value = rlp.encode(account)
+
+  let proofResult = verifyMptProof(proof.MptProof, trustedStateRoot, key, value)
+
+  case proofResult.kind
+  of ValidProof:
+    ok()
+  of MissingKey:
```
web3-developer, line 35: Do we need to support validating the absence of a value? Removing it might reduce the risk of verification bugs.

fluffy/tests/test_helpers.nim
```diff
@@ -108,3 +112,34 @@ func buildHeadersWithProof*(
...
+
+  cn.genesis.alloc
+
+proc toState*(alloc: GenesisAlloc): 
+    (AccountState, Table[EthAddress, StorageState]) {.raises: [RlpError].} =
+  var accountTrie = initHexaryTrie(newMemoryDB())
+  var storageStates = initTable[EthAddress, StorageState]()
+
+  for address, genAccount in alloc:
+    var account = newAccount(genAccount.nonce, genAccount.balance)
```
kdeme, line 129: Nothing functionally wrong here (I think), but we generally prefer to use `let` and construct immediately the full object to its correct values when possible (see also https://status-im.github.io/nim-style-guide/language.objconstr.html). This avoids possible introduction of half initialized objects whether now or potentially in the future when code gets altered.

I do realize that the existing `newAccount` func is quite misleading on this. While I don't know the history of this code, it does seem to be mostly used for creating actual new (fresh) accounts.

In any case, I think here we can move the account creation after storage root creation, and in case `genAccount.code.len() == 0`  one `keccakHash`  for the codeHash can also be avoided this way?

fluffy/tests/test_helpers.nim
```diff
@@ -108,3 +112,34 @@ func buildHeadersWithProof*(
...
+
+proc getGenesisAlloc*(filePath: string): GenesisAlloc =
+  var cn: NetworkParams
+  if not loadNetworkParams(filePath, cn):
+    quit(1)
+
+  cn.genesis.alloc
+
+proc toState*(alloc: GenesisAlloc): 
+    (AccountState, Table[EthAddress, StorageState]) {.raises: [RlpError].} =
```
kdeme, line 124: What raises the `RlpError`? Does it come from the `storageTrie.put` and `accountTrie.put`?

fluffy/tests/test_state_proof_verification.nim
```diff
@@ -0,0 +1,179 @@
...
+proc checkValidProofsForExistingLeafs(
+    genAccounts: GenesisAlloc, 
+    accountState: AccountState, 
+    storageStates: Table[EthAddress, StorageState]) {.raises: [KeyError, RlpError].} = 
+
+  for address, account in genAccounts:
+    var acc = newAccount(account.nonce, account.balance)
+    acc.codeHash = keccakHash(account.code)
+    let codeResult = verifyContractBytecode(acc.codeHash, account.code)
+    #echo codeResult
```
kdeme, line 29: debugging leftovers here and further down?

fluffy/network/state/experimental/state_proof_verification.nim
web3-developer: The MissingKey scenario supports proving that a leaf value doesn't exist in the state. The proof needs to be generated from a valid state where the account doesn't exist. I'm guessing that we will end up removing this but I guess we can wait and see until the spec if finalized. Happy to leave it in for now until we know.

fluffy/network/state/experimental/state_proof_verification.nim
```diff
@@ -0,0 +1,67 @@
...
+  of InvalidProof:
+    err(proofResult.errorMsg)
+
+func verifyContractBytecode*(
+    trustedCodeHash: KeccakHash, 
+    bytecode: openArray[byte]): Result[void, string] =
+  if trustedCodeHash == keccakHash(bytecode):
+    ok()
+  else:
+    err("hash of bytecode doesn't match the expected code hash")
```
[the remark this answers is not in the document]
web3-developer: Do you mean that I need to add a newline?

fluffy/tests/test_helpers.nim
```diff
@@ -7,10 +7,14 @@
...
 import
   stew/shims/net,
-  eth/[keys, rlp],
+  eth/[common, keys, rlp, trie, trie/db],
   eth/p2p/discoveryv5/[enr, node, routing_table],
   eth/p2p/discoveryv5/protocol as discv5_protocol,
-  ../network/history/[accumulator, history_content]
+  ../network/history/[accumulator, history_content],
+  ../network/state/experimental/state_proof_generation,
+  ../../nimbus/[config, db/core_db],
```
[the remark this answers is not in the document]
web3-developer: Good catch. Looks like it's not needed. I've now removed it.

fluffy/tests/test_helpers.nim
web3-developer: Yes thats right. It looks like this file doesn't include a `{.push raises: [].}` at the top of file so the raises declaration can be removed I guess.

fluffy/tests/test_helpers.nim
```diff
@@ -108,3 +112,34 @@ func buildHeadersWithProof*(
...
+        let value = rlp.encode(slotValue)
+        storageTrie.put(key, value)
+      storageStates[address] = storageTrie.StorageState
+      account.storageRoot = storageTrie.rootHash()
+
+    let key = keccakHash(address).data
+    let value = rlp.encode(account)
+    accountTrie.put(key, value)
+
+  (accountTrie.AccountState, storageStates)
```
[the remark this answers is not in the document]
web3-developer: Will add the newline if that's what you meant.

fluffy/tests/test_state_proof_verification.nim
web3-developer: Yep, good point. Will remove.

fluffy/tests/test_helpers.nim
web3-developer: Sure, I've updated the code to create the account after the storage root. I've also updated the code hashing so that we default to the empty hash value and only hash the bytecode if is is not empty.

fluffy/network/state/experimental/state_proof_verification.nim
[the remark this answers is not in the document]
kdeme: Yes, github was indicating the "no newline" marker.

fluffy/tests/test_helpers.nim
kdeme: > Yes thats right.

Ok, I've noticed that before and it always seemed weird to me that those `put` calls could raise `RlpError`. In fact, in history_network code I added an assertion on that. Something to investigate better at some point.


> It looks like this file doesn't include a `{.push raises: [].}` at the top of file so the raises declaration can be removed I guess.

You can keep the declaration there.

We should add `{.push raises: [].}` and whichever additional ones also in test code, to set the good example :)

fluffy/network/state/experimental/state_proof_verification.nim
kdeme: ok, lets leave it as is for now and we can iterate over it as we see fit when we get to actually use it and/or when the specification gets more developed.

fluffy/tests/test_state_proof_verification.nim
```diff
@@ -0,0 +1,172 @@
+# Nimbus
```
kdeme, line 1: I just realized that this test is not yet added to be run under `make fluffy-test`. You need to add it to the `all_fluffy_tests.nim` else it won't actually be tested in CI.

web3-developer: Thanks, will do.

PR #1958, opened 2023-12-29
Initial implementation of block witness verification.

web3-developer:


fluffy/network/state/experimental/state_proof_types.nim
```diff
@@ -0,0 +1,39 @@
...
+# Copyright (c) 2023 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [].}
+
+import eth/[common, trie]
+
```
web3-developer, line 11: I moved the proof types into this file.

fluffy/network/state/experimental/state_proof_types.nim
```diff
@@ -0,0 +1,39 @@
...
+  StorageProof* = distinct MptProof
+
+  Witness* = seq[byte]
+
+  AccountData* = object
+    account*: Account
+    code*   : seq[byte]
+    storage*: TableRef[UInt256, UInt256]
+
+
```
web3-developer, line 28: Borrowing the below procs so that they can be used on the AccountState and StorageState types which are distinct from HexaryTrie.

fluffy/network/state/experimental/state_proof_verification.nim
```diff
@@ -35,8 +33,7 @@ proc verifyAccount*(
   of ValidProof:
     ok()
   of MissingKey:
-    # For an account that doesn't exist yet, which is fine.
-    ok()
+    err("missing key")
```
web3-developer, line 36: Fixing a bug here which could have in theory lead to storing invalid leafs in the portal state.

fluffy/network/state/experimental/state_proof_verification.nim
```diff
@@ -55,15 +52,67 @@ proc verifyContractStorageSlot*(
...
+          storage[slotKey] = slotValue
+
+    accounts[key.address] = AccountData(
+        account: account,
+        code: code,
+        storage: storage)
+
+  return accounts
+
+proc verifyWitness*(
```
web3-developer, line 94: This new proc supports verifying a block witness against a trusted state root and then returning the data embedded in the witness as a table of AccountData which contains accounts, code, and storage slots which were updated a part of running a block of transactions.

Unit tests are still in progress. The block witness is likely the best format for transferring the data from Nimbus to the bridge node (ignoring Verkle tries for now). Sending individual MPT proofs for each updated account, code and storage slot would likely lead to duplicated interim nodes being sent and therefore would use more space.

fluffy/network/state/experimental/state_proof_verification.nim
```diff
@@ -55,15 +52,67 @@ proc verifyContractStorageSlot*(
...
+    trustedStateRoot: KeccakHash,
+    witness: Witness): Result[TableRef[EthAddress, AccountData], string] =
+
+  let db: CoreDbRef = newCoreDbRef(LegacyDbMemory)
+  var tb = initTreeBuilder(witness, db, {wfEIP170}) # what flags to use here?
```
web3-developer, line 99: Wasn't sure which flags we should set here?

fluffy/network/state/experimental/state_proof_verification.nim
```diff
@@ -55,15 +52,67 @@ proc verifyContractStorageSlot*(
...
+  except Exception as e:
+    err(e.msg)
+
+
+
+
+
+
+
+
```
web3-developer, line 118: I will remove this whitespace in the next update.

stateless/tree_from_witness.nim
```diff
@@ -35,7 +35,7 @@ type
     db: DB
     root: KeccakHash
     flags: WitnessFlags
-    keys: seq[AccountAndSlots]
+    keys*: seq[AccountAndSlots]
```
web3-developer, line 38: This minor change was required to get access to the account and storage keys collected after building the in memory trie from the block witness.

fluffy/network/state/experimental/state_proof_types.nim
```diff
@@ -0,0 +1,39 @@
...
+  MptProof* = seq[seq[byte]]
+  AccountProof* = distinct MptProof
+  StorageProof* = distinct MptProof
+
+  Witness* = seq[byte]
+
+  AccountData* = object
+    account*: Account
+    code*   : seq[byte]
+    storage*: TableRef[UInt256, UInt256]
```
web3-developer, line 26: One thing I wasn't sure about is whether to use ref (heap) or stack storage for the above types. I understand that Nim by default uses value semantics so most types will be copied on assignment by default. For larger objects this isn't ideal so my reasoning here was to use TableRef to prevent copying. Not sure what the best practices are for Nim regarding this topic? Are there any best practices that I should follow?

kdeme: See here about ref usage, it depends, see more here: https://status-im.github.io/nim-style-guide/language.refobject.html

We do typically prefer to make it ref at usage, but we do have a lot of code not (yet?) doing that.

kdeme (commented):
Looks good in general, but with one important remark which is that Portal/Fluffy won't use full block witnesses (the multiproofs), at least not in any current planning. So I'm not fully sure if the location for this code is appropriate. 

Fluffy/Portal will however need similar generation of the proofs.

fluffy/network/state/experimental/state_proof_types.nim
web3-developer: Thanks for sharing that. I've now changed the storage field to a Table instead of TableRef because I think its cleaner to have the full object behave as a value type.

web3-developer:
> Looks good in general, but with one important remark which is that Portal/Fluffy won't use full block witnesses (the multiproofs), at least not in any current planning. So I'm not fully sure if the location for this code is appropriate.
> 
> Fluffy/Portal will however need similar generation of the proofs.

Thanks. Yes I understand what you mean, the portal network proofs are single MPT proofs not block witnesses. I guess I was thinking that either Fluffy or the bridge node could use the block witness data to update the touched account state. We talked about needing a solution for this before. Do you have any thoughts on how the bridge will work in terms of getting the state from Nimbus into the portal network?

I guess we have a few questions to think through as a team:
- Will Fluffy, the bridge and Nimbus all run as separate processes/binaries?
- If so how will the state data get in the portal network? For this we need to cover the two main scenarios; generating proofs for all accounts, and generated proofs for only touched account state.

kdeme:
> 
> I guess we have a few questions to think through as a team:
> 
>     * Will Fluffy, the bridge and Nimbus all run as separate processes/binaries?
> 
>     * If so how will the state data get in the portal network?

Yes, I think the preferred architecture would be:

```mermaid
flowchart LR
    Nimbus-EL<-- custom json-rpc -->Bridge-- Portal json-rpc -->Fluffy 
```
This allows for more flexibility in setup (e.g. re-use an existing EL node, bridge, etc.). And, in theory allows for more client interoperability/compatibility. Although, on the EL side it is unlikely that we get this custom json-rpc easily added in other clients until Portal is a proven light client for state data.

But in general, I think it makes sense to keep the bridge as light as possible, and rely on "external" ways of getting the data (full node, file dumps, etc.). It is what we have been doing with the history & beacon bridge also.


Whether we want to send the individual single proofs or the multiproof (block witness), I'm not sure.
One puts less work on the bridge side, while the other saves more on data size. I guess we could experiment on how easy it is to move from one to the other.

kdeme (approved):

PR #1959, opened 2023-12-30
Verify account mod for wire data

mynameisdaniil:
So, the idea behind this PR is to adapt current implementation of the proof verification to the one that is better mapped to the real data. This is pretty much an open discussion and work in progress, but I feel we have to steer development into this direction.

### DIFF
--- a/fluffy/network/state/state_network.nim
+++ b/fluffy/network/state/state_network.nim
@@ -14,7 +14,8 @@ import
   ../../database/content_db,
   ../wire/[portal_protocol, portal_stream, portal_protocol_config],
   ./state_content,
-  ./state_distance
+  ./state_distance,
+  ./state_proof_verification
 
 logScope:
   topics = "portal_state"
@@ -83,16 +84,9 @@ proc validateContent(
         return false
       let
         proof = decodedProof.asSeq().map((p: ByteList) => p.toSeq())
-        trieKey = keccakHash(key.accountTrieProofKey.address).data.toSeq()
-        value = proof[^1].decode(seq[seq[byte]])[^1]
         stateRoot = MDigest[256](data: key.accountTrieProofKey.stateRoot)
-        verificationResult = verifyMptProof(proof, stateRoot, trieKey, value)
-      case verificationResult.kind:
-        of ValidProof:
-          true
-        else:
-          warn "Received invalid account trie proof"
-          false
+        verificationResult = verifyAccount(stateRoot, key.accountTrieProofKey.address, AccountProof(proof))
+      verificationResult.isOk()
     of contractStorageTrieProof:
       true
     of contractBytecode:
--- a/fluffy/network/state/state_proof_verification.nim
+++ b/fluffy/network/state/state_proof_verification.nim
@@ -5,8 +5,6 @@
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-{.push raises: [].}
-
 import
   std/sequtils,
   stint,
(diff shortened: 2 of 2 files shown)


fluffy/network/state/state_proof_verification.nim
```diff
@@ -5,10 +5,8 @@
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-{.push raises: [].}
```
mynameisdaniil, line 8: Or should we keep it and handle wrap `decode` into try/catch?

fluffy/network/state/state_proof_verification.nim
```diff
@@ -17,19 +15,19 @@ export results
 
 type
   MptProof = seq[seq[byte]]
-  AccountProof* = distinct MptProof
+  AccountProof* = distinct MptProof # I'm not really sure how I feel about this
```
mynameisdaniil, line 18: I'm not really sure what's the purpose of making this distinct

fluffy/network/state/state_proof_verification.nim
```diff
@@ -17,19 +15,19 @@ export results
...
+    proof: AccountProof
+    ): Result[void, string] =
   
-  let key = toSeq(keccakHash(address).data)
-  let value = rlp.encode(account)
-
-  let proofResult = verifyMptProof(proof.MptProof, trustedStateRoot, key, value)
+  let 
+    key = keccakHash(address).data.toSeq()
+    value = proof.MptProof[^1].decode(seq[seq[byte]])[^1] # TODO this is undefined in spec. Update when the specs gets updated
```
mynameisdaniil, line 29: I'm not sure about this line. And the current draft doesn't specify how this is going to be encoded. But this is how it is encoded in data provided by Milos, so it is probably how this will end up in the spec.

mynameisdaniil:
Also, I've moved `state_proof_verification` out of experimental.
Also also, I'm not sure what to do with tests written by Ben. We can probably modify them to still pass (I'd ask Ben to help me with that), but what is the point of custom genesis when we have test vectors based on real data?

mynameisdaniil:
I'll close it for now, as I discovered spec updates that I wasn't aware of.

fluffy/network/state/state_proof_verification.nim
web3-developer: I believe the best practice is to keep these in every file so that we correctly deal with exceptions. Generally the result type is preferred based on what I've read so we should try to catch exceptions and wrap into a Result error.

fluffy/network/state/state_proof_verification.nim
web3-developer: My reasoning was to enforce strictness using the type system. Making it distinct makes it behave like a separate type that can only be used with the API that support validation of AccountProofs. It can be converted to a MptProof if required if there are exceptions.

web3-developer:
> Also, I've moved `state_proof_verification` out of experimental. Also also, I'm not sure what to do with tests written by Ben. We can probably modify them to still pass (I'd ask Ben to help me with that), but what is the point of custom genesis when we have test vectors based on real data?

The reason for putting the code in the experimental folder was because the spec is still preliminary and likely to change. I would suggest leaving it there until we get a more concrete spec defined.

fluffy/network/state/state_proof_verification.nim
web3-developer: I would suggest leaving this code as is for now and doing this part in the calling code above. You can just get the last element of the proof then do rlp.decode. That way we don't need to worry about refactoring tests now when the spec isn't yet fully defined. In the future we could refactor as needed.

fluffy/network/state/state_network.nim
```diff
@@ -83,16 +84,9 @@ proc validateContent(
         return false
       let
         proof = decodedProof.asSeq().map((p: ByteList) => p.toSeq())
-        trieKey = keccakHash(key.accountTrieProofKey.address).data.toSeq()
-        value = proof[^1].decode(seq[seq[byte]])[^1]
         stateRoot = MDigest[256](data: key.accountTrieProofKey.stateRoot)
```
web3-developer, line 87: I would suggest using the KeccakHash type. It makes it more clear what the type of hash is.

PR #1960, opened 2024-01-08
Implementation of Nimbus eth_getProof RPC Endpoint.

web3-developer:
This PR includes changes to implement the eth_getProof RPC endpoint implemented as follows:
- The account and storage proofs contain an array of RLP encoded merkle patrica trie nodes starting from the root node, following the path of the trie key until reaching the leaf node. The leaf node is also included as the final node in the proof.
- When getting an account that doesn't exist in the state we return values set with defaults. Balance, nonce, codeHash and storageHash will be zero in this case.
- When returning an account that doesn't exist in the state we also return a proof that proves the absence of the value in the state. In this case the proof will contain all nodes that do exist following the path of the provided key.
- If an existing account is not a contract account then the code hash will be equal to the keccak256 hash of an empty string.
- If an account has no storage slots populated then it will have a storageRoot equal to the keccak256 hash of the empty root node.
- If the slots array parameter is empty then we return an empty list for the storageProof.
- If we supply a non empty array of slots as a parameter then we return a storage proof for each non existent slot with the slot value equal to zero. These proofs will prove that the slot doesn't exist in the state. If the storage state is empty then the proof will be empty. If the storage state is not empty then the proof will contain all rlp nodes following the path of the slot key.

Here is the link to the spec:  https://eips.ethereum.org/EIPS/eip-1186. Note that the spec is in the 'Stagnant' state so it was never moved to final.


nimbus/db/distinct_tries.nim
```diff
@@ -1,5 +1,5 @@
 # Nimbus
-# Copyright (c) 2023 Status Research & Development GmbH
+# Copyright (c) 2024 Status Research & Development GmbH
```
web3-developer, line 2: Is this the correct way to update the copyright notice? Or should it be 2023-2024?

nimbus/db/state_db.nim
```diff
@@ -29,6 +29,8 @@ export
   isDeadAccount,
   isEmptyAccount,
   newAccountStateDB,
-  rootHash
+  rootHash,
+  getAccountProof,
+  getStorageProof
```
web3-developer, line 34: Exporting the new proof procs to be used with the ReadOnlyStateDB type.

nimbus/db/trie_get_branch.nim
```diff
@@ -0,0 +1,94 @@
...
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or
+#    http://www.apache.org/licenses/LICENSE-2.0)
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT) or
+#    http://opensource.org/licenses/MIT)
+# at your option. This file may not be copied, modified, or distributed except
+# according to those terms.
+
+# This implementation of getBranch on the CoreDbPhkRef type is a temporary solution
+# which can be removed once we get an equivient proc defined on the CoreDbPhkRef type
+# in the db layer.
```
web3-developer, line 13: This code was copied from the HexaryTrie library in order to implement getBranch on the CoreDbPhkRef type.

nimbus/db/state_db/base.nim
```diff
@@ -249,6 +253,29 @@ proc isDeadAccount*(db: AccountStateDB, address: EthAddress): bool =
   else:
     result = true
 
+proc removeEmptyRlpNode(branch: var seq[MptNodeRlpBytes]) =
+  if branch.len() == 1 and branch[0] == emptyRlp:
+    branch.del(0)
```
web3-developer, line 258: The getBranch code returns a single node with a value of emptyRlp (zero encoded as rlp) if the trie state is empty. We remove this because this isn't a valid proof and causes the proof verification to fail. Also Geth returns an empty list in this case so we do the same.

nimbus/rpc/p2p.nim
```diff
@@ -503,6 +503,49 @@ proc setupEthRpc*(
...
+            balance: acc.balance,
+            nonce: w3Qty(acc.nonce),
+            codeHash: w3Hash(acc.codeHash),
+            storageHash: w3Hash(acc.storageRoot),
+            storageProof: storage)
+    else:
+      ProofResponse(
+            address: w3Addr(address),
+            accountProof: seq[RlpEncodedBytes](accountProof),
+            storageProof: storage)
```
web3-developer, line 547: We return zero values (default values) for the account if the account doesn't exist, including the codeHash and storageHash.

nimbus/db/distinct_tries.nim
jangko: 2023-2024

jangko:
looks good to me

kdeme (approved):

nimbus/db/distinct_tries.nim
web3-developer: Thanks, I'll update in that case.

web3-developer:
> looks good to me

Great. Planning to merge it after I update the copyright notices.

PR #1961, opened 2024-01-09
Fix vmflags for witness generation 1934.

web3-developer:
Changes included in this PR:
- Fix issue where vmflags were being reset in the middle of processBlocksImpl. I needed these vmflags to work so I could enable witness generation in the test.
- Turned on witness generation and verification in the test_blockchain_json test. The witness generation was previously disabled. Note that witnesses generation is still disabled by default (in production) but can be enabled by setting the GenerateWitness vmflag.
- Fixed a minor issue where witness generation would fail for blocks with no transactions. The code has been updated to return an empty witness for a block that has no transactions. This should almost never happen in practice I'm guessing but probably a good idea to cover anyway.

### DIFF
--- a/tests/test_blockchain_json.nim
+++ b/tests/test_blockchain_json.nim
@@ -176,6 +176,12 @@ proc parseTestCtx(fixture: JsonNode, testStatusIMPL: var TestStatus): TestCtx =
 proc blockWitness(vmState: BaseVMState, chainDB: CoreDbRef) =
   let rootHash = vmState.stateDB.rootHash
   let witness = vmState.buildWitness()
+
+  if witness.len() == 0:
+    if vmState.stateDB.makeMultiKeys().keys.len() != 0:
+      raise newException(ValidationError, "Invalid trie generated from block witness")
+    return
+
   let fork = vmState.fork
   let flags = if fork >= FKSpurious: {wfEIP170} else: {}
 
@@ -221,6 +227,7 @@ proc importBlock(ctx: var TestCtx, com: CommonRef,
       com,
       tracerInst,
     )
+    ctx.vmState.generateWitness = true # Enable saving witness data
 
   let
     chain = newChain(com, extraValidation = true, ctx.vmState)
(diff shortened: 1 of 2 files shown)


tests/test_blockchain_json.nim
```diff
@@ -176,6 +176,12 @@ proc parseTestCtx(fixture: JsonNode, testStatusIMPL: var TestStatus): TestCtx =
 proc blockWitness(vmState: BaseVMState, chainDB: CoreDbRef) =
   let rootHash = vmState.stateDB.rootHash
   let witness = vmState.buildWitness()
+
+  if witness.len() == 0:
+    if vmState.stateDB.makeMultiKeys().keys.len() != 0:
+      raise newException(ValidationError, "Invalid trie generated from block witness")
+    return
```
web3-developer, line 183: Returning here because we can't validate an empty witness. The witness should only be empty if the multikeys are empty due to having no transactions in a block.

tests/test_blockchain_json.nim
```diff
@@ -221,6 +227,7 @@ proc importBlock(ctx: var TestCtx, com: CommonRef,
       com,
       tracerInst,
     )
+    ctx.vmState.generateWitness = true # Enable saving witness data
```
web3-developer, line 230: I'm enabling this for all tests. Will this be a problem because its not covering the default code path where witness generation is disabled? If so I can create a separate test which has the witness generation enabled and leave this one as is.

I guess it depends if there are already other tests already covering this.

kdeme: I'm assuming this is just an additional code path, and the default is still taken too?

But I don't know this code well at all, @jangko can probably answer better.

jangko: It's ok. Like @kdeme said, the witness generation is is just additional code path. And `test_blockchain_json` actually don't care about block witness. I piggyback the block witness test in `test_blockchain_json` because the required code is simply there, instead of crafting another similar test code just to test block witness.

jangko (approved):

tests/test_blockchain_json.nim
web3-developer: Ok thanks for the explanation. I'll leave it as is in that case.

PR #1962, opened 2024-01-10
Fix clashing import of AccountProof type.

web3-developer:


fluffy/network/state/experimental/state_proof_verification.nim
```diff
@@ -14,7 +14,7 @@ import
   stew/results,
   ./state_proof_types,
   ../../../../stateless/[tree_from_witness, witness_types],
-  ../../../../nimbus/db/[core_db, state_db, state_db/base]
+  ../../../../nimbus/db/[core_db, state_db]
```
web3-developer, line 17: state_db/base contains another type named 'AccountProof' which is used in the Nimbus getProof RPC endpoint.

fluffy/network/state/experimental/state_proof_verification.nim
```diff
@@ -77,7 +77,7 @@ proc buildAccountsTableFromKeys(
   for key in keys:
     let account = db.getAccount(key.address)
     let code = if key.codeLen > 0:
-        db.AccountStateDB.kvt().get(account.codeHash.data)
+        db.getTrie().parent().kvt().get(account.codeHash.data)
```
web3-developer, line 80: Now just using the types defined for the ReadOnlyStateDB type from the state_db import to avoid using AccountStateDB directly.

kdeme (approved):

PR #1971, opened 2024-01-15
Nimbus light client integration with status-go rebased version

vitvly:


arnetheduck:
This version still looks like it's missing the separate library "main" file, suggested here: https://github.com/status-im/nimbus-eth1/pull/1572#issuecomment-1838885345

With that approach, we can also un-declare metrics for that file specifically and the rest of the nimbus-eth1 project is not affected.

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -39,14 +41,22 @@ func getConfiguredChainId(networkMetadata: Eth2NetworkMetadata): Quantity =
   else:
     return networkMetadata.cfg.DEPOSIT_CHAIN_ID.Quantity
 
-proc run(config: VerifiedProxyConf) {.raises: [CatchableError].} =
+proc run*(config: VerifiedProxyConf, callbacks: varargs[OnHeaderCallback]) {.raises: [CatchableError], gcsafe.} =
```
arnetheduck, line 44:
```suggestion
proc run*(config: VerifiedProxyConf, headerCallback: OnHeaderCallback = nil) {.raises: [CatchableError], gcsafe.} =
```

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -167,6 +177,13 @@ proc run(config: VerifiedProxyConf) {.raises: [CatchableError].} =
       when lcDataFork > LightClientDataFork.None:
         info "New LC finalized header",
           finalized_header = shortLog(forkyHeader)
+        if headerCallback != nil:
+          notice "### Invoking finalizedHeaderCallback"
```
arnetheduck, line 181: debug leftover?

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -253,10 +277,9 @@ proc run(config: VerifiedProxyConf) {.raises: [CatchableError].} =
   while true:
     poll()
 
-when isMainModule:
+when isMainModule and appType != "lib":
```
arnetheduck, line 280: no longer needed

nimbus_verified_proxy/nimbus_verified_proxy.nim
vitvly: Replaced it with `ptr Context` param.

nimbus_verified_proxy/nimbus_verified_proxy.nim
vitvly: Removed.

nimbus_verified_proxy/nimbus_verified_proxy.nim
vitvly: Removed.

vitvly:
Rebased with latest master, added cleanup code.

vitvly:
Changed nim-confutils dependency to point to updated upstream.

kdeme (commented):
Not so much a review, but I'm more curious about the planned architecture here.

nimbus_verified_proxy is supposed to be the trustless EL json-rpc API drop-in replacement. But I noticed in this PR also that new optimistic and finalized headers are being passed along via a callback. What is the architectural idea here exactly?

vitvly:
For now the headers are only used to extract block roots in status-go, in order to persist sync progress. Otherwise not needed (yet?).

kdeme:
> For now the headers are only used to extract block roots in status-go, in order to persist sync progress. Otherwise not needed (yet?).

Do you mean, so that when restarting, you can provided a much newer block root, to have a very short sync?

If so, this is a feature that I've been wanting to add to the nimbus_verified_proxy.

vitvly:
> > For now the headers are only used to extract block roots in status-go, in order to persist sync progress. Otherwise not needed (yet?).
> 
> Do you mean, so that when restarting you can provide a much newer block root in order to have a very short sync?
> 
> If so, this is a feature that I've been wanting to add to the nimbus_verified_proxy.

Yes, it'll check finalized headers for a `state_root` value and store it in status-go DB. Would be great to have it in nimbus_verified_proxy i think.

PR #1976, opened 2024-01-17
New state network types update

mynameisdaniil:
~~UPD: Do not waste time reviewing. WIP~~
This PR contains main types contained in current spec + some tests for serialization/de-serialization of said types.
Test vectors are still in the test files, since I believe we'll be modifying types, tests and everything else few more times before we arrive to the 'stable' spec. Then, I'll move test vectors in https://github.com/status-im/portal-spec-tests/
Also, I've removed some outdated tests. I'll re-introduce them later as/if they needed.


mynameisdaniil:
ContentKey selectors and test vectors going to be updated soon due to discussion in specs PR.

kdeme:
fyi I rebased as there were a lot of module bumps

fluffy/network/state/state_content.nim
```diff
@@ -49,52 +50,92 @@ type
...
     accountTrieNode = 0x20
-    contractStorageTrieNode = 0x21
-    accountTrieProof = 0x22
+    contractTrieNode = 0x21
+    contractCode = 0x22
+    # NOTE unused
     contractStorageTrieProof = 0x23
-    contractBytecode = 0x24
+    # NOTE unused
+    accountTrieProof = 0x24
```
kdeme, line 58: You can remove these types considering the functionality regarding them has also been removed. No more point of keeping them around. Same for the cases that use them.

fluffy/network/state/state_content.nim
```diff
@@ -49,52 +50,92 @@ type
...
+
+  NibblePair* = byte
+  Nibbles* = object
+    packedNibbles*: List[NibblePair, 32]
+    isOddLength*: bool
```
kdeme, line 63: Order seems reversed here vs spec.

fluffy/network/state/state_content.nim
```diff
@@ -49,52 +50,92 @@ type
...
+
+  WitnessNode* = List[byte, 1024]
```
kdeme, line 65: Could use the ByteList alias here also,  there is no differences ofcourse, but reads better when putting next to the spec.

```suggestion
  WitnessNode* = ByteList
```

fluffy/network/state/state_content.nim
```diff
@@ -122,46 +163,10 @@ template computeContentId*(digestCtxType: type, body: untyped): ContentId =
...
+  let idHash = sha2.sha256.digest(contentKey.asSeq())
+  readUintBE[256](idHash.data)
+
 func toContentId*(contentKey: ContentKey): ContentId =
-  case contentKey.contentType:
-  of unused:
-    raiseAssert "Should not be used and fail at decoding"
-  of accountTrieNode: # sha256(path | node_hash)
-    let key = contentKey.accountTrieNodeKey
-    computeContentId sha256:
```
kdeme, line 131: `template computeContentId` can also be removed as it is no longer used

fluffy/tests/state_network_tests/test_state_content.nim
```diff
@@ -9,36 +9,230 @@ import
   std/[os, json, sequtils],
   testutils/unittests,
   stew/[byteutils, io2],
-  eth/keys,
+  eth/keys, eth/common/eth_types,
```
kdeme, line 12: There should be no need to import this. The module that uses types from here in their public API should export those.

fluffy/tests/state_network_tests/test_state_content.nim
```diff
@@ -9,36 +9,230 @@ import
...
+      encoded = contentKey.encode()
+    # echo ">>>", $encoded
+    check $encoded == accountTrieNodeKeyEncoded
+
+    let decoded = encoded.decode().valueOr:
+      raiseAssert "Cannot decode AccountTrieNodeKey"
+    check decoded.contentType == accountTrieNode
+    check decoded.accountTrieNodeKey.nodeHash == NodeHash.fromHex(nodeHash)
+    check decoded.accountTrieNodeKey.path.isOddLength == isOddLength
+    check decoded.accountTrieNodeKey.path.packedNibbles == Nibbles.packedNibbles.init(packedNibbles)
```
kdeme, line 59: When doing many `check`s on consecutive lines it is my personal preference to write one `check:` 

Or/And, perhaps better, here just compare the full `accountTrieNodeKey`. No need to "rebuild" the values on the right side again.

fluffy/tests/state_network_tests/test_state_content.nim
```diff
@@ -9,36 +9,230 @@ import
...
+    let decoded = decode(encoded)
+
+    check decoded.isNone()
 
-    check decodedProof == accountTrieProof
+  test "Invalid key - empty input":
+    let encoded = ByteList.init(@[])
+    let decoded = decode(encoded)
 
+    check decoded.isNone()
```
kdeme, line 238: I think it would be good to put two test `suite`s, 1 for key and 1 for content. Or at least group these with the other keys.

fluffy/tests/state_network_tests/test_state_network.nim
```diff
@@ -1,270 +0,0 @@
...
-
-      check foundContent.isSome()
-
-      check decodeSsz(foundContent.get(), AccountTrieProof).isOk()
-
-    await node1.closeWait()
-    await node2.closeWait()
-
-
-  asyncTest "Test Share Full State":
```
kdeme, line 118: This test should be fairly easy to adapt and still use

fluffy/tests/state_network_tests/test_state_network.nim
```diff
@@ -1,270 +0,0 @@
...
-      check:
-        foundContent.isSome()
-
-      let hash = keccakHash(foundContent.get())
-      check hash.data == key
-
-    await node1.closeWait()
-    await node2.closeWait()
-
-  asyncTest "Find content in the network via content lookup":
```
kdeme, line 173: Same for this one

fluffy/tests/state_network_tests/test_state_content.nim
```diff
@@ -9,36 +9,230 @@ import
...
+      encoded = SSZ.encode(nibbles)
+    check encoded.toHex() == oddNibbles
+    # echo ">>>", encoded.toHex()
+
+  const accountTrieNodeKeyEncoded = "2024000000cafebabe000000000000000000000000000000000000000000000000000000000500000000123456789abc"
+  test "Encode/decode AccountTrieNodeKey":
+    let
+      packedNibbles = @[NibblePair 0x12, 0x34, 0x56, 0x78, 0x9A, 0xBC]
+      isOddLength = false
+      nodeHash = "cafebabe"
```
kdeme, line 45: Use the proper hash len here? It seems (from your result output) that the fromHex pads it with zeroes, but if you want to share this as test vector it will be confusing.

kdeme: You can also immediately do the `NodeHash.fromHex(nodeHash)` here (and make them const)

Basically use the deleted `test_state_content.nim` as example.

fluffy/network/state/state_content.nim
mynameisdaniil: The reason I went with this alias is to explicitly show max length. Also, ByteList is defined with max_length of 2048 which is not to the spec.
Also, I think it is just easier to read:
```
  NibblePair* = byte
  Nibbles* = object
    isOddLength*: bool
    packedNibbles*: List[NibblePair, 32]

  WitnessNode* = List[byte, 1024]
  Witness* = List[WitnessNode, 1024]
```
all the sizes and basic types are here

kdeme: > The reason I went with this alias is to explicitly show max length. Also, ByteList is defined with max_length of 2048 which is not to the spec.

Right, that ByteList is defined as ByteList(2048) in portal wire protocol. Ok then, I think I'll add a ByteList with limit alias to the ssz module.

kdeme (approved):

PR #1977, opened 2024-01-18
Experimental RPC endpoints for generating block witnesses

web3-developer:
This change adds two new custom RPC endpoints which may be used by the Portal Network bridge node to get the changed account state for each new block. It would be impractical to query Nimbus for every account and storage slot for every block so these endpoints allow us to get just the updated state and then feed it into the portal state network. Here are the interfaces for the new endpoints:
```
proc exp_getWitnessByBlockNumber(blockId: BlockIdentifier, statePostExecution: bool): seq[byte]
proc exp_getProofsByBlockNumber(blockId: BlockIdentifier, statePostExecution: bool): seq[ProofResponse]
```

The first endpoint returns a block witness which is a binary format which follows the spec here: https://github.com/ethereum/portal-network-specs/blob/01a49a8c9bf08121ecde1b9270a6f2f679cb2568/witness.md.
The second endpoint returns a list of proofs for accounts and storage slots in the same format as the eth_getProof endpoint except it returns a list instead of a single proof.

Both endpoints support returning the state from before or after executing the transactions in the block. Each of these options would be useful in different scenarios. For example, stateless block execution would require getting the block witness data from before execution of the transactions in order to execute the transactions against the witness. For the portal network we will likely want to get the list of proofs from after execution of the block because the bridge will simply be forwarding the proofs into the portal state network and it will want the latest updated state after execution.

The new endpoints are disabled by default and can be enabled by supplying the `--rpc-api=exp` flag. `exp` is the new JSON-RPC namespace which has been added for experimental endpoints.

This implementation doesn't yet store the block witnesses. It simply fetches the transactions from the requested block, then re-runs the transactions (without persisting to the db) in order to collect the keys of the updated account and storage state which are used to look up and return the account state from before or after the block execution. We only support returning witnesses/proofs for blocks that have been persisted to disk. I believe this is fine at least for now because the portal network only supports feeding in data from the canonical chain, therefore feeding in block data that may be a from a forks etc probably won't be required. Storage of block witnesses in the database will be coming next perhaps in a separate PR.


fluffy/network/state/experimental/state_proof_types.nim
```diff
@@ -18,14 +18,6 @@ type
   AccountProof* = distinct MptProof
   StorageProof* = distinct MptProof
 
-  BlockWitness* = seq[byte]
-
-  AccountData* = object
-    account*: Account
-    code*   : seq[byte]
-    storage*: Table[UInt256, UInt256]
-
```
web3-developer, line 27: Moved into the stateless folder because it likely won't be used by Fluffy and I'm now using these types in Nimbus.

fluffy/network/state/experimental/state_proof_verification.nim
```diff
@@ -67,50 +65,3 @@ func verifyContractBytecode*(
     ok()
   else:
     err("hash of bytecode doesn't match the expected code hash")
-
```
web3-developer, line 70: Moved into the stateless folder because it likely won't be used by Fluffy and I'm now using these types in Nimbus.

fluffy/tests/test_state_proof_verification.nim
```diff
@@ -214,26 +166,4 @@ suite "State Proof Verification Tests":
     for file in genesisFiles:
       let accounts = getGenesisAlloc("fluffy" / "tests" / "custom_genesis" / file)
       var state = accounts.toState()
-      checkInvalidProofsWithBadValue(accounts, state[0], state[1])
```
web3-developer, line 217: Tests moved into stateless folder.

nimbus/core/executor/process_block.nim
```diff
@@ -175,12 +176,13 @@ proc processBlock*(
...
-  # block will still get a pruned trie. If trie pruning deactivated,
-  # `applyDeletes` have no effects.
-  dbTx.commit(applyDeletes = false)
+  if commit:
+    # `applyDeletes = false`
+    # If the trie pruning activated, each of the block will have its own state
+    # trie keep intact, rather than destroyed by trie pruning. But the current
+    # block will still get a pruned trie. If trie pruning deactivated,
+    # `applyDeletes` have no effects.
+    dbTx.commit(applyDeletes = false)
```
web3-developer, line 185: I'm calling processBlock from the new RPC endpoints with commit = false.

nimbus/evm/state.nim
```diff
@@ -290,10 +290,10 @@ proc `generateWitness=`*(vmState: BaseVMState, status: bool) =
   if status: vmState.flags.incl GenerateWitness
   else: vmState.flags.excl GenerateWitness
 
-proc buildWitness*(vmState: BaseVMState): seq[byte]
-    {.raises: [CatchableError].} =
+proc buildWitness*(
+    vmState: BaseVMState,
+    mkeys: MultikeysRef): seq[byte] {.raises: [CatchableError].} =
```
web3-developer, line 295: Added this new proc which supports getting a blockWitness by passing in a MultikeysRef. This was required in order to support returning a block witness from before execution of the transactions. Previously getting a block witness would always return the state of the updated accounts from after executing the block of transactions.

nimbus/evm/state.nim
```diff
@@ -304,6 +304,11 @@ proc buildWitness*(vmState: BaseVMState): seq[byte]
   var wb = initWitnessBuilder(vmState.com.db, rootHash, flags)
   wb.buildWitness(mkeys)
 
+proc buildWitness*(
+    vmState: BaseVMState): seq[byte] {.raises: [CatchableError].} =
+  let mkeys = vmState.stateDB.makeMultiKeys()
+  buildWitness(vmState, mkeys)
```
web3-developer, line 310: This is the existing proc refactored to use the new proc above.

nimbus/rpc/p2p.nim
```diff
@@ -29,6 +29,40 @@ type
   BlockHeader = eth_types.BlockHeader
   Hash256 = eth_types.Hash256
```
web3-developer, line 31: Refactored this code out of the RPC endpoint in order to make testing easier.

stateless/tree_from_witness.nim
```diff
@@ -196,12 +196,12 @@ proc writeCode(t: var TreeBuilder, code: openArray[byte]): Hash256 =
...
-proc accountNode(t: var TreeBuilder, depth: int): NodeKey
-proc accountStorageLeafNode(t: var TreeBuilder, depth: int): NodeKey
-proc hashNode(t: var TreeBuilder, depth: int, storageMode: bool): NodeKey
-proc treeNode(t: var TreeBuilder, depth: int = 0, storageMode = false): NodeKey
+proc branchNode(t: var TreeBuilder, depth: int, storageMode: bool): NodeKey {.gcsafe.}
+proc extensionNode(t: var TreeBuilder, depth: int, storageMode: bool): NodeKey {.gcsafe.}
+proc accountNode(t: var TreeBuilder, depth: int): NodeKey {.gcsafe.}
+proc accountStorageLeafNode(t: var TreeBuilder, depth: int): NodeKey {.gcsafe.}
+proc hashNode(t: var TreeBuilder, depth: int, storageMode: bool): NodeKey {.gcsafe.}
+proc treeNode(t: var TreeBuilder, depth: int = 0, storageMode = false): NodeKey {.gcsafe.}
```
web3-developer, line 204: I couldn't get my code to compile without adding these {.gcsafe.} pragmas for some reason. It might be related to the conditional compilation using global variables but I didn't get a chance to confirm. Not sure if this will be an issue or not.

stateless/witness_verification.nim
```diff
@@ -0,0 +1,74 @@
...
+        if slotExists:
+          storage[slotKey] = slotValue
+
+    accounts[key.address] = AccountData(
+        account: account,
+        code: code,
+        storage: storage)
+
+  return accounts
+
```
web3-developer, line 54: This proc is used to verify the witness data against a stateRoot and then return the data contained in the witness in a table.

tests/test_blockchain_json.nim
```diff
@@ -198,6 +199,36 @@ proc blockWitness(vmState: BaseVMState, chainDB: CoreDbRef) =
   if root != rootHash:
     raise newException(ValidationError, "Invalid trie generated from block witness")
```
web3-developer, line 201: Testing the getBlockWitness proc which contains the code to execute the block of transactions. Adding it here because we get full test coverage for all scenarios covered by the test_blockchain_json test.

nimbus/rpc/experimental.nim
```diff
@@ -0,0 +1,124 @@
...
+  doAssert processBlockResult == ValidationResult.OK
+
+  let mkeys = vmState.stateDB.makeMultiKeys()
+
+  if statePostExecution:
+    (vmState.stateDB.rootHash, vmState.buildWitness(mkeys), flags)
+  else:
+    # Reset state to what it was before executing the block of transactions
+    let initialState = BaseVMState.new(blockHeader, com)
+    (initialState.stateDB.rootHash, initialState.buildWitness(mkeys), flags)
```
web3-developer, line 59: We use the statePostExecution boolean flag to determine whether to return the witness using the initial state from before execution of the transactions or the state from after.

nimbus/rpc/experimental.nim
```diff
@@ -0,0 +1,124 @@
...
+
+proc getBlockProofs*(
+    accDB: ReadOnlyStateDB,
+    witnessRoot: KeccakHash,
+    witness: BlockWitness,
+    flags: WitnessFlags): seq[ProofResponse] {.raises: [RlpError].} =
+
+  if witness.len() == 0:
+    return @[]
```
web3-developer, line 68: Blocks with no transactions will generate an empty witness.

nimbus/core/executor/process_block.nim
jangko, line 185: Database 'beginTransaction' should always paired with 'dispose', 'commit', or 'rollback'. If you don't call them in pair, the db tx will stay in memory.

I think what you want to achive is add another layer of db transaction and then dispose it.

Example:
'''Nim
In your rpc method
Tx=Db.BeginTransaction
Do something
Tx.dispose
'''

tests/rpc/experimental_rpc_client.nim
```diff
@@ -0,0 +1,20 @@
...
+  json_rpc/errors,
+  web3/eth_api
+
+export
+  rpcclient,
+  errors,
+  eth_api_types,
+  conversions
+
+createRpcSigs(RpcClient, currentSourcePath.parentDir / "experimental_rpc_calls.nim")
```
jangko, line 20: Nim-json-rpc have a new set of create rpc signature without the need to load it from file. Try it out.😁

stateless/tree_from_witness.nim
jangko: It actually good to add gcsafe pragma. They are useful if you call them in a closure proc.

tests/test_rpc_experimental_json.nim
```diff
@@ -0,0 +1,200 @@
...
+      "block46400.json",
+      "block46402.json",
+      "block47205.json",
+      "block47216.json",
+      "block48712.json"
+      ]
+
+    let RPC_PORT = 8545
+    var
+      rpcServer = newRpcSocketServer(["127.0.0.1:" & $RPC_PORT])
```
jangko, line 109: For testing purpose, better to use port zero and let the OS choose the port for you. The new json-rpc library have server.localAddress to get the active port for your client. Doing so will prevent the next test in queue will not fail because of used port(if they use the same port).

jangko (changes requested):

nimbus/core/executor/process_block.nim
web3-developer: Yes good point. Thanks for catching this. I remember seeing that code in the eth_call RPC endpoint. I'll update.

tests/rpc/experimental_rpc_client.nim
web3-developer: Okay great, can you direct me to where I can see an example? Is Nimbus updated to use this new version?

tests/test_rpc_experimental_json.nim
web3-developer: Thanks. Sure, I'll update.

tests/rpc/experimental_rpc_client.nim
jangko: It is being used in nim-json-rpc test and in nim-web3: 'eth_api.nim' and 'engine_api.nim'.

nimbus/core/executor/process_block.nim
web3-developer: Hey @jangko, actually this processBlock proc has the following lines near the top:
```
  var dbTx = vmState.com.db.beginTransaction()
  defer: dbTx.dispose()
```
It looks like dispose() will roll back the transaction already in this proc unless I'm missing something?

web3-developer: @jangko If I add another transaction at a higher layer with a begin and then dispose, will the child transaction still be committed if it has a begin and commit?

web3-developer: I also noticed that when the transactions are being processed the vmState.stateDB.persist is called on the AccountsCache. Will this be a problem? Or is this just pushing it to the database to be handled by the db transaction at the higher layer?

jangko: > It looks like dispose() will roll back the transaction already in this proc unless I'm missing something?

It is designed to simplify the api, `dispose` will only rollback the transaction if there is exception raised and `commit` is not called. But when the transaction already commited, `dispose` is a noop. Well, actually it is a bit confusing, but we are too lazy to create a better api. It is a legacy from the beginning of nimbus project.

> If I add another transaction at a higher layer with a begin and then dispose, will the child transaction still be committed if it has a begin and commit? Or will the parent transaction prevent the child transaction from being committed?

When a database transaction started, it gives you guarantee, no children transaction will reach the database if the parent call `rollback` or `dispose`. The answer is no.
In `tracer.nim` you may see the process can be intercepted and diverted by a custom database to somewhere else. That is the only possible way to alter the database behavior. But as far as I know, block processing only use regular database transaction.

jangko: > I also noticed that when the transactions are being processed the vmState.stateDB.persist is called on the AccountsCache. Will this be a problem? Or is this just pushing it to the database to be handled by the db transaction at the higher layer?

It will be handled by db transaction at higher layer. There is no backdoor or bypass mechanism. Nim is good at performing some magic, but we don't use it for state db.

jangko (approved):
LGTM

nimbus/rpc/experimental.nim
```diff
@@ -0,0 +1,129 @@
...
+    com: CommonRef,
+    blockHeader: BlockHeader,
+    statePostExecution: bool): (KeccakHash, BlockWitness, WitnessFlags)
+    {.raises: [CatchableError].} =
+
+  let
+    chainDB = com.db
+    blockHash = chainDB.getBlockHash(blockHeader.blockNumber)
+    blockBody = chainDB.getBlockBody(blockHash)
+    vmState = BaseVMState.new(blockHeader, com)
```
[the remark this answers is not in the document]
web3-developer: Good question. I did have a think about this part and basically yes we would need to be running an archive node for this to work. If pruning is enabled (in the future) and you query for an older block which doesn't exist in the db then an RPC error would be returned. It's basically a similar scenario to the existing getBlockByNumber RPC endpoint. I did test this manually by running Nimbus locally and confirmed it returns an informative RPC error.

nimbus/rpc/experimental.nim
```diff
@@ -0,0 +1,129 @@
...
+
+type
+  BlockHeader = eth_types.BlockHeader
+  ReadOnlyStateDB = state_db.ReadOnlyStateDB
+
+proc getBlockWitness*(
+    com: CommonRef,
+    blockHeader: BlockHeader,
+    statePostExecution: bool): (KeccakHash, BlockWitness, WitnessFlags)
+    {.raises: [CatchableError].} =
```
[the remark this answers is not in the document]
web3-developer: Good catch, thanks. I believe those exceptions shouldn't occur in practice but I'll update it to use the more specific types in my next PR anyway.

nimbus/rpc/experimental.nim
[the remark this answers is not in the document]
kdeme: My point was rather regarding the pruning of state data.
But yes, pruning of block data will/can also become applicable after EIP-4444.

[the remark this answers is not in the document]
web3-developer: Oh I see what you mean. I haven't tested this but the current implementation will likely just fail when trying to initialize the vm to the state of the parent block if the state doesn't exist. It appears to throw the CatchableError from that line. What would you expect the endpoint to return in this scenario? I think a reasonable assumption is that for this endpoint to work you will either need to run an archive node or only query recent blocks as they get created. Maybe I should try catching the CatchableError and rethrowing it with a message that explains the above assumption.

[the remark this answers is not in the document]
web3-developer: Ok sure sounds good. I'll check to make sure it doesn't throw a Defect on that line.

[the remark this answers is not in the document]
web3-developer: It looks like it does throw a defect if the stateRoot doesn't exist on initialization. I'm not sure if it's worth trying to handle this case just yet because it shouldn't be possible to trigger it until we have pruning. Perhaps I should just leave a comment in the code as a reminder to implement the check you suggested above once we have those parameters available?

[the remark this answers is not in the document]
kdeme: Yes, a comment sounds good. For now a Defect is fine, as it shouldn't be a case that can occur unless the db is corrupt or so. But when pruning gets added, that additional check may not be forgotten.

PR #1981, opened 2024-01-23
Witness Generation Bug Fixes

web3-developer:


nimbus/db/ledger/accounts_cache.nim
```diff
@@ -654,7 +654,9 @@ proc getStorageRoot*(ac: AccountsCache, address: EthAddress): Hash256 =
   else: acc.account.storageRoot
 
 func update(wd: var WitnessData, acc: RefAccount) =
-  wd.codeTouched = CodeChanged in acc.flags
+  # once the code is touched make sure it doesn't get reset back to false in another update
+  if not wd.codeTouched:
+    wd.codeTouched = CodeChanged in acc.flags or CodeLoaded in acc.flags
```
web3-developer, line 659: This gets called after every transaction just before persist is called on the account cache. The WitnessData is held in a HashSet and updated in place if it already exists. The codeTouched flag was incorrectly being reset to false which was causing the bytecode to be excluded from the witness.

web3-developer: CodeLoaded should also trigger the codeTouched flag because if we need to read the code during the transaction then we will need it in the block witness for stateless execution.

nimbus/db/ledger/accounts_ledger.nim
```diff
@@ -657,7 +657,9 @@ proc getStorageRoot*(ac: AccountsLedgerRef, address: EthAddress): Hash256 =
   else: acc.account.storageVid.hash().valueOr: EMPTY_ROOT_HASH
 
 func update(wd: var WitnessData, acc: RefAccount) =
-  wd.codeTouched = CodeChanged in acc.flags
+  # once the code is touched make sure it doesn't get reset back to false in another update
+  if not wd.codeTouched:
+    wd.codeTouched = CodeChanged in acc.flags or CodeLoaded in acc.flags
```
web3-developer, line 662: I noticed that we have this in the AccountsLedger code as well. Assuming this will be used at some point as well, if not already.

stateless/test_block_witness.nim
```diff
@@ -24,7 +24,7 @@ type
 
 proc testGetBranch(tester: Tester, rootHash: KeccakHash, testStatusIMPL: var TestStatus) =
   var trie = initAccountsTrie(tester.memdb, rootHash)
-  let flags = {wfEIP170}
+  let flags = {wfNoFlag}
```
web3-developer, line 27: The {wfEIP170} flag turns on a check in the witness building code which throws an exception if the bytecode to be read or written is larger than the max valve as specified in EIP170. I've updated the tests to use {wfNoFlag} to allow the tests to cover scenarios from before EIP170.

stateless/tree_from_witness.nim
```diff
@@ -1,5 +1,5 @@
 # Nimbus
-# Copyright (c) 2020-2023 Status Research & Development GmbH
+# Copyright (c) 2020-2024 Status Research & Development GmbH
```
web3-developer, line 2: Fixing copyright notices from my previous PR.

kdeme (approved):

nimbus/db/ledger/accounts_cache.nim
kdeme: Do we have a test case that specifically tests this?

jangko: Good catch

jangko (approved):

PR #1986, opened 2024-01-25
Storage of block witnesses

web3-developer:
It would be good to get some early feedback on this change. 

I've added a new flag to Nimbus `--generate-witness` which will enable generation and storage of block witnesses in the database. The feature is disabled by default. The block witness is stored in the database by block hash and is persisted during the persistBlocks proc.

### DIFF
--- a/nimbus/db/core_db/core_apps_newapi.nim
+++ b/nimbus/db/core_db/core_apps_newapi.nim
@@ -921,6 +921,21 @@ proc haveBlockAndState*(db: CoreDbRef, headerHash: Hash256): bool =
   # see if stateRoot exists
   db.exists(header.stateRoot)
 
+proc getBlockWitness*(db: CoreDbRef, blockHash: Hash256): seq[byte] {.gcsafe.} =
+  let data = db.newKvt(Shared)
+               .get(blockHashToBlockWitnessKey(blockHash).toOpenArray).valueOr:
+    if error.error != KvtNotFound:
+      warn logTxt "getBlockWitness()", blockHash, action="get()", error=($$error)
+    return
+
+  return data
+
+proc setBlockWitness*(db: CoreDbRef, blockHash: Hash256, witness: seq[byte]) =
+  let witnessKey = blockHashToBlockWitnessKey blockHash
+  db.newKvt.put(witnessKey.toOpenArray, witness).isOkOr:
+    warn logTxt "setBlockWitness()", witnessKey, action="put()", error=($$error)
+    return
+
 # ------------------------------------------------------------------------------
 # End
 # ------------------------------------------------------------------------------
--- a/nimbus/db/core_db/core_apps_newapi.nim
+++ b/nimbus/db/core_db/core_apps_newapi.nim
@@ -921,17 +921,17 @@ proc haveBlockAndState*(db: CoreDbRef, headerHash: Hash256): bool =
   # see if stateRoot exists
   db.exists(header.stateRoot)
 
-proc getBlockWitness*(db: CoreDbRef, blockHash: Hash256): seq[byte] {.gcsafe.} =
-  let data = db.newKvt(Shared)
-               .get(blockHashToBlockWitnessKey(blockHash).toOpenArray).valueOr:
-    if error.error != KvtNotFound:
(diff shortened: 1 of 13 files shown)


jangko:
Why store the witness in database? We can always build the witness from the state right?

web3-developer:
> Why store the witness in database? We can always build the witness from the state right?

Yes that's true, we can generate the witness on demand from the block as I've done in the experimental RPC endpoints. Zahary asked me to implement storage of the block witnesses because we will likely need them for the future stateless client. This change isn't needed right now but I guess its more of a forward looking change which is why it's turned off by default. My understanding is that in the future full nodes will generate and store block witnesses to be shared over the network to other stateless Ethereum clients which will be able to verify blocks by using the witness as the initial state. Storing the witness and then just fetching it would likely perform better if serving the data to a number of other clients rather then regenerating it for each request. Of course all this is dependent on the verkle tries update which should in theory reduce the size of the witness enough so that it can be propagated across the network efficiently and within the block time.

web3-developer:
Ready for final review and feedback.

nimbus/db/core_db/core_apps_newapi.nim
```diff
@@ -921,6 +921,21 @@ proc haveBlockAndState*(db: CoreDbRef, headerHash: Hash256): bool =
   # see if stateRoot exists
   db.exists(header.stateRoot)
 
+proc getBlockWitness*(db: CoreDbRef, blockHash: Hash256): seq[byte] {.gcsafe.} =
+  let data = db.newKvt(Shared)
+               .get(blockHashToBlockWitnessKey(blockHash).toOpenArray).valueOr:
+    if error.error != KvtNotFound:
+      warn logTxt "getBlockWitness()", blockHash, action="get()", error=($$error)
+    return
```
kdeme, line 929: Would it not make more sense to return here a `Result` or `results.Opt`?

web3-developer: It appears that the db.newKvt.get proc returns the CoreDbRc type which is defined deeper inside the database layer code. It looks like none of the other procs in this file are returning this type perhaps to hide this type from the calling code. I copied this code pattern from the getScore proc defined further above. I agree that it doesn't look great to just return and appears to ignore errors apart from the warning. Perhaps it would be better let the exception be thrown up the stack here if one occurs.

web3-developer: Hi @mjfh , are you able to help with addressing the question above? I see you are the original author of this file so I'm wondering what the reasoning is for this code pattern?

kdeme: > Perhaps it would be better let the exception be thrown up the stack here if one occurs?

I had a look at this code and it doesn't seem any exception (CatchableError) can occur here. The `db.newKvt.get` code does not raise at least (see `{.push raises: [].}` on top of that file). And the `CoreDbRc` returned is a Result type (You are already using it like that, with the `valueOr`).

So you could just return that same result type. But perhaps better would be to remap the current Result type to a simpler one without the `CoreDbErrorRef` error part and instead something more digestible for the API user (like just an error string). Unless it would be required for the layer up to have that error information, but I don't think so.

> I copied this code pattern from the getScore proc defined further above.

Yes, so `getScore` results in 0 basically when the getter fails. I don't know of this functionality, but perhaps there the idea is to get 0 if no score exists yet in the db (no idea really, someone familiar with this code could say if this is an issue or not).
In the case of `getBlockWitness` it would require to do a `len>0` test after the call or rely on failing at the moment of verification (but that could be after a network transfer for example).

web3-developer: Sure, I'll update it to use the regular result type and will map the error code to a plain string.

web3-developer:
I'm now made the witness generation parameter hidden in the Nimbus CLI. 

@jangko @kdeme Do either of you have any objections to me merging this change?

kdeme (approved):

kdeme:
> I've now made the witness generation parameter hidden in the Nimbus CLI.
> 
> @jangko @kdeme Do either of you have any objections to me merging this change?

I do not.

PR #1998, opened 2024-01-31
Initial implementation of era1 + export + verify

kdeme:
Initial implementation of era1 + export and verify command in the existing eth_data_exporter.

Implementation of era1 file format as currently described in: https://github.com/ethereum/go-ethereum/pull/26621

Main issue currently is: https://github.com/status-im/nimbus-eth1/commit/3c67911c1b7121c57abad03d1c295d097a4624b9#diff-1948da0f3a7cfa79607319be10ae226dae3c9d9bbbd6ff35faea52df948c571cR40-R43
```
# - Snappy does not give the same compression result as the implementation used
# by geth for some block headers and block bodies. This is an issue if we want
# to rely on sha256sum as checksum for the individual era1 files as is suggested
# in https://github.com/ethereum/go-ethereum/pull/26621
```

Other TODOs:
- [x] General clean-up / rework some code
- [x] Validation of blocks and receipts in verify command
- [x] Testing of last (incomplete) era.
- [x] Testing index (Later to be used when Portal bridge uses era files)

Regarding the exportEra1 command: The export happens by requesting the block data over EL json-rpc API. So it is not directly integrated in Nimbus. This is done because it is difficult to currently get Nimbus EL to sync until merge point on mainnet, and with this architecture it is/should be compatible with other clients. Downside is that it will obviously be slower than directly accessing the db.

### DIFF
--- a/fluffy/eth_data/era1.nim
+++ b/fluffy/eth_data/era1.nim
@@ -9,7 +9,7 @@
 
 import
   std/[strformat, typetraits],
-  results, stew/[endians2, io2, byteutils],
+  results, stew/[endians2, io2, byteutils, arrayops],
   stint, snappy,
   eth/common/eth_types_rlp,
   beacon_chain/spec/beacon_time,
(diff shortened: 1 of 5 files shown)


lightclient:
I'm just curious if you verify the checksums do not match? I actually haven't checked any impls to verify if this is actually an issue in practice or just theory.

kdeme:
> I'm just curious if you verify the checksums do not match? I actually haven't checked any impls to verify if this is actually an issue in practice or just theory.

Yes, the checksums do not match.
When I compared some era1 files created from your branch with some of this nimbus implementation, I noticed occasionally (not all, in fact a minority) records of `BlockHeader`s and `BlockBody`s to be different. When I further investigated, it turned out the the compressed byte string was different, but once I decompressed both, the resulting RLP encoded byte string was the same. I will further investigate when this occurs and whether it is some sort of ambiguity in snappy.

My era1 verification code can however perfectly decompress geth created era1 files and verify them (accumulator creation, txRoot, ommersHash and receiptsRoot checks).

lightclient:
Okay fantastic. I never really came up with an appropriate solution to the checksum issue. I think it is still valuable to provide per-client because of how fast it can be verified. But it is good to know the accumulator is in agreement.

kdeme:
One more issue was noticed when decoding the later geth era1 files, fixed here: https://github.com/status-im/nim-eth/pull/672
Invalid encoder/decoder on Receipt lists on our side.

tersec:
> Okay fantastic. I never really came up with an appropriate solution to the checksum issue. I think it is still valuable to provide per-client because of how fast it can be verified. But it is good to know the accumulator is in agreement.

How much slower is `checksum(decode_ssz(ssz_encoded_blob))` than `checksum(ssz_encoded_blob)`?

The thing which should match is the decoded, not encoded, version.

fluffy/eth_data/era1.nim
```diff
@@ -0,0 +1,374 @@
...
+  for i in 0..<count:
+    ? f.readFileExact(buf)
+
+    let
+      offset = uint64.fromBytesLE(buf)
+      absolute =
+        if offset == 0: 0'i64
+        else:
+          # Wrapping math is actually convenient here
+          cast[int64](cast[uint64](startPos) + offset)
```
kdeme, line 189: This is the change that I need to currently read the era1 files produced by go-ethereum.

```suggestion
      offset = uint64.fromBytesLE(buf)
      currentPos = ? f.getFilePos().mapErr(toString)
      absolute =
        if offset == 0: 0'i64
        else:
          # Wrapping math is actually convenient here
          cast[int64](cast[uint64](currentPos) + offset)
```

kdeme:
This first version is ready to be merged if https://github.com/status-im/nimbus-eth2/pull/5866 gets in.

Has been tested manually by exporting some of the biggest (in data size) eras. And also the latest, incomplete era.
Verify has been tested both with self generated ones and with the ones from geth, the latter does still need the small adaption in index code.

Exporting of era1 files does take very long (due to the current architecture) and is thus not that practical to use for the full range.

kdeme:
This has now also been tested on an era1 file generated with the latest (corrected) version in go-ethereum.

PR #2003, opened 2024-02-02
state network test vectors update

mynameisdaniil:
Latest spec update. Added real test vectors and moved them out to https://github.com/status-im/portal-spec-tests/. Also added codec for nibbles. Decided not to use methods that we already have, since it will require us to change some types to make it work. And conceptually it is wrong, since encoding we use here might (and will) differ from the one that we already have.


fluffy/network/state/state_content.nim
```diff
@@ -13,11 +13,20 @@
 import
   nimcrypto/[hash, sha2, keccak], stew/results, stint,
   eth/common/eth_types,
+  eth/trie/nibbles,
```
kdeme, line 16: Can be removed as you are not using it in the end, or did I miss some usage?

fluffy/network/state/state_content.nim
```diff
@@ -125,3 +128,46 @@ func toContentId*(contentKey: ByteList): ContentId =
 
 func toContentId*(contentKey: ContentKey): ContentId =
   toContentId(encode(contentKey))
+
+func packNibbles*(nibbles: seq[byte]): Nibbles {.raises: ValueError} =
+  if nibbles.len() > MAX_UNPACKED_NIBBLES_LEN:
+    raise newException(ValueError, "Can't pack more than 64 nibbles")
```
kdeme, line 134: Generally we prefer more explicit error handling mechanisms such are `Result`. See also: https://status-im.github.io/nim-style-guide/errors.html

fluffy/network/state/state_content.nim
```diff
@@ -125,3 +128,46 @@ func toContentId*(contentKey: ByteList): ContentId =
...
+    else:
+      output.add(NibblePair(currentByte or nibble))
+      currentByte = 0
+    highNibble = not highNibble
+
+  Nibbles(isOddLength: isOddLength, packedNibbles: Nibbles.packedNibbles.init(output))
+
+func unpackNibbles*(nibbles: Nibbles): seq[byte] {.raises: ValueError} =
+  if nibbles.packedNibbles.len() > MAX_PACKED_NIBBLES_LEN:
+    raise newException(ValueError, "Can't unpack more than 32 nibbles")
```
kdeme, line 157: Idem as above. Although here one could probably assert, as this data would have been SSZ decoded and failed on that step.

Or created by packNibbles and failed.

fluffy/network/state/state_content.nim
mynameisdaniil: Agree. But this is here just as a failsafe and to explicitly show limitation. As you mentioned below, these checks should never fail in normal operation and using `Result` will force us to unwrap it each time.

fluffy/network/state/state_content.nim
mynameisdaniil: I think `doAssert` is a way to go in both cases, since 64 nibbles is the limit, so in case of normal operation we should not hit it in any way.

fluffy/network/state/state_content.nim
kdeme: Well, an exception is even worse than the unwrap, as you need to catch it or it will bubble all the way up.

If it should never occur a `doAssert` should be added or it should be coded in such way that it cannot occur.

mynameisdaniil: Totally agree on `doAssert`

kdeme (approved):
ok to merge for me when CI is green

PR #2009, opened 2024-02-05
More witness fixes

web3-developer:
After improving the tests for the exp_getProofsByBlockNumber endpoint to validate the list of state updates against a local test state, I discovered the following issues which are fixed in this PR:
- State roots didn't match for some of the tests and further investigation showed that not all the state updates were coming through in the block proofs. Values were not being returned when a storage slot was updated back to a zero value.
- AccountCache code has been updated to allow zero value storage slots in the witness cache data.
- Storage slots are now not returned if the account doesn't exist (for example due to a self destruct)
- exp_getProofsByBlockNumber endpoint refactored to build the list of proofs directly from the MultikeyRef. This was required to return the zero value slots which don't get returned in the block witness due to how the witness is built from the state tries. For zero value storage slots there is no leaf because it is deleted during the call to persist. I believe the block witness doesn't need to store zero value storage slots because this is the default value for a slot.

### DIFF
--- a/nimbus/db/ledger/accounts_cache.nim
+++ b/nimbus/db/ledger/accounts_cache.nim
@@ -664,11 +664,6 @@ func update(wd: var WitnessData, acc: RefAccount) =
       wd.storageKeys.incl k
 
   for k, v in acc.overlayStorage:
-    if v.isZero and k notin wd.storageKeys:
-      continue
-    if v.isZero and k in wd.storageKeys:
-      wd.storageKeys.excl k
-      continue
     wd.storageKeys.incl k
 
 func witnessData(acc: RefAccount): WitnessData =
--- a/nimbus/db/ledger/accounts_cache.nim
+++ b/nimbus/db/ledger/accounts_cache.nim
@@ -1,5 +1,5 @@
 # Nimbus
-# Copyright (c) 2023 Status Research & Development GmbH
+# Copyright (c) 2023-2024 Status Research & Development GmbH
 # Licensed under either of
 #  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or
 #    http://www.apache.org/licenses/LICENSE-2.0)
--- a/nimbus/rpc/experimental.nim
+++ b/nimbus/rpc/experimental.nim
@@ -16,14 +16,13 @@ import
   ../[transaction, vm_state, constants, vm_types],
   ../db/state_db,
   rpc_types, rpc_utils,
-  ../core/tx_pool,
   ../common/common,
   ../utils/utils,
   ../beacon/web3_eth_conv,
   ./filters,
   ../core/executor/process_block,
   ../db/ledger,
-  ../../stateless/[witness_verification, witness_types],
+  ../../stateless/[witness_verification, witness_types, multi_keys],
   ./p2p
 
 type
@@ -33,7 +32,7 @@ type
 proc getBlockWitness*(
     com: CommonRef,
     blockHeader: BlockHeader,
-    statePostExecution: bool): (KeccakHash, BlockWitness, WitnessFlags)
+    statePostExecution: bool): (MultikeysRef, BlockWitness)
     {.raises: [RlpError, BlockNotFound, ValueError, CatchableError].} =
 
   let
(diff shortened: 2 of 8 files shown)


nimbus/db/ledger/accounts_cache.nim
```diff
@@ -664,11 +664,6 @@ func update(wd: var WitnessData, acc: RefAccount) =
       wd.storageKeys.incl k
 
   for k, v in acc.overlayStorage:
-    if v.isZero and k notin wd.storageKeys:
-      continue
-    if v.isZero and k in wd.storageKeys:
-      wd.storageKeys.excl k
-      continue
```
web3-developer, line 671: This change is required so that state updates for storage slots from a non zero value to a zero value would be recorded. We want to know about these updates in the new exp_getProofsByBlockNumber endpoint. This change actually doesn't effect the block witness generation because zero value storage slots are deleted in the AccountCache.persist call and therefore the block witness won't contain any zero value storage slots.

tests/test_rpc_experimental_json.nim
```diff
@@ -56,37 +57,77 @@ proc importBlockData(node: JsonNode): (CommonRef, Hash256, Hash256, UInt256) {.
...
+      check stateDB.getCodeHash(address) == EMPTY_SHA3
+    else:
+      check stateDB.getCodeHash(address) == codeHash
+
+    if storageHash == ZERO_HASH256 or storageHash == EMPTY_ROOT_HASH:
+      check stateDB.getStorageRoot(address) == EMPTY_ROOT_HASH
+    else:
+      check stateDB.getStorageRoot(address) == storageHash
+
+  check stateDB.rootHash == expectedStateRoot
```
web3-developer, line 130: Improved the validation here to apply the state changes in the test and check that the test state root matches the expected value.

nimbus/rpc/experimental.nim
```diff
@@ -59,34 +58,27 @@ proc getBlockWitness*(
...
   else:
     # Reset state to what it was before executing the block of transactions
     let initialState = BaseVMState.new(blockHeader, com)
-    result = (initialState.stateDB.rootHash, initialState.buildWitness(mkeys), flags)
+    result = (mkeys, initialState.buildWitness(mkeys))
 
   dbTx.rollback()
 
-
 proc getBlockProofs*(
```
web3-developer, line 69: Refactored this to simply use the MultikeysRef to build the list of proofs from the state. Avoid using the witness for simplicity.

nimbus/rpc/experimental.nim
```diff
@@ -59,34 +58,27 @@ proc getBlockWitness*(
...
 
-  for address, account in verifyWitnessResult.value():
-    var slots = newSeqOfCap[UInt256](account.storage.len())
+  for keyData in mkeys.keys:
+    let address = keyData.address
+    var slots = newSeq[UInt256]()
 
-    for slotKey, _ in account.storage:
-      slots.add(slotKey)
+    if not keyData.storageKeys.isNil and accDB.accountExists(address):
```
web3-developer, line 79: Storage slots won't be returned if the account doesn't exist.

web3-developer:
I'm using the rpc_getproofs_track_state_changes test to apply the state changes from the getProofs endpoints for each block. While running the test locally against Nimbus I've been able to successfully verify that the state roots are correct for the first 700 000 blocks. 

This one is ready for final feedback and review.

tests/test_rpc_getproofs_track_state_changes.nim
```diff
@@ -0,0 +1,148 @@
...
+    if storageHash == ZERO_HASH256 or storageHash == EMPTY_ROOT_HASH:
+      check stateDB.getStorageRoot(address) == EMPTY_ROOT_HASH
+    else:
+      check stateDB.getStorageRoot(address) == storageHash
+
+  check stateDB.rootHash == expectedStateRoot
+
+proc rpcGetProofsTrackStateChangesMain*() =
+
+  suite "rpc getProofs track state changes tests":
```
kdeme, line 106: I understand this is a semi-manual test where you also need to have Nimbus up and running and populated with the state until that tested block?

Might want to put a little comment somewhere here about that and/or how to run it.

tests/test_rpc_getproofs_track_state_changes.nim
```diff
@@ -0,0 +1,148 @@
...
+          check storageData[slotProof.key] == slotProof.value
+
+      if code.len() > 0:
+        stateDB.setCode(address, code)
+
+    if (balance == 0 and nonce == 0 and codeHash == ZERO_HASH256 and storageHash == ZERO_HASH256):
+      stateDB.setCode(address, @[])
+      stateDB.clearStorage(address)
+      stateDB.deleteAccount(address)
+    elif (balance == 0 and nonce == 0 and codeHash == EMPTY_SHA3 and storageHash == EMPTY_ROOT_HASH):
```
kdeme, line 76: Perhaps explain in a comment the differences between these?

kdeme (commented):
lgtm, I did skim rather fast over it but it makes sense to me . Perhaps @jangko who knows this code better wants to have a look still.

tests/test_rpc_getproofs_track_state_changes.nim
jangko: I agree with Kim, this should be explained with comments or at least a self describe template.

jangko:
No objections. The critical parts in accounts_cache is not modified. It is relatively safe to modify the witness creation part. The tests looks good.

tests/test_rpc_getproofs_track_state_changes.nim
web3-developer: Sure, will do.

tests/test_rpc_getproofs_track_state_changes.nim
web3-developer: Ok, will add a comment here.

web3-developer:
> No objections. The critical parts in accounts_cache is not modified. It is relatively safe to modify the witness creation part. The tests looks good.

Thanks. In that case, I'll merge it after adding the suggested comments in the test.

tests/test_rpc_getproofs_track_state_changes.nim
```diff
@@ -5,6 +5,23 @@
 #  * MIT license ([LICENSE-MIT](LICENSE-MIT) or http://opensource.org/licenses/MIT)
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
+# This test is intended to be run manually against a running instance of Nimbus-Eth1
+# so it should not be added to the test runner or to the CI test suite.
+#
```
kdeme, line 10: It would be good to still have this test compile in CI to avoid bit rot.

kdeme: Oh, I just noticed there is already a `test_tools_build.nim` file which builds all these tools in one go: https://github.com/status-im/nimbus-eth1/blob/master/tests/test_tools_build.nim

This test/tool could be added there. (This is done all in one go to make it faster fyi).

web3-developer: Thanks, good point. Sure, will do.

PR #2020, opened 2024-02-10
Formatting fluffy with nph v0.5.1 + CI check

kdeme:


fluffy/database/content_db.nim
```diff
@@ -83,56 +81,80 @@ template expectDb(x: auto): untyped =
...
-    "Custom function isInRadius creation OK")
+    "Custom function isInRadius creation OK"
+  )
 
-  let sizeStmt = db.prepareStmt(
-    "SELECT page_count * page_size as size FROM pragma_page_count(), pragma_page_size();",
-    NoParams, int64).get()
+  let sizeStmt =
+    db
+    .prepareStmt(
```
kdeme, line 110: This does get a bit spacious.

fluffy/database/content_db_custom_sql_functions.nim
```diff
@@ -7,25 +7,21 @@
...
-  eth/db/kvstore_sqlite3
+import stew/ptrops, stint, sqlite3_abi, eth/db/kvstore_sqlite3
 
 func xorDistance(a: openArray[byte], b: openArray[byte]): seq[byte] =
   doAssert(a.len == b.len)
 
   let length = a.len
   var distance: seq[byte] = newSeq[byte](length)
-  for i in 0..<length:
+  for i in 0 ..< length:
```
kdeme, line 17: I read https://arnetheduck.github.io/nph/style.html#infix-operators and I am a bit surprised that a lot of the code uses spaces there. But yes, both with and without spaces are used and I don't really care which one we select.

fluffy/database/content_db.nim
arnetheduck: `.get()` -> `[]` will get you back into non-dot-constructor-mode

fluffy/network/wire/portal_protocol.nim
```diff
@@ -70,23 +79,44 @@ declareCounter portal_gossip_without_lookup,
...
+  248,
+  249,
+  250,
+  251,
+  252,
+  253,
+  254,
+  255,
+  256
+]
```
kdeme, line 110: This is also still an unexpected formatting result. It seems to be caused by the `float64` on the first index.

Solvable by adding specifically the array definition, or by making every number a `.0`

fluffy/network/wire/portal_protocol.nim
```diff
@@ -1381,22 +1421,23 @@ proc neighborhoodGossip*(
...
+    let contentKV = ContentKV(contentKey: contentKeys[i], content: contentItem)
     discard contentList.add(contentKV)
 
   # Just taking the first content item as target id.
   # TODO: come up with something better?
-  let contentId = p.toContentId(contentList[0].contentKey).valueOr:
-    return 0
+  let contentId =
+    p.toContentId(contentList[0].contentKey).valueOr:
+      return 0
```
kdeme, line 1440: I don't fully understand why here the `let contentId =` gets split from the `p.toContentId(contentList[0].contentKey).valueOr:`. I've it often get split, but also seen it remain on the same line in other occasions.

kdeme: Actually, after looking at more cases, it pretty much always gets split.

It seem sthe only locations where it doesn't is when it get `(...) around it, e.g. https://github.com/status-im/nimbus-eth1/pull/2020/files#diff-b1c3e1df30ba38fbb2809053a0bffd888ee51681435cf333ac4a80d4026f97eeR324-R336

fluffy/network/wire/portal_protocol.nim
arnetheduck: `[239.float, ...]` should flatten it - hm, this is almost worth a bug report and/or a special case maybe - the problem is that `float64 239` is considered a complex expression and therefore the whole list gets complex rendering - have to think a bit about this one

fluffy/network/wire/portal_protocol.nim
arnetheduck: right - it might be because the `valueOr` introduces a forced newline at which point we lose track of the line length and perform a pessimistic formatting - this could be considered a bug to report as well

fluffy/network/history/accumulator.nim
```diff
@@ -29,17 +31,17 @@ const
...
-  # `mergeBlockNumber`, but it could be used as value to double check on at
-  # merge block.
-  # TODO: Could also be used as value to actual finish the accumulator, instead
-  # of `mergeBlockNumber`, but:
-  # - Still need to store the actual `mergeBlockNumber` and run-time somewhere
-  # as it allows for each pre vs post merge block header checking.
-  # - Can't limit `historicalEpochs` SSZ list at `preMergeEpochs` value.
-  # - Should probably be stated in the portal network specs.
-  # TERMINAL_TOTAL_DIFFICULTY = u256"58750000000000000000000"
+    # TODO:
```
kdeme, line 34: hmm, curious as to why this gets indentation added

fluffy/network/wire/portal_protocol.nim
cheatfate: Modern Nim usually uses first value to get value type, so type could be avoided for next values. And there could be a lot of different variants, not only `<simple type><value>`, you also should remember about `0'u` and so on.

fluffy/tools/beacon_lc_bridge/beacon_lc_bridge.nim
```diff
@@ -217,57 +215,49 @@ proc asPortalBlockData*(
...
-      fee: some(payload.baseFeePerGas),
-      withdrawalsRoot: withdrawalsRoot,
-      blobGasUsed: options.none(uint64),
-      excessBlobGas: options.none(uint64) # TODO: adjust later according to deneb fork
+      timestamp: payload.timestamp.EthTime, extraData: bytes payload.extraData,
+      mixDigest: payload.prevRandao.asEthHash, nonce: default(BlockNonce),
+      fee: some(payload.baseFeePerGas), withdrawalsRoot: withdrawalsRoot,
+      blobGasUsed: options.none(uint64), excessBlobGas: options.none(uint64)
+        # TODO: adjust later according to deneb fork
     )
```
kdeme, line 256: Odd that here these get grouped together. bug?

fluffy/tools/eth_data_exporter/exporter_conf.nim
```diff
@@ -41,26 +41,19 @@ const
...
-    exportHeadersWithProof =
-      "Export block headers with proof from *.e2s headers file and epochAccumulator files"
+    exportBlockData = "Export block data (headers, bodies and receipts) to a json format or a database. Some of this functionality is likely to get deprecated"
+    exportEpochHeaders = "Export block headers from an Ethereum JSON RPC Execution endpoint to *.e2s files arranged per epoch (8192 blocks)"
+    verifyEpochHeaders = "Verify *.e2s files containing block headers. Verify currently only means being able to RLP decode the block headers"
+    exportAccumulatorData = "Build and export the master accumulator and historical epoch accumulators. Requires *.e2s block header files generated with the exportHeaders command up until the merge block"
+    printAccumulatorData = "Print the root hash of the master accumulator and of all historical epoch accumulators. Requires data generated by exportAccumulatorData command"
+    exportHeaderRange = "Export block headers from an Ethereum JSON RPC Execution endpoint to *.e2s files (unlimited amount)"
+    exportHeadersWithProof = "Export block headers with proof from *.e2s headers file and epochAccumulator files"
```
kdeme, line 56: For enums the text should probably also go on the next line?

fluffy/conf.nim
```diff
@@ -60,215 +62,240 @@ type
...
+      defaultValue: StdoutLogKind.Auto,
+      name: "log-format"
+    .}: StdoutLogKind
 
-    udpPort* {.
-      defaultValue: 9009
-      desc: "UDP listening port"
-      name: "udp-port" .}: uint16
+    udpPort* {.defaultValue: 9009, desc: "UDP listening port", name: "udp-port".}:
+      uint16
```
kdeme, line 81: These that fit on one line look a bit weird within the rest. But this has more to do with confutils dsl I guess.
Perhaps I should use `#!fmt: off` here, or just not care about it :)

fluffy/network_metadata.nim
```diff
@@ -10,46 +10,42 @@
 import
   std/[sequtils, strutils, os, macros],
   stew/results,
-  chronos/timer#,
-  # eth/common/eth_types
+  chronos/timer
+    #,
+    # eth/common/eth_types
```
kdeme, line 15: These leftover comments shouldn't be here in the first place. But it is strange that for the last import item the comment goes as postfix (?) even when it fits on the line?

arnetheduck: there's a preference to bundle all postfix comments on a single indent level which makes sense for documentation - therefore: end-of-line if it's a single line comment that fits, else indented newline - there's a few stragglers like this where that rule doesn't make sense but for the vast majority of code, it's a big improvement

fluffy/conf.nim
arnetheduck: write a longer help text :stuck_out_tongue:

fluffy/tools/beacon_lc_bridge/beacon_lc_bridge.nim
arnetheduck: looks like a bug, yes

fluffy/network/history/accumulator.nim
arnetheduck: This is related to postfix detection heuristics (which are a bit wonky) - if there is no more variable in a `const` section, there's no natural space in the AST to make it a prefix comment, so it becomes a postfix instead, and in this case it became anchored to `preMergeEpochs`.

This could be considered a bug assuming it's possible to anchor it to `const` instead which would need a bit of digging. Bug, maybe :) 

That said, now that's become indented with respect to `preMergeEpochs` it'll stay that way even if a future version would anchor it to `const` instead, assuming that's possible.

fluffy/network/wire/portal_protocol.nim
kdeme: Created an issue: https://github.com/arnetheduck/nph/issues/52

fluffy/network/wire/portal_protocol.nim
kdeme: Created issue: https://github.com/arnetheduck/nph/issues/53

fluffy/tools/eth_data_exporter/exporter_conf.nim
kdeme: https://github.com/arnetheduck/nph/issues/54

.github/workflows/fluffy.yml
```diff
@@ -315,6 +315,18 @@ jobs:
...
+        # Pin nph to a specific version to avoid sudden style differences.
+        # Updating nph version should be accompanied with running the new
+        # version on the fluffy directory.
+        run: |
+          VERSION="v0.4.1"
+          ARCHIVE="nph-linux_x64.tar.gz"
+          curl -L "https://github.com/arnetheduck/nph/releases/download/${VERSION}/${ARCHIVE}" -o ${ARCHIVE}
+          tar -xzf ${ARCHIVE}
+          ./nph fluffy/
+          git diff --exit-code
```
arnetheduck, line 328: hm, now that I think about this check, it should probably only check the files that changed in a particular PR, like @etan-status did with the copyright nag

kdeme: Perhaps yes. But after this first initial big PR it would normally always only hit the files changed, unless:
a. Some PR got merged with failed linter
b. On update of nph version if some style was adjusted in nph.

In case of b, I think I like it more that all future style changes are bundled in one PR, compared to having it happen on each PR related to change of actual code functionality.

fluffy/tools/beacon_lc_bridge/beacon_lc_bridge.nim
kdeme: Seems like the cause is the TODO comment, worked around it for now but I'll file an issue for this.

kdeme: Created an issue for this: https://github.com/arnetheduck/nph/issues/64

kdeme:
I'm read to merge this as most concerning issues are resolved.

I will make a secondary PR where I add nph to the vendor list and create an optional make target for manual formatting. Whether or not that is the way forward can be discussed there then.

arnetheduck:
> I'm ready to merge this as most concerning issues are resolved.

I have a fix for the comment thing - turns out it was completely miscalculating the length in this case, so I'll push a 0.5.1 in a minute that addreses this - the length thing causes a few other things to get pushed into new lines apart from the long-list format itself.

arnetheduck:
> add nph to the vendor list

yeah, this is something I've been thinking about: how to give projects a way to tell developers which version they should be formatting with, similar to how a particular nim version is needed for compiling. for us, vendor works well - for broader nimble use, it's an open question.

PR #2031, opened 2024-02-16
Add flag to explicitly enable state network

mynameisdaniil:
Let's disable state network by default until everything is worked out


fluffy/conf.nim
```diff
@@ -246,6 +246,11 @@ type
       defaultValueDesc: $defaultDisablePoke
       name: "disable-poke" .}: bool
 
+    stateNetworkEnabled* {.
+      desc: "Enable State Network"
```
kdeme, line 250: As this is not an option we want to keep, It would be best to make it hidden

PR #2032, opened 2024-02-17
Fix usage of deprecated startMetricsHttpServer call

kdeme:


fluffy/fluffy.nim
```diff
@@ -244,13 +244,19 @@ proc run(config: PortalConf) {.raises: [CatchableError].} =
...
+        error "Could not start metrics HTTP server", url, error
+        quit QuitFailure
+
+    info "Starting metrics HTTP server", url
     try:
-      chronos_httpserver.startMetricsHttpServer($address, port)
-    except CatchableError as exc: raise exc
-    # TODO: Ideally we don't have the Exception here
-    except Exception as exc: raiseAssert exc.msg
+      waitFor server.start()
```
arnetheduck, line 255: unlike startMetricsHttpServer, this will run the server on the same thread as the rest of the app, which may or may not be what you want

kdeme: Thanks, I was not aware of that (I had a quick glance and noticed still thread creation, but didn't realize the difference of the http server running on main thread). I'm unsure whether we want this in Fluffy or not.

I did notice that the `nimbus_validator_client` uses this new way. Did we consider yet what i the way forward with our applications, considering the old way did get deprecated?

arnetheduck: actually, I was wrong ;) it indeed still creates a thread and runs metrics there, just with a different setup

arnetheduck (approved):

fluffy/fluffy.nim
kdeme: Right, I was still waiting with merging this as iirc the difference is/was that the http server gets moved to the main thread (but metrics accounting itself stay on the secondary thread). 

Shouldn't be much of an issue so I'll rebase and merge this.

PR #2033, opened 2024-02-18
State network content retrieval

mynameisdaniil:
### What was the problem?
State network spec states that there are two distinct flavors of each data type:
1) Offer data that is gossiped via `Offer/Accept`
2) Retrieval data that is stored locally and provided via `Find/Found Content`

We already had data types for gossip as well as basic gossip implemented for `Account Trie Node`. We lacked types describing retrieval data, as well as ways to convert Offer data into Retrieval data.

### How this PR fixes that
By implementing ADTs describing this two flavours: `OfferContentValue`, `RetrievalContentValue`. Which allowed to implement functions that work on one and convert it into another: `offerContentToRetrievalContent`. Which allowed us to modify offer data into retrieval data right before storing it into local DB.
This PR also fixes gossip test. That is now checks if the next node in gossip chain stored it's gossiped data in retrieval format.

### Bonus
This PR also gets rid of multiple encoding/decoding calls by moving decoding operation into separate step, and then operating mainly on decoded data. Except for gossip function, as it needs both encoded and decoded data. Encoded is for `ContractCodeOffer` as it is not being modified when gossiped. And decoded is for the rest of the types.

### DIFF
--- a/fluffy/network/state/state_network.nim
+++ b/fluffy/network/state/state_network.nim
@@ -87,19 +87,18 @@ proc recursiveGossipAccountTrieNode(
     p: PortalProtocol,
     maybeSrcNodeId: Opt[NodeId],
     decodedKey: ContentKey,
-    decodedValue: OfferContentValue
+    decodedValue: AccountTrieNodeOffer
     ): Future[void] {.async.} =
-      let value = decodedValue.accountTrieNode
       var
         nibbles = decodedKey.accountTrieNodeKey.path.unpackNibbles()
-        proof = value.proof
+        proof = decodedValue.proof
 
       discard nibbles.pop()
       discard (distinctBase proof).pop()
       let
         updatedValue = AccountTrieNodeOffer(
           proof: proof,
-          blockHash: value.blockHash,
+          blockHash: decodedValue.blockHash,
         )
         updatedNodeHash = keccakHash(distinctBase proof[^1])
         encodedValue = SSZ.encode(updatedValue)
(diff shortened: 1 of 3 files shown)


fluffy/network/state/state_network.nim
```diff
@@ -201,13 +166,32 @@ proc processContentLoop(n: StateNetwork) {.async.} =
...
-          n.portalProtocol,
-          maybeSrcNodeId,
-          contentKeys,
-          contentValues
+      for i, contentValue in contentValues:
+        let
+          contentKey = contentKeys[i]
+          (decodedKey, decodedValue) = decodeKV(contentKey, contentValue).valueOr:
+            error "Unable to decode offered Key/Value"
+            continue
```
kdeme, line 174: I think it would be better to break/return here. When one item fails decoding, better just stop decoding and wasting CPU cycles on anything else from this node.

fluffy/network/state/state_network.nim
```diff
@@ -201,13 +166,32 @@ proc processContentLoop(n: StateNetwork) {.async.} =
...
+        if validateContent(decodedKey, decodedValue):
+          let
+            valueForRetrieval = decodedValue.offerContentToRetrievalContent().encode()
+            contentId = n.portalProtocol.toContentId(contentKey).valueOr:
+              error "Received offered content with invalid content key", contentKey
+              continue
```
kdeme, line 180: Idem as comment above.

fluffy/network/state/state_network.nim
```diff
@@ -201,13 +166,32 @@ proc processContentLoop(n: StateNetwork) {.async.} =
...
+          await gossipContent(
+            n.portalProtocol,
+            maybeSrcNodeId,
+            contentKey,
+            decodedKey,
+            contentValue,
+            decodedValue
           )
+        else:
+          error "Received offered content failed validation", contentKey
```
kdeme, line 194: Idem ^

PR #2081, opened 2024-03-18
tree embeddings from accounts from EIP-6800

advaita-saha:
## Verkle tree embeddings, from the accounts from EIP-6800

- [EIP-6800](https://eips.ethereum.org/EIPS/eip-6800)


nimbus/db/verkle/verkle_accounts.nim
```diff
@@ -0,0 +1,245 @@
...
+    tree/operations, 
+    tree/commitment
+  ],
+  "../../../vendor/nim-eth-verkle/constantine/constantine"/[
+    serialization/codecs_banderwagon
+  ]
+
+type
+  ChunkedCode = seq[byte]
+  VerkleTrie = ref object
```
jangko, line 25: Better if `VerkleTrie` become `VerkleTrieRef`, I know some of object naming in nimbus-eth1 doesn't follow this convention, but they are written before we realize the importance of correct naming.

nimbus/db/verkle/verkle_accounts.nim
```diff
@@ -0,0 +1,245 @@
...
+    root: BranchesNode
+  # db: <something>
+  # persistcheck: <some-flag>
+
+const
+  VersionLeafKey = 0
```
jangko, line 31: I think `VersionLeafKey..CodeSizeLeafKey` is a good candidate for enum because they don't have holes, using enum will make the type stronger. If enum is cumbersome, perhaps `distinct int/uint`  can help.

nimbus/db/verkle/verkle_accounts.nim
```diff
@@ -0,0 +1,245 @@
...
+  var poly: array[3, Field]
+  
+  poly[0] = zeroField()
+  poly[1].fromLEBytes(newAddr[0..15])
+  poly[2].fromLEBytes(newAddr[16..^1])
+
+  var ret = poly.ipaCommitToPoly()
+
+  var getTreePolyIndex0Point: Point
+  var a = fromHex(Bytes32, "0x22196df2c10590e04c34bd5cc57e09911b98c782a503d21bc1838e1c6e1a10bf")
```
jangko, line 72: `const a = fromHex(Bytes32, "0x22196df2c10590e04c34bd5cc57e09911b98c782a503d21bc1838e1c6e1a10bf")`
will give you benefits of compile time evaluation. if it is not modified later.

nimbus/db/verkle/verkle_accounts.nim
```diff
@@ -0,0 +1,245 @@
...
+  # poly = [2+256*64, address_le_low, address_le_high, tree_index_le_low, tree_index_le_high]
+  var poly: array[256, Field]
+
+  # 32-byte address, interpreted as two little endian
+  # 16-byte numbers.
+  poly[1].fromLEBytes(newAddr[0..15])
+  poly[2].fromLEBytes(newAddr[16..^1])
+
+  var getTreePolyIndex0Point {.noinit.} : Point
+  var a = fromHex(Bytes32, "0x22196df2c10590e04c34bd5cc57e09911b98c782a503d21bc1838e1c6e1a10bf")
```
jangko, line 111: idem for `const a`

nimbus/db/verkle/verkle_accounts.nim
```diff
@@ -0,0 +1,245 @@
...
+  discard getTreePolyIndex0Point.deserialize(a)
```
jangko, line 112: I don;t know how deserialize implemented, but `getTreePolyIndex0Point` is also a good candidate for compile time evaluation and become a const.

nimbus/db/verkle/verkle_accounts.nim
```diff
@@ -0,0 +1,245 @@
...
+
+proc getTreeKeyCodeChunkWithEvaluatedAddress*(addressPoint: Point, chunk: UInt256): Bytes32 =
+  let (treeIndex, subIndex) = getTreeKeyCodeChunkIndices(chunk)
+  return getTreeKeyWithEvaluatedAddress(addressPoint, treeIndex, subIndex)
+
+
+proc newVerkleTrie*(): VerkleTrie =
+  result = VerkleTrie(root: newTree())
+
+proc updateAccount*(trie: var VerkleTrie, address: EthAddress, acc: Account) =
```
jangko, line 170: if `VerkleTrie` become `VerkleTrieRef`, we can omit `var` modifier. because `var` modifier essentially not needed for pointer of ref in this case. Except we use that var param as output to create a new ref.

nimbus/db/verkle/verkle_accounts.nim
```diff
@@ -0,0 +1,245 @@
...
+      nonceValfull[i] = 0
+
+  var balanceVal = acc.balance.toBytesLE()
+  trie.root.setValue(verKey, varVal)
+  trie.root.setValue(nonceKey, nonceValfull)
+  trie.root.setValue(balanceKey, balanceVal)
+  trie.root.setValue(codeKeccakKey, acc.codeHash.data)
+
+    
+proc hashVerkleTrie*(trie: var VerkleTrie): Bytes32 =
```
jangko, line 195: ditto for the `var VerkleTrie`

nimbus/db/verkle/verkle_accounts.nim
```diff
@@ -0,0 +1,245 @@
...
+  return trie.root.commitment.serializePoint()
+
+# ChunkifyCode generates the chunked version of an array representing EVM bytecode
+proc chunkifyCode*(code: openArray[byte]) : ChunkedCode =
+  var
+    chunkOffset = 0 # offset in the chunk
+    chunkCount  = len(code) div 31
+    codeOffset  = 0 # offset in the code
+
+  if len(code) mod 31 != 0:
```
jangko, line 206: `len(code)` -> `code.len` is more readable and one char less.

nimbus/db/verkle/verkle_accounts.nim
agnxsh: deserialize() does an inplace alteration to the Point variable (`getTreePolyIndex0Point`), and returns a boolean.

jangko: Does it call any C functions? or pure Nim?, if it is pure Nim, probably can use const.

nimbus/db/verkle/verkle_accounts.nim
advaita-saha: Yeah right, got it

nimbus/db/verkle/verkle_accounts.nim
advaita-saha: unable to use const : Seems like it does call some C functions internally in constantine implementation of marshaling/unmarshaling

agnxsh:
We can probably migrate the test vectors later to some other file, just to make it look cleaner, otherwise everything seems OK to me.

nimbus/db/verkle/verkle_accounts.nim
agnxsh: If we take to enum, then in order to retrieve the value do we need to typecast it to uint or something similar? How do we retrieve the number if it's an enum?

PR #2100, opened 2024-03-22
Remove redundant words

depthlending:


jangko:
Please update the copyright year, you can see which file need update in the linter CI log.

PR #2132, opened 2024-04-16
evm: Reject contract creation if the storage is non-empty(EIP-7610)

### LINKED ISSUE #2128: nim t8n to reject contract creation if destination storage is non empty
now geth team says https://github.com/ethereum/EIPs/pull/8161/files
contract creation must be reverted if storage is not empty. can you please also do it in nimbus t8n and I see if you fail the same tests as geth to correct the tests. 
this eip sucks as it breaks a few legacy tests

jangko:
ping @winsvega, this is nimbus-eth1 EIP-7610 implementation.
fix #2128


mjfh (approved):

winsvega:
Confirm. This patch is in line with current geth implementation.

PR #2152, opened 2024-04-24
Remove usages of CoreDb in Fluffy.

web3-developer:
Considering that we don't need the complexity of CoreDb and will likely never need the CoreDb aristo implemention in Fluffy, lets keep it simple and use the HexaryTrie implementation when doing in memory merkle root calculations.

### DIFF
--- a/fluffy/network/history/history_network.nim
+++ b/fluffy/network/history/history_network.nim
@@ -11,7 +11,7 @@ import
   stew/results,
   chronos,
   chronicles,
-  eth/[common/eth_types_rlp, rlp],
+  eth/[common/eth_types_rlp, rlp, trie, trie/db],
   eth/p2p/discoveryv5/[protocol, enr],
   ../../common/common_types,
   ../../database/content_db,
@@ -190,7 +190,7 @@ func encode*(receipts: seq[Receipt]): seq[byte] =
 # for if/when peer scoring/banning is added.
 
 proc calcRootHash(items: Transactions | PortalReceipts | Withdrawals): Hash256 =
-  var tr = newCoreDbRef(LegacyDbMemory).mptPrune
+  var tr = initHexaryTrie(newMemoryDB(), isPruning = false)
   for i, item in items:
     try:
       tr.put(rlp.encode(i), item.asSeq())
--- a/fluffy/network/history/history_network.nim
+++ b/fluffy/network/history/history_network.nim
@@ -16,7 +16,7 @@ import
   ../../common/common_types,
   ../../database/content_db,
   ../../network_metadata,
-  ../../../nimbus/[constants, db/core_db],
+  ../../../nimbus/constants,
   ../wire/[portal_protocol, portal_stream, portal_protocol_config],
   "."/[history_content, accumulator, beacon_chain_historical_roots]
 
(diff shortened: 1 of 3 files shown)


web3-developer:
Note that I also reviewed the usage of CoreDb in the test_state_network test but I don't yet understand the details of that test. I'll update it if possible once I've implemented more of the state network code and have a better understanding.

kdeme (approved):

kdeme:
> Note that I also reviewed the usage of CoreDb in the test_state_network test but I don't yet understand the details of that test. I'll update it if possible once I've implemented more of the state network code and have a better understanding.

Yeah that's a very old test where a JSON genesis file is read and a trie is created from that. Next all the trie nodes are stored in the database and then some retrieval test is done.

Here too that CoreDb could be replaced, but it is fine to do it in an iteration where you look into the tests.

This type of test that starts from this genesis file is however pretty useful as first starting point to start testing the state network I think.

fluffy/network/history/history_network.nim
```diff
@@ -11,7 +11,7 @@ import
   stew/results,
   chronos,
   chronicles,
-  eth/[common/eth_types_rlp, rlp],
+  eth/[common/eth_types_rlp, rlp, trie, trie/db],
```
kdeme, line 14: Can remove the `db/core_db` import lower?

fluffy/tools/beacon_lc_bridge/beacon_lc_bridge.nim
```diff
@@ -41,7 +41,7 @@ import
   chronicles,
   chronos,
   confutils,
-  eth/[keys, rlp],
+  eth/[keys, rlp, trie, trie/db],
```
kdeme, line 44: Can remove the `  ../../nimbus/db/core_db,` import lower?

fluffy/network/history/history_network.nim
web3-developer: Yes, will do. Thanks

fluffy/tools/beacon_lc_bridge/beacon_lc_bridge.nim
web3-developer: Yes, will do.

web3-developer:
> > Note that I also reviewed the usage of CoreDb in the test_state_network test but I don't yet understand the details of that test. I'll update it if possible once I've implemented more of the state network code and have a better understanding.
> 
> Yeah that's a very old test where a JSON genesis file is read and a trie is created from that. Next all the trie nodes are stored in the database and then some retrieval test is done.
> 
> Here too that CoreDb could be replaced, but it is fine to do it in an iteration where you look into the tests.
> 
> This type of test that starts from this genesis file is however pretty useful as first starting point to start testing the state network I think.

I ran into some issues when I tried to update the test. I updated the `genesisToTrie` proc to this:
```
proc genesisToTrie(filePath: string): HexaryTrie =
  let alloc = getGenesisAlloc(filePath)
  toState(alloc)[0].HexaryTrie
```

and I'm getting this error:
```
 [91mERR[0m 2024-04-25 14:21:34.690+08:00 [1mUnable to decode offered Key/Value        [0m topics="[93mportal_state[0m"
none()
```

Basically the test breaks when I try to update it to use a trie state generated directly from the json file. No idea why this is effecting the decoding.

PR #2166, opened 2024-05-06
Add retrieval content validation to state network.

web3-developer:


fluffy/network/state/state_validation.nim
```diff
@@ -0,0 +1,33 @@
+# Fluffy
+# Copyright (c) 2024 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+import eth/common, ./state_content
+
```
web3-developer, line 9: Is there anything else to validate after retrieving a state value apart from checking that the value returned matches the expected hash contained in the key?

fluffy/tests/state_network_tests/test_state_validation.nim
```diff
@@ -0,0 +1,66 @@
...
+      var contentValueRetrieval = SSZ.decode(
+        testData.content_value_retrieval.hexToSeqByte(), AccountTrieNodeRetrieval
+      )
+
+      contentValueRetrieval.node[^1] += 1 # Modify node hash
+
+      check:
+        validateAccountTrieNodeHash(
+          contentKey.accountTrieNodeKey, contentValueRetrieval
+        ) == false
```
web3-developer, line 66: More tests are being added. Currently in progress.

fluffy/network/state/state_validation.nim
kdeme: For the retrieval part, no, I don't think so

kdeme (approved):

PR #2170, opened 2024-05-10
Fluffy state offer validation

web3-developer:


fluffy/network/state/state_validation.nim
```diff
@@ -5,30 +5,192 @@
...
+  (firstNibble.byte, isLeaf, packedNibbles)
+
+proc isValidTrieProof(
+    expectedRootHash: KeccakHash, path: Nibbles, proof: TrieProof
+): bool =
+  if proof.len() == 0:
+    return false
+
+  if not isValidTrieNode(expectedRootHash, proof[0]):
+    return false
```
web3-developer, line 56: @kdeme Do you think it's worth adding a warning log to each of the locations where validation can fail? Might reduce readability but could help when debugging issues down the track.

kdeme: This would be mostly for the case if during initial debug / interop with other clients, there is invalid data provided over the network?
I think in that case, we can log ore verbose for now (`warning` or `error` log level) and then once things have mostly settled, move them to `debug` log level.

To make that an easier process it might be a better idea to use a `Result[void, string]` instead of the `bool` and do the logging at a higher level.

web3-developer: Yes that's right, during integration testing it would be helpful. Ok, I'll refactor to use the Result type then log the error message.

fluffy/network/state/state_validation.nim
```diff
@@ -5,30 +5,197 @@
...
+  ../../common/common_types,
+  ./state_content
+
+# private functions
+
+proc hashEquals(value: TrieNode | Bytecode, expectedHash: KeccakHash): bool {.inline.} =
+  keccakHash(value.asSeq()) == expectedHash
+
+proc isValidNextNode(nodeRlp: Rlp, rlpIdx: int, nextNode: TrieNode): bool =
+  let nextHashRlp = nodeRlp.listElem(rlpIdx)
```
kdeme, line 21: I think before doing `listElem` an `isList()` check is required (there is a ` doAssert self.isList()` in there, yes, perhaps not the best API).

Or are we certain that these RLP bytes are a list at this point?

fluffy/tests/state_network_tests/test_state_validation.nim
```diff
@@ -137,7 +156,467 @@ suite "State Validation":
...
+        res.error() == "hash of fetched bytecode doesn't match the expected code hash"
+
+  # Account offer validation tests
+
+  test "Validate valid AccountTrieNodeOffer nodes":
+    const file = testVectorDir / "account_trie_node.yaml"
+    const stateRoots = [
+      "0x1ad7b80af0c28bc1489513346d2706885be90abb07f23ca28e50482adb392d61".hexToSeqByte(),
+      "0x1ad7b80af0c28bc1489513346d2706885be90abb07f23ca28e50482adb392d61".hexToSeqByte(),
+      "0xd7f8974fb5ac78d9ac099b9ad5018bedc2ce0a72dad1827a1709da30580f0544".hexToSeqByte(),
```
kdeme, line 172: So the state roots are missing data from the yaml test vectors?

fluffy/tests/state_network_tests/test_state_validation_json.nim
```diff
@@ -0,0 +1,142 @@
+# Nimbus
```
kdeme, line 1: Perhaps a better name for this file would be `test_state_validation_genesis.nim`

fluffy/tests/state_network_tests/test_state_validation_json.nim
```diff
@@ -0,0 +1,142 @@
...
+          )
+          proofResult = validateOfferedContractTrieNode(
+            accountState.rootHash(), contractTrieNodeKey, contractTrieOffer
+          )
+        check proofResult.isErr()
+
+suite "State Proof Verification Tests":
+  let genesisFiles = [
+    "berlin2000.json", "calaveras.json", "chainid1.json", "chainid7.json",
+    "devnet4.json", "devnet5.json", "holesky.json", "mainshadow1.json", "merge.json",
```
kdeme, line 129: You got all these testnet genesis files from https://github.com/eth-clients repos I assume?

I thought we had access to most of these in the vendors git submodules, but apparently not. That's only the case for nimbus-eth2 currently.

kdeme (approved):

fluffy/network/state/state_network.nim
```diff
@@ -13,6 +13,7 @@ import
   eth/common,
   eth/p2p/discoveryv5/[protocol, enr],
   ../../database/content_db,
+  ../history/[history_network],
```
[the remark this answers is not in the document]
web3-developer: No problem, will update.

fluffy/network/state/state_network.nim
```diff
@@ -137,34 +144,57 @@ proc getContent*(n: StateNetwork, key: ContentKey): Future[Opt[seq[byte]]] {.asy
...
-  of accountTrieNode:
-    validateAccountTrieNode(n, contentKey, contentValue)
-  of contractTrieNode:
-    validateContractTrieNode(n, contentKey, contentValue)
-  of contractCode:
-    validateContractCode(n, contentKey, contentValue)
+): Future[Result[void, string]] {.async.} =
+  doAssert(contentKey.contentType == contentValue.contentType)
+
+  let res =
```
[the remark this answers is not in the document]
web3-developer: Good point. I believe I refactored this one, there was previously some code before the return. I'll clean it up.

fluffy/network/state/state_network.nim
```diff
@@ -266,22 +301,25 @@ proc processContentLoop(n: StateNetwork) {.async.} =
...
-
-          await gossipContent(
-            n.portalProtocol, maybeSrcNodeId, contentKey, decodedKey, contentValue,
-            decodedValue,
-          )
-        else:
-          error "Received offered content failed validation", contentKey
+
+        let res = await n.validateContent(decodedKey, decodedValue)
+        res.isOkOr:
```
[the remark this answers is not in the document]
web3-developer: Yep that works

fluffy/network/state/state_validation.nim
web3-developer: We are sure its a list because it only gets called from validateTrieProof which calls listLen() which internally checks if the data is a list. If its not then it return len of zero.

fluffy/network/state/state_validation.nim
```diff
@@ -5,30 +5,197 @@
...
+  if nextHashRlp.isEmpty:
+    return false
+
+  let nextHash = nextHashRlp.toBytes()
+  if nextHash.len() != 32:
+    return false
+
+  # is there a better way to build a KeccakHash from bytes?
```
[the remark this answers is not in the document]
web3-developer: This doesn't work unfortunately because nextHash is a seq and not an array.

fluffy/tests/state_network_tests/test_state_validation_json.nim
web3-developer: No actually these are from the Nimbus-Eth1 test directory here: tests/customgenesis. I just copied a few more over as a part of this change.

fluffy/tests/state_network_tests/test_state_validation_json.nim
web3-developer: Sure, I'll update

fluffy/tests/state_network_tests/test_state_validation.nim
web3-developer: Well the test vectors contain the blockHash inside the encoded offer data which can be used to get the stateRoot in theory. But this test doesn't run the history network and simply passes in the stateRoot so I had to provide it. I can open a PR on the test vectors repo to add them if you think its a good idea?

fluffy/network/state/state_validation.nim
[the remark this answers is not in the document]
kdeme: Ah, right.

Then I think it is better to use `initCopyFrom`: https://github.com/status-im/nim-stew/blob/master/stew/arrayops.nim#L57 or `toArray`: https://github.com/status-im/nim-stew/blob/master/stew/objects.nim#L24

Example usage on `initCopyFrom`: https://github.com/status-im/nimbus-eth1/blob/8767bbd10a880147bc0ac5cf6ab4f01b0419ca06/fluffy/eth_data/era1.nim#L414

fluffy/tests/state_network_tests/test_state_validation_json.nim
kdeme: Ah, ok then.

fluffy/network/state/state_validation.nim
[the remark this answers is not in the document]
web3-developer: Thanks, will do

PR #2180, opened 2024-05-13
Workaround for 0 gas price estimation

etan-status:
The transaction spammer from Kurtosis estimates the gas price of its transactions using 'eth_gasPrice'. Our implementation returns 0 when no transactions have been executed yet, not taking into account the EIP-1559 block base fee. Force a hardcoded minimum for now to unstuck.


jangko:
I have no problem with this one.

Issue #2094, opened 2024-03-22
Give RPC method access to latest state and pending block/tx being assembled in txpool


jangko:
Currently, RPC methods cannot return "pending" block to client.
Transactions in txPool also cannot be returned to client.
Some methods such as `eth_createAccessList` requires latest `nonce` from an account cannot get it from txPool.

jangko:
Also remember to fix #2181 after this issue is solved.

jangko:
fixed by #4687

PR #2181, opened 2024-05-13
Workaround for disallowed transaction superseding

etan-status:
The transaction spammer from Kurtosis keeps spamming transactions with the same nonce because report 'pending' account nocne based on 'latest' rather than actually considering the mempool. To avoid errors from rejecting those, disable the required gas price bump when replacing a transaction with a new nonce.


tersec:
This seems reasonable as a temporary workaround, but have the ethereum-package maintainers commented on whether this is something they consider a bug? Especially in terms of being able to test handling of pathological/malicious conditions, requiring such workarounds seem counterproductive for them.

tersec:
Is this strictly a workaround for Kurtosis behavior which should probably be fixed, or does it point to a `nimbus-eth1` issue? If the former, seems like something the Kurtosis panda ops people should address

jangko:
This will be merged, but a note is added to #2094

PR #2182, opened 2024-05-13
Bump nimbus-eth2 to 87605d08a7f9cfc3b223bd32143e93a6cdf351ac

etan-status:

PR #2185, opened 2024-05-14
Bump nim-libp2p to 21cbe3a91a70811522554e89e6a791172cebfef2

etan-status:


etan-status:
Integrating into #2182

PR #2198, opened 2024-05-20
Unified mode for undumping gzip-ed or era1-ed encoded block dumps

mjfh:
ackn:
  Built on Daniel's work


nimbus/db/era1_db/db_desc.nim
```diff
@@ -0,0 +1,95 @@
...
+  Era1DbBlocks* = object
+    ## Capability of an `era1` file, to be indexed by starting block num
+    fileName*: string                         # File name on disk
+    nBlocks*: uint                            # Number of blocks available
+
+  Era1DbRef* = ref object
+    dir*: string                              # Database folder
+    blocks*: KeyedQueue[uint64,Era1File]      # Era1 block on disk
+    byBlkNum*: SortedSet[uint64,Era1DbBlocks] # File access info
+    ranges*: IntervalSetRef[uint64,uint64]    # Covered ranges
```
arnetheduck, line 35: an era1 file is exactly 8192 blocks (except the lasts one), so this code is not needed - in fact, the era files do not need to be indexed at all on startup (this is very slow) - all that's needed is the list of roots in the accumulator used for verifying era1 consistency (see `getAccumulator` in fluffy code) - from there, the names of the files and the ranges of blocks contained can be computed) cc @daniellm

arnetheduck:
Somewhat unfortunate that the binaries were added to the main repo - we tend to keep large files out from the main repos to keep cloning times reasonable - instead, a git submodule with a dedicated large-files repo should have been used.

nimbus/db/era1_db/db_desc.nim
kdeme: Yeah, this implementation seems overly complex to me. When I was planning on implementing this, it looked like I could do so very similar as `EraDB` is implemented, but with usage of the Accumulator instead.

PR #2200, opened 2024-05-21
Fluffy state network refactor

web3-developer:
Summary of changes:
- Removed unneeded object variant types.
- Moved state_content code into separate files.
- Added helper functions and encode/decode to state_content types.
- Refactored/cleaned up state_network code.
- Moved all recursive gossip code to a separate file state_gossip.


fluffy/network/state/state_gossip.nim
```diff
@@ -0,0 +1,97 @@
...
+  ../history/history_network,
+  ../wire/[portal_protocol, portal_stream],
+  ./state_content
+
+export results
+
+logScope:
+  topics = "portal_state"
+
+proc gossipOffer*(
```
kdeme, line 25: `gossipOffer` is perhaps not the best name for this call? Perhaps just `recursiveGossip` as it is also named in https://github.com/ethereum/portal-network-specs/blob/master/state-network.md#gossip ?

web3-developer:
> Great refactor, looks much cleaner! :shipit:

Thanks!

fluffy/network/state/state_gossip.nim
web3-developer: Sure, I'll review the naming in my next change when I implement the full recursive gossip.

PR #2205, opened 2024-05-22
trivial memory-based speedups

arnetheduck:
* HashKey becomes non-ref
* use openArray instead of seq in lots of places
* avoid sequtils.reversed when unnecessary
* add basic perf stats to test_coredb


arnetheduck:
fwiw, this represents ~40% time savings on block import

mjfh:
> fwiw, this represents ~40% time savings on (early) block import
cool

PR #2209, opened 2024-05-22
Attempt to roll back stateless mode implementation in a single PR

mjfh:
why:
+ Stateless mode is not fully working and in the way
+ Single PR should make it feasible to investigate for a possible re-implementation


nimbus/core/executor/process_block.nim
```diff
@@ -28,10 +28,12 @@ import
...
-proc processTransactions*(vmState: BaseVMState;
-                          header: BlockHeader;
-                          transactions: seq[Transaction]): Result[void, string]
-    {.gcsafe, raises: [CatchableError].} =
+proc processTransactions*(
+    vmState: BaseVMState;
+    header: BlockHeader;
+    transactions: seq[Transaction];
+      ): Result[void, string]
+      {.gcsafe, raises: [CatchableError].} =
```
arnetheduck, line 36: where is this strange code formatting coming from, and why is it part of this revert? This is not how we format them, we should generally move in the direction of the style guide / nep1 / nph.

PR #2210, opened 2024-05-23
Implementation of Fluffy state network gossip.

web3-developer:
Summary of changes:
- Implemented `getParent` helper functions to calculate the parent path and proof for recursive gossip.
- We now handle cases where multiple nibbles should be removed when calculating the parent path for recursive gossip.
- Fixed/re-implemented recursive gossip for account trie proof offers. 
- Implemented recursive gossip for contract trie proof offers.
- Now always gossip the received offer after validation.
- Some other minor refactors/cleanups.
- Moved some helper functions to state_utils.nim
- Implemented typed lookups for getting state network content: `getAccountTrieNode`, `getContractTrieNode` and `getContractCode`

Tests still in progress.

### DIFF
--- a/fluffy/network/state/state_gossip.nim
+++ b/fluffy/network/state/state_gossip.nim
@@ -9,89 +9,122 @@ import
   results,
   chronos,
   chronicles,
-  eth/common/eth_hash,
   eth/common,
-  eth/p2p/discoveryv5/[protocol, enr],
-  ../../database/content_db,
-  ../history/history_network,
-  ../wire/[portal_protocol, portal_stream],
-  ./state_content
+  ../wire/portal_protocol,
+  ./state_content,
+  ./state_utils
 
-export results
+export results, state_content
 
 logScope:
   topics = "portal_state"
 
+func getParent(nibbles: Nibbles, proof: TrieProof): (Nibbles, TrieProof) =
+  doAssert(nibbles.len() > 0, "nibbles too short")
+  doAssert(proof.len() > 1, "proof too short")
+
+  let
+    parentProof = TrieProof.init(proof[0 ..^ 2])
+    parentEndNode = rlpFromBytes(parentProof[^1].asSeq())
+
+  # the trie proof should have already been validated when receiving the offer content
+  doAssert(parentEndNode.listLen() == 2 or parentEndNode.listLen() == 17)
+
+  var unpackedNibbles = nibbles.unpackNibbles()
+
+  if parentEndNode.listLen() == 17:
+    # branch node so only need to remove a single nibble
+    unpackedNibbles.setLen(unpackedNibbles.len() - 1)
+    return (unpackedNibbles.packNibbles(), parentProof)
+
+  # leaf or extension node so we need to remove one or more nibbles
+  let (_, isEven, prefixNibbles) = decodePrefix(parentEndNode.listElem(0))
+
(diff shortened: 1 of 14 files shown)


kdeme (approved):
lgtm, fine with adding tests in follow-up PR also. As you see fit.

fluffy/network/state/state_gossip.nim
```diff
@@ -9,89 +9,122 @@ import
...
+    (parentNibbles, parentProof) = getParent(key.path, offer.storageProof)
+    parentKey = ContractTrieNodeKey.init(
+      key.address, parentNibbles, keccakHash(parentProof[^1].asSeq())
+    )
+    parentOffer =
+      ContractTrieNodeOffer.init(parentProof, offer.accountProof, offer.blockHash)
+
+  (parentKey, parentOffer)
+
 proc gossipOffer*(
```
web3-developer, line 74: @kdeme By the way I opted to leave the name as gossipOffer because in the case of bytecode there is no recursive gossip. The more generic name makes sense to me for this reason.

fluffy/network/state/state_gossip.nim
```diff
@@ -9,89 +9,122 @@ import
...
-  return
+    srcNodeId: Opt[NodeId],
+    keyBytes: ByteList,
+    offerBytes: seq[byte],
+    key: ContractTrieNodeKey,
+    offer: ContractTrieNodeOffer,
+) {.async.} =
+  asyncSpawn p.neighborhoodGossipDiscardPeers(
+    srcNodeId, ContentKeysList.init(@[keyBytes]), @[offerBytes]
+  )
```
web3-developer, line 107: @kdeme In this code I'm gossiping out the offer without await and then gossiping the parent offer below. Is there reason that you can think of to try and enforce the order of gossip by using an await here?

fluffy/network/state/state_gossip.nim
kdeme: Yeah, that's fine. The call is practically also doing more than just the recursive gossip as it also does the regular gossip.

fluffy/network/state/state_gossip.nim
```diff
@@ -9,89 +9,122 @@ import
...
-    encodedKey =
-      ContentKey(accountTrieNodeKey: updatedKey, contentType: accountTrieNode).encode()
-
-  await p.neighborhoodGossipDiscardPeers(
-    maybeSrcNodeId, ContentKeysList.init(@[encodedKey]), @[encodedValue]
+  let (parentKey, parentOffer) = getParent(key, offer)
+  asyncSpawn p.neighborhoodGossipDiscardPeers(
+    srcNodeId,
+    ContentKeysList.init(@[parentKey.toContentKey().encode()]),
+    @[parentOffer.encode()],
```
kdeme, line 94: Perhaps a little comment in front here that indicates that this is the recursive gossip part.

fluffy/network/state/state_gossip.nim
web3-developer: Ok, I'll leave it for now and review later if and when needed once we have the state network up and running.

fluffy/network/state/state_gossip.nim
web3-developer: Sure, will do

PR #2219, opened 2024-05-25
rm Clique consensus method support and Goerli network

tersec:
- https://github.com/ethereum/go-ethereum/releases/tag/v1.14.0
> Geth v1.14.0 stops automatically constructing the pending block (https://github.com/ethereum/go-ethereum/pull/28623). Performance wise there is a significant cost to creating a potential pending block, yet most node operators do not care about it. Not even validators need the pending block. This work also drops support for mining/signing a block "off-demand" (i.e. not via the engine API, rather by Geth itself), meaning Clique signing is also removed going forward.
>
>    With Clique mining removed, Geth v1.14.0 cannot be a Clique signer any more. This however goes hand-in-hand with the removal or pre-merge chain selection and synchronisation support described earlier.

- https://besu.hyperledger.org/private-networks/how-to/configure/consensus/clique
>  danger
>
> Clique is not suitable for production environments. Use only in development environments.

Regarding Goerli, https://blog.ethereum.org/2023/11/30/goerli-lts-update and https://github.com/status-im/nimbus-eth2/pull/6057


hive_integration/nodocker/engine/engine_env.nim
```diff
@@ -93,7 +91,7 @@ proc newEngineEnv*(conf: var NimbusConf, chainFile: string, enableAuth: bool): E
     chain = newChain(com)
 
   com.initializeEmptyDb()
-  let txPool = TxPoolRef.new(com, conf.engineSigner)
+  let txPool = TxPoolRef.new(com, ZERO_ADDRESS)
```
arnetheduck, line 94: can we get rid of this last arg?

arnetheduck (approved):

hive_integration/nodocker/engine/engine_env.nim
tersec: Sure, as a follow-up PR.

tersec: In particular, it appears, after POW has been removed more thoroughly, because this `miner` variable is also used for that purpose, and some tests still therefore refer to it. This requires more care than POA, because currently it appears to start in POW in several instances, then transition to POS almost, but not quite, immediately on startup.

It might be cleaner to address POW first, but ultimately, either ordering works, as long as there's not an expectation that POW mode produces blocks with non-zero miners in the meantime, which should be fine insofar as it doesn't seem important it produces block at all at this point.

PR #2233, opened 2024-05-28
Implement engine_getClientVersionV1

### LINKED ISSUE #2230: Implement `engine_getClientVersionV1`
https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/identification.md#engine_getclientversionv1

jangko:
fix #2230

### DIFF
--- a/nimbus/version.nim
+++ b/nimbus/version.nim
@@ -1,4 +1,4 @@
-# Copyright (c) 2018-2022 Status Research & Development GmbH
+# Copyright (c) 2018-2024 Status Research & Development GmbH
 # Licensed under either of
 #  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
 #  * MIT license ([LICENSE-MIT](LICENSE-MIT))
@@ -6,7 +6,9 @@
 # This file may not be copied, modified, or distributed except according to
 # those terms.
 
-import strutils
+import
+  strutils,
+  stew/byteutils
 
 const
   NimbusName* = "nimbus-eth1"
@@ -24,7 +26,10 @@ const
   NimbusVersion* = $NimbusMajor & "." & $NimbusMinor & "." & $NimbusPatch
   ## is the version of Nimbus as a string.
 
-  GitRevision* = strip(staticExec("git rev-parse --short HEAD"))[0..5]
+  GitRevisionString* = strip(staticExec("git rev-parse --short HEAD"))
+
+  GitRevisionBytes* = hexToByteArray[4](GitRevisionString)
+
+  GitRevision* = GitRevisionString[0..5]
 
   NimVersion* = staticExec("nim --version | grep Version")
-
--- a/nimbus/rpc/engine_api.nim
+++ b/nimbus/rpc/engine_api.nim
@@ -13,7 +13,8 @@ import
   web3/[conversions, execution_types],
   ../beacon/api_handler,
   ../beacon/beacon_engine,
-  ../beacon/web3_eth_conv
+  ../beacon/web3_eth_conv,
+  ../version
 
 {.push raises: [].}
 
@@ -33,6 +34,7 @@ const supportedMethods: HashSet[string] =
     "engine_forkchoiceUpdatedV3",
     "engine_getPayloadBodiesByHashV1",
     "engine_getPayloadBodiesByRangeV1",
+    "engine_getClientVersionV1",
   ])
 
 # I'm trying to keep the handlers below very thin, and move the
@@ -94,3 +96,13 @@ proc setupEngineAPI*(engine: BeaconEngineRef, server: RpcServer) =
   server.rpc("engine_getPayloadBodiesByRangeV1") do(
       start: Quantity, count: Quantity) -> seq[Option[ExecutionPayloadBodyV1]]:
     return engine.getPayloadBodiesByRange(start.uint64, count.uint64)
+
+  server.rpc("engine_getClientVersionV1") do(version: ClientVersionV1) ->
+                                         seq[ClientVersionV1]:
+    # TODO: what should we do with the `version` parameter?
+    return @[ClientVersionV1(
+      code: "NB",
+      name: NimbusName,
+      version: NimbusVersion,
+      commit: FixedBytes[4](GitRevisionBytes),
+    )]


nimbus/version.nim
```diff
@@ -24,7 +26,11 @@ const
...
-  GitRevision* = strip(staticExec("git rev-parse --short HEAD"))[0..5]
+  # strip    : remove spaces
+  # --short=8: ensure we get 8 chars of commit hash
+  # [0..7]   : remove trailing chars(e.g. on Github Windows CI)
+  GitRevision* = strip(staticExec("git rev-parse --short=8 HEAD"))[0..7]
 
-  NimVersion* = staticExec("nim --version | grep Version")
+  GitRevisionBytes* = hexToByteArray[4](GitRevision)
 
+  NimVersion* = staticExec("nim --version | grep Version")
```
tersec, line 36: `NimMajor`/`NimMinor`/`NimPatch` seem to avoid the need for this (perhaps unfortunately necessary, given the spec) `staticExec` and dependence within the code on specific paths?

jangko: It's only for decoration purpose.

tersec: That's its use, yes, but to source the data, e.g., which looks like
```
Nim Compiler Version 1.6.18 [Linux: amd64]
```
on this system, one could directly use the `NimMajor`, etc constants provided by Nim rather than shelling out to Nim, which introduces various possible failure modes. Especially that pipe to `grep`, that's not particularly portable across systems.

jangko: Okay, will remove the `staticExec`.

nimbus/version.nim
```diff
@@ -1,12 +1,14 @@
...
 # Licensed under either of
 #  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
 #  * MIT license ([LICENSE-MIT](LICENSE-MIT))
 # at your option.
 # This file may not be copied, modified, or distributed except according to
 # those terms.
 
-import strutils
+import
+  strutils,
```
tersec, line 10:
```suggestion
  std/strutils,
```

nimbus/version.nim
tersec: https://github.com/status-im/nimbus-eth1/pull/2233/commits/cca66f0ca6984217a08bf7fdf252864e064cbd06

nimbus/version.nim
tersec: https://github.com/status-im/nimbus-eth1/pull/2233/commits/cca66f0ca6984217a08bf7fdf252864e064cbd06

nimbus/version.nim
```diff
@@ -26,16 +31,11 @@ const
   NimbusVersion* = $NimbusMajor & "." & $NimbusMinor & "." & $NimbusPatch
   ## is the version of Nimbus as a string.
 
-  # strip    : remove spaces
+  # strip: remove spaces
   # --short=8: ensure we get 8 chars of commit hash
-  # [0..7]   : remove trailing chars(e.g. on Github Windows CI)
-  GitRevision* = strip(staticExec("git rev-parse --short=8 HEAD"))
+  # -C sourcePath: get the correct git hash no matter where the current dir is.
+  GitRevision* = strip(staticExec("git -C " & sourcePath & " rev-parse --short=8 HEAD"))
```
tersec, line 37: Might this need something similar to [strutils.escape(...)](https://nim-lang.org/docs/strutils.html#escape%2Cstring%2Cstring%2Cstring) if paths contain spaces or other non-alphanumeric characters?

jangko: no, I've found the problem. To add more spaces, the .git folder is deleted from in Windows CI.

.github/workflows/ci.yml
```diff
@@ -267,7 +267,7 @@ jobs:
           mingw32-make ${DEFAULT_MAKE_FLAGS}
           build/nimbus.exe --help
           # give us more space
-          find . -type d -name ".git" -exec rm -rf {} +
+          # find . -type d -name ".git" -exec rm -rf {} +
```
tersec, line 270: You introduced this last year: https://github.com/status-im/nimbus-eth1/commit/48930970ce9e6ac2ce6ac4865b9f139aebbc1d12

What were the conditions then such that it made sense, and have they persisted?

tersec:
General question, if someone has a non-`git` source archive which is otherwise buildable (perhaps they use GitHub's "Download ZIP" functionality, or perhaps they download a release source tarball), and run this, it seems like they'll get something similar to what
```nim
const foo = staticExec("git -C /tmp rev-parse --short=8 HEAD") # an extant directory, but not a git repo directory
echo "commithash is: " & foo
```
produces:
```
fatal: not a git repository (or any of the parent directories): .git
```

This seems less than optimal.

.github/workflows/ci.yml
jangko: iirc at that time windows runner have less than 14GB of disk space, although the documentation said all OS are given the same disk space. Don't know today.

tersec: As long as it the CI builds still work, however much disk space the CI runners have, not an issue, then. If reverting this change triggers disk exhaustion in Windows CI runners, can try other approaches.

nimbus/version.nim
tersec: The .git folder being deleted in the Windows CI evidently indeed was the key issue here, but this still won't work when the `sourcePath` contains spaces or similar characters.

nimbus/version.nim
```diff
@@ -1,12 +1,33 @@
...
+import
+  std/[strutils, os, sequtils],
+  stew/byteutils
+
+const
+  sourcePath  = currentSourcePath.rsplit({DirSep, AltSep}, 1)[0]
+  nimbusRevision {.strdefine.} = "00000000"
+
+static:
+  doAssert(nimbusRevision.len == 8, "nimbusRevision must be consist of 8 characters")
```
tersec, line 18:
```suggestion
  doAssert(nimbusRevision.len == 8, "nimbusRevision must consist of 8 characters")
```

tersec: https://github.com/status-im/nimbus-eth1/pull/2233/commits/877325844b28505c4cbcf5db57300729db9b2d0e

nimbus/version.nim
tersec: The current approach isn't sufficient either:

```
$ ls -l /tmp/staging/
total 4
drwxrwxr-x 3 user user 4096 May 28 18:23 'path with " in it'
$ git -C /tmp/staging/path\ with\ \"\ in\ it/ rev-parse --short=8 HEAD
8dc8fb68
$ cat a.nim
const sourcePath = "/tmp/staging/path with \" in it"
from std/os import dirExists
doAssert dirExists(sourcePath)
echo "directory exists named: " & sourcePath
const foo = staticExec("git -C \"" & sourcePath & "\" rev-parse --short=8 HEAD")
echo "commithash is " & foo
$ ./env nim c -r a
...
directory exists named: /tmp/staging/path with " in it
commithash is /bin/sh: 1: Syntax error: Unterminated quoted string
```

tersec:
> Now they will get zero git revision if they compile from source tarball or download github zip. And we can configure the build system to give some number to replace git commit hash.

Sure, as reasonable as anything

nimbus/version.nim
jangko: ha ha forget to escape

tersec: https://github.com/status-im/nimbus-eth1/pull/2233/commits/7e284ddafc27194c800f75a4c7afffedd5323e4f seems to add extra quotation marks now? `strutils.escape` already adds them:
```nim
import std/strutils
echo strutils.escape("some string")
echo strutils.escape("some")
echo strutils.escape("embedded \" character")
```
results in
```
"some string"
"some"
"embedded \" character"
```

jangko: huh?, I thought `strutils.escape` only escape chars, turn out it also add prefix and suffix.

PR #2242, opened 2024-05-30
use --styleCheck:error

tersec:


jangko:
to compile evmc code path, use `-d:evmc_enabled`

tersec:
Yeah, that uncovered one instance of this.

I'm not sure `evmcStatus` is the best one of these, and it's unfortunate Nim doesn't have a more direct way of allowing snake/camel cases within `--styleCheck:error`, but ensuring `grep`/`rg`/`ag`/GitHub web UI/etc will find symbols properly is worthwhile enough to change one symbol.

Open to suggestions of better alternatives to `evmcStatus`.

tersec:
I'm also not sure why the Fluffy CI sometimes runs and sometimes doesn't; presumably it's detecting the lack of applicable changes or similar, but `config.nims` in the root should affect it. It's possible that there are instances there, since it imports `nimbus-eth2`, which definitely can't directly use `--styleCheck:error` in places where the CL and EL worlds merge, especially around the engine API implementation, because precisely of this camel/snake case distinction between them.

kdeme:
> 'm also not sure why the Fluffy CI sometimes runs and sometimes doesn't; presumably it's detecting the lack of applicable changes or similar, but `config.nims` in the root should affect it.

Yes, that's selected here in the CI: https://github.com/status-im/nimbus-eth1/blob/master/.github/workflows/fluffy.yml#L12-L21

Seems like I never added config.nims in there.

kdeme:
FYI, Fluffy's code will still fail on styleCheck error currently. But because of https://github.com/status-im/nimbus-eth1/blob/master/fluffy/nim.cfg#L17, this change will still make it pass.

I can fix the style check errors separately .

tersec:
Ok, merging, and the variable name can be updated later if someone prefers. The rest is fairly mechanical.

PR #2247, opened 2024-05-30
Use --styleCheck:error for Fluffy + fixes

kdeme:
There seems to be a clash between the names of an object field and a proc here. As workaround names of the procs are changed.

### DIFF
--- a/fluffy/nim.cfg
+++ b/fluffy/nim.cfg
@@ -15,6 +15,6 @@
 -d:PREFER_BLST_SHA256=false
 
 --styleCheck:usages
---styleCheck:hint
+--styleCheck:error
 --hint[ConvFromXtoItselfNotNeeded]:off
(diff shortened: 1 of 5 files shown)


fluffy/nim.cfg
```diff
@@ -15,6 +15,6 @@
 -d:PREFER_BLST_SHA256=false
 
 --styleCheck:usages
```
tersec, line 17: These are already specified in `config.nims` at this point -- is it useful to have them here too?

(Maybe the answer is yes, e.g., it would enable Fluffy to be able to move repos more portably)

kdeme: Yeah, not that useful right now, but somewhat in the case that fluffy needs to be moved out the repo.

I had them in there since long time because fluffy had this added way before it went in the config.nims for the hole repo.

PR #2251, opened 2024-05-30
Aristo avoid storage trie update race conditions

mjfh:


nimbus/db/aristo/TODO.md
```diff
@@ -0,0 +1,9 @@
+* Check whether `HashKey` can be reduced to a simple 32 byte array (see
+  *desc_identifiers.nim*)
```
arnetheduck, line 2: also NibblesSeq

PR #2254, opened 2024-05-30
import: read from era files

arnetheduck:
This PR extends the `nimbus import` command to also allow reading from era files - this command allows creating or topping up an existing database with data coming from era files instead of network sync.

* add `--era1-dir` and `--max-blocks` options to command line
* make `persistBlocks` report basic stats like transactions and gas
* improve error reporting in several API
* allow importing multiple RLP files in one go
* clean up logging options to match nimbus-eth2
* make sure database is closed properly on shutdown

### DIFF
--- a/nimbus/nimbus.nim
+++ b/nimbus/nimbus.nim
@@ -221,6 +221,19 @@ proc run(nimbus: NimbusNode, conf: NimbusConf) =
   when defined(evmc_enabled):
     evmcSetLibraryPath(conf.evm)
 
+  # Trusted setup is needed for processing Cancun+ blocks
+  if conf.trustedSetupFile.isSome:
+    let fileName = conf.trustedSetupFile.get()
+    let res = Kzg.loadTrustedSetup(fileName)
+    if res.isErr:
+      fatal "Cannot load Kzg trusted setup from file", msg=res.error
+      quit(QuitFailure)
+  else:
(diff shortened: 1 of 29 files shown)


arnetheduck:
currently, when running the command twice, it fails sometimes depending on the chunk size, ie `build/nimbus import --era1-dir:/mnt/share/eth --max-blocks=1024` fails on the second launch while `build/nimbus import --era1-dir:/mnt/share/eth --max-blocks:16385` can be resumed.

nimbus/nimbus.nim
```diff
@@ -241,26 +233,29 @@ proc start(nimbus: NimbusNode, conf: NimbusConf) =
...
-    if res.isErr:
-      fatal "Cannot load baked in Kzg trusted setup", msg=res.error
-      quit(QuitFailure)
+  com.initializeEmptyDb()
 
   case conf.cmd
   of NimbusCmd.`import`:
     importBlocks(conf, com)
   else:
+    if conf.trustedSetupFile.isSome:
```
jangko, line 245: `loadKzgTrustedSetup` need to be loaded before importing blocks or subsequent call to kzg functions in block validation  will return error.

arnetheduck: ah - it was explicitly disabled for `import`, for some reason. I'll move it back then.

PR #2256, opened 2024-05-31
drop empty accounts earlier

arnetheduck:
Empty accounts degenerate to `DoNothing` and have no other use, meaning that we can avoid keeping them in memory


arnetheduck:
Maybe useful but also not worth the complexity for now

PR #2266, opened 2024-05-31
Run coredb without journal

mjfh:


nimbus/db/aristo/aristo_blobify.nim
```diff
@@ -420,12 +439,40 @@ proc deblobify*(data: openArray[byte]; vGen: var seq[VertexID]): Result[void,Ari
...
+      ): Result[void,AristoError] =
+  ## De-serialise the last saved state data record previously encoded with
+  ## `blobify()`.
+  if data.len != 73:
+    return err(DeblobWrongSize)
+  if data[^1] != 0x7f:
+    return err(DeblobWrongType)
+  (addr lSst.src.data[0]).copyMem(unsafeAddr data[0], 32)
+  (addr lSst.trg.data[0]).copyMem(unsafeAddr data[32], 32)
+  lSst.serial = uint64.fromBytesBE data[64..72]
```
arnetheduck, line 463: the `[a..b]` call should never be used when a view of the data is sufficient - it allocates a new temporary sequence on the heap - the correct call in cases like this is `toOpenArray` cc @mjfh

PR #2278, opened 2024-06-01
Validate header timestamp in engine_forkchoiceUpdated

### LINKED ISSUE #2273: `engine_forkchoiceUpdatedV1`/`V2`/`V3` incorrectly only check for correct engine API call version when payload attributes present
It does these checks in `validateVersion`, but this is only called when `attrsOpt.isSome`:
https://github.com/status-im/nimbus-eth1/blob/bda760f41dac22f66c9a3ffa56d25bd2f274f4d1/nimbus/beacon/api_handler/api_forkchoice.nim#L188-L193

whereas https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/shanghai.md#specification-1 requires:
> 2. Consensus layer client **MUST** call this method instead of `engine_forkchoiceUpdatedV1` under any of the following conditions:
>    - `headBlockHash` references a block which `timestamp` is greater or equal to the Shanghai timestamp,
>    - `payloadAttributes` is not `null` and `payloadAttributes.timestamp` is greater or equal to the Shanghai timestamp.

There's no check for the `headBlockHash` block `timestamp` mismatching the fork case.

jangko:
fix #2273

### DIFF
--- a/nimbus/beacon/api_handler/api_forkchoice.nim
+++ b/nimbus/beacon/api_handler/api_forkchoice.nim
@@ -57,7 +57,7 @@ template validateHeaderTimestamp(header, com, apiVersion) =
   if com.isCancunOrLater(header.timestamp):
     if apiVersion != Version.V3:
       raise invalidAttr("forkChoiceUpdated" & $apiVersion &
-          " doesn't support head block with Cancun timestamp")
+          " doesn't support head block with timestamp >= Cancun")
   # See fCUV2 specification No.2 bullet 1
   # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/shanghai.md#specification-1
   elif com.isShanghaiOrLater(header.timestamp):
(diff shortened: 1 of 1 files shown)


nimbus/beacon/api_handler/api_forkchoice.nim
```diff
@@ -51,6 +51,24 @@ template validateVersion(attr, com, apiVersion) =
         raise invalidParams("if timestamp is earlier than Shanghai," &
           " payloadAttributes must be PayloadAttributesV1")
 
+template validateHeaderTimestamp(header, com, apiVersion) =
+  # See fCUV3 specification No.2 bullet iii
+  # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/cancun.md#specification-1
+  if com.isCancunOrLater(header.timestamp):
+    if apiVersion != Version.V3:
+      raise invalidAttr("forkChoiceUpdated" & $apiVersion &
+          " doesn't support head block with Cancun timestamp")
```
tersec, line 60: This could be either a Cancun or Prague timestamp, since there's no distinct fcUV4 for Prague

tersec: https://github.com/status-im/nimbus-eth1/pull/2278/commits/0e1d924ca321c11127b5539d6dc01fed6dcf2665

PR #2279, opened 2024-06-01
Speed up account ledger a little

arnetheduck:
`persist` is a hotspot when processing blocks because it is run at least once per transaction and loops over the entire account cache every time.

Here, we introduce an extra `dirty` map that keeps track of all accounts that need checking during `persist` which fixes the immediate inefficiency, though probably this could benefit from a more thorough review - we also get rid of the unused clearCache flag - we start with a fresh cache on every fresh vmState.

* avoid unnecessary code hash comparisons
* avoid unnecessary copies when iterating
* use EMPTY_CODE_HASH throughout for code hash comparison

### DIFF
--- a/hive_integration/nodocker/engine/node.nim
+++ b/hive_integration/nodocker/engine/node.nim
@@ -74,7 +74,7 @@ proc processBlock(
 
   vmState.mutateStateDB:
     let clearEmptyAccount = vmState.determineFork >= FkSpurious
-    db.persist(clearEmptyAccount, ClearCache in vmState.flags)
+    db.persist(clearEmptyAccount)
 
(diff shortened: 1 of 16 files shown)


jangko:
Improvement on my machine: 
* Windows:
  - test_generalstate_json from 2m15s become 1m56s
* WSL Linux:
  - test_generalstate_json from 2m10s become 2m6s

A real linux machine should be faster. But this already set a new record. It never went below 2 minutes before on my machine.

arnetheduck:
> test_generalstate_json from 2m15s become 1m56s

interesting - I would not have expected it to matter on the general state tests since they're mostly small ..

PR #2281, opened 2024-06-02
Kurtosis testing for nimbus eth1 and eth2

### LINKED ISSUE #172: Create a sharable debugging and testing environment for Ethereum
We can create a Vagrant recipe that simplifies the complex setup required for testing Nimbus against other Ethereum clients. 

Such a VM would provide a GUI environment where all kind of debugging and system monitoring tools are pre-installed (e.g. a pre-configured copy of VS Code with tools for debugging Parity, Geth and py-evm, tools such as Wireshark with additional [Ethereum-related dissectors](https://media.consensys.net/releasing-wireshark-dissectors-for-ethereum-%C3%B0%CE%BEvp2p-protocols-215c9656dd9c) and so on).

Over time, the shared effort to improve and document the capabilities of this environment is expected to pay out in increased productivity of the entire team.

advaita-saha:
Fixes #172 
The kurtosis package that will spin up a private Ethereum testnet over Docker or Kubernetes with multi-client support, Flashbot's mev-boost infrastructure for PBS-related testing/validation, and other useful network tools (transaction spammer, monitoring tools, etc). Kurtosis packages are entirely reproducible and composable, so this will work the same way over Docker or Kubernetes, in the cloud or locally on your machine. For more details https://github.com/kurtosis-tech/ethereum-package

This PR adds easier support to spin up test networks and run testnets both locally and over the github CI. For running the testnet and then the checks run the `run-kurtosis-check.sh` script. This will spin up a testnet with your local `nimbus-eth1` build and pull the `nimbus-eth2` build from `statusim` dockerhub, then run the following tests

- Check if atleast one client is ready
- Check if all EL and CL clients are synced
  - Check if CL clients are synced
  - Check if EL clients are synced
- Check consensus finality
- Check consensus attestation stats
- Check consensus reorgs
- Check consensus forks

- Check if all clients propose blocks with legacy EOA transactions
- 10 EOA transactions per block
- check block proposals
- Execute all opcodes as contract deployment 
- Execute all opcodes as contract call
- Check precompile calls
- Check chain stability ( finality checks )

The networks configs can be easily changes and multi-client testing can be introduced in this whole setup only, by just changing the network configs in `kurtosis-network-params.yml`

> NOTE: This kurtosis CI check is now set to fail, when the checks pass


run-kurtosis-check.sh
```diff
@@ -0,0 +1,202 @@
...
+echo "Checking docker installation"
+if command -v docker &> /dev/null; then
+    echo "Docker installation found"
+else
+    echo "Docker installation not found. Please install docker."
+    exit 1
+fi
+
+# Install jq if not installed already
+if [ "$(dpkg -l | awk '/jq/ {print }'|wc -l)" -ge 1 ]; then
```
tersec, line 27: This is not a reliable check for the `jq` package, e.g:
```
$ dpkg -l | awk '/jq/ {print }'
ii  jq                                   1.7.1-3                        amd64        lightweight and flexible command-line JSON processor
ii  jqp                                  0.5.0-1                        amd64        TUI playground to experiment with jq (program)
ii  libjq1:amd64                         1.7.1-3                        amd64        lightweight and flexible command-line JSON processor - shared library
ii  yq                                   3.2.3-1                        all          Command-line YAML processor - jq wrapper for YAML documents
```
in general, it will pick up `jq` as a substring in either the name or one-line description of the package. `jq` is an unusual enough bigram it's not that common, but it happens.

One can use `dpkg -l jq` to query for the `jq` package directly, or other similar more parseable commands, e.g,
```
$ dpkg --get-selections jq
jq						install
```
which doesn't need the `awk`/`grep` step here.

But this still will fail on non-Debian/Ubuntu/Mint/etc systems. Probably better to detect whether a `jq` binary is in the path directly, regardless of potential package source, e.g., `command -v jq` (technically more POSIX-compliant) or `which jq` (more common).

kurtosis-network-params.yml
```diff
@@ -0,0 +1,32 @@
...
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or
+#    http://www.apache.org/licenses/LICENSE-2.0)
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT) or
+#    http://opensource.org/licenses/MIT)
+# at your option. This file may not be copied, modified, or distributed except
+# according to those terms.
+
+participants:
+  - el_type: nimbus
+    el_image: advaitanimbus/nimbus-builder-ci:master
```
tersec, line 13: This is fine for a while, to have this live in a semi-personal Docker account, but at some point would ideally become part of one of the official Docker accounts.

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+#    http://opensource.org/licenses/MIT)
+# at your option. This file may not be copied, modified, or distributed except
+# according to those terms.
+
+FROM debian:testing-slim AS build
+
+SHELL ["/bin/bash", "-c"]
+
+RUN apt-get clean && apt update \
+ && apt -y install build-essential git-lfs librocksdb-dev libpcre3-dev
```
tersec, line 16: Is `libpcre3-dev` actually necessary?

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+ && make -j$(nproc) V=1 LOG_LEVEL=TRACE nimbus
+
+# --------------------------------- #
+# Starting new image to reduce size #
+# --------------------------------- #
+FROM debian:testing-slim as deploy
+
+SHELL ["/bin/bash", "-c"]
+RUN apt-get clean && apt update \
+ && apt -y install build-essential librocksdb-dev libpcre3-dev
```
tersec, line 33: Ditto, `libpcre3-dev`

run-kurtosis-check.sh
```diff
@@ -0,0 +1,202 @@
...
+# Use sed to replace the el_image value in the file
+cat kurtosis-network-params.yml | envsubst > assertoor.yaml
+sed -i "s/el_image: .*/el_image: $new_el_image/" assertoor.yaml
+
+sudo kurtosis run \
+  --enclave nimbus-localtestnet \
+  github.com/kurtosis-tech/ethereum-package \
+  --args-file assertoor.yaml
+
+assertoor_url=$(echo "$enclave_dump" | grep assertoor | grep http | sed 's/.*\(http:\/\/[0-9.:]\+\).*/\1/')
```
tersec, line 50: How does this work? `enclave_dump` is only defined two lines later.

If the variable really is missing, this is one of the reasons people use `set -eu`, so these triggering shows up visibly

tersec:
In general, it's useful to run these shell script through `shellcheck`. Some of what it turns up (useless use of `cat`) is essentially cosmetic/aesthetic, and some of it is in practice ok (using `$foo` rather than `"$foo"` when it's a URL which will in practice typically be well-behaved regarding spaces, etc), but still worth running it and seeing if it's worth addressing, e.g., a sample of its output:
```
In run-kurtosis-check.sh line 42:
cat kurtosis-network-params.yml | envsubst > assertoor.yaml
    ^-------------------------^ SC2002 (style): Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.


In run-kurtosis-check.sh line 72:
docker logs -f $assertoor_container &
               ^------------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
docker logs -f "$assertoor_container" &


In run-kurtosis-check.sh line 76:
  tasks=$(curl -s ${assertoor_url}/api/v1/test_run/$1 | jq -c ".data.tasks[] | {index, parent_index, name, title, status, result}")
                  ^--------------^ SC2086 (info): Double quote to prevent globbing and word splitting.
                                                   ^-- SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
  tasks=$(curl -s "${assertoor_url}"/api/v1/test_run/"$1" | jq -c ".data.tasks[] | {index, parent_index, name, title, status, result}")


In run-kurtosis-check.sh line 80:
  while read task; do
        ^--^ SC2162 (info): read without -r will mangle backslashes.


In run-kurtosis-check.sh line 90:
    if [ ! -z "$task_graph" ]; then
         ^-- SC2236 (style): Use -n instead of ! -z.


In run-kurtosis-check.sh line 111:
  done <<< $(echo "$tasks")
           ^--------------^ SC2046 (warning): Quote this to prevent word splitting.
           ^--------------^ SC2116 (style): Useless echo? Instead of 'cmd $(echo foo)', just use 'cmd foo'.


In run-kurtosis-check.sh line 128:
  tests=$(curl -s ${assertoor_url}/api/v1/test_runs | jq -c ".data[] | {run_id, test_id, name, status}")
                  ^--------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
  tests=$(curl -s "${assertoor_url}"/api/v1/test_runs | jq -c ".data[] | {run_id, test_id, name, status}")


In run-kurtosis-check.sh line 129:
  while read test; do
        ^--^ SC2162 (info): read without -r will mangle backslashes.
```

.dockerignore
```diff
@@ -0,0 +1,10 @@
+# Nimbus
+# Copyright (c) 2018-2024 Status Research & Development GmbH
```
tersec, line 2: None of these files existed before this year, not sure if it makes sense to include that in the copyright year set or just have it only have 2024 for now -- not sure whether the per-file copyright header is supposed to apply more broadly to Nimbus as a whole (which, sure, 2018 sounds right) or these specific files.

run-kurtosis-check.sh
```diff
@@ -0,0 +1,202 @@
...
+    fi
+    
+    echo -e " $(printf '%-4s' "$task_id")\t$task_status\t$task_result\t$(printf '%-50s' "$task_graph$task_name") \t$task_title"
+  done <<< $(echo "$tasks")
+}
+
+# poll & check test status
+final_test_result=""
+failed_test_id=""
+while true
```
tersec, line 117: I guess this is designed for running locally in a loop, rather than a finite number of times in CI

kurtosis-network-params.yml
advaita-saha: This params file is more like a template, and the image names will get overwritten by the script, so we can put any placeholder, doesn't matter

run-kurtosis-check.sh
advaita-saha: Ohhh right, this shouldn't be here

run-kurtosis-check.sh
advaita-saha: Yeah that makes sense.
Shifted to `which`

Dockerfile
advaita-saha: Actually this was suggested by the ethpandaops team during interop. I will check once if it works without this ( should work apparently )

.dockerignore
advaita-saha: Okk then let's keep the copyright year to be only 2024 for now

Dockerfile
arnetheduck: this used to be required by nim-metrics but the pcre dep was entirely removed from there - I think it should be removeable, worth testing

advaita-saha: Not it's not necessary. It can work without PCRE as expected

advaita-saha:
Removed the need for pushing the temporary image built to dockerhub, now it will be able to use a local docker image

run-kurtosis-check.sh
```diff
@@ -0,0 +1,202 @@
...
+
+# The new el_image value
+new_el_image="localtestnet"
+
+# Use sed to replace the el_image value in the file
+cat kurtosis-network-params.yml | envsubst > assertoor.yaml
+sed -i "s/el_image: .*/el_image: $new_el_image/" assertoor.yaml
+
+sudo kurtosis run \
+  --enclave nimbus-localtestnet \
```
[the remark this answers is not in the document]
tersec: After some discussions, this seems fine for now. Might have to be revisited.

tersec:
One general concern is, adding the Kurtosis tests right now means CI will always be failing. While this is to some extent an accurate reflection of, well, `nimbus-eth1` doesn't really function as an EL currently, the CI also functions as

- a set of regression tests so at least what works now doesn't stop working without people noticing; and
- when people look at a list of commits or PRs, it results in everything looking "failing". Which, see above, and it should be made to work when it can be, but until it's actionable information, it just masks other CI failures in these summary views

Merging this PR as-is would hinder these goals. We've used a few approaches here:
- not running it CI (but it's kind of useful to do so, good to make sure it stays working)
- reduce the standards it needs to run, so it runs and meets a lower standard (e.g., just finalizes in a way that's allowed to be optimistic, for example, and does not contain any odd Kurtosis-injected behavior)
- add the moral equivalent of ` || true` to the end of the relevant parts of the scripts, with a noted `TODO` to fix it
- some kind of "skip" status, not sure if that's well supported by GitHub Actions.

So for example
> Check if all clients propose blocks with legacy EOA transactions
>
> 10 EOA transactions per block
>
> check block proposals
>
> Execute all opcodes as contract deployment
>
> Execute all opcodes as contract call
>
> Check precompile calls
>
> Check chain stability ( finality checks )

There's a lot of flexibility here, which is useful -- might be able to find a useful subset of those which work today, and merge it with those working (to get the regression properties in `master`), and then create a PR on top of which adds the next checks one wants to get green in CI, then merge that when it is.

But arguably, just adding guaranteed-fails to the CI is a regression unless otherwise mitigated.

Apropos the the `shellcheck` warnings, at least some of them are worth addressing:

- the specific usage of a `bash` variable before declaration was removed, but that's something I noticed by hand, and it would still be better to add `set -eu` (or the other "bash strict modes" to taste, `set -euo pipefail`, etc, but the specific error would have been caught automatically by `set -eu`)

- some of the things it points out should be quoted probably really should be quoted. Decimal numbers aren't likely to suddenly become `bash`-unfriendly, but `$assertoor_container`, well, yes, that might; there's no intrinsic reason the those container names must be `bash`-friendly, nor is there there an intrinsic reason URLs are. Just an example, but it might be easier to just unconditionally add all the quotes it suggests. Sometimes one is intentionally aiming for some wildcard/globbing behavior, but that doesn't appear to be the case here.

We've run into random "standard task doesn't work when a user on Windows with, horror, spaces in their path tries to build/etc" things it's worth being careful about the quoting. Considered creating a lint job which runs/enforces some version of `shellcheck`.

arnetheduck:
>  CI will always be failing. 
we can run them without reporting the error to the runner - or even beer, report an error when they start working so we know when that happens

tersec:
> > CI will always be failing.
> 
> we can run them without reporting the error to the runner - or even beer, report an error when they start working so we know when that happens

Yeah, especially if there's no useful subset that is interesting. But it depends where it fails. If all but one of the listed Kurtosis checks works, that's might be worth genuinely checking as part of CI. I suspect this is not the case though:
```
time="2024-06-05T21:53:59Z" level=warning msg="upstream client error: beacon node is optimistic, retrying in 10 sec..." client=1-nimbus-nimbus module=consensus
time="2024-06-05T21:53:59Z" level=info msg="Check result: false, Failed Clients: [1-nimbus-nimbus]" RunID=1 TestID=external-0 module=test task=check_consensus_sync_status taskidx=4
time="2024-06-05T21:53:59Z" level=info msg="Check result: false, Failed Clients: [1-nimbus-nimbus]" RunID=1 TestID=external-0 module=test task=check_execution_sync_status taskidx=5
time="2024-06-05T21:54:04Z" level=info msg="Check result: false, Failed Clients: [1-nimbus-nimbus]" RunID=1 TestID=external-0 module=test task=check_consensus_sync_status taskidx=4
time="2024-06-05T21:54:04Z" level=info msg="Check result: false, Failed Clients: [1-nimbus-nimbus]" RunID=1 TestID=external-0 module=test task=check_execution_sync_status taskidx=5
time="2024-06-05T21:54:09Z" level=info msg="Check result: false, Failed Clients: [1-nimbus-nimbus]" RunID=1 TestID=external-0 module=test task=check_consensus_sync_status taskidx=4
time="2024-06-05T21:54:09Z" level=info msg="Check result: false, Failed Clients: [1-nimbus-nimbus]" RunID=1 TestID=external-0 module=test task=check_execution_sync_status taskidx=5
time="2024-06-05T21:54:09Z" level=warning msg="upstream client error: beacon node is optimistic, retrying in 10 sec..." client=1-nimbus-nimbus module=consensus
time="2024-06-05T21:54:14Z" level=info msg="Check result: false, Failed Clients: [1-nimbus-nimbus]" RunID=1 TestID=external-0 module=test task=check_consensus_sync_status taskidx=4
time="2024-06-05T21:54:14Z" level=info msg="Check result: false, Failed Clients: [1-nimbus-nimbus]" RunID=1 TestID=external-0 module=test task=check_execution_sync_status taskidx=5
time="2024-06-05T21:54:19Z" level=info msg="Check result: false, Failed Clients: [1-nimbus-nimbus]" RunID=1 TestID=external-0 module=test task=check_execution_sync_status taskidx=5
time="2024-06-05T21:54:19Z" level=info msg="Check result: false, Failed Clients: [1-nimbus-nimbus]" RunID=1 TestID=external-0 module=test task=check_consensus_sync_status taskidx=4
time="2024-06-05T21:54:19Z" level=warning msg="upstream client error: beacon node is optimistic, retrying in 10 sec..." client=1-nimbus-nimbus module=consensus
```
and since most/all of the listed tests depend on the engine API and non-optimistic block processing to be interesting (in the context of `nimbus-eth1`, whether `nimbus-eth2` can maintain an optimistic sync without an EL isn't especially interesting), and this appears not to be functioning from the beginning, agree that right now running and triggering error on success would be a good approach.

.github/workflows/kurtosis.yml
```diff
@@ -0,0 +1,261 @@
...
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    name: Nimbue eth1 - eth2 compatibility check
```
tersec, line 23:
```suggestion
    name: Nimbus eth1 - eth2 compatibility check
```

.github/workflows/kurtosis.yml
```diff
@@ -0,0 +1,261 @@
...
+        key: ${{ runner.os }}-buildx-${{ github.sha }}
+        restore-keys: |
+          ${{ runner.os }}-buildx-
+
+    - name: Build Docker image
+      shell: bash
+      run: docker build . -t localtestnet
+        
+
+    - name: Setup Kurtosis
```
tersec, line 50:
```suggestion
    - name: Set up Kurtosis
```

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+FROM debian:testing-slim as deploy
+
+SHELL ["/bin/bash", "-c"]
+RUN apt-get clean && apt update \
+ && apt -y install build-essential librocksdb-dev
+RUN apt update && apt -y upgrade
+
+RUN ldd --version ldd
+
+RUN rm -rf /home/user/nimbus-eth1/build/nimbus
```
tersec, line 38: Isn't `/home/user/nimbus-eth1/build/nimbus` a file, not a directory (`rm -r`)?

run-kurtosis-check.sh
```diff
@@ -0,0 +1,253 @@
...
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT) or
+#    http://opensource.org/licenses/MIT)
+# at your option. This file may not be copied, modified, or distributed except
+# according to those terms.
+
+# ------------------------------------------------
+#          Inputs on how to run checks
+# ------------------------------------------------
+echo
+echo -n "Do you want to run the checks in terminal or visit the assertoor URL? (terminal/url) "
```
tersec, line 17: `echo -n` is [not available on FreeBSD](https://man.freebsd.org/cgi/man.cgi?echo) which also notes:
```

       The  newline may	also be	suppressed by appending	`\c' to	the end	of the
       string, as is done by iBCS2 compatible systems.	Note that the  -n  op-
       tion  as	 well as the effect of `\c' are	implementation-defined in IEEE
       Std 1003.1-2001 ("POSIX.1") as amended by Cor. 1-2002.	For  portabil-
       ity, echo should	only be	used if	the first argument does	not start with
       a  hyphen (`-') and does	not contain any	backslashes (`\').  If this is
       not sufficient, printf(1) should	be used.
```

run-kurtosis-check.sh
```diff
@@ -0,0 +1,253 @@
...
+read reply
+
+echo
+echo -n "Build new changes (yes/no)? "
```
tersec, line 21: `echo -n` again; `printf` would be more portable

run-kurtosis-check.sh
```diff
@@ -0,0 +1,253 @@
...
+  done
+
+  # save test results & status to github output
+  echo "test_result=$(echo "$final_test_result")"
+  echo "test_status"
+  for status_line in "${status_lines[@]}"
+  do
+    echo -e "$status_line"
+  done
+  echo ""
```
tersec, line 239: Sometimes this PR uses
```sh
  echo
```
and other times it uses
```sh
  echo ""
```
to echo an empty line. Would probably be preferable to pick one and use it consistently

.github/workflows/kurtosis.yml
```diff
@@ -0,0 +1,261 @@
...
+
+            if [ "$task_result" == "none" ]; then
+              task_result="${GRAY}none   ${NC}"
+            elif [ "$task_result" == "success" ]; then
+              task_result="${GREEN}success${NC}"
+            elif [ "$task_result" == "failure" ]; then
+              task_result="${RED}failure${NC}"
+            fi
+            
+            echo -e " $(printf '%-4s' "$task_id")\t$task_status\t$task_result\t$(printf '%-50s' "$task_graph$task_name") \t$task_title"
```
tersec, line 128: FreeBSD [does not support `echo -e`](https://man.freebsd.org/cgi/man.cgi?echo)

.github/workflows/kurtosis.yml
```diff
@@ -0,0 +1,261 @@
...
+            else
+              status_name="$test_status"
+            fi
+            status_lines+=("  $(printf '%-3s' "$test_id") $status_name \t$test_name")
+            total_tests=$(expr $total_tests + 1)
+          done <<< $(echo "$tests")
+
+          for status_line in "${status_lines[@]}"
+          do
+            echo -e "$status_line"
```
tersec, line 179: `echo -e` nonportable

.github/workflows/kurtosis.yml
advaita-saha: Are these changes necessary?
Since this code is for the github action
And this script will be running inside a docker container ( environment is fixed by us )

Dockerfile
advaita-saha: For some instances it was caching, so just wanted to force remove, so wrote `-rf` as old habbit.
Surely this can be changed to just `-f`

.github/workflows/kurtosis.yml
tersec: Ideally, it also supports running locally outside of the GitHub Action containers, so people can run as much of the CI as feasible locally if they want.

If https://unix.stackexchange.com/questions/730403/echo-e-e-does-not-print-an-escape-from-bash-script-on-macos is to be believed, then the `bash` 3.2.x version shipped by Apple has buggy support for `echo -e` as well, and while we don't really support FreeBSD, we do support macOS in this context, including Docker.

My understanding is that while newer macOS releases have switched to `zsh` by default and should not have these issues, if people upgraded from older macOS releases their shell might not have been switched from `bash` to `zsh`, even if they're on the most current macOS version.

So, they're not critical, no. But my suspicion is that not doing them will lead to bugs on macOS.

advaita-saha: Ok then what would be the better way to handle backlash escapes, in a platform independent manner ?

tersec (approved):

PR #2299, opened 2024-06-05
Core db disable legacy api n remove distinct tries

mjfh:


nimbus/core/chain/persist_blocks.nim
```diff
@@ -167,7 +167,7 @@ proc persistBlocksImpl(c: ChainRef; headers: openArray[BlockHeader];
     c.com.syncCurrent = header.blockNumber
 
     # Done with this block
-    lapTx.commit()
+    # lapTx.commit()
```
web3-developer, line 170: @mjfh Was this a mistake?

Issue #2302, opened 2024-06-05
Implement EIP-2537 for Prague


tersec:
https://eips.ethereum.org/EIPS/eip-2537

tersec:
This largely exists already -- https://github.com/status-im/nimbus-eth1/pull/2296 shows where -- but the gas costs, fork assignment, and test cases need some updating

PR #2315, opened 2024-06-07
EIP-2537 BLS12-381 G1 add/mul/exp and G2 add/mul support with tests

tersec:
https://github.com/status-im/nimbus-eth1/issues/2302

Mostly re-enabling existing code with modified gas costs

Upstream test vectors now live at https://github.com/ethereum/EIPs/tree/master/assets/eip-2537 but there's some slight reformatting already in this repo, so keeping that

These are easier to verify the test cases for. The G2 exp cases are tricker, but they're based on the same base gas cost as G2 mul, so good to have G2 mul in place for regression/consistency checks


nimbus/evm/precompiles.nim
```diff
@@ -35,35 +35,29 @@ type
...
-
-proc getMaxPrecompileAddr(fork: EVMFork): PrecompileAddresses =
+func getMaxPrecompileAddr(fork: EVMFork): PrecompileAddresses =
   if fork < FkByzantium: paIdentity
   elif fork < FkIstanbul: paPairing
-  # EIP 2537: disabled
-  # reason: not included in berlin
-  # elif fork < FkBerlin: paBlake2bf
-  elif fork < FkCancun: paBlake2bf
+  elif fork < FkPrague: paBlake2bf
```
jangko, line 54:
```Nim
elif fork < FkCancun: paBlake2bf
elif fork < FkPrague: paPointEvaluation
```

mratsim:
Would it be also possible to add Mgas/s similar to what I do there: https://github.com/mratsim/constantine/blob/7ea21a2e45c5e02f8aff2cfc1d94cd3191108559/benchmarks/bench_eth_evm_precompiles.nim#L63-L76

Can be tested vs Constantine with

```
git clone https://github.com/mratsim/constantine
cd constantine
CC=clang nimble bench_eth_evm_precompiles
```


The BN254 and BLS12-381 seem significantly mispriced to me given the current state of cryptographic software.

https://ethereum-magicians.org/t/eip-2537-bls12-precompile-discussion-thread/4187/76

tersec:
> Would it be also possible to add Mgas/s similar to what I do there: https://github.com/mratsim/constantine/blob/7ea21a2e45c5e02f8aff2cfc1d94cd3191108559/benchmarks/bench_eth_evm_precompiles.nim#L63-L76
> 
> Can be tested vs Constantine with
> 
> ```
> git clone https://github.com/mratsim/constantine
> cd constantine
> CC=clang nimble bench_eth_evm_precompiles
> ```
> 
> The BN254 and BLS12-381 seem significantly mispriced to me given the current state of cryptographic software.
> 
> https://ethereum-magicians.org/t/eip-2537-bls12-precompile-discussion-thread/4187/76

It seems like various gas prices were already reduced, e.g., G1 add from 600 to 500, G2 add from 4500 to 800, and G2 mul from 55,000 to 45,000, so there's already been some consideration of this? But, alright, this seems reasonable to look at.

nimbus/evm/precompiles.nim
tersec: fixed

PR #2318, opened 2024-06-07
Implementation of procs to support Fluffy state JSON-RPC endpoints.

web3-developer:


fluffy/network/history/history_network.nim
```diff
@@ -610,17 +610,23 @@ proc getEpochAccumulator(
...
       return err("Cannot retrieve epoch accumulator for given block number")
-    blockHash = epoch[epochData.blockRelativeIndex].blockHash
 
+  ok(epoch[epochData.blockRelativeIndex].blockHash)
+
+proc getBlock*(
+    n: HistoryNetwork, bn: UInt256
+): Future[Result[Opt[Block], string]] {.async.} =
+  let
+    blockHash = ?(await n.getBlockHashByNumber(bn))
```
web3-developer, line 629: @kdeme Please let me know if this update to the history network looks ok.

kdeme: Yeah, that's fine.

fluffy/rpc/rpc_eth_api.nim
```diff
@@ -139,7 +139,7 @@ proc installEthApiHandlers*(
 
   rpcServerWithProxy.registerProxyMethod("eth_feeHistory")
 
-  rpcServerWithProxy.registerProxyMethod("eth_getBalance")
+  # rpcServerWithProxy.registerProxyMethod("eth_getBalance")
```
kdeme, line 142: Not really a big deal, as I don't think anybody uses/tests it in that way, but I wanted to note that currently with these registerProxyMethods enabled you can let the requests be forwarded to another endpoint until implemented.

kdeme (approved):

fluffy/rpc/rpc_eth_api.nim
web3-developer: Good point. I'm planning to implement these endpoints shortly but I'll revert this part just to prevent any issues in the mean time.

PR #2319, opened 2024-06-07
Revert "Coredb fix storage tree issues"

tersec:
try to avoid this by a more targeted approach but this is fallback

Reverts status-im/nimbus-eth1#2317

2024-06-07T14:52:16,020152656+02:00

have this ready in case a better approach doesn't happen


tersec:
https://github.com/status-im/nimbus-eth1/pull/2320 will probably suffice to fix this

PR #2320, opened 2024-06-07
re-disable test_tracer_json

tersec:
might be the cause of the CI build failure just introduced


tersec:
sure, works locally, can't be worse than status quo, merging

PR #2328, opened 2024-06-09
avoid initTable

arnetheduck:
`initTable` is obsolete since nim 0.19 and can introduce significant memory overhead while providing no benefit (since the table will be grown to the default initial size on first use anyway).

In particular, aristo layers will not necessarily use all tables they initialize, for exampe when many empty accounts are being created.


tersec:
The CI failures here aren't from this PR, they inherit from `master`

tersec (approved):

kdeme (approved):

PR #2333, opened 2024-06-10
Updates to Fluffy to support hive tests.

bhartnett:
This PR includes the following changes:
- Updated the fluffy-with-portal-hive page of the Fluffy book to align with the updates steps to run hive for portal.
- Add an optional and hidden parameter to Fluffy to support disabling validation of state content against a state root. This is required for now because the portal hive tests don't support setting up history network with block headers data. 
- Updated the portal_stateStore JSON-RPC endpoint so that it will decode content values which are expected to be in the form of a offer. It then converts the offer to an lookup value (without the proof) before storing. No other validations are done here because this is only receiving the data locally and storing locally. Not part of the usual offer flow.


web3-developer:
All the hive state tests are now passing locally after making these changes.

kdeme (approved):

PR #2343, opened 2024-06-12
Bump nim-eth for uTP async raises, part 1

kdeme:
Didn't add latest changes in uTP as there is a lot of breaking stuff in-between to resolve


kdeme:
Closing in favour of the bigger PR that includes this also: https://github.com/status-im/nimbus-eth1/pull/2344

PR #2344, opened 2024-06-13
Bump nim-eth, nim-web3, nimbus-eth2

jangko:
This PR may looks massive, but actually it only deals with: 
- Replace std.Option with results.Opt from nim-web3, but because it affected a lot of areas, then I decide to replace all `Option` usage.
- Fields name changes from nim-eth
- BlockNumber from UInt256 to uint64

While the GasInt conversion to uint64 have been cancelled because it is too complex on its own.

Fluffy failing CI is related to #2343


jangko:
Looks like no more conflicts between commits and base branch.

PR #2351, opened 2024-06-13
Triggered write event for kvt

mjfh:


nimbus/db/core_db/base/base_desc.nim
```diff
@@ -28,6 +28,7 @@ type
     Ooops
     AristoDbMemory            ## Memory backend emulator
     AristoDbRocks             ## RocksDB backend
+    AristoDbDualRocks         ## Dual RocksDB backends for `Kvt` and `Aristo`
```
arnetheduck, line 31: let's not add another backend - we should be running tests in the environment exactly how they will appear in production and this looks like a trap for synchronization bugs because of the non-atomicity between updates.

PR #2352, opened 2024-06-13
Add push raises to Fluffy state network code.

web3-developer:


kdeme (approved):
There are quite a bit of calls that assume proper sanitized RLP input.
I think it would be good to have those well documented as what type of input is expected.

I didn't verify if all that input is checked upon at a higher layer, but even if it is currently, in the future some of these calls might be used wrongly.

web3-developer:
> There are quite a bit of calls that assume proper sanitized RLP input. I think it would be good to have those well documented as what type of input is expected.
> 
> I didn't verify if all that input is checked upon at a higher layer, but even if it is currently, in the future some of these calls might be used wrongly.

Sure, I'll add more details in the comments. Basically the state validation should handle checking the data coming into Fluffy is valid. Once the data is validated then we can assume that these rlp errors should never happen. If they do happen then it means some sort of logic or programmer bug. Even when fetching the data in lookups over the network, we also validate that the data matches the expected hash which means the nodes must be valid. 

Having said all this, the validation code should be more careful when handing these exceptions so I'll put in another change to do that part.

PR #2364, opened 2024-06-14
Implement JSON-RPC endpoints that use state network data.

web3-developer:


fluffy/rpc/rpc_eth_api.nim
```diff
@@ -343,65 +346,125 @@ proc installEthApiHandlers*(
...
+      # TODO: Implement
+      raise newException(ValueError, "tag not yet implemented")
+    else:
+      let
+        blockNumber = quantityTag.number.uint64.u256
+        blockHash = (await historyNetwork.getBlockHashByNumber(blockNumber)).valueOr:
+          raise newException(ValueError, error)
+
+        balance = (await stateNetwork.get().getBalance(blockHash, data.EthAddress)).valueOr:
+          # Should we return 0 here or throw a more detailed error?
```
kdeme, line 372: If I understand it right, there are many errors that can occur here and it would be difficult to be certain that the account simply doesn't exist (and thus return 0).

fluffy/rpc/rpc_eth_api.nim
```diff
@@ -343,65 +346,125 @@ proc installEthApiHandlers*(
...
+    else:
+      let
+        blockNumber = quantityTag.number.uint64.u256
+        blockHash = (await historyNetwork.getBlockHashByNumber(blockNumber)).valueOr:
+          raise newException(ValueError, error)
+
+        nonce = (
+          await stateNetwork.get().getTransactionCount(blockHash, data.EthAddress)
+        ).valueOr:
+          # Should we return 0 here or throw a more detailed error?
```
kdeme, line 400: idem ^

kdeme (approved):

fluffy/rpc/rpc_eth_api.nim
web3-developer: Yeah that was my thought as well.

PR #2368, opened 2024-06-15
Remove AccountStateDB

jangko:
AccountStateDB should no longer be used.
It's usage have been reduce to read only operations. Replace it with LedgerRef to reduce maintenance burden.


nimbus/db/ledger/base/api_tracking.nim
```diff
@@ -78,7 +79,7 @@ type
     LdgAdressesIt              = "addresses"
     LdgCachedStorageIt         = "cachedStorage"
     LdgPairsIt                 = "pairs"
-    LdgStorageIt               = "storage"
+    LdgStorageIt               = "storage"
```
tersec, line 82:
```suggestion
    LdgStorageIt               = "storage"
```

PR #2371, opened 2024-06-15
update newBlockchainTests.md with results of make test

tersec:
This has no code changes at all. It literally is the result of running `git reset --hard && make test` on `master`.

It's a potentially interesting question why all those tests changed from OK to skip, but, that's out of scope of this PR.


jangko:
OK -> Skip: #2346

PR #2380, opened 2024-06-17
avoid computing state root just to know if storage is empty

arnetheduck:
The state root computation here is one of the major hotspots in block processing - in the cases the code only needs to know if it's empty or not, it can be done a lot faster.

Adding a separate function for this looks fragile and should probably be revisited.


arnetheduck:
30% perf increase!

```
blocks: 4750848, baseline: 31h39m56s, contender: 21h29m12s
Time (total): -10h10m43s, -32.14%
```

PR #2382, opened 2024-06-17
Fix lowest-hanging fruit in VM

arnetheduck:
* replace set with bitseq for code validity test
* remove unusued code from CodeStream
* avoid unnecessary byte-by-byte copies


arnetheduck:
These are the trivially fixable ones from https://github.com/status-im/nimbus-eth1/pull/2381

nimbus/db/ledger/accounts_ledger.nim
```diff
@@ -613,7 +613,7 @@ proc clearEmptyAccounts(ac: AccountsLedgerRef) =
     ac.ripemdSpecial = false
 
 proc persist*(ac: AccountsLedgerRef,
-              clearEmptyAccount: bool = false) {.deprecated.} =
```
arnetheduck, line 616: oops from previous PR :)

PR #2398, opened 2024-06-19
Coredb maintenance

mjfh:


nimbus/db/trie_get_branch.nim
```diff
@@ -88,7 +88,12 @@ proc getBranchAux(
     raise newException(RlpError, "node has an unexpected number of children")
 
 proc getBranch*(
-    self: CoreDxPhkRef;
+    # self: CoreDxPhkRef;
```
arnetheduck, line 91: broadly, it's better to document the codebase from a point of view of the current design - historical information belongs in the commit message, but when reading the current code, the assumption is that the past is not known (nobody approaching a codebase first reads the old code and then the new code)

PR #2400, opened 2024-06-19
Update data directory location based on network selected.

web3-developer:


fluffy/fluffy.nim
```diff
@@ -60,6 +60,12 @@ proc onOptimisticHeader(
     when lcDataFork > LightClientDataFork.None:
       info "New LC optimistic header", optimistic_header = shortLog(forkyHeader)
 
+proc getDbDirectory(network: PortalNetwork): string =
+  if network == PortalNetwork.mainnet:
+    "db"
+  else:
+    "db_" & network.symbolName()
```
web3-developer, line 67: @kdeme Note that this will use a different db directory and therefore a different database when using 'none' as the network. Does this sound ok? Will this break any deployments that you are aware of?

kdeme (approved):

PR #2405, opened 2024-06-21
ForkedChain implementation

### LINKED ISSUE #2260: Mising operations of Aristo: handle sidechain in memory
Copied from discord discussion:

One of the test case in test_blockchain_json doing something like this:

`nimbus-eth1/tests/fixtures/eth_tests/BlockchainTests/TransitionTests/bcHomesteadToDao/DaoTransactions.json`
(several other failing test cases also doing similar things)

```
Genesis-D1-D2-D3-D4-D5X-H1-H2-H3-H4-H5X-D5-D6-D7-D8-...-D14
```

where D is default chain, and H is side chain. Block with a 'X' is an invalid block. When executing block H1, there is an error message like this because H1 request for genesis stateRoot but cannot find it anymore.

```
Error: unhandled exception: AccountLedger.init(): RootNotFound(Aristo, ctx=ctx/newColFn(), error=GenericError)
```

comments from @arnetheduck :
This is something we should be able to support with the pure in-memory option - ie as long as D4 is the finalization point, it's supported, but that also means that only D4 has been written to disk.

so it sounds like it's a set of operations that we need are not captured in the current API design, but for the sake of argument, here's how it can be done (based on what I've seen in persistBlocks at least:

when processing D1, we open transaction T1
then for D2..D15, we open T1.2..T1.15 - ie we never commit T1 - this is the state in which persistBlocks is iterating over the blocks we give it
in this state, we can still process forks because we've not committed T1 and we have all the "diff" layers in memory

Let's say that D4 is finalized. A naive way to implement this is:
roll back all transactions including T1
re-play D1..D4 and commit that
re-play D5..D15 and the remaining branches without committing them

then when D8 is finalized, repeat - this time, H1 is not replayed, because it forks off a finalized section of the graph

jangko:
- revamp test_blockchain_json using ForkedChain
- re-enable previously failing test cases.

No optimization

The new test_blockchain_json total run time is not slower than the old one although we employ chain replay. Probably because we also skip descendants of invalid chains altogether instead of try to validate/execute it.

Also the_blockchain_json not relying on total difficulty/chain score anymore even when testing pre PoS chains.

~note: uncles, withdrawals, transactions, and receipts are not written to database.~

fix #2260

### DIFF
--- a/nimbus/core/chain/forked_chain.nim
+++ b/nimbus/core/chain/forked_chain.nim
@@ -130,9 +130,10 @@ proc validateBlock(c: var ForkedChain,
 
   ok()
 
-proc replaySegment(c: var ForkedChain, head: Hash256) =
+proc replaySegment(c: var ForkedChain, target: Hash256) =
+  # Replay from base+1 to target block
   var
-    prevHash = head
+    prevHash = target
     chain = newSeq[EthBlock]()
 
   while prevHash != c.baseHash:
@@ -147,8 +148,9 @@ proc replaySegment(c: var ForkedChain, head: Hash256) =
       updateCursor = false).expect("have been validated before")
     c.cursorHeader = chain[i].header
 
-proc writeBaggage(c: var ForkedChain, blockHash: Hash256) =
-  var prevHash = blockHash
+proc writeBaggage(c: var ForkedChain, target: Hash256) =
+  # Write baggage from base+1 to target block
+  var prevHash = target
   while prevHash != c.baseHash:
     let blk =  c.blocks[prevHash]
     c.db.persistTransactions(blk.blk.header.number, blk.blk.transactions)
@@ -257,8 +259,10 @@ proc importBlock*(c: var ForkedChain, blk: EthBlock): Result[void, string] =
     return err("Block is not part of valid chain")
 
   # TODO: If engine API keep importing blocks
-  # but not finalized it, e.g. current chain length > 128 blocks
-  # Should we move the base forward?
+  # but not finalized it, e.g. current chain length > StagedBlocksThreshold
+  # We need to persist some of the in-memory stuff
+  # to a "staging area" or disk-backed memory but it must not afect `base`.
+  # `base` is the point of no return, we only update it on finality.
 
   c.replaySegment(header.parentHash)
   c.validateBlock(c.cursorHeader, blk)
--- a/nimbus/core/chain/forked_chain.nim
+++ b/nimbus/core/chain/forked_chain.nim
@@ -18,7 +18,7 @@ import
   ../executor/process_block
 
 type
-  HeadDesc = object
+  CursorDesc = object
     number: BlockNumber
     hash: Hash256
 
@@ -35,11 +35,11 @@ type
     db: CoreDbRef
     com: CommonRef
     blocks: Table[Hash256, BlockDesc]
-    headHash: Hash256
     baseHash: Hash256
     baseHeader: BlockHeader
-    headHeader: BlockHeader
-    heads: seq[HeadDesc]
+    cursorHash: Hash256
+    cursorHeader: BlockHeader
+    cursorHeads: seq[CursorDesc]
 
 # ------------------------------------------------------------------------------
 # Private
@@ -75,40 +75,40 @@ proc processBlock(c: ForkedChain,
 
   ok(move(vmState.receipts))
 
-proc updateHeads(c: var ForkedChain,
+proc updateCursorHeads(c: var ForkedChain,
                  hash: Hash256,
                  header: BlockHeader) =
-  for i in 0..<c.heads.len:
-    if c.heads[i].hash == header.parentHash:
-      c.heads[i] = HeadDesc(
+  for i in 0..<c.cursorHeads.len:
+    if c.cursorHeads[i].hash == header.parentHash:
+      c.cursorHeads[i] = CursorDesc(
         hash: hash,
         number: header.number,
       )
       return
 
-  c.heads.add HeadDesc(
+  c.cursorHeads.add CursorDesc(
     hash: hash,
     number: header.number,
   )
 
-proc updateHead(c: var ForkedChain,
+proc updateCursor(c: var ForkedChain,
                 blk: EthBlock,
                 receipts: sink seq[Receipt]) =
   template header(): BlockHeader =
     blk.header
 
-  c.headHeader = header
-  c.headHash = header.blockHash
-  c.blocks[c.headHash] = BlockDesc(
+  c.cursorHeader = header
+  c.cursorHash = header.blockHash
+  c.blocks[c.cursorHash] = BlockDesc(
     blk: blk,
     receipts: move(receipts)
   )
-  c.updateHeads(c.headHash, header)
+  c.updateCursorHeads(c.cursorHash, header)
 
-proc validatePotentialHead(c: var ForkedChain,
+proc validateBlock(c: var ForkedChain,
           parent: BlockHeader,
           blk: EthBlock,
-          updateHead: bool = true)  =
+          updateCursor: bool = true): Result[void, string] =
   let dbTx = c.db.newTransaction()
   defer:
     dbTx.dispose()
@@ -116,11 +116,13 @@ proc validatePotentialHead(c: var ForkedChain,
   var res = c.processBlock(parent, blk)
   if res.isErr:
     dbTx.rollback()
-    return
+    return err(res.error)
 
   dbTx.commit()
-  if updateHead:
-    c.updateHead(blk, move(res.value))
+  if updateCursor:
+    c.updateCursor(blk, move(res.value))
+
+  ok()
 
 proc replaySegment(c: var ForkedChain,
                    head: Hash256) =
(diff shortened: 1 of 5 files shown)


nimbus/core/chain/forked_chain.nim
```diff
@@ -0,0 +1,177 @@
...
+  let headHash = head.blockHash
+  result.head = headHash
+  result.base = headHash
+
+proc addBlock*(c: var ForkedChain, blk: EthBlock) =
+  template header(): BlockHeader =
+    blk.header
+
+  if header.parentHash == c.head:
+    c.validatePotentialHead(blk)
```
arnetheduck, line 132: pass in the parent header here so that it does not have to be loaded from the database - this way, we never have to load parent header from database when processing blocks _except_ to load the `base` header (on startup and when updating the base) - all blocks will be descendants of the `base` header _or_ descendants of one of the blocks in `c.blocks`

arnetheduck: this also prevents an important abuse early: that of trying to process a block that is older than `base` or from an unknown chain

nimbus/core/chain/forked_chain.nim
```diff
@@ -0,0 +1,177 @@
...
+  c.stagingTx = c.db.newTransaction()
+  for i in countdown(chain.high, chain.low):
+    c.validatePotentialHead(chain[i], updateHead = false)
+
+  chain[^1].header.number
+
+proc updateBase(c: var ForkedChain,
+                head: Hash256, headBlockNumber: BlockNumber) =
+  c.base = head
+  c.head = head
```
arnetheduck, line 110: updating the `base` should not update the head. 

typically, the base is a block that is 128 blocks behind head - if there is no finality, it is older - in particular, when we wire this to the json-rpc, we will be able to query state information newer than the base but not older

nimbus/core/chain/forked_chain.nim
```diff
@@ -0,0 +1,177 @@
...
+
+  if header.parentHash notin c.blocks:
+    # if it's parent is an invalid block
+    # there is no hope the descendant is valid
+    return
+
+  discard c.replaySegment(header.parentHash)
+  c.validatePotentialHead(blk)
+
+proc finalizeSegment*(c: var ForkedChain,
```
arnetheduck, line 149: when we finalize a segment, we should not yet write it to the database - instead, we should compute `baseBlockNumber = min(finalized.blockNumber, head.blockNumber - 128)` - this is the new base - the block at this block number is also what we give to `c.db.persistent` and we should fetch the root by loading the finalized block by hash and then walking the parent chain until we find the base block number (so we don't accidentally make an orphaned block the base).

arnetheduck: another way to look at it: right now, this function should be called `forceUpdateBase` because what it does is to reset the on-disk database to a state where `head`, `finalized` and `base` are the same (as happens on startup) - this is unusual for a "live" chain.

nimbus/core/chain/forked_chain.nim
```diff
@@ -0,0 +1,177 @@
...
+    c.updateBase(finalized, c.headBlockNumber)
+    c.stagingTx = c.db.newTransaction()
+    return ok()
+
+  if finalized notin c.blocks:
+    return err("Finalized head not in segments list")
+
+  c.stagingTx.rollback()
+  c.stagingTx = c.db.newTransaction()
+  let headBlockNumber = c.replaySegment(finalized)
```
arnetheduck, line 168: since here we are replaying the in-memory blocks to make them permanently stored, we could also write tx:s, receipts etc at this point - also, we can remove them from the in-memory `c.blocks` because we will never need to replay that part again

```suggestion
  let newBaseBlockNumber = c.replaySegment(newBase)
  # TODO clear c.blocks of the old blocks
```

nimbus/core/chain/forked_chain.nim
```diff
@@ -0,0 +1,177 @@
...
+
+  c.stagingTx.commit()
+  c.db.persistent(headBlockNumber).isOkOr:
+    return err("Failed to save state: " & $$error)
```
arnetheduck, line 172:
```suggestion
  c.db.persistent(newBaseBlockNumber).isOkOr:
    return err("Failed to save state: " & $$error)
  c.stagingTx.commit()
```

per above discussion, we save the new base to the persistent database - this way, when reloading the state from the database, we load the base state, not the "temporary" head that was part of a fork at some point. Also, we must make sure that the `persistent` call is inside the "permanent" transaction so that it gets written to disk at this point

jangko: ` c.stagingTx.commit()` have to be called before `c.db.persistent` or it will crash.

nimbus/core/chain/forked_chain.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/2405/commits/d40529ebef0d973a322218ed8c71e0fd083eda83

nimbus/core/chain/forked_chain.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/2405/commits/7c679a8ea941a2dc89bd3081d9f6ce4dad80850a

nimbus/core/chain/forked_chain.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/2405/commits/7c679a8ea941a2dc89bd3081d9f6ce4dad80850a

nimbus/core/chain/forked_chain.nim
jangko: write txs, receipts, etc: https://github.com/status-im/nimbus-eth1/pull/2405/commits/73a0f7caeb7d27f8d2bb31dcad7f49e227f71952

remove written blocks from memory: https://github.com/status-im/nimbus-eth1/pull/2405/commits/7c679a8ea941a2dc89bd3081d9f6ce4dad80850a

nimbus/core/chain/forked_chain.nim
```diff
@@ -0,0 +1,285 @@
...
+  result.headHeader = result.baseHeader
+
+proc addBlock*(c: var ForkedChain, blk: EthBlock) =
+  if c.stagingTx.isNil:
+    c.stagingTx = c.db.newTransaction()
+
+  template header(): BlockHeader =
+    blk.header
+
+  if header.parentHash == c.headHash:
```
arnetheduck, line 208: In the execution api, `newPayload` does not set the head - I think it's best we model this function (and all tests) around that idea that setting the head is always a separate action (aka forkChoiceUpdated - cc @tersec) - this will ensure that testing corresponds to "real-world" usage of the API.

tersec: Indeed, `newPayload` does not and should not set the head. `forkchoiceUpdated` is when/how a CL commits to an execution payload/EL block. `newPayload` does not, and cannot, represent any kind of commitment to this block by the CL except that it might be superficially valid from the CL perspective.

jangko: https://github.com/status-im/nimbus-eth1/pull/2405/commits/273fea033b5085de7be70a90b4ff05ffc13f95d8

nimbus/core/chain/forked_chain.nim
```diff
@@ -0,0 +1,325 @@
...
+    return c.validateBlock(c.baseHeader, blk)
+
+  if header.parentHash notin c.blocks:
+    # If it's parent is an invalid block
+    # there is no hope the descendant is valid
+    return err("Block is not part of valid chain")
+
+  # TODO: If engine API keep importing blocks
+  # but not finalized it, e.g. current chain length > 128 blocks
+  # Should we move the base forward?
```
arnetheduck, line 261: no - to deal with this properly, we'll need to persist some of the in-memory stuff to a "staging area" or disk-backed memory but it must not affect `base` - base is the point of no return, we only update it on finality.

Also, the 128 should be configurable on startup.

nimbus/core/chain/forked_chain.nim
```diff
@@ -0,0 +1,325 @@
...
+  # and possibly switched to other chain beside the one with cursor
+  doAssert(finalizedHeader.number <= headHeader.number)
+
+  # Write segment from base+1 to finalized into database
+  c.stagingTx.rollback()
+  c.stagingTx = c.db.newTransaction()
+  c.replaySegment(finalizedHash)
+  c.writeBaggage(finalizedHash)
+  c.stagingTx.commit()
+  # Update base forward to finalized
```
arnetheduck, line 311: here, base should move to `min(finalized.blockNumber, head - baseDistance)` where `baseDistance` is the 128 constant - later we'll make that configurable as well - it's important to have base at least that much behind so we can answer state queries about history that deep.

arnetheduck: this probably means moving the computation of the base block up a bit and separating `finalizedHash` and `baseHash` - they are two separate concepts - finalizedHash will typically be 64-95 blocks behind head depending on how the voting goes

nimbus/core/chain/forked_chain.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/2405/commits/8b80c62bfc6ded7fb34e1656a6cb0fbd935c1cd6

nimbus/core/chain/forked_chain.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/2405/commits/d40f007889f5917907d6b533b114a48937b3826d

arnetheduck:
Alright! I think this is enough to cover the MVP needs for this part, we'll obviously need to revisit a lot of it but we can start wiring it up to the execution API and the rest of nimbus :partying_face:

arnetheduck (approved):

PR #2407, opened 2024-06-23
Fix t8n: blobGasUsed exceeds allowance issue

jangko:


tools/t8n/transition.nim
```diff
@@ -252,6 +252,14 @@ proc exec(ctx: var TransContext,
       )
       continue
 
+    blobGasUsed += tx.getTotalBlobGas
+    if blobGasUsed > MAX_BLOB_GAS_PER_BLOCK:
```
arnetheduck, line 256: shouldn't t8n be calling `validateEip4844Header` or something like that so that we use the production code in the test?

arnetheduck: or rather, shouldn't this check be somewhere else? it seems that we could introduce this check in the transaction processing pipeline (since the limit applies to a block, it must also apply to each transaction) - we should fail transaction validation in this case so that we don't process blocks that contain transactions like this

jangko: yes we can use something similar to `gasPool` to check for block gas limit.

jangko: https://github.com/status-im/nimbus-eth1/pull/2407/commits/7ef6317aec9468b131fc0a23c97be00eeba91ef4

PR #2409, opened 2024-06-24
Remove pre and post opcode handlers from EVM

jangko:


arnetheduck:
Fun fact: these no-op:s were about 0.5-1% of execution time when importing blocks

PR #2411, opened 2024-06-24
Clear account cache after each block

arnetheduck:
When processing long ranges of blocks, the account cache grows unbounded which causes huge memory spikes.

Here, we move the cache to a second-level cache after each block - the second-level cache is cleared on the next block after that which creates a simple LRU effect.

There's a small performance cost of course, though overall the freed-up memory can now be reassigned to the rocksdb row cache which not only makes up for the loss but overall leads to a performance increase.

The bump to 2gb of rocksdb row cache here needs more testing but is slightly less and loosely based on the savings from this PR and the circular ref fix in #2408 - another way to phrase this is that it's better to give rocksdb more breathing room than let the memory sit unused until circular ref collection happens ;)

### DIFF
--- a/nimbus/core/executor/process_block.nim
+++ b/nimbus/core/executor/process_block.nim
@@ -124,7 +124,11 @@ proc procBlkEpilogue(
     if vmState.collectWitnessData:
       db.collectWitnessData()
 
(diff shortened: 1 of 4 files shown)


arnetheduck:
Pre:
image


Post:
image

PR #2415, opened 2024-06-26
wip: on-demand mpt construction

arnetheduck:
Given that actual data is stored in the `Vertex` structure, it's useful to think of the MPT as a cache for computing roots rather than being a functional requirement on its own.

This PR engenders this line of thinking by incrementally computing the MPT only when it's needed, ie when a state (or similar) root is needed.

This has the effect of siginficantly reducing memory usage as well as improving performance:

* no need for dirty-mpt-node book-keeping
* no need to build complex forest of upcoming hashing work
* only hashes that are functionally needed are ever computed - intermediate nodes whose MTP root is not observed are never computed / processed


arnetheduck:
Merged as #2426

PR #2424, opened 2024-06-27
Only use pre hashed addresses as account keys

mjfh:


nimbus/db/aristo/aristo_api.nim
```diff
@@ -326,20 +308,14 @@ type
 
   AristoApiMergeStorageDataFn* =
     proc(db: AristoDbRef;
-         stoKey: openArray[byte];
+         accPath: openArray[byte];
```
arnetheduck, line 311: it would probably make sense to change these to pathid if they represent that kind of paths - the usage of openArray here opens up for errors that shouldn't be possible (like giving paths longer than 32 bytes etc)

mjfh: I ran into that. This is checked by default in non-release mode (see `base.nim`.) The problem was rather the use of an `EthAddress` instead of its hash.

mjfh: Is it possible to have a working distinct `openArray[byte]` type?

arnetheduck: > Is it possible to have a working distinct openArray[byte] type?

Not that I know.

Broadly though, it's about leaving a trace of understanding as to what kind of key we're working with - openArray could be anything whereas ethereum paths are distinctly limited

PR #2425, opened 2024-06-28
Combine smaller tests into bigger one

jangko:
1. test_state_db and test_ledger -> test_ledger. They are the same thing now.
2. stack, memory, code_stream, gas_meter, misc, overflow -> test_evm_support. They are small tests and fall into the same area.


arnetheduck:
ideally, our tests would be named after the module they test, makes them a lot easier to find, ie if I'm changing module xxx, I'd go looking for `test_xxx`.

PR #2427, opened 2024-06-28
Bump RocksDb version and enable autoClose on opt types to prevent memory leaks

web3-developer:
It turns out that most of these option types were not being cleaned up. By using the autoClose flag in the library they will all be cleaned up when the database is closed.

autoClose is always false by default so that if a user of an API passes a value they are expected to close the resource when finished. This seems the most intuitive but if we think otherwise for any reason, we could make true the default.

### DIFF
--- a/nimbus/db/core_db/backend/aristo_rocksdb.nim
+++ b/nimbus/db/core_db/backend/aristo_rocksdb.nim
@@ -45,7 +45,7 @@ proc toRocksDb*(
   #      data model itself has settled down as their optimal values will depend
   #      on the shape of the data - it'll also be different per column family..
 
-  let tableOpts = defaultTableOptions()
+  let tableOpts = defaultTableOptions(autoClose = true)
   # This bloom filter helps avoid having to read multiple SST files when looking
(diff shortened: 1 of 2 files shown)


jangko:
Does it have to be explicitly to set `autoClose = true`? What if user forget to set `autoClose = true`, will it leak?

web3-developer:
> Does it have to be explicitly to set `autoClose = true`? What if user forget to set `autoClose = true`, will it leak?

`autoClose = false` is the default so if the parameter is not provided then the user needs to call close on each type manually after usage. If `autoClose = true` then the resources are cleaned up inside the parent type's close function. So there is basically a resource hierarchy ending with the database type at the top which will clean up all resources in the close function if and only if all types have `autoClose = true` set.

I'm still not sure what is the best default behavior. On the one hand, it might be unintuitive to have `autoClose = true` as the default because if you close another type, then the child type is also closed unexpectedly. Making `autoClose = false` the default makes it clear to the user that if they turn the feature on by setting to true then they know and can expect the type to be closed automatically.

PR #2429, opened 2024-06-29
Era support for nimbus import

### LINKED ISSUE #2306: Add era support to block importer
The [current version](https://github.com/status-im/nimbus-eth1/blob/0268093fcc6efecba40301a8e079ed204bbbd3d0/nimbus/nimbus_import.nim#L96) of the block importer only supports era1 files, meaning we can only cover the range of blocks up to the merge on mainnet.

After the merge, we can use consensus layer era blocks instead - this is slightly more involved:

* we must decode the consensus layer block based on its fork schedule, meaning we need access to the consensus layer runtime configuration
* we must map slot number to block number - slot numbers may have gaps in which there is no (execution) block, meaning that the slot will grow faster than the execution block number

We must also take into account networks which don't have a pre-merge period in era1. 

To load era files, we should have a new `--era-dir` option similar to `--era1-dir` - the importer will then load the base state of the database, figure out whether to load era1 or era, then go about topping up blocks the same way as is done with era1.

advaita-saha:
Fixes #2306

### DIFF
--- a/nimbus/nimbus_import.nim
+++ b/nimbus/nimbus_import.nim
@@ -465,16 +465,18 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
         importedSlot = firstSlotAfterMerge
 
       while running and imported < conf.maxBlocks and importedSlot < endSlot:
-        var clblock = getBlockFromEra(
+        let clblock = getBlockFromEra(
           eraDB,
           historical_roots.asSeq(),
           historical_summaries.asSeq(),
           Slot(importedSlot),
           clConfig.cfg,
-        )
-        if clblock.isSome:
-          blocks.add getEth1Block(clblock.get())
-          imported += 1
+        ).valueOr:
+          importedSlot += 1
+          continue
+
+        blocks.add getEth1Block(clblock)
+        imported += 1
 
         importedSlot += 1
         if blocks.lenu64 mod conf.chunkSize == 0:
@@ -502,16 +504,18 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
         importedSlot, dataDir = conf.dataDir.string, eraDir = conf.eraDir.string
 
       while running and imported < conf.maxBlocks and importedSlot <= endSlot:
-        var clblock = getBlockFromEra(
+        let clblock = getBlockFromEra(
           eraDB,
           historical_roots.asSeq(),
           historical_summaries.asSeq(),
           Slot(importedSlot),
           clConfig.cfg,
-        )
-        if clblock.isSome:
-          blocks.add getEth1Block(clblock.get())
-          imported += 1
+        ).valueOr:
+          importedSlot += 1
+          continue
+
+        blocks.add getEth1Block(clblock)
+        imported += 1
 
         importedSlot += 1
         if blocks.lenu64 mod conf.chunkSize == 0 and blocks.len > 0:
--- a/nimbus/nimbus_import.nim
+++ b/nimbus/nimbus_import.nim
@@ -395,7 +395,7 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
 
   if isDir(conf.era1Dir.string):
     doAssert conf.networkId == MainNet or conf.networkId == SepoliaNet,
-      "Only mainnet/sepolia era1 current supported"
+      "Only mainnet/sepolia era1 currently supported"
 
     let
       # TODO the merge block number could be fetched from the era1 file instead,
--- a/nimbus/nimbus_import.nim
+++ b/nimbus/nimbus_import.nim
@@ -47,7 +47,7 @@ proc latestEraFile*(eraDir: string, cfg: RuntimeConfig): Result[(string, Era), s
       if parts.len() == 3 and parts[0] == cfg.CONFIG_NAME:
         let era =
           try:
-            parseInt(parts[1])
+            parseBiggestInt(parts[1])
           except ValueError:
             return err("Invalid era number")
         if era > latestEra:
--- a/nimbus/nimbus_import.nim
+++ b/nimbus/nimbus_import.nim
@@ -276,37 +276,56 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
       boolFlag({PersistBlockFlag.NoFullValidation}, not conf.fullValidation) +
       boolFlag(NoPersistBodies, not conf.storeBodies) +
       boolFlag({PersistBlockFlag.NoPersistReceipts}, not conf.storeReceipts)
-    clConfig =
-      if conf.networkId == HoleskyNet:
-        getMetadataForNetwork("holesky")
-      elif conf.networkId == SepoliaNet:
-        getMetadataForNetwork("sepolia")
-      elif conf.networkId == MainNet:
-        getMetadataForNetwork("mainnet")
-      else:
-        error "Unsupported network", network = conf.networkId
-        quit(QuitFailure)
-    genesis_validators_root =
-      if conf.networkId == MainNet:
-        Eth2Digest.fromHex(
-          "0x4b363db94e286120d76eb905340fdd4e54bfe9f06bf33ff6cf5ad27f511bfe95"
-        ) # Mainnet Validators Root
-      elif conf.networkId == HoleskyNet:
-        Eth2Digest.fromHex(
-          "0x9143aa7c615a7f7115e2b6aac319c03529df8242ae705fba9df39b79c59fa8b1"
-        ) # Holesky Validators Root
-      elif conf.networkId == SepoliaNet:
-        Eth2Digest.fromHex(
-          "0xd8ea171f3c94aea21ebc42a1ed61052acf3f9209c00e4efbaaddac09ed9b8078"
-        ) # Sepolia Validators Root
-      else:
-        error "Unsupported network", network = conf.networkId
-        quit(QuitFailure)
     blocks: seq[EthBlock]
+    clConfig: Eth2NetworkMetadata
+    genesis_validators_root: Eth2Digest
+    lastEra1Block: uint64
+    firstSlotAfterMerge: uint64
 
   defer:
     if csv != nil:
       close(csv)
+
+  # Network Specific Configurations
+  # TODO: the merge block number could be fetched from the era1 file instead,
+  #       specially if the accumulator is added to the chain metadata
+  if conf.networkId == MainNet:
+    doAssert isDir(conf.era1Dir.string), "Era1 directory not found"
+    clConfig = getMetadataForNetwork("mainnet")
+    genesis_validators_root = Eth2Digest.fromHex(
+      "0x4b363db94e286120d76eb905340fdd4e54bfe9f06bf33ff6cf5ad27f511bfe95"
+    ) # Mainnet Validators Root
+    lastEra1Block = 15537393'u64 # Mainnet
+    firstSlotAfterMerge =
+      if isDir(conf.eraDir.string):
+        4700013'u64 # Mainnet
+      else:
+        notice "No eraDir found for Mainnet, block loading will stop after era1"
+        0'u64 # No eraDir for Mainnet
+  elif conf.networkId == SepoliaNet:
+    doAssert isDir(conf.era1Dir.string), "Era1 directory not found"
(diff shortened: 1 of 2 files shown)


nimbus/nimbus_import.nim
```diff
@@ -202,6 +366,74 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
...
+        importedSlot += 1
+        if blocks.lenu64 mod conf.chunkSize == 0:
+          process()
+      
+      if blocks.len > 0:
+        process()
+
+  if isDir(conf.eraDir.string) and conf.networkId == HoleskyNet:
+    
+    let
```
tersec, line 406: This is almost duplicate code as for mainnet era1 loading (and/vice versa)

nimbus/nimbus_import.nim
```diff
@@ -202,6 +366,74 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
...
+      clConfig = getMetadataForNetwork("holesky")
```
tersec, line 407: This is almost duplicate code as for mainnet era loading (and/vice versa)

tersec: e.g. adding https://sepolia.era.nimbus.team/ era (putting aside era1) importing this way would require another nearly but not quite duplicate 30 lines of code

advaita-saha: Addresses the issue, now `sepolia` doesn't need to be handled separately

nimbus/nimbus_import.nim
```diff
@@ -113,79 +304,96 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
...
+          ].join(",")
+        )
+        csv.flushFile()
+      except IOError as exc:
+        warn "Could not write csv", err = exc.msg
+    blocks.setLen(0)
+
+  ## ##################################################### 
+  ##                                                
+  ##   Networks with pre-merge and post-merge history
```
arnetheduck, line 368: let's maybe avoid the massive comment blocks since they take up a lot of valuable vertical space (it's not _that_ important)

nimbus/nimbus_import.nim
```diff
@@ -83,15 +239,26 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
 
   var
     imported = 0'u64
+    importedSlot = 
+      if conf.csvStats.isSome:
```
arnetheduck, line 243: the csv file is purely for debugging - it should not be read for this purpose

nimbus/nimbus_import.nim
```diff
@@ -83,15 +239,26 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
...
+        try:
+          let file = readFile(conf.csvStats.get())
+          let lines = file.splitLines()
+          lines[lines.len-2].split(",")[2].parseInt().uint64
+        except:
```
arnetheduck, line 248: we don't use `except:` - instead, we catch more specific errors

nimbus/nimbus_import.nim
```diff
@@ -35,6 +38,159 @@ declareCounter nec_imported_gas,
...
+      return err("Unable to read state: " & exc.msg)
+
+  withState(state[]):
+    when consensusFork >= ConsensusFork.Capella:
+      return ok((forkyState.data.historical_roots, forkyState.data.historical_summaries, slot+8192))
+    else:
+      return ok((forkyState.data.historical_roots, HashList[HistoricalSummary, Limit HISTORICAL_ROOTS_LIMIT](), slot+8192))
+
+proc getBlockFromEra*(
+    db: EraDB, historical_roots: openArray[Eth2Digest], historical_summaries: openArray[HistoricalSummary], slot: Slot, cfg: RuntimeConfig): Opt[ForkedTrustedSignedBeaconBlock] =
```
arnetheduck, line 96: if you auto-format this with `nph`, it'll remove these excessively long lines

nimbus/nimbus_import.nim
tersec: https://github.com/status-im/nimbus-eth1/pull/2429/commits/c3e84a1b879d6be534cf8847bcc4e14debc80adc

nimbus/nimbus_import.nim
```diff
@@ -340,21 +363,37 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
...
-    doAssert conf.networkId == MainNet, "Only mainnet era1 current supported"
+    doAssert conf.networkId == MainNet or conf.networkId == SepoliaNet, "Only mainnet/sepolia era1 current supported"
 
-    const
+    let
       # TODO the merge block number could be fetched from the era1 file instead,
       #      specially if the accumulator is added to the chain metadata
-      lastEra1Block = 15537393
-      firstSlotAfterMerge = 4700013
+      lastEra1Block =
```
tersec, line 377: These network-dependent constants would be easier to read if they were consolidated per network, so, e.g., if each network needs
- last Era1 block
- first slot after merge
- genesis validators root

they could be combined into one place each, which would avoid the need for
- multiple `if conf.networkId == MainNet`,
- this construction where `doAssert conf.networkId == MainNet or conf.networkId == SepoliaNet` is what makes the `if mainnet else: it must be sepolia` pattern safe (i.e. it's not "locally" safe, it needs reasoning on a larger scale, without being particularly intrinsically conceptually complicated)

nimbus/nimbus_import.nim
```diff
@@ -408,12 +445,14 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
       if blocks.len > 0:
         process()
 
+  ## ################################################
+  ##                                                
+  ##      Networks with not pre-merge history
```
tersec, line 450: The giant comment block aside, "without pre-merge history"

nimbus/nimbus_import.nim
```diff
@@ -35,6 +38,159 @@ declareCounter nec_imported_gas,
...
+    latestEraFile = ""
+
+  try:
+    for kind, obj in walkDir eraDir:
+      let (_, name, _) = splitFile(obj)
+      let parts = name.split('-')
+      if parts.len() == 3 and parts[0] == cfg.CONFIG_NAME:
+        let era =
+          try:
+            parseInt(parts[1])
```
tersec, line 54: It often makes sense use [`parseBiggestInt`](https://nim-lang.org/docs/strutils.html#parseBiggestInt%2Cstring) instead of `parseInt` because the former isn't dependent on whether the target system is 32-bit or 64-bit, so doesn't invite tricky-to-debug bit size problems. And, for 64-bit platforms, it's the same, so no efficiency loss.

nimbus/nimbus_import.nim
```diff
@@ -35,6 +38,159 @@ declareCounter nec_imported_gas,
...
+    try:
+      newClone(readSszForkedHashedBeaconState(cfg, slot, bytes))
+    except SerializationError as exc:
+      return err("Unable to read state: " & exc.msg)
+
+  withState(state[]):
+    when consensusFork >= ConsensusFork.Capella:
+      return ok((forkyState.data.historical_roots, forkyState.data.historical_summaries, slot+8192))
+    else:
+      return ok((forkyState.data.historical_roots, HashList[HistoricalSummary, Limit HISTORICAL_ROOTS_LIMIT](), slot+8192))
```
tersec, line 93: Era's 8192 slots per file is derived from various constants in the consensus specs, e.g., `SLOTS_PER_HISTORICAL_ROOT`, or `SLOTS_PER_EPOCH * SHARD_COMMITTEE_PERIOD`. Not sure if makes sense to re-derive these now that from this code's perspective, they're just what Era files do, but it's a bit of a magical constant at the moment.

nimbus/nimbus_import.nim
```diff
@@ -113,79 +304,96 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
...
+    blocks.setLen(0)
+
+  ## ##################################################### 
+  ##                                                
+  ##   Networks with pre-merge and post-merge history      
+  ##
+  ## #####################################################
   if isDir(conf.era1Dir.string):
-    doAssert conf.networkId == MainNet, "Only mainnet era1 current supported"
+    doAssert conf.networkId == MainNet or conf.networkId == SepoliaNet, "Only mainnet/sepolia era1 current supported"
```
tersec, line 372:
```suggestion
    doAssert conf.networkId == MainNet or conf.networkId == SepoliaNet, "Only mainnet/sepolia era1 currently supported"
```

nimbus/nimbus_import.nim
```diff
@@ -202,6 +410,76 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
...
+      ).valueOr:
+        error "Error loading historical summaries", error
+        quit QuitFailure
+
+    if importedSlot <= endSlot:
+      notice "Importing era archive HoleskyNet",
+        importedSlot, dataDir = conf.dataDir.string, eraDir = conf.eraDir.string
+
+      while running and imported < conf.maxBlocks and importedSlot <= endSlot:
+        var clblock = getBlockFromEra(
```
tersec, line 470: - why is this `var`, not `let`?
- assuming this is using `Opt`/`Result` as it should be then the `if clblock.isSome`/`clblock.get` can be replaced with more statically-verifiably safer patterns

nimbus/nimbus_import.nim
```diff
@@ -202,6 +410,76 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
...
         process() # last chunk, if any
+    
+    if start > lastEra1Block:
+      doAssert isDir(conf.eraDir.string), "Era directory not found"
+
+      notice "Importing era archive",
+        start, dataDir = conf.dataDir.string, eraDir = conf.eraDir.string
+
+      let
+        eraDB = EraDB.new(
```
tersec, line 421: This still seems like quite a lot of duplicated logic (logging, database opening, the chunk size buffering/checking/resetting, the error checking for if the database does not open, etc) between these two `notice "Importing era archive",` and `notice "Importing era archive HoleskyNet",` sections

nimbus/nimbus_import.nim
tersec: A decent heuristic here is to start looking for block `n` at slot `n` and then scan forward. It usually won't take long, and guaranteed is never behind.

nimbus/nimbus_import.nim
```diff
@@ -35,6 +38,159 @@ declareCounter nec_imported_gas,
...
+proc getTxs*(txs: seq[bellatrix.Transaction]): seq[common.Transaction] =
+  result = newSeqOfCap[common.Transaction](txs.len)
+  for tx in txs:
+    try:
+      result.add(rlp.decode(tx.asSeq(), common.Transaction))
+    except RlpError:
+      return @[]
+
+proc getWithdrawals*(x: seq[capella.Withdrawal]): seq[common.Withdrawal] =
+  result = newSeqOfCap[common.Withdrawal](x.len)
```
tersec, line 121: https://status-im.github.io/nim-style-guide/language.result.html

nimbus/nimbus_import.nim
tersec: https://github.com/status-im/nimbus-eth1/pull/2429/commits/7f92388628d17af8f8ecf01752c586bc228c94fd

nimbus/nimbus_import.nim
tersec: https://github.com/status-im/nimbus-eth1/pull/2429/commits/8c112a075410494d913b41c9bcaf8fc117422e06

nimbus/nimbus_import.nim
tersec: https://github.com/status-im/nimbus-eth1/pull/2429/commits/36df849a3a7090a86c1aaddbff92d1024ae96d55

nimbus/nimbus_import.nim
tersec: https://github.com/status-im/nimbus-eth1/pull/2429/commits/21d64d6703d105d3ae4cb8f0424758399ce2014f

nimbus/nimbus_import.nim
advaita-saha: In my opinion doesn't make sense doing that for the import process

PR #2423, opened 2024-06-27
Wiring ForkedChainRef to other components

jangko:
- Disable majority of hive simulators
- Only enable pyspec_sim for the moment
- The pyspec_sim is using a smaller RPC service wired to ForkedChainRef
- The RPC service will gradually grow

PR #2441, opened 2024-07-03
Enable consensus_sim

jangko:
This PR meant to be merged to #2423

### DIFF
--- a/hive_integration/nodocker/consensus/consensus_sim.nim
+++ b/hive_integration/nodocker/consensus/consensus_sim.nim
@@ -12,6 +12,7 @@ import
   stew/byteutils,
   results,
   chronicles,
+  ../../../nimbus/core/chain,
   ../../../nimbus/core/block_import,
   ../../../nimbus/common,
   ../../../nimbus/core/eip4844,
@@ -28,11 +29,12 @@ proc processChainData(cd: ChainData): TestStatus =
 
   com.initializeEmptyDb()
 
+  let c = newForkedChain(com, com.genesisHeader)
   for bytes in cd.blocksRlp:
     # ignore return value here
     # because good blocks maybe interleaved with
     # bad blocks
-    discard importRlpBlock(bytes, com, "consensus_sim")
+    discard importRlpBlocks(bytes, c, finalize = true)
 
   let head = com.db.getCanonicalHead()
   let blockHash = "0x" & head.blockHash.data.toHex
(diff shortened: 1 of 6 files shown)


nimbus/core/block_import.nim
```diff
@@ -14,47 +14,55 @@ import
...
+  var head: BlockHeader
+  if not com.db.getCanonicalHead(head):
+    error "cannot get canonical head from db"
+    quit(QuitFailure)
+
+  let chain = newForkedChain(com, head, baseDistance = 0)
+
+  # success or not, we quit after importing blocks
+  for i, blocksFile in conf.blocksFile:
+    if isFile(string blocksFile):
```
tersec, line 63: Potential TOCTOU issue, also, doesn't check for permissions and some other issues? I'm not sure why not let the `io2.readAllBytes`-based error checking handle this, since this up-front check is both insufficient on its own and doesn't seem to provide particularly useful UX feedback.

Indeed, it silently skips `blockFiles` which aren't files or don't exist, but that seems to be debatably useful. If someone specifies a file to import from and that file does not exist, that seems like an unexpected condition.

jangko: https://github.com/status-im/nimbus-eth1/pull/2441/commits/e48855f437f60bd7767076ee83236207cdb4219c

hive_integration/nodocker/consensus/consensus_sim.nim
```diff
@@ -45,8 +47,23 @@ proc processChainData(cd: ChainData): TestStatus =
...
+  "ForkStressTest.json",
+  "blockChainFrontierWithLargerTDvsHomesteadBlockchain.json",
+  "blockChainFrontierWithLargerTDvsHomesteadBlockchain2.json",
+  "lotsOfLeafs.json",
+  "loopMul.json"
+  ]
+
 proc main() =
-  const basePath = "tests" / "fixtures" / "eth_tests" / "BlockchainTests"
+  const basePath = "tests/fixtures/eth_tests/BlockchainTests"
```
tersec, line 66: Does this preserve functionality on Windows? [`/`](https://nim-lang.org/docs/ospaths2.html#/%2Cstring%2Cstring) uses [`DirSep`](https://nim-lang.org/docs/osseps.html#DirSep), which might also be `\` on Windows.

While for non-`\\?\` based paths in Windows, or those for many purposes for the Win32 APIs, they'll handle either `/` or `\`, it means that if one has to switch to the style of paths which allow more than `MAX_PATH == 260` length, e.g, because of a source directory with a long name, then using `/` in a hardcoded way would cause this to fail.

jangko: So far the blockchain tests folder only contains shallow recursion and the file names also short.

PR #2450, opened 2024-07-04
Enable LTO compilation

arnetheduck:
Similar to nimbus-eth2, LTO gives a significant boost for any CPU-bound operations such as the EVM.

The options are copied straight from nimbus-eth2 - for example at block height 1.7M there's a computation-heavy section where we can see a 25% improvement (!) in block processing time.

```
                       bps_x     bps_y     tps_x     tps_y time_x time_y     bpsd     tpsd    timed

(1722223, 1733334]    102.52    138.90  1,049.67  1,420.61  2m41s  1m58s   35.78%   35.78%  -26.32%
```

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -20,10 +20,41 @@ if getEnv("NIMBUS_BUILD_SYSTEM") == "yes" and
    system.fileExists(currentDir & "nimbus-build-system.paths"):
   include "nimbus-build-system.paths"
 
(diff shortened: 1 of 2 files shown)


jangko:
a month ago: GST 2m+
#2449: GST 1m7s
this: GST 1m3s

and we still have room to go below 1 minute.

Issue #2452, opened 2024-07-04
Avoid resizing EVM stack unnecessarily


arnetheduck:
Many (computational) EVM operations pop values from the stack and push the result back - popping and pushing is costly not only because we must re-check size constraints but also  because `seq.setLen` is (comparatively) slow - these instructions would benefit from operating in-place so that they overwrite the operand rather than popping and pushing it:

![image](https://github.com/status-im/nimbus-eth1/assets/1382986/f0db076e-f4fb-424b-88d3-21204e65f716)

Example:
https://github.com/status-im/nimbus-eth1/blob/b82dcdcc7633068381d40518e99c916adf455b6a/nimbus/evm/interpreter/op_handlers/oph_arithmetic.nim#L193

arnetheduck:
https://github.com/status-im/nimbus-eth1/pull/2481

PR #2461, opened 2024-07-06
Convert GasInt to uint64, bump nim-eth and nimbus-eth2

### LINKED ISSUE #2292: Switch `GasInt` to `uint64`
The initial choice in https://github.com/status-im/nimbus-eth1/issues/35#issuecomment-391726518 is argued to simplify development by raising a `Defect` whenever an out-of-bounds computation happens.

While this might have been an appropriate choice in early-stages design, it is not a reliable choice for production code, also because types where `GasInt` is used are RLP-serialized and RLP doesn't support signed integers.

Since `geth` uses `uint64`, this will also ensure that our handling is similar to that of geth:

https://github.com/ethereum/go-ethereum/blob/682ae838b2312a4ec8e5581069039b567e33c7c2/core/types/block.go#L76C2-L76C10

As an alternative to plain `uint64`, we could also develop a special "checked uint64" in which all operations return an overflow flag that is checked where applicable.

jangko:
txpool test has to be disabled again because txpool gaslimit, gasprice, and gastip requires more detailed attention.
fix #2292

### DIFF
--- a/nimbus/constants.nim
+++ b/nimbus/constants.nim
@@ -48,7 +48,7 @@ const
   GENESIS_MIX_HASH* =                       ZERO_HASH256
   GENESIS_EXTRA_DATA* =                     ""
   GAS_LIMIT_MINIMUM* =                      5000
-  GAS_LIMIT_MAXIMUM* =                      0x7fffffffffffffff.GasInt # Maximum the gas limit (2^63-1).
+  GAS_LIMIT_MAXIMUM* =                      int64.high.GasInt # Maximum the gas limit (2^63-1).
   DEFAULT_GAS_LIMIT* =                      8_000_000
 
   EMPTY_SHA3* =                             "c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470".toDigest
--- a/nimbus/constants.nim
+++ b/nimbus/constants.nim
@@ -48,7 +48,7 @@ const
   GENESIS_MIX_HASH* =                       ZERO_HASH256
   GENESIS_EXTRA_DATA* =                     ""
   GAS_LIMIT_MINIMUM* =                      5000
-  GAS_LIMIT_MAXIMUM* =                      high(GasInt)
+  GAS_LIMIT_MAXIMUM* =                      0x7fffffffffffffff.GasInt # Maximum the gas limit (2^63-1).
   DEFAULT_GAS_LIMIT* =                      8_000_000
 
   EMPTY_SHA3* =                             "c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470".toDigest
--- a/nimbus/evm/interpreter/gas_costs.nim
+++ b/nimbus/evm/interpreter/gas_costs.nim
@@ -54,7 +54,7 @@ type
     GasInitcodeWord     # Payment for each word (rounded up) for initcode
     GasWarmStorageRead  # Transient storage read and write cost.
 
-  GasFeeSchedule = array[GasFeeKind, GasInt]
+  GasFeeSchedule = array[GasFeeKind, Natural]
 
   GasParams* = object
     # Yellow Paper, Appendix H - https://ethereum.github.io/yellowpaper/paper.pdf
@@ -97,8 +97,7 @@ type
   # gasRefund of sstore can be a negative number
   SStoreGasResult* = object
     gasCost*: GasInt
-    gasRefund*: GasInt
-    reduceRefund*: GasInt
+    gasRefund*: int64
 
   CallGasResult = tuple[gasCost, childGasLimit: GasInt]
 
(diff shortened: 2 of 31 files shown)


nimbus/evm/interpreter/gas_costs.nim
```diff
@@ -321,36 +320,36 @@ template gasCosts(fork: EVMFork, prefix, ResultGasCostsName: untyped) =
   func `prefix gasLog0`(currentMemSize, memOffset, memLength: GasNatural): GasInt {.nimcall.} =
     result = `prefix gasMemoryExpansion`(currentMemSize, memOffset, memLength)
 
-    result += static(FeeSchedule[GasLog]) +
-      static(FeeSchedule[GasLogData]) * memLength
+    result += GasInt(static(FeeSchedule[GasLog]) +
+      static(FeeSchedule[GasLogData]) * memLength)
```
tersec, line 324: Some of the changes in `GasInt(static(...))` while others use `static(GasInt(...))`. The latter seems preferable, being as it is verifies and computes at least as much, probably more, at compile-time in terms of ensuring allowed integer type conversions. So remaining consistent with
https://github.com/status-im/nimbus-eth1/blob/0d9b0d9d6854c09603edbd1bddf087e42aef8fed/nimbus/evm/interpreter/gas_costs.nim#L431-L465
(except line 448, which does `GasInt(static(...))`) might be an improvement

nimbus/constants.nim
```diff
@@ -48,7 +48,7 @@ const
   GENESIS_MIX_HASH* =                       ZERO_HASH256
   GENESIS_EXTRA_DATA* =                     ""
   GAS_LIMIT_MINIMUM* =                      5000
-  GAS_LIMIT_MAXIMUM* =                      high(GasInt)
+  GAS_LIMIT_MAXIMUM* =                      0x7fffffffffffffff.GasInt # Maximum the gas limit (2^63-1).
```
tersec, line 51: `int64.high.GasInt` maybe clearer, don't have to trust the exact number of `f`s.

nimbus/evm/interpreter/gas_costs.nim
```diff
@@ -54,7 +54,7 @@ type
     GasInitcodeWord     # Payment for each word (rounded up) for initcode
     GasWarmStorageRead  # Transient storage read and write cost.
 
-  GasFeeSchedule = array[GasFeeKind, GasInt]
+  GasFeeSchedule = array[GasFeeKind, Natural]
```
tersec, line 57: See [`language.integers`](https://status-im.github.io/nim-style-guide/language.integers.html) and [`language.range`](https://status-im.github.io/nim-style-guide/language.range.html). That part about possible `Defect`s is probably relatively harmless here, since these gas costs are mostly constants, and even the computed ones are computed fairly straightforwardly from the same underlying constants, with bounds provided by Ethereum of, e.g., 30M (might become higher, but it's probably not going to become 30B gas/block limit meaning they're fairly well-behaved at this point in the code flow.

Two more specific questions though:
- `Natural - Natural` and some other operations decay into `int` (e.g., `typeof(1.Natural - 2.Natural)` is `int`), whereas in general `nimbus-eth1` for these purposes tends to use specific-sized integers. So both (a) it doesn't really stay `Natural` anyway, in an implicit/never-written-out way, so it means that the downstream `uint64` types have to handle potential conversions from negative numbers as far as Nim's type system is concerned, and therefore using `Natural hasn't bought much, and (b) the type it adds (`int`) to the mix is particularly unfortunate; and
- since `GasInt` is `uint64` now, and, see above, the `Natural` doesn't really need to be there to bound the values below `high(int64).GasInt`, doesn't using a mix of `Natural` and `uint64` (with the silent addition of `int`, per first point) mostly add pointless `Natural` <-> `int` <-> `uint64` conversions, vs keeping everything in the native `GasInt` type?

nimbus/constants.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/2461/commits/d831eee4dc3301de394a575425180c5f720635c2

nimbus/evm/interpreter/gas_costs.nim
jangko: I will not fix this in this PR, see the reason below

nimbus/evm/interpreter/gas_costs.nim
jangko: I will not fix this in this PR too. The focus of this PR is converting `GasInt` to `uint64` in `nim-eth`.
If I fix this and the above one in this PR, that will be too much for one PR.
My goal is to eliminate as much as possible `GasInt(..)` usage. That that will require overhaul of the EVM.
I have plan to fix it along with #2452 after I return from my vacation.

tersec: fair enough, sure

nimbus/evm/interpreter/gas_costs.nim
tersec: right, makes sense

PR #2462, opened 2024-07-09
don't write hash->slot map

arnetheduck:
This map is unused but because it's written for every storage update it causes a lot of database churn - even though the hashes are mostly the same, they are written both to the rocksdb WAL and SST files and then have to be "compacted away" which causes a lot of unnecessary disk activity.

Should we wish to re-implement it, it should (likely) be done in a way that only new hashes are written since most slots are the same.

In the same vein, avoid writing trivially detectable A-B-A storage changes which happen with surprising frequency.


arnetheduck:
https://github.com/status-im/nimbus-eth1/pull/2463 - turns out it was used after all..

PR #2475, opened 2024-07-11
Shift era helpers to a different file

advaita-saha:
- Shifts all the helper function for handling era file under a file name `utils/era_helpers.nim`
- Few minor logic fixes for the transition

### DIFF
--- a/nimbus/utils/era_helpers.nim
+++ b/nimbus/utils/era_helpers.nim
@@ -1,0 +1,194 @@
+# Nimbus
+# Copyright (c) 2024 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or
+#    http://www.apache.org/licenses/LICENSE-2.0)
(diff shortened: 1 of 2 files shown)


tersec (approved):

nimbus/utils/era_helpers.nim
```diff
@@ -0,0 +1,194 @@
...
+  ?f.getStateSSZ(slot, bytes)
+
+  if bytes.len() == 0:
+    return err("State not found")
+
+  let state =
+    try:
+      newClone(readSszForkedHashedBeaconState(cfg, slot, bytes))
+    except SerializationError as exc:
+      return err("Unable to read state: " & exc.msg)
```
kdeme, line 78: I missed the first PR, but considering `latestEraFile` is 99% copied and `loadHistoricalRootsFromEra` is for a big part the same as the existing code in `./fluffy/eth_data_exporter/cl_data_exporter`, it would have been nice to have these somewhere common. But okay, in the current setup it is not clear where that would go.

PR #2485, opened 2024-07-14
Store extension nodes together with the branch

arnetheduck:
Extension nodes must be followed by a branch - as such, it makes sense to store the two together both in the database and in memory:

* fewer reads, writes and updates to traverse the tree
* simpler logic for maintaining the node structure
* less space used, both memory and storage, because there are fewer nodes overall

There is also a downside: hashes can no longer be cached for an extension - instead, only the extension+branch hash can be cached - this seems like a fine tradeoff since computing it should be fast.

TODO: fix commented code

### DIFF
--- a/tests/test_aristo/test_misc.nim
+++ b/tests/test_aristo/test_misc.nim
@@ -11,6 +11,7 @@
 ## Aristo (aka Patricia) DB trancoder test
 
 import
+  std/algorithm,
   eth/common,
   results,
   stew/byteutils,
@@ -33,7 +34,7 @@ proc testShortKeys*(
 
   let samples = [
     # From InvalidBlocks/bc4895-withdrawals/twoIdenticalIndex.json
-    [("80".x,
+    @[("80".x,
       "da808094c94f5374fce5edbc8e2a8697c15331677e6ebf0b822710".x,
       "27f166f1d7c789251299535cb176ba34116e44894476a7886fe5d73d9be5c973".k),
      ("01".x,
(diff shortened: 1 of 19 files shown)


tests/test_aristo/test_misc.nim
```diff
@@ -44,9 +45,24 @@ proc testShortKeys*(
...
+      ("0100".x, "0100".x,
+      "d1c0699fe7928a536e0183c6400ae34eb1174ce6b21f9d117b061385034743ad".k),
+      ("0101".x, "0101".x,
+      "74ddb98cb56e2dd7e8fa090b9ce740c3de589b72403b20136af75fb6168b1d19".k),
+      ("0200".x, "0200".x,
+      "2e777f06ab2de1a460a8412b8691f10fdcb162077ab5cbb1865636668bcb6471".k),
+      ]]
 
-  let gossip = false # or noisy
+  let gossip = true # or noisy
```
arnetheduck, line 65:
```suggestion
  let gossip = false # or noisy
```

tests/test_aristo/test_misc.nim
```diff
@@ -67,12 +83,32 @@ proc testShortKeys*(
...
+
+      #let rc = sig.db.check
+      # xCheckRc rc.error == (0,0)
       xCheck r == w
+      #debugEcho rc
 
   true
-
+var v = testShortKeys()
+echo v
```
arnetheduck, line 111:
```suggestion
```

a really nice feature of `unittest2` is the ability to compile and run each module independently

arnetheduck:
Merged as #2494, thanks @mjfh !

PR #2486, opened 2024-07-15
Fluffy State Bridge - Building state using state diffs.

web3-developer:
This is a proof of concept implementation of the Fluffy state bridge. Not yet completed. Currently state bridge reads the state diffs using trace_replayBlockTransctions. The stateDiffs are read and applied to the state in the database. After applying the state, the state root is checked against the block header to verify that it is correct. 

Still to be implemented:
- Caching updated trie nodes between blocks.
- Convert each updated trie node to a proof.
- Convert each proof to an offer.
- Send each offer into the portal state network.


fluffy/tools/portal_bridge/portal_bridge_state.nim
```diff
@@ -7,12 +7,180 @@
...
+    let worldState = db.withTransaction:
+      let
+        # Requires an active transaction because it writes an emptyRlp node to tries on initialization
+        worldState = WorldStateRef.init(db)
+        genesisAccounts = genesisBlockForNetwork(MainNet).alloc
+      worldState.applyGenesisAccounts(genesisAccounts)
+      worldState
+
+    var currentBlockNumber = startBlockNumber
+    echo "Starting from block number: ", currentBlockNumber
```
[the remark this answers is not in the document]
web3-developer: Yep will do. I'm planning to review all the logging a bit later once I'm finished the basic implementation.

fluffy/tools/portal_bridge/portal_bridge_state.nim
```diff
@@ -7,12 +7,180 @@
...
+
+      db.withTransaction:
+        for stateDiff in stateDiffs:
+          applyStateDiff(worldState, stateDiff)
+        applyBlockRewards(worldState, blockObject, uncleBlocks)
+
+      doAssert(blockObject.stateRoot.bytes() == worldState.stateRoot.data)
+
+      inc currentBlockNumber
+  except CatchableError as e:
```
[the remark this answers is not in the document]
web3-developer: Sure, will do. I'm planning to review all the exception handling a bit later.

fluffy/tools/portal_bridge/portal_bridge_state.nim
```diff
@@ -7,12 +7,180 @@
...
+
+    let worldState = db.withTransaction:
+      let
+        # Requires an active transaction because it writes an emptyRlp node to tries on initialization
+        worldState = WorldStateRef.init(db)
+        genesisAccounts = genesisBlockForNetwork(MainNet).alloc
+      worldState.applyGenesisAccounts(genesisAccounts)
+      worldState
+
+    var currentBlockNumber = startBlockNumber
```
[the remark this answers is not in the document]
web3-developer: Yes thats correct. We need the pre-state of the start block in order to progress. I'm thinking to put in a check to see if the trie contains a root node matching the start block parent and if not then exit with an error message telling the user to start from an earlier block or something similar.

web3-developer:
@kdeme I've refactored and cleaned up the code. Improved the logging and error handling. Also using an async queue to separate the part that collects the block data via rpc and the part that generates the state using the state diffs. If you have any final thoughts before I merge, feedback welcome.

fluffy/tools/portal_bridge/portal_bridge_state.nim
```diff
@@ -7,12 +7,139 @@
...
+
+    db.withTransaction:
+      for stateDiff in blockData.stateDiffs:
+        worldState.applyStateDiff(stateDiff)
+      let
+        minerData =
+          (EthAddress(blockData.blockObject.miner), blockData.blockObject.number.uint64)
+        uncleMinersData =
+          blockData.uncleBlocks.mapIt((EthAddress(it.miner), it.number.uint64))
+      worldState.applyBlockRewards(minerData, uncleMinersData)
```
kdeme, line 124: I hadn't realized that we needed to get all the uncles also, but makes sense considering they get rewards pre-merge.

It's interesting though as in history network we do not store the uncles.

fluffy/tools/portal_bridge/portal_bridge_state.nim
```diff
@@ -7,12 +7,139 @@
...
+
+  while true:
+    await sleepAsync(10.seconds)
+    info "Block data queue length: ", queueLen = blockDataQueue.len()
 
 proc runState*(config: PortalBridgeConf) =
   let
-    portalClient = newRpcClientConnect(config.portalRpcUrl)
-    web3Client = newRpcClientConnect(config.web3Url)
+    #portalClient = newRpcClientConnect(config.portalRpcUrl)
```
kdeme, line 141: Adding the conversion to Portal state content and gossiping it into the Portal network is for a second PR I assume?

kdeme (approved):
lgtm. The split up in retrieval of data an applying it makes sense I think. I assume the retrieval part takes the longest?

One question I have is what happens now if you start with an existing database? It will just start from scratch still?

And then also, do you have some early measurements of how many blocks or transactions per minutes we reach with this architecture? And what is the bottleneck? And perhaps a rough guess how that could compare to executing it inside the bridge?

fluffy/tools/portal_bridge/portal_bridge_state.nim
web3-developer: Yeah the uncles are required to calculate the block rewards. Not sure if there is a better way. I found that getting this data via the eth RPC calls still performs well using websocket which reasonably fast

fluffy/tools/portal_bridge/portal_bridge_state.nim
web3-developer: Yeah more changes coming. I'm thinking better to merge in multiple PRs as I go.

web3-developer:
> lgtm. The split up in retrieval of data an applying it makes sense I think. I assume the retrieval part takes the longest?
> 
> One question I have is what happens now if you start with an existing database? It will just start from scratch still?
> 
> And then also, do you have some early measurements of how many blocks or transactions per minutes we reach with this architecture? And what is the bottleneck? And perhaps a rough guess how that could compare to executing it inside the bridge?

The split up using the queues is just an experiment to see if the performance can be improved by allowing the CPU intensive parts to continue while waiting on networking IO. Its something that we can play with but may or may not be needed in the end. Currently yes the rpc retrieval of block data takes the longest and so the queue is always empty as applying the state updates is very fast. The performance of this second part may change soon though once I finish building the proofs from the state and convert to offers.

Currently we can only start from block 1 because the existing HexaryTrie library has some issues which I don't fully understand just yet. Long story short it currently only works with a single state (not working as archive of state). I think we need a better HexaryTrie library to be honest. Anyway yes it will only sync from block 1 for now. Having existing data in the database doesn't cause any issues and the state root calculations will still be correct etc when resyncing. 

I don't have any recorded measurements but when using a websocket connection with RocksDb as the backend, the performance is extremely fast. I can sync the first 1 million blocks in about 10-15 minutes. From memory this is way faster than when I ran Nimbus for the first million blocks. Currently the bottleneck is fetching the data over rpc. I'd guess that the bottleneck might become sending the data to the portal state network if we need to wait in order to audit that the data was received over the network. In my opinion this architecture will likely be faster than executing the blocks using era1 files. The downside is that it currently requires having a node that can produce the state diffs. I'm thinking that this might be a short term architecture for us if we end up integrating Fluffy into Nimbus. We could implement trace_replayBlockTransactions and all RPC calls could become direct function calls inside Nimbus. Perhaps Nimbus could act as a bridge node while syncing in the future.

arnetheduck:
> I can sync the first 1 million blocks in about 10-15 minutes. From memory this is way faster than when I ran Nimbus for the first million blocks. 

```
INF 2024-07-18 11:27:16.746+02:00 Imported blocks                            blockNumber=1000001 blocks=1000000 importedSlot=1 txs=1674264 mgas=68098.967 bps=2443.262 tps=13790.009 mgps=367.947 avgBps=4847.571 avgTps=8116.113 avgMGps=330.115 elapsed=3m26s288ms
```

you might want to update that mental image of sync speed ;)

I _suspect_ we'd be better off reusing the aristo-based setup in fluffy also, including using its trie support (which now has gained a lot of performance-oriented improvements).

kdeme:
> Its something that we can play with but may or may not be needed in the end. Currently yes the rpc retrieval of block data takes the longest and so the queue is always empty as applying the state updates is very fast. The performance of this second part may change soon though once I finish building the proofs from the state and convert to offers.

Yes, depending on what is the slowest part we might have to activate multiple runners. E.g. concurrent calls for the retrieval part. Of course I'm not sure if that would help in any way if you use the same EL node for them, probably not.

> From memory this is way faster than when I ran Nimbus for the first million blocks.

Things have changed quite a bit here, you might want to have a look at some more recent measurements, e.g.: https://discord.com/channels/613988663034118151/809765453052248094/1252340877804834888

Things will also be very different for the first 1M or 2M blocks until later when there are actually txs.
I assume not running the execution ourselves will definitely be slower at the first blocks (as no txs and thus a lot of network overhead for not that many state updates?), and perhaps be beneficial later on. But I'm not sure about that, the json-rpc overhead might be too much. I think it is important to test this soon so we can decide on best architecture.

web3-developer:
> > I can sync the first 1 million blocks in about 10-15 minutes. From memory this is way faster than when I ran Nimbus for the first million blocks.
> 
> ```
> INF 2024-07-18 11:27:16.746+02:00 Imported blocks                            blockNumber=1000001 blocks=1000000 importedSlot=1 txs=1674264 mgas=68098.967 bps=2443.262 tps=13790.009 mgps=367.947 avgBps=4847.571 avgTps=8116.113 avgMGps=330.115 elapsed=3m26s288ms
> ```
> 
> you might want to update that mental image of sync speed ;)
> 
> I _suspect_ we'd be better off reusing the aristo-based setup in fluffy also, including using its trie support (which now has gained a lot of performance-oriented improvements).

Oh nice, that is fast. I was referring to when I ran Nimbus using the legacy db quite a while ago. For these numbers are you calculating the state root and building the rlp encoded mpt trie for every block? Fluffy requires this data for each block in order to send the offers containing the mpt proofs.

Also can you point me to the relevant file/type in Aristo so I can have a look at the trie support?

web3-developer:
> > Its something that we can play with but may or may not be needed in the end. Currently yes the rpc retrieval of block data takes the longest and so the queue is always empty as applying the state updates is very fast. The performance of this second part may change soon though once I finish building the proofs from the state and convert to offers.
> 
> Yes, depending on what is the slowest part we might have to activate multiple runners. E.g. concurrent calls for the retrieval part. Of course I'm not sure if that would help in any way if you use the same EL node for them, probably not.
> 
> > From memory this is way faster than when I ran Nimbus for the first million blocks.
> 
> Things have changed quite a bit here, you might want to have a look at some more recent measurements, e.g.: https://discord.com/channels/613988663034118151/809765453052248094/1252340877804834888
> 
> Things will also be very different for the first 1M or 2M blocks until later when there are actually txs. I assume not running the execution ourselves will definitely be slower at the first blocks (as no txs and thus a lot of network overhead for not that many state updates?), and perhaps be beneficial later on. But I'm not sure about that, the json-rpc overhead might be too much. I think it is important to test this soon so we can decide on best architecture.

The first part can be done (mostly) concurrently because the data for each block can be fetched independantly so that is something worth looking into I think when trying to optimize the performance. 

Yes I agree, in theory executing the state diffs could be faster compared to actually executing every transactions especially during the DOS attacks block ranges. Sure, once I've finished the basic implementation. I can try comparing the performance to Trin's state bridge and or Nimbus era1 sync and see how it goes.

kdeme:
> Sure, once I've finished the basic implementation, I can try comparing the performance to Trin's state bridge and or Nimbus era1 sync and see how it goes.

I think you could already start doing a comparison with the current code. 
In the end you mostly want to compare local execution versus this version, which is basically comparing what you have now with the current numbers from nimbus EL, but for a longer block range.

PR #2491, opened 2024-07-15
faster slot finding in nimbus import

advaita-saha:


nimbus/nimbus_import.nim
```diff
@@ -226,9 +226,10 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
       historical_summaries: openArray[HistoricalSummary],
   ) =
     if blockNumber > 1:
-      importedSlot = blockNumber
+      importedSlot = ( blockNumber - lastEra1Block ) + firstSlotAfterMerge
```
arnetheduck, line 229:
```suggestion
      importedSlot = (blockNumber - lastEra1Block) + firstSlotAfterMerge
```

nimbus/nimbus_import.nim
```diff
@@ -237,7 +238,11 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
         if clBlock.isSome:
           let ethBlock = getEth1Block(clBlock.get())
           parentHash = ethBlock.header.parentHash
-
+          if checkOnce:
```
arnetheduck, line 241: what's this for? and why are we computing hashes and not simply comparing block numbers? Computing hashes takes a lot of cpu processing

advaita-saha: While iterating through the slot numbers, when we find a block in a slot number I am calculating the difference between the `blockNumber` and the `block_number` from the executionPayload from the slot, then adding the difference to the `importedSlot`. This pushes the lower bound more, making the search way smaller

We need to do it only once after we find the slot from the from the initial lower bound `(blockNumber - lastEra1Block) + firstSlotAfterMerge`. This is kept inside the loop since we don't know when we will find a slot with a block after the initial lower bound. So when it does, it toggles the flag and updates the lower bound then keeps searching from there on, resulting in a much smaller search

advaita-saha: Yeah shifting from hashes to blockNumber would be better

advaita-saha:
Now in the slot finding, no type of hashing is performed

nimbus/nimbus_import.nim
```diff
@@ -220,26 +220,36 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
...
+  # making the search way smaller
   template updateLastImportedSlot(
       era: EraDB,
       historical_roots: openArray[Eth2Digest],
       historical_summaries: openArray[HistoricalSummary],
   ) =
     if blockNumber > 1:
-      importedSlot = blockNumber
+      # Setting the initial lower bound 
+      importedSlot = (blockNumber - lastEra1Block) + firstSlotAfterMerge
```
tersec, line 236: This logic can in be generalized -- insofar as this exploits the ever-growing gap between slot and block numbers post-merge, each individual era (CL) `BeaconBlock` has an `ExecutionPayload` with the block number, from which one can infer the most recent version of this change, or bound it (it's monotonic), etc. So https://beaconcha.in/slot/9526568 shows that slot 9526568 has block number 20319895, and similar information is independently in every post-merge `BeaconBlock` era files contain.

nimbus/utils/era_helpers.nim
```diff
@@ -134,6 +134,12 @@ proc getWithdrawals*(x: seq[capella.Withdrawal]): seq[common.Withdrawal] =
     )
   return withdrawals
 
+proc getEth1BlockNumber*(blck: ForkedTrustedSignedBeaconBlock): uint64 =
+  ## Get the eth1 block number from a beacon block.
+  withBlck(blck):
+    when consensusFork >= ConsensusFork.Bellatrix:
```
tersec, line 140: This only truly holds once the merge happens, which is no earlier than the Bellatrix fork but could be later, or could never have happened. For Sepolia and mainnet it did, and that's probably enough not to care about the what-if-Capella-merge case. But there's still a part of Bellatrix which is unmerged. The condition is Bellatrix and non-default `ExecutionPayload`.

See https://github.com/status-im/nimbus-eth2/blob/ebd0217a10ff8d529307157f476385c4d82d8b14/beacon_chain/spec/helpers.nim#L390-L435 for how `is_merge_transition_complete()`, `is_merge_transition_complete()`, et cetera are defined for precise conditions.

nimbus/nimbus_import.nim
advaita-saha: Yes the process you described is done in later steps to further reduce the search

nimbus/utils/era_helpers.nim
advaita-saha: Yes that can be done
My concern would be `is_execution_block()` is internally also doing the same as checking if the fork is greater than bellatrix, instead of having some chain specific checking

Other functions are doing that, but then I need to load the beacon_state, which I don't want to do inside the slot finding mechanism

tersec: > Yes that can be done My concern would be `is_execution_block()` is internally also doing the same as checking if the fork is greater than bellatrix, instead of having some chain specific checking

It's also doing
```nim
    const defaultExecutionPayload =
      default(typeof(blck.body.execution_payload))
    blck.body.execution_payload != defaultExecutionPayload
```

> Other functions are doing that, but then I need to load the beacon_state, which I don't want to do inside the slot finding mechanism

Sure, agree, the beacon state-dependent ones should be avoided here.

nimbus/utils/era_helpers.nim
```diff
@@ -138,7 +138,8 @@ proc getEth1BlockNumber*(blck: ForkedTrustedSignedBeaconBlock): uint64 =
   ## Get the eth1 block number from a beacon block.
   withBlck(blck):
     when consensusFork >= ConsensusFork.Bellatrix:
-      return forkyBlck.message.body.execution_payload.block_number
+      if forkyBlck.message.is_execution_block:
+        return forkyBlck.message.body.execution_payload.block_number
```
tersec, line 142: Well, this will happen to return 0 (because that's `default(uint64)`) when this isn't true.

Maybe that's correct, in which case, it'd be easier to do what you originally did -- no `is_execution_block` check, and pre-merge it returns 0, because the execution payload will be empty. But the intent should be documented/commented, otherwise it looks like a potential oversight

But it does mean `getEth1BlockNumber` behaves in a way which has to be anticipated by the rest of the codebase

PR #2493, opened 2024-07-16
Rework beacon block proofs to better structure

kdeme:
The 3 proofs can be reworked to two proofs as we can use the BeaconBlock directly instead of BeaconBlockHeader and BeaconBlockBody. This is possible because the HTR of the BeaconBlock is the same as the one of the BeaconBlockHeader.

This results in 32 bytes less as an intermediate hash can be removed. But more importantly looks more clean and compact in structure and code.


kdeme:
Requires test vectors to get merged first: https://github.com/ethereum/portal-spec-tests/pull/20

kdeme:
Disabled the tests for now that depend on the updated test vectors PR to get merged

PR #2496, opened 2024-07-17
Use block number or timestamp to determine fork rules

### LINKED ISSUE #2322: Use block number / information to determine fork rules
In code like https://github.com/status-im/nimbus-eth1/blob/4e50b05564f03126ee6bc85192811a7d8aced105/nimbus/core/withdrawals.nim#L25, the global `com` state is used to validate data found in the block itself - this is confusing - the block rules should be decided from the given block header since there may be a mismatch between the two.

jangko:
Avoid confusion raised by `forkGTE` usage where block informations are present.

fix #2322

### DIFF
--- a/nimbus/beacon/api_handler/api_newpayload.nim
+++ b/nimbus/beacon/api_handler/api_newpayload.nim
@@ -159,11 +159,16 @@ proc newPayload*(ben: BeaconEngineRef,
   let ttd = com.ttd.get(high(UInt256))
 
   if version == Version.V1:
-    let td  = db.getScore(header.parentHash).valueOr:
+    let ptd  = db.getScore(header.parentHash).valueOr:
       0.u256
-    if (not com.forkGTE(MergeFork)) and td < ttd:
+    let gptd  = db.getScore(parent.parentHash)
+    if ptd < ttd:
       warn "Ignoring pre-merge payload",
-        number = header.number, hash = blockHash, td, ttd
+        number = header.number, hash = blockHash.short, ptd, ttd
+      return invalidStatus()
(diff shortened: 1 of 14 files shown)


tools/t8n/transition.nim
```diff
@@ -459,7 +459,9 @@ proc transitionAction*(ctx: var TransContext, conf: T8NConf) =
       # un-set it if it has been set too early
       ctx.env.parentBeaconBlockRoot = Opt.none(Hash256)
 
-    if com.forkGTE(MergeFork):
+    let isMerged = config.terminalTotalDifficulty.isSome and
```
jangko, line 462: This part is taken from geth t8n tool. And it makes sense since t8n doesn't deals with block but only transactions transition. Hence the merge condition is based on config, and not from block information. Especially not from TD in database.

nimbus/beacon/api_handler/api_newpayload.nim
```diff
@@ -159,11 +159,16 @@ proc newPayload*(ben: BeaconEngineRef,
...
+    let ptd  = db.getScore(header.parentHash).valueOr:
       0.u256
-    if (not com.forkGTE(MergeFork)) and td < ttd:
+    let gptd  = db.getScore(parent.parentHash)
+    if ptd < ttd:
       warn "Ignoring pre-merge payload",
-        number = header.number, hash = blockHash, td, ttd
+        number = header.number, hash = blockHash.short, ptd, ttd
+      return invalidStatus()
+    if parent.difficulty > 0.u256 and gptd.isSome and gptd.value >= ttd:
```
jangko, line 169: This gptd additional check also taken from geth.

arnetheduck:
can we remove forkGTE completely, and the state it uses?

PR #2502, opened 2024-07-18
Optimize EVM stack usage

jangko:
Although minuscule, but the improvement is visible. The GST now done in 51s from previous 52s.

### DIFF
--- a/nimbus/evm/interpreter/op_handlers/oph_arithmetic.nim
+++ b/nimbus/evm/interpreter/op_handlers/oph_arithmetic.nim
@@ -39,106 +39,60 @@ func slt(x, y: UInt256): bool =
 
 proc addOp (k: var VmCtx): EvmResultVoid =
   ## 0x01, Addition
-  ? k.cpt.stack.lsCheck(2)
-  let
-    lhs = k.cpt.stack.lsPeekInt(^1)
-    rhs = k.cpt.stack.lsPeekInt(^2)
-  k.cpt.stack.lsShrink(1)
-  k.cpt.stack.lsTop(lhs + rhs)
-  ok()
+  k.cpt.stack.binaryOp(`+`)
 
 proc mulOp(k: var VmCtx): EvmResultVoid =
   ## 0x02, Multiplication
-  ? k.cpt.stack.lsCheck(2)
-  let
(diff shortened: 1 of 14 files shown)


nimbus/evm/interpreter/op_handlers/oph_arithmetic.nim
```diff
@@ -39,35 +39,56 @@ func slt(x, y: UInt256): bool =
...
 proc addOp (k: var VmCtx): EvmResultVoid =
   ## 0x01, Addition
-  let (lhs, rhs) = ? k.cpt.stack.popInt(2)
-  k.cpt.stack.push(lhs + rhs)
+  ? k.cpt.stack.lsCheck(2)
+  let
+    lhs = k.cpt.stack.lsPeekInt(^1)
+    rhs = k.cpt.stack.lsPeekInt(^2)
+  k.cpt.stack.lsShrink(1)
+  k.cpt.stack.lsTop(lhs + rhs)
```
arnetheduck, line 47: a safer option here would perhaps be to retain the checks, but do something like:
```
template withTop(stack, sym, body) = 
  if stack.len > 0:
    template sym = stack[^1]
    body

withStack(top):
  top = top + rhs
```

arnetheduck: then the whole operation becomes:
```
let a = ?stack.popInt(0)
? stack.withTop(top):
  top = top + a
```

jangko: Thanks for  the `template sym` trick.

arnetheduck:
see also https://github.com/status-im/nimbus-eth1/issues/2452 - the improvement here will be very contract-dependent - later contracts have a lot more arithmetic in them as they start doing actual cryptography in the contract and these kinds of optimizations become a lot more valuable.

nimbus/evm/interpreter/op_handlers/oph_arithmetic.nim
jangko: Both unary and binary ops are optimized with specialized templates.
https://github.com/status-im/nimbus-eth1/pull/2502/commits/a26a16533b2f0202aa5016601a721fb375310c74

PR #2507, opened 2024-07-19
Add a size limit to content items received over portal_stream

kdeme:
This limit of 2MB was somewhat roughly chosen based on:

The total list recommended soft limit of 2MB defined in devp2p: https://github.com/ethereum/devp2p/blob/master/caps/eth.md#blockbodies-0x06

> The recommended soft limit for BlockBodies responses is 2 MiB.

The limit that geth actually applies of 10MB: https://github.com/ethereum/go-ethereum/blob/master/eth/protocols/eth/protocol.go#L49

Note that these are limits of total protocol messages, which can included several block headers or block bodies or block receipts, which is why I basically reduce the 10MB to 2MB for individual content items.
The total of message is already restricted to 64 and we know up front how many we are supposed to receive.

All blocks to date are also well below this limit: https://etherscan.io/chart/blocksize

If you would however apply the different constants in portal spec for this, these limits would get way to high for block body and receipts: https://github.com/ethereum/portal-network-specs/blob/659e7c79577c1cf01283e447032ac4346c56bc84/history/history-network.md#constants

### DIFF
--- a/fluffy/network/wire/portal_stream.nim
+++ b/fluffy/network/wire/portal_stream.nim
@@ -35,6 +35,11 @@ const
   talkReqOverhead = getTalkReqOverhead(utpProtocolId)
   utpHeaderOverhead = 20
   maxUtpPayloadSize = maxDiscv5PacketSize - talkReqOverhead - utpHeaderOverhead
+  maxContentItemSize = 2 * 1024 * 1024
+    # 2MB, a general content size maxium to avoid endlessly long invalid data to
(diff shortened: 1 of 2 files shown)


kdeme:
I realized the goal of avoiding endlessly long invalid data to be read is already achieved with the per content item applied `defaultContentReadTimeout`. Hence, in retrospect, I don't think this size limit adds anything extra. It makes more sense to properly tune that timeout to a throughput that we find acceptable that it would take to receive a 2MB content item for example.

PR #2512, opened 2024-07-22
Update nim-rocksdb to latest. Fixes re-building issue on MacOS.

web3-developer:


kdeme (commented):
I didn't further investigate yet, but with this bump, I'm still hitting the rebuild on each `make portal_bridge` on macOS.

web3-developer:
After running make update, it will rebuild rocksdb one time, then after that it shouldn't do any further rebuilds. I'll do a bit more testing to confirm it is working.

PR #2535, opened 2024-07-29
Fluffy State Bridge - State Gossip via Portal JSON-RPC 

web3-developer:
This PR contains the following changes:
- Use account, storage and bytecode updated caches to build state proofs.
- Build state offers from state proofs.
- Store block data with stateDiffs in database to improve performance after the first run. RPC is only used to collect data when there is no data in database for each block.
- Refactored and improved structure of state diff types.
- Add support to start state backfill from specific block. This works as long as the state exists in the db from a previous run.
- Support DAO hard fork at block 1920000.
- Gossip state offers for each block via the portal JSON-RPC portal_stateGossip endpoint.


fluffy/tools/portal_bridge/portal_bridge_state.nim
```diff
@@ -12,134 +12,293 @@ import
...
+      )
+
+  # Load the world state using the parent state root
+  let worldState = WorldStateRef.init(db, firstBlock.parentStateRoot)
 
   while true:
     let blockData = await blockDataQueue.popFirst()
+    if blockData.blockNumber == 2.uint64:
+      # exit
+      break
```
web3-developer, line 183: This will be removed shortly. I'm currently doing some testing and forgot to remove this.

fluffy/tools/portal_bridge/portal_bridge_state.nim
```diff
@@ -12,134 +12,293 @@ import
...
+    portalClient: RpcClient,
+    offerWithKey:
+      AccountTrieOfferWithKey | ContractTrieOfferWithKey | ContractCodeOfferWithKey,
+) {.async: (raises: [CancelledError]).} =
+  let
+    keyBytes = offerWithKey.key.toContentKey().encode().asSeq()
+    offerBytes = offerWithKey.offer.encode()
+  try:
+    let numPeers =
+      await portalClient.portal_stateGossip(keyBytes.to0xHex(), offerBytes.to0xHex())
```
web3-developer, line 247: I thought it might be a good idea to use the standard portal_stateGossip RPC endpoint in order to allow the Fluffy state bridge to be used with any portal client. If we use a custom RPC endpoint then the bridge would only support sending data into the portal state network via Fluffy.

PR #2541, opened 2024-08-06
Fluffy state bridge - Add CLI parameters and implement workers for offer gossip.

web3-developer:
This PR includes the following changes:
- Tried using RPC batching for sending offers for entire block. This method doesn't work very well due to hitting the max HTTP payload size because each offer payload is rather large. Fluffy also isn't able to keep up when sending offers this fast in large chunks. 
- For now we continue to send offers one at a time but instead use multiple workers to process sending the offers for multiple blocks concurrently. There is no dependency between blocks for this part so it doesn't need to be done sequential.
- Fixed a bug in the contract trie offer building when constructing the path.
- Improved the Fluffy logging to reduce the amount of info logs when processing offers.
- Added debugging/testing function to verify correctness of state proofs before converting to offers.
- Added CLI parameters to support configuring the Fluffy state bridge.

### DIFF
--- a/fluffy/network/state/state_gossip.nim
+++ b/fluffy/network/state/state_gossip.nim
@@ -97,7 +97,7 @@ proc gossipOffer*(
   let req1Peers = await p.neighborhoodGossip(
     srcNodeId, ContentKeysList.init(@[keyBytes]), @[offerBytes]
   )
-  info "Offered content gossipped successfully with peers", keyBytes, peers = req1Peers
+  debug "Offered content gossipped successfully with peers", keyBytes, peers = req1Peers
 
 proc gossipOffer*(
     p: PortalProtocol,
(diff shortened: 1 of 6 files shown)


kdeme (commented):
We can increase that limit. I assume it is the limit on the fluffy node receiving side?

That one was already increased slightly: https://github.com/status-im/nimbus-eth1/blob/master/fluffy/fluffy.nim#L222

Could make it higher or configurable.


If fluffy doesn't manage to keep up, there are some adjustments that could be done, but not sure if it is sufficient:
- Could gossip it to less neighbours
- Could test some adjustment in gossip algo, e.g. remove the potential request of closest nodes and just send to what we know.

web3-developer:
> We can increase that limit. I assume it is the limit on the fluffy node receiving side?
> 
> That one was already increased slightly: https://github.com/status-im/nimbus-eth1/blob/master/fluffy/fluffy.nim#L222
> 
> Could make it higher or configurable.
> 
> If fluffy doesn't manage to keep up, there are some adjustments that could be done, but not sure if it is sufficient:
> 
> * Could gossip it to less neighbours
> * Could test some adjustment in gossip algo, e.g. remove the potential request of closest nodes and just send to what we know.

Yes, this is the HTTP request limit in the Fluffy HTTP RPC server. 

I noticed the increased HTTP request body limit and I actually did temporarily increase it locally. After doing so, I ran into other errors coming from Fluffy which indicated that Fluffy wasn't keeping up.

I'm running a local testnet with 2 nodes so we can't gossip to less neighbours in this case but removing the additional request for closest nodes would certainly speed things up.

PR #2547, opened 2024-08-06
Return empty list instead of error in getPooledTxs handler

jangko:


arnetheduck:
fwiw, warnings are still printed because https://github.com/status-im/nimbus-eth1/blob/3cef119b789aa9adccdd13e4dc4161911eca4728/nimbus/sync/handlers/eth.nim#L452 returns nothing

Issue #2544, opened 2024-08-06
Implement logging options from eth2


arnetheduck:
`nimbus` should support the same logging options as eth2 by reusing the shared binary support, similar to https://github.com/status-im/nimbus-eth1/pull/2488 - in particular, this allows autodetecting things like tty color support etc

PR #2560, opened 2024-08-13
Aristo lazily delete larger subtrees

mjfh:


nimbus/db/aristo/aristo_constants.nim
```diff
@@ -38,6 +38,19 @@ const
     ## functions with fixed assignments of the type of a state root (e.g. for
     ## a receipt or a transaction root.)
 
+  ACC_LRU_SIZE* = 1024 * 1024
+    ## LRU cache size for accounts that have storage, see `.accLeaves` and
+    ## `.stoLeaves` fields of the main descriptor.
+
+  DELETE_SUBTREE_VERTICES_MAX* = 25
```
arnetheduck, line 45: I don't think this logic makes much sense - the vertex id of a subtree is the prefix of the key of every entry meaning that we can delete the entire subtree with a single [database operation](https://rocksdb.org/blog/2018/11/21/delete-range.html).

As mentioned on the calls, deleting subtrees is not a bottleneck that I can see - have you measured the impact of this additional complexity?

nimbus/db/aristo/aristo_constants.nim
```diff
@@ -38,6 +38,19 @@ const
...
+    ## Maximum number of vertices for a tree to be deleted instantly. If the
+    ## tree is larger, only the sub-tree root will be deleted immediately and
+    ## subsequent entries will be deleted not until the cache layers are saved
+    ## to the backend.
+    ##
+    ## Set to zero to disable in which case all sub-trees are deleted
```
arnetheduck, line 51: broadly, we either want to have logic or not to have it - have it, but disabled, is a lose-lose because it adds bug-hunting surface without adding value - this is an important feature of nimbus in general: keep things minimal

arnetheduck:
This PR adds 500 lines of code, but I'm not convinced the added complexity is worth the benefit, specially given that there are more efficient ways of implementing the same thing - it should probably be reconsidered.

nimbus/db/aristo/aristo_constants.nim
mjfh: Yes I have in the `delete_subtree_lazy.nim` source against 5m blocks. It was never a bottleneck but avoids to pull lots of key/vertex entries from the persistent database into the cache layers as documented below the `DELETE_SUBTREE_VERTICES_MAX` setting.

mjfh: DELETE_SUBTREE_VERTICES_MAX* = 25
    ## Maximum number of vertices for a tree to be deleted instantly. If the
    ## tree is larger, only the sub-tree root will be deleted immediately and
    ## subsequent entries will be deleted not until the cache layers are saved
    ## to the backend.
    ##
    ## Set to zero to disable in which case all sub-trees are deleted
    ## immediately.

nimbus/db/aristo/aristo_constants.nim
mjfh: No problem to behave as set `DELETE_SUBTREE_VERTICES_MAX` to one. This constant was mainly for having several variants for testing. There was a removal planned of all conditional clauses in a subsequent PR.

mjfh:
> This PR adds 500 lines of code, but I'm not convinced the added complexity is worth the benefit, specially given that there are more efficient ways of implementing the same thing - it should probably be reconsidered.

Actually not. After cleaning up conditional code there will be about 150 or less lines left (much of it comments).

nimbus/db/aristo/aristo_constants.nim
arnetheduck: > Yes I have in the delete_subtree_lazy.nim source against 5m blocks.

And what is the outcome of the measurement, in terms of benefits? ie import time, memory usage .. ?

fluffy/fluffy.nim.cfg
```diff
@@ -1,4 +1,11 @@
--d:"chronicles_sinks=textlines[dynamic],json[dynamic]"
+#
+# For some reason `json[dynamic]` causes problems with subsequent modules from
+# `Aristo` when compiling `fluffy`. There might be a `chronicles` inport missing
+# but it is not obvious where. -- jordan
+#
+#-d:"chronicles_sinks=textlines[dynamic],json[dynamic]"
+
+-d:"chronicles_sinks=textlines[dynamic]"
```
arnetheduck, line 8: we should not be changing the logging setup - it is common to all nimbus projects except nimbus-eth1 - see also https://github.com/status-im/nimbus-eth1/issues/2544

arnetheduck: typically, if `json` is not working it's because there's a json converter import missing somewhere

nimbus/db/aristo/aristo_constants.nim
arnetheduck: > into the cache layers

for scan operations of this sort, the common option is to implement a flag that avoids the cache - for rocksdb:

https://github.com/facebook/rocksdb/blob/d458331ee90d0e8b5abd90e9f340d6857f7f679d/include/rocksdb/options.h#L1711C1-L1711C26

for filesystems, it's done via `madvise` and various `fcntl` flags - a similar thing could be done here thus avoiding the complexity of introducing special handling that depends on the tree size - in fact, it _never_ makes sense to evict other items from the cache when deleting verticies - this is not limited to subtrees or "small" subtrees so such an option would be more broadly applicable. 

In the  same way, an iterator that iterates over significant parts of the trie should also not evict items from the cache.

PR #2569, opened 2024-08-19
avoid exceptions, tx copy

arnetheduck:
* avoid some exceptions using the new compile-time strformat
* avoid copying the full transaction only to normalise 2 fields (expensive)


arnetheduck:
@jangko the tx copy changes are not the cleanest code and could probably be applied to the rest of the codebase to remove the normalization code entirely! this part is the one that appears on the hot block import path.

PR #2574, opened 2024-08-20
External syncer

advaita-saha:
External script to sync the client upto the head of chain using a CL 

```bash
make -j12 nrpc
./build/nrpc external_sync --network:<eg sepolia> --beacon-api:"http://127.0.0.1:5052" --el-engine-api:"http://127.0.0.1:8550" --jwt-secret:<path-to-secret>
```


nimbus/external_sync.nim
```diff
@@ -0,0 +1,224 @@
...
+        let
+          time2 = Moment.now()
+          diff1 = (time2 - time1).nanoseconds().float / 1000000000
+          speed = hashChain.len.float / diff1
+          remaining = diff.float / speed
+        notice "Remaining blocks to be covered", diff = diff
+        notice "Speed of fetching blocks", speed = speed
+        notice "Time remaining", remaining = remaining
+
+      #let chainLen = hashChain.len
```
tersec, line 147: ?

nimbus/external_sync.nim
```diff
@@ -0,0 +1,224 @@
...
+      let (test, _) = client.getBlockFromBeaconChain(
+        BlockIdent.init(hashChain[hashChain.len - 1]), clConfig
+      )
+      notice "Blocks loaded from CL", chainLen = hashChain.len
+      #let testNumber = test.header.number
```
tersec, line 152: ?

nimbus/external_sync.nim
```diff
@@ -0,0 +1,224 @@
...
+
+  # Download the blocks from CL and load into the EL
+  template sync() =
+    while running and hashChain.len > 0:
+      if hashChain.len <= 1:
+        blocks.add(curBlck)
+        hashChain.setLen(0)
+      else:
+        # Load 2000 block at a time
+        for i in 0 ..< (if hashChain.len <= 2000: hashChain.len else: 2000):
```
tersec, line 164: This `2000` constant is repeated; probably better to pull it out to at least within this localized `else` block (doesn't have to be some big global constant, just, in same scope as 2000 is already used) as the maximum number of blocks at a time. Right now, if this is changed, it's possible to introduce bugs by updating one but not the other `2000`.

nimbus/external_sync.nim
```diff
@@ -0,0 +1,224 @@
...
+    (eth1Block, prev_hash)
+  else:
+    error "CL head is not available"
+    quit(QuitFailure)
+
+proc loadBlocksFromBeaconChain(conf: NimbusConf) {.async.} =
+  let coreDB =
+    # Resolve statically for database type
+    case conf.chainDbMode
+    of Aristo, AriPrune:
```
tersec, line 78: There don't seem to be any other options:
https://github.com/status-im/nimbus-eth1/blob/4db9c5c2d5551446daf0fab9eadba0a51dbfa2b3/nimbus/config.nim#L107-L110

so what does it accomplish/achieve to use a case expression here?

nimbus/external_sync.nim
advaita-saha: Would be like ?
```nim
let maxBlocks = 2000
for i in 0 ..< (if hashChain.len <= maxBlocks: hashChain.len else: maxBlocks):
    let (eth1blck, _) =
      client.getBlockFromBeaconChain(BlockIdent.init(hashChain.pop()), clConfig)
    blocks.add(eth1blck)
  blocks.add(curBlck)
  ```

File: nimbus/external_sync.nim [hunk 5]
```diff
@@ -0,0 +1,224 @@
...
+
+  template boolFlag(flags, b): PersistBlockFlags =
+    if b:
+      flags
+    else:
+      {}
+
+  let
+    chain = com.newChain()
+    clConfig = getMetadataForNetwork("sepolia").cfg
```
tersec, line 99: Is this meant to be specialized this way for Sepolia or more broadly applicable to Holesky and mainnet?

nimbus/external_sync.nim
```diff
@@ -0,0 +1,224 @@
...
+        hashChain.setLen(0)
+      else:
+        # Load 2000 block at a time
+        for i in 0 ..< (if hashChain.len <= 2000: hashChain.len else: 2000):
+          let (eth1blck, _) =
+            client.getBlockFromBeaconChain(BlockIdent.init(hashChain.pop()), clConfig)
+          blocks.add(eth1blck)
+        blocks.add(curBlck)
+
+      #let numberOfBlocks = blocks.len
```
tersec, line 170: ?

nimbus/external_sync.nim
advaita-saha: Need to be fixed, it is network agnostic

nimbus/external_sync.nim
advaita-saha: Okk other types have been removed

nimbus/external_sync.nim
```diff
@@ -0,0 +1,226 @@
...
+    if b:
+      flags
+    else:
+      {}
+
+  let
+    chain = com.newChain()
+    clConfig =
+      if conf.networkId == MainNet:
+        getMetadataForNetwork("mainnet").cfg
```
tersec, line 97: Might be better to split this into two parts:
- the map from `MainNet`/`SepoliaNet`/`HoleskyNet` to `"mainnet"`/`"sepolia"`/`"holesky"`; and
- the getMetadataForNetwork(foo).cfg

Otherwise it's a little repetitive/redundant

nimbus/external_sync.nim
```diff
@@ -0,0 +1,227 @@
...
+      notice "Current block number", currentBlockNumber
+
+  # Download the blocks from CL and load into the EL
+  template sync() =
+    while running and hashChain.len > 0:
+      if hashChain.len <= 1:
+        blocks.add(curBlck)
+        hashChain.setLen(0)
+      else:
+        let maxBlocks = 2000 # Load 2000 block at a time
```
tersec, line 167:
```suggestion
        const maxBlocks = 2000 # Load 2000 block at a time
```

arnetheduck:
let's maybe call this `nrpc` with a `sync` subcommand, so we can extend it with other commands - ie one can imagine making a cli for other json-rpc functionality

nimbus/external_sync.nim
```diff
@@ -0,0 +1,227 @@
...
+        "holesky"
+      else:
+        error "Unsupported network", network = conf.networkId
+        quit(QuitFailure)
+    clConfig = getMetadataForNetwork(networkName).cfg
+
+  var
+    currentBlockNumber = com.db.getSavedStateBlockNumber()
+    flags =
+      boolFlag({PersistBlockFlag.NoValidation}, true) +
```
tersec, line 110: I don't understand how/why these `boolFlags` work? It seems like they resolve basically statically, so `boolFlag({PersistBlockFlag.NoValidation}, true)` resolves to just `{PersistBlockFlag.NoValidation}`, et cetera.

nimbus/external_sync.nim
```diff
@@ -0,0 +1,227 @@
...
+      boolFlag({PersistBlockFlag.NoFullValidation}, true) +
+      boolFlag(NoPersistBodies, true) +
```
tersec, line 112: Why is NoPersistBodies` not a `{NoPersistBodies}` but the are `{Foo}`?

nimbus/external_sync.nim
```diff
@@ -0,0 +1,227 @@
...
+
+    (eth1Block, prev_hash)
+  else:
+    error "CL head is not available"
+    quit(QuitFailure)
+
+proc loadBlocksFromBeaconChain(conf: NimbusConf) {.async.} =
+  let coreDB = AristoDbRocks.newCoreDbRef(string conf.dataDir, conf.dbOptions())
+
+  let com = CommonRef.new(
```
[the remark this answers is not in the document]
arnetheduck: (only one application may ever have write access to the db, else corruption will ensue)

nimbus/external_sync.nim
```diff
@@ -0,0 +1,227 @@
...
+    # Check if CL head is behind the EL head
+    if currentBlockNumber > head_number:
+      error "CL head is behind the EL head"
+      quit(QuitFailure)
+
+    if diff != 1:
+      let time1 = Moment.now()
+      while running and diff > 2:
+        prevHash = client.getRootFromHeader(BlockIdent.init(prevHash))
+        hashChain.add(prevHash)
```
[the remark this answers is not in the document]
arnetheduck: the risk with building a chain is that the CL changes head thus making the chain obsolete

nimbus/external_sync.nim
```diff
@@ -92,109 +68,93 @@ proc loadBlocksFromBeaconChain(conf: NimbusConf) {.async.} =
...
-      else:
-        const maxBlocks = 2000 # Load 2000 block at a time
-        for i in 0 ..< min(hashChain.len, maxBlocks):
-          let (eth1blck, _) =
-            client.getBlockFromBeaconChain(BlockIdent.init(hashChain.pop()), clConfig)
-          blocks.add(eth1blck)
-        
-        if hashChain.len == 0:
-          blocks.add(curBlck)
+  if conf.networkId == MainNet:
```
tersec, line 88:
```nim
let (clConfig, lastEra1Block, firstSlotAfterMerge) =
  case conf.networkId
  of MainNet:
    (...)
  of Sepolia:
    (...)
  of Holesky:
    (..)
  else:
    (...)
    quit(...)
```
avoids
- the period where `clConfig`, `lastEra1Block`, and `firstSlotAfterMerge` are not initialized to any useful value
- the possibility that one of those gets theoretically initialized but the others do not because they're separate assignment statements
- the repeated checks of `if conf.networkId == ....`
- because of being able to couple declaration and initialization, allows them all to be `let` rather than `var` with the associated easier debugging

and allows for the Nim and C compilers to figure out the best way to do the branching, which might or might not be this sequential `==` (e.g., maybe it does perfect hashing, up to it).

e.g.
```nim
let s = "a"
let h = case s
of "s": "FOO"
of "a": "BAR"
else: quit(0)
echo h
```
```

File: nimbus/config.nim [hunk 14]
```diff
@@ -391,6 +391,11 @@ type
       defaultValue: ""
       name: "beacon-api" .}: string
 
+    eth1EngineApi* {.
```
tersec, line 394: While `nimbus-eth1` is the name of the `status-im` project, the broader ecosystem branding of this is the execution layer/EL.

nrpc/config.nim
```diff
@@ -0,0 +1,394 @@
...
+
+      rocksdbBlockCacheSize {.
+        hidden
+        defaultValue: defaultBlockCacheSize
+        defaultValueDesc: $defaultBlockCacheSize
+        name: "debug-rocksdb-block-cache-size".}: int
+
+    of `external_sync`:
+
+      # github.com/ethereum/execution-apis/
```
tersec, line 285: This breaks up the URL unusably so it has to be reconstructed by hand. Yes, 80 character lines where practical but that doesn't reasonably apply to single URLs like this.

Also the current version of the spec link for this is https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/authentication.md#key-distribution

nrpc/nrpc.nim
```diff
@@ -0,0 +1,329 @@
...
+    engineUrl = EngineApiUrl.init(
+        conf.eth1EngineApi,
+        jwtSecret
+      )
+
+    web3 = await engineUrl.newWeb3()
+    rpcClient = web3.provider
+    data = 
+      try:
+        await rpcClient.exchangeCapabilities(@["engine_exchangeTransitionConfigurationV1","engine_forkchoiceUpdatedV1","engine_getPayloadBodiesByHash","engine_getPayloadBodiesByRangeV1","engine_getPayloadV1","engine_newPayloadV1"])
```
tersec, line 126: And unlike the URL, this could readily be split across multiple lines to reduce it to under 80 characters

nrpc/config.nim
```diff
@@ -0,0 +1,394 @@
...
+    if not loadNetworkParams(p, result):
+      raise newException(ValueError, "failed to load customNetwork")
+  except CatchableError:
+    raise newException(ValueError, "failed to load customNetwork")
+
+func completeCmdArg(T: type NetworkParams, val: string): seq[string] =
+  return @[]
+
+
+proc getNetworkId(conf: NRpcConf): Option[NetworkId] =
```
tersec, line 325: This can/probably should be using an `Opt[NetworkId]`rather than `Option[NetworkId]`, which enables several idioms to use it more safely without the explicit `unwrap`-moral-equivalent requirement (i.e. also therefore `Opt.some Mainnet`, `Opt.some SepoliaNet`, etc).

nrpc/nrpc.nim
```diff
@@ -0,0 +1,315 @@
...
+      continue
+
+    clNum = blk.header.number
+    # decreasing the lower bound with each iteration
+    importedSlot += currentBlockNumber - clNum
+
+  notice "Found the slot to start with", slot = importedSlot
+  importedSlot
+
+proc syncToEngineApi*(conf: NRpcConf) {.async.} =
```
tersec, line 102: Why is this exported?

nrpc/nrpc.nim
```diff
@@ -0,0 +1,315 @@
...
+    (curBlck, isAvailable) = client.getCLBlockFromBeaconChain(
+      BlockIdent.init(Slot(importedSlot)), clConfig
+    )
+    importedSlot += 1
+    if not isAvailable:
+      importedSlot += 1
+      continue
+
+    withBlck(curBlck):
+      when consensusFork >= ConsensusFork.Deneb:
```
tersec, line 173: Is the `consensusFork < ConsensusFork.Deneb` case handled at all, or should it be? It seems like this just ignores blocks older than `Deneb`. Maybe that's fine though, it assumes the EL is synced to a certain extent. If it's supposed to be able to sync an EL still in Shapella or older, I'm not sure how this can work though.

nrpc/nrpc.nim
```diff
@@ -0,0 +1,315 @@
...
+          finalizedHash = finalizedBlck.header.parentHash
+          (finalizedBlck, _) = client.getELBlockFromBeaconChain(BlockIdent.init(BlockIdentType.Finalized), clConfig)
+
+    
+    var
+      state = ForkchoiceStateV1(
+        headBlockHash: headHash.asBlockHash,
+        safeBlockHash: finalizedHash.asBlockHash,
+        finalizedBlockHash: finalizedHash.asBlockHash)
+      fcudata = await rpcClient.forkchoiceUpdated(state, Opt.none(PayloadAttributesV3))
```
tersec, line 196: Similarly, this assumes Deneb or Electra (both use fcUV3). Assuming fcUV3 isn't a general-purpose approach, though it does work for now. But combine https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#engine_forkchoiceupdatedv2 which notes
> 2. Consensus layer client **MUST** call this method instead of `engine_forkchoiceUpdatedV1` under any of the following conditions:
>    - `headBlockHash` references a block which `timestamp` is greater or equal to the Shanghai timestamp,
>    - `payloadAttributes` is not `null` and `payloadAttributes.timestamp` is greater or equal to the Shanghai timestamp.

with https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/cancun.md#update-the-methods-of-previous-forks which notes:
> This document defines how Cancun payload should be handled by the [`Shanghai API`](./shanghai.md).
>
> For the following methods:
>
> - [`engine_forkchoiceUpdatedV2`](./shanghai.md#engine_forkchoiceupdatedv2)
> - [`engine_newPayloadV2`](./shanghai.md#engine_newpayloadV2)
> - [`engine_getPayloadV2`](./shanghai.md#engine_getpayloadv2)
>
> a validation **MUST** be added:
>
> 1. Client software **MUST** return `-38005: Unsupported fork` error if the `timestamp` of payload or payloadAttributes greater or equal to the Cancun activation timestamp.

and one can't just use a single `forkchoiceUpdated` version, but have to use the correct one for each block, even if the actual arguments are identical (always `ForkchoiceStateV1`) compliantly with the engine API as specified.

Incidentally, if it happens to work on `nimbus-eth1` as an EL currently to use incorrect fcU version, that's also a bug. I found/filed a bunch of these issues a couple months ago, though there might be more.

nrpc/nrpc.nim
```diff
@@ -0,0 +1,315 @@
...
+    flags =
+      boolFlag({PersistBlockFlag.NoValidation}, conf.noValidation) +
+      boolFlag({PersistBlockFlag.NoFullValidation}, not conf.fullValidation) +
+      boolFlag(NoPersistBodies, not conf.storeBodies) +
+      boolFlag({PersistBlockFlag.NoPersistReceipts}, not conf.storeReceipts) +
+      boolFlag({PersistBlockFlag.NoPersistSlotHashes}, not conf.storeSlotHashes)
+    blocks: seq[EthBlock]
+    curBlck: EthBlock
+    client = RestClientRef.new(conf.beaconApi).valueOr:
+      error "Cannot connect to server"
```
tersec, line 237: Maybe include `conf.beaconApi` in the error message, as well as note in the text that it's the beacon API server not engine API server.

Sometimes people have logs of the error messages but not the command-line argument which started the process. It also reduces ambiguity in the case of multiple potential servers (as, indeed, there are -- beacon API and engine API). Right now only by reading the code could one figure out that the "server" here is the beacon API server/CL rather than engine API server/EL.

nrpc/nrpc.nim
```diff
@@ -0,0 +1,315 @@
...
+  ../nimbus/core/chain,
+  ./config,
+  ../nimbus/db/core_db/persistent,
+  ../nimbus/utils/era_helpers,
+  kzg4844/kzg_ex as kzg,
+  ../nimbus/core/eip4844,
+  web3, web3/[engine_api, primitives, conversions],
+  beacon_chain/spec/digest,
+  beacon_chain/el/el_conf,
+  beacon_chain/el/el_manager,
```
tersec, line 23: Does this still need the full `el_manager` or can it use `engine_api_conversions`?

nrpc/nrpc.nim
```diff
@@ -0,0 +1,315 @@
...
+template getELBlockFromBeaconChain(
+    client: RestClientRef, blockIdent: BlockIdent, clConfig: RuntimeConfig
+): (EthBlock, bool) =
+  let (clBlock, isAvailable) = getCLBlockFromBeaconChain(client, blockIdent, clConfig)
+
+  # Constructing the EL block from the CL block
+  var eth1block: EthBlock
+  if isAvailable:
+    eth1Block = clBlock.asTrusted().getEthBlock().valueOr:
+        error "Failed to get EL block from CL head"
```
tersec, line 63: should be 2-space indentation here

nrpc/nrpc.nim
```diff
@@ -0,0 +1,315 @@
...
+      uint64(await rpcClient.eth_blockNumber())
+    except CatchableError as exc:
+      error "Error getting block number", error = exc.msg
+      0'u64
+
+  var
+    currentBlockNumber = elBlockNumber()
+    curBlck: ForkedSignedBeaconBlock
+    client = RestClientRef.new(conf.beaconApi).valueOr:
+      error "Cannot connect to server"
```
tersec, line 144: See comment elsewhere about ambiguity of "server" (beacon API or engine API) and utility of including exactly which server (`conf.beaconApi`).

nrpc/nrpc.nim
```diff
@@ -0,0 +1,315 @@
...
+      error "Failed to persist blocks", error = statusRes.error
+      quit(QuitFailure)
+    info "Persisted blocks", blocks = blocks.len
+    blocks.setLen(0)
+
+  var (headBlck, _) = client.getELBlockFromBeaconChain(BlockIdent.init(BlockIdentType.Finalized), clConfig)
+
+  if headBlck.header.number <= currentBlockNumber:
+    notice "CL head is behind of EL head, or in sync", head = headBlck.header.number
+    quit(QuitSuccess)
```
tersec, line 252: I'm not sure if it's an initial goal, or -- but is the idea to keep restarting this external syncer or have a `tail -f`-like mode, or?

Because as written, it seems like it won't really keep the EL perpetually in sync but rather sync once and have to be restarted to sync more.

Maybe that's the design goal, I'm not sure. But in that case there still needs to be some way to maintain sync.

nrpc/nrpc.nim
tersec: I'm not really sure though, why call `exchangeCapabilities` at all. It's not necessary and the one place it's used is kind of misleading (`notice "Connected to Beacon Chain", data = data`, but `data`is about the engine API server/EL, not beacon API server/CL). Maybe better to skip the whole `exchangeCapabilities` thing and just use core engine API endpoints which have to be there anyway, which is what this code does already.

It was a good way of bootstrapping a test of, does the engine API calling work at all, but it's not obviously useful beyond that.

nrpc/nrpc.nim
```diff
@@ -0,0 +1,315 @@
...
+    web3 = await engineUrl.newWeb3()
+    rpcClient = web3.provider
+    data = 
+      try:
+        await rpcClient.exchangeCapabilities(@["engine_exchangeTransitionConfigurationV1","engine_forkchoiceUpdatedV1","engine_getPayloadBodiesByHash","engine_getPayloadBodiesByRangeV1","engine_getPayloadV1","engine_newPayloadV1"])
+      except CatchableError as exc:
+        error "Error Connecting to the EL Engine API", error = exc.msg
+        @[]
+
+  notice "Connected to Beacon Chain", data = data
```
tersec, line 131: See above, `data` here has nothing to do with the beacon chain

nrpc/config.nim
```diff
@@ -0,0 +1,394 @@
...
+
+# KLUDGE: The `load()` template does currently not work within any exception
+#         annotated environment.
+{.pop.}
+
+proc makeConfig*(cmdLine = commandLineParams()): NRpcConf
+    {.raises: [CatchableError].} =
+  ## Note: this function is not gc-safe
+
+  # The try/catch clause can go away when `load()` is clean
```
tersec, line 357: I'm a bit confused what it's accomplishing now even. It catches and re-raises whatever it gets anyway, so the result isn't in any obvious way more exception-safe or even narrowing the scope of exceptions (from `CatchableError`) to begin with.

nrpc/nrpc.nim
```diff
@@ -0,0 +1,315 @@
...
+    (getMetadataForNetwork("mainnet").cfg, 15537393'u64, 4700013'u64)
+  of SepoliaNet:
+    (getMetadataForNetwork("sepolia").cfg, 1450408'u64, 115193'u64)
+  of HoleskyNet:
+    (getMetadataForNetwork("holesky").cfg, 0'u64, 0'u64)
+  else:
+    error "Unsupported network", network = conf.networkId
+    quit(QuitFailure)
+
+template findSlot(client: RestClientRef, currentBlockNumber: uint64, lastEra1Block: uint64, firstSlotAfterMerge: uint64): uint64 =
```
tersec, line 82: Long line

nrpc/nrpc.nim
advaita-saha: For now these calls are kept for testing purposes since, the debugging for the segmentation fault is not yet completed

nrpc/config.nim
advaita-saha: Ahhh right makes sense
Migrating from `Option` to `Opt`

nrpc/nrpc.nim
advaita-saha: This assumption for now is solely based on my system. Once the segmentation is sorted, handling all the other cases is very easy as the structure would be exactly same, except for the payload version

nrpc/nrpc.nim
```diff
@@ -0,0 +1,315 @@
...
+    clNum = blk.header.number
+    # decreasing the lower bound with each iteration
+    importedSlot += currentBlockNumber - clNum
+
+  notice "Found the slot to start with", slot = importedSlot
+  importedSlot
+
+proc syncToEngineApi*(conf: NRpcConf) {.async.} =
+
+  template boolFlag(flags, b): PersistBlockFlags =
```
tersec, line 104: Is this particular `boolFlag` template used at all? The one at (currently) line 216 seems to be the one used.

nrpc/nrpc.nim
advaita-saha: Yeah makes sense to include that in the error message

nrpc/nrpc.nim
advaita-saha: Using the `ELConnection` for the jwt secret handling

nrpc/nrpc.nim
tersec: ok

nrpc/nrpc.nim
tersec: Sure, alright

nrpc/nrpc.nim
advaita-saha: As far as I understand, once the EL is synced to the tip
The CL can take over and make the engineAPI calls driving the CL, without the need for this external_sync

nrpc/nrpc.nim
tersec: Oh, indeed.

Well, points to more refactoring to be done in ELManager at some point, but for this PR's purposes, out of scope.

nrpc/nrpc.nim
tersec: In theory, yes. Switching over properly might be a bit tricky to time, but it should work.

nrpc/config.nim
```diff
@@ -0,0 +1,360 @@
...
+
+func defaultDataDir*(): string =
+  when defined(windows):
+    getHomeDir() / "AppData" / "Roaming" / "Nimbus"
+  elif defined(macosx):
+    getHomeDir() / "Library" / "Application Support" / "Nimbus"
+  else:
+    getHomeDir() / ".cache" / "nimbus"
+
+func defaultKeystoreDir*(): string =
```
tersec, line 40: Does anything here use this?

nrpc/config.nim
```diff
@@ -0,0 +1,360 @@
...
+      desc: "Enable metrics logging"
+      defaultValue: false
+      name: "log-metrics" .}: bool
+
+    logMetricsInterval* {.
+      desc: "Interval at which to log metrics, in seconds"
+      defaultValue: 10
+      name: "log-metrics-interval" .}: int
+
+    metricsEnabled* {.
```
tersec, line 135: Regarding `defaultMetricsServerPort`, `metricsEnabled`, `metricsPort`, `and `metricsAddress`, if I'm not mistaken, there appears to be no metrics support to begin with, so I'm not sure what the purpose of these command-line options is

nrpc/config.nim
```diff
@@ -0,0 +1,360 @@
...
+    beaconApi* {.
+      desc: "Beacon API url"
+      defaultValue: ""
+      name: "beacon-api" .}: string
+
+    network {.
+      desc: "Name or id number of Ethereum network(mainnet(1), sepolia(11155111), holesky(17000), other=custom)"
+      longDesc:
+        "- mainnet: Ethereum main network\n" &
+        "- sepolia: Test network (proof-of-work)\n" &
```
tersec, line 91: Sepolia hasn't been proof of work in years, though, sure, it began that way

Makefile
```diff
@@ -213,13 +216,17 @@ update-from-ci: | sanity-checks update-test
...
 		echo -e $(BUILD_MSG) "build/$@" && \
-		$(ENV_SCRIPT) nim c $(NIM_PARAMS) -o:build/$@ "$${TOOL_DIR}/$@.nim"
+		$(ENV_SCRIPT) nim c $(NIM_PARAMS) -d:chronicles_log_level=TRACE -o:build/$@ "$${TOOL_DIR}/$@.nim"
 
 # a phony target, because teaching `make` how to do conditional recompilation of Nim projects is too complicated
 nimbus: | build deps rocksdb
 	echo -e $(BUILD_MSG) "build/$@" && \
 		$(ENV_SCRIPT) nim c $(NIM_PARAMS) -d:chronicles_log_level=TRACE -o:build/$@ "nimbus/$@.nim"
 
+# external_sync: | build deps rocksdb
```
tersec, line 226: Some commented code -- is the intent to keep it commented, to remove it, to uncomment it, something else?

nrpc/config.nim
advaita-saha: Actually was planing to add metrics support to the sync-db, thus kept the metrics option

nimbus/config.nim
```diff
@@ -377,6 +377,16 @@ type
...
 
+    beaconApi* {.
+      desc: "Beacon API url"
+      defaultValue: ""
+      name: "beacon-api" .}: string
+
+    eth1EngineApi* {.
+      desc: "Eth1 Engine API url"
+      defaultValue: ""
+      name: "eth1-engine-api" .}: string
```
arnetheduck, line 388:
```suggestion
      name: "engine-api" .}: string
```
`eth1` has been deprecated as a name

advaita-saha: Nimbus config shouldn't have this only, should be removed fully 
It lies in the `nrpc/config.nim`

PR #2599, opened 2024-09-09
Re-enable eth_getProof implementation

web3-developer:
This PR re-enables the eth_getProof endpoint which was previously disabled when switching over from the legacy db to Aristo. 

The Aristo proof and slotProof functions are working for the success case when the values exist and are passing all the tests. Proofs for non-existent values as described [here](https://eips.ethereum.org/EIPS/eip-1186#proofs-for-non-existent-values) are not yet working and needs to be fixed in the Aristo implementation.


web3-developer:
This can be merged once the Aristo fixes are completed.

nimbus/db/ledger/backend/accounts_ledger.nim
```diff
@@ -858,6 +858,22 @@ proc getEthAccount*(ac: AccountsLedgerRef, address: EthAddress): Account =
     raiseAssert "getAccount(): cannot convert account: " & $$rc.error
   rc.value
 
+proc getAccountProof*(ac: AccountsLedgerRef, address: EthAddress): seq[seq[byte]] =
```
arnetheduck, line 861: why `seq[byte]` in the inner layers? aren't the proof items constant-size?

web3-developer: Not constant size in this case. The inner seq[byte] is a rlp encoded node from the mpt trie. The length depends on a number of factors such as the type of node, leaf, extension or branch node and the content of the node.

nimbus/db/ledger/backend/accounts_ledger.nim
```diff
@@ -862,15 +862,19 @@ proc getAccountProof*(ac: AccountsLedgerRef, address: EthAddress): seq[seq[byte]
...
 proc getStorageProof*(ac: AccountsLedgerRef, address: EthAddress, slots: openArray[UInt256]): seq[seq[seq[byte]]] =
   var storageProof = newSeqOfCap[seq[seq[byte]]](slots.len)
 
   for slot in slots:
     let slotProof = ac.ledger.slotProof(keccakHash(address), keccakHash(toBytesBE(slot))).valueOr:
-      raiseAssert "getStorageProofs() cannot get slotProof: " & $$error
-    storageProof.add(slotProof)
+      if error.aErr == FetchPathNotFound:
+        storageProof.add(@[])
+        continue
```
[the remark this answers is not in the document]
mjfh: It should. Internally it tries to fetch the account leaf first which might not exist. So `FetchPathNotFound` should be returned.

PR #2600, opened 2024-09-09
Jordan

mjfh:
Status:
Proof of concept. While functional, this code is only useful when completed by fetching block bodies and executing blocks.

Throughput (anecdotal):
Observed download times for 14m `MainNet` headers varies between 30min  and 1h (Era1 size truncated to 66m blocks.), full download ~52min. The number of peers downloading concurrently is crucial  here.


mjfh:
Not meant to be pushed

PR #2607, opened 2024-09-10
Update flare header download mechanism

mjfh:


nimbus/sync/flare/worker.nim
```diff
@@ -8,6 +8,7 @@
 # at your option. This file may not be copied, modified, or distributed
 # except according to those terms.
 
+
```
tersec, line 11: ?

nimbus/sync/flare/worker/staged/linked_hchain.nim
```diff
@@ -20,7 +20,7 @@ const
   extraTraceMessages = false # or true
     ## Enabled additional logging noise
 
-  verifyLinkedHChainOk = not defined(release) # or true
+  verifyDataStructureOk = not defined(release) # or true
```
tersec, line 23: These should not depend on `defined(release)` at all. If one wants a debugging define, ok, but it should be separately controlled, not conflated with optimization level, et cetera. Otherwise, when one tries to narrow down an issue by switching to debug-mode compiles (because it disables the C optimizations) suddenly one is debugging just an entirely different piece of software.

mjfh: ok

nimbus/sync/flare/worker/staged/headers.nim
```diff
@@ -24,6 +24,44 @@ logScope:
 const extraTraceMessages = false # or true
   ## Enabled additional logging noise
 
+# ------------------------------------------------------------------------------
+# Private functions
+# ------------------------------------------------------------------------------
+
+# Copied from `nimbus_import`
+func shortLog(a: chronos.Duration, parts = int.high): string {.inline.} =
```
tersec, line 32: Any particular reason it should be `{.inline.}`? It's not that lightweight a function anyway once the template instantiations expand.

nimbus/sync/flare/worker/staged/headers.nim
```diff
@@ -24,6 +24,44 @@ logScope:
 const extraTraceMessages = false # or true
   ## Enabled additional logging noise
 
+# ------------------------------------------------------------------------------
+# Private functions
+# ------------------------------------------------------------------------------
+
+# Copied from `nimbus_import`
```
tersec, line 31: Perhaps worthwhile to refactor and reuse? It's a basically 30-line function, it's not that short.

nimbus/sync/flare/worker/staged/headers.nim
mjfh: No, copied from `jacek`

nimbus/sync/flare/worker/staged.nim
```diff
@@ -27,14 +27,14 @@ const
   extraTraceMessages = false # or true
     ## Enabled additional logging noise
 
-  verifyStagedQueueOk = not defined(release) or true
+  verifyDataStructureOk = false or true
```
tersec, line 30: Is this meant to be always on or always off by default? Because right now it's pinned to being on.

`const extraTraceMessages = false # or true` does the opposite.

If this is intentional, ok. But it's kind of unclear.

mjfh: true, for debugging I only insert a `#`; once I go into production I will remove the tail

PR #2608, opened 2024-09-10
Migrate `keyed_queue` to `minilru`

arnetheduck:
Compared to `keyed_queue`, `minilru` uses significantly less memory, in particular for the 32-byte hash keys where `kq` stores several copies of the key redundantly.


arnetheduck:
~200mb savings in "cache overhead" (everything that was not part of cache overhead removed) - we can also simplify the code and remove the `AccountKey` hack because the cache no longer stores multiple copies of the key - further refactoring also possible

image

PR #2609, opened 2024-09-11
Fluffy: Prevent removal of bootstrap nodes in local networks.

web3-developer:
When network=none we don't connect to mainnet or the testnet so we likely need to use our own bootstrap nodes in this case rather than use the default ones. 

I found that Fluffy would remove its peer nodes from the routing table if any timeout occurs which breaks things in a small local network such as when there are only two nodes running. In this case the nodes would need to be restarted to force them to reconnect to each other.

This change prevents the nodes from being removed if they are defined as bootstrap nodes and network=none.


web3-developer:
On second thought, I won't merge this because I've discovered that by using the launch_local_testnet.sh script I'm not hitting the timeout then disconnect issue as long as I run with at least 3 nodes, and so this change isn't required. The testnet script starts the first node then uses the enr file as the bootnode after the enr update has occurred which is then used by the other nodes to connect.

PR #2611, opened 2024-09-11
Prague types conversion

jangko:


nimbus/beacon/payload_conv.nim
```diff
@@ -35,6 +35,13 @@ func txRoot(list: openArray[Web3Tx]): common.Hash256
   {.noSideEffect.}:
     calcTxRoot(ethTxs(list))
 
+func requestsRoot(p: ExecutionPayload): Opt[common.Hash256]
+             {.gcsafe, raises:[].} =
+  {.noSideEffect.}:
```
tersec, line 40: This is presumably because
https://github.com/status-im/nimbus-eth1/blob/75808bc03b5d59d904a052427d7131ba65b79cd3/nimbus/utils/utils.nim#L24-L28
has side effects, but it's unclear to me why it should, in principle.

jangko: function pointer call?

PR #2615, opened 2024-09-12
Fluffy testnet script updates

web3-developer:


fluffy/scripts/launch_local_testnet.sh
```diff
@@ -34,10 +34,10 @@ if [ ${PIPESTATUS[0]} != 4 ]; then
 fi
 
 OPTS="h:n:d"
-LONGOPTS="help,nodes:,data-dir:,run-tests,log-level:,base-port:,base-rpc-port:,trusted-block-root:,portal-bridge,base-metrics-port:,reuse-existing-data-dir,timeout:,kill-old-processes"
+LONGOPTS="help,nodes:,data-dir:,run-tests,log-level:,base-port:,base-rpc-port:,trusted-block-root:,portal-bridge,base-metrics-port:,reuse-existing-data-dir,timeout:,kill-old-processes,skip-build,portal-subnetworks:,disable-state-root-validation"
 
 # default values
-NUM_NODES="64"
```
web3-developer, line 40: 64 as the default is high. Is there a specific reason for the higher default?

fluffy/scripts/launch_local_testnet.sh
```diff
@@ -195,16 +218,21 @@ if [[ "${HAVE_LSOF}" == "1" ]]; then
...
 BINARIES="fluffy"
 if [[ "${PORTAL_BRIDGE}" == "1" ]]; then
   BINARIES="${BINARIES} portal_bridge"
 fi
-$MAKE -j ${NPROC} LOG_LEVEL=TRACE ${BINARIES}
 
-if [[ "$RUN_TESTS" == "1" ]]; then
-  TEST_BINARIES="test_portal_testnet"
-  $MAKE -j ${NPROC} LOG_LEVEL=INFO ${TEST_BINARIES}
+if [[ "${SKIP_BUILD}" == "1" ]]; then
```
web3-developer, line 226: If found it useful to be able to skip the build when rerunning and the binary already exists.

fluffy/scripts/launch_local_testnet.sh
```diff
@@ -218,13 +246,6 @@ cleanup() {
   for BINARY in ${BINARIES}; do
     pkill -f -9 -P $$ ${BINARY} &>/dev/null || true
   done
-
-  # Delete the binaries we just built, because these are with none default logs.
-  # TODO: When fluffy gets run time log options a la nimbus-eth2 we can keep
-  # the binaries around.
-  for BINARY in ${BINARIES}; do
-    rm build/${BINARY}
-  done
```
web3-developer, line 227: My guess was that this isn't needed anymore because fluffy now supports setting the log level?

fluffy/scripts/launch_local_testnet.sh
kdeme: Not really, just that it was mostly used for running in CI and there it was being tested at 64.
When used for more manual testing (which I already do for beacon network, i usually use 4 or 8 nodes there), it does make sense to set it lower.

fluffy/scripts/launch_local_testnet.sh
kdeme: Yes, seems like this can be removed now.

kdeme (approved):

PR #2620, opened 2024-09-12
Add BlockHeader by number to Portal history network

kdeme:
- Add new content + content key functionality for header by number
- Remove EpochRecords from the network
- Add pruning call for the EpochRecords + required deprecated functionality
- Adjust getBlock and getBlockHashByNumber to make use of the new functionality instead
- Delete content_verifier as it was only verifying the now deprecated EpochRecord


kdeme:
This implements this spec change: https://github.com/ethereum/portal-network-specs/pull/334

Follow-up PR will be done for support of injecting block by number in our `portal_bridge`

PR #2627, opened 2024-09-13
Flare sync

mjfh:
Status
* Headers download works
* Bodies download + execution/import works
* Metrics available

Todo
* More quantitative tests (seems ok so far)


nimbus/sync/flare/worker_desc.nim
```diff
@@ -207,6 +141,32 @@ proc `$`*(w: BnRange): string =
 proc bnStr*(w: BlockNumber): string =
   "#" & $w
 
+# Source: `nimbus_import.shortLog()`
```
tersec, line 144: Should be refactored and reused, not copy/pasted

nimbus/sync/flare/worker/update.nim
```diff
@@ -17,12 +17,12 @@ import
...
   ../worker_desc,
   ./update/metrics,
-  "."/[db, unproc]
+  "."/[db, headers_unproc]
 
 logScope:
   topics = "flare update"
 
-const extraTraceMessages = false # or true
+const extraTraceMessages = false
```
tersec, line 25: The codebase already has a lot of these, I realize, but the existence of `extraTraceMessages` seems like a hack, and it would be nicer not to add more of it. It's basically creating an additional ad-hoc Chronicles logging level.

`trace` is already intentionally/allowed to be very verbose. An alternative goal might be to realign logging levels so `trace` can accommodate whatever useful `extraTraceMessages` exist, and that `notice`, `info`, `warn`, `error`, etc levels are also more aligned with end-users.

nimbus/sync/flare/worker/start_stop.nim
```diff
@@ -11,20 +11,19 @@
...
+  "."/[headers_staged, headers_unproc]
 
 when enableTicker:
   import ./start_stop/ticker
 
 logScope:
   topics = "flare start/stop"
 
-const extraTraceMessages = false or true
+const extraTraceMessages = false
```
tersec, line 26: See other comment about `extraTraceMessages`

nimbus/sync/flare/worker/headers_staged.nim
```diff
@@ -295,82 +283,70 @@ proc stagedProcess*(ctx: FlareCtxRef; info: static[string]): int =
 
     result.inc # count records
 
-    when extraTraceMessages:
-      trace info & ": staged record saved", iv, layout=ok, nSaved=result
+    #when extraTraceMessages:
```
tersec, line 286: And `when foo`, then being commented out, well, the point of `when foo` is to conditional compilation to begin with.

Better to avoid the build-up of random commented-out code which had plagued Nimbus until recently.

nimbus/sync/flare/worker.nim
```diff
@@ -56,15 +75,29 @@ proc release*(ctx: FlareCtxRef) =
 
 proc start*(buddy: FlareBuddyRef): bool =
   ## Initialise worker peer
-  if buddy.startBuddy():
+  const info = "RUNSTART"
+
+  when runOnSinglePeerOnly:
+    if 0 < buddy.ctx.pool.nBuddies:
+      debug info & " single peer already connected",
```
tersec, line 82: Shouldn't the `maxPeers` parameter to `init` be able to handle this case more generally, with a value of 1?

nimbus/sync/flare/worker/start_stop.nim
arnetheduck: indeed - one of the steps to bring the codebase to beta quality will be to remove all such temporary debug code and either promote it to "real" code, move it into unit tests or simply remove it - these hacks may be useful during development but they should eventually result in a decision to either keep or remove them.

nimbus/sync/flare/worker/update.nim
mjfh: It is a hack which will eventually vanish

nimbus/sync/flare/worker/headers_staged.nim
mjfh: Ditto

nimbus/sync/flare/worker.nim
mjfh: I have planned already for limiting the `flare` part only as `maxPeers` seems to control the hole peer pool in a way.

It seems that double digit numbers of peers slow down response times considerably (will continue observe that.) So it makes sense to limit the peers accepted for sync interdependently of the max number of peers available for service.

nimbus/sync/flare/worker_desc.nim
mjfh: ?? this is the only place where it is defined

tersec: https://github.com/status-im/nimbus-eth1/blob/5cd02974621d70a952b801af13a2ce0cc9260465/nimbus/nimbus_import.nim#L36-L61

nimbus/sync/flare/worker.nim
tersec: To be more concrete:
https://github.com/status-im/nimbus-eth1/blob/6dc8ab8393502d43bee3243340bf7db01be54490/nimbus/nimbus.nim#L146-L148

Would replacing that with:
```nim
     of SyncMode.Flare: 
       nimbus.flareSyncRef = FlareSyncRef.init( 
         nimbus.ethNode, nimbus.chainRef, 1)
```
be mostly equivalent to `runOnSinglePeerOnly`?

nimbus/sync/flare/worker/headers_staged.nim
tersec: Well, in this case though it's doubly odd. Even if one stipulates utility for `extraTraceMessages`, this is a _commented-out_ `when extraTraceMessages:` block. Why both comment it out and use `when extraTraceMessages:`?

nimbus/sync/flare/worker/update.nim
tersec: > It is a hack which will eventually vanish

Is there a plan for how/under what conditions?
```
nimbus-eth1$ rg "when extraTraceMessages|const extraTraceMessages" | wc -l
110
```

nimbus/sync/flare/worker_desc.nim
```diff
@@ -140,28 +88,19 @@ type
 
   FlareBuddyData* = object
     ## Local descriptor data extension
-    fetchBlocks*: BnRange
+    #fetchBlocks*: BnRange
```
tersec, line 91: Is this commented-out `fetchBlocks` still useful, or?

nimbus/sync/flare/worker_desc.nim
```diff
@@ -140,28 +88,19 @@ type
...
-    coins*: uint64                   ## Sequence of fliopped coins
+    # Debugging and logging.
+    nMultiLoop*: int                 ## Number of runs
+    stoppedMultiRun*: chronos.Moment ## Time when run-multi stopped
+    multiRunIdle*: chronos.Duration  ## Idle time between runs
 
   FlareCtxData* = object
     ## Globally shared data extension
-    rng*: ref HmacDrbgContext        ## Random generator, pre-initialised
     lhcSyncState*: LinkedHChainsSync ## Syncing by linked header chains
```
tersec, line 101: Is `lhcSyncState` used anywhere?

nimbus/sync/flare/worker_config.nim
```diff
@@ -0,0 +1,72 @@
...
+
+  metricsUpdateInterval* = chronos.seconds(10)
+    ## Wait at least this time before next update
+
+  daemonWaitInterval* = chronos.seconds(30)
+    ## Some waiting time at the end of the daemon task which always lingers
+    ## in the background.
+
+  runNoHeadersIdleWaitInterval* = chronos.seconds(30)
+    ## Sllep some time in multi-mode if there is nothing to do
```
tersec, line 33: `Sleep some ...`

nimbus/sync/flare/worker_config.nim
```diff
@@ -0,0 +1,72 @@
...
+
+  # ----------------------
+
+  nFetchHeadersRequest* = 1_024
+    ## Number of headers that will be requested with a single `eth/xx` message.
+    ##
+    ## On `Geth`, responses to larger requests are all truncted to 1024 header
```
tersec, line 40: `truncated`

nimbus/sync/flare/worker_desc.nim
```diff
@@ -11,64 +11,13 @@
 {.push raises:[].}
 
 import
-  pkg/[bearssl/rand, chronos, chronos/timer],
+  pkg/[bearssl/rand, chronos], # chronos/timer],
```
tersec, line 14:
```
nimbus-eth1/nimbus/sync/flare/worker_desc.nim(14, 15) Warning: imported and not used: 'rand' [UnusedImport]
```

nimbus/sync/flare/worker.nim
```diff
@@ -133,36 +175,75 @@ proc runMulti*(buddy: FlareBuddyRef) {.async.} =
...
 
-  if ctx.unprocTotal() == 0 and ctx.stagedChunks() == 0:
-    # Save cpu cycles waiting for something to do
-    when extraTraceMessages:
-      debug info & ": idle wasting time", peer
-    await sleepAsync runMultiIdleWaitInterval
-    return
+  else:
+    let
+      ctx = buddy.ctx
```
tersec, line 183:
```
nimbus-eth1/nimbus/sync/flare/worker.nim(183, 7) Hint: 'ctx' is declared but not used [XDeclaredButNotUsed]
```

nimbus/sync/flare/worker/headers_staged.nim
```diff
@@ -295,82 +283,70 @@ proc stagedProcess*(ctx: FlareCtxRef; info: static[string]): int =
```
tersec, line 283:
```
nimbus-eth1/nimbus/sync/flare/worker/headers_staged.nim(282, 9) Hint: 'ok' is declared but not used [XDeclaredButNotUsed]
```

nimbus/sync/flare/worker.nim
```diff
@@ -25,6 +25,28 @@ logScope:
...
+  0 < buddy.ctx.headersUnprocTotal() and
+    buddy.ctrl.running and
+    not buddy.ctx.poolMode
+
+proc napUnlessHeadersToFetch(
+    buddy: FlareBuddyRef;
+    info: static[string];
+      ): Future[bool] {.async.} =
+  ## When idle, save cpu cycles waiting for something to do.
+  let ctx = buddy.ctx
```
tersec, line 42:
```
nimbus-eth1/nimbus/sync/flare/worker.nim(42, 7) Hint: 'ctx' is declared but not used [XDeclaredButNotUsed]
```

nimbus/sync/flare/worker/db.nim
```diff
@@ -139,31 +156,37 @@ proc dbStoreLinkedHChainsLayout*(ctx: FlareCtxRef): bool =
...
+    return false
+
+  when extraTraceMessages:
+    trace info & ": saved pesistently on DB"
   true
 
 
 proc dbLoadLinkedHChainsLayout*(ctx: FlareCtxRef) =
   ## Restore chain layout from persistent db
   const info = "dbLoadLinkedHChainsLayout"
```
tersec, line 177:
```
nimbus-eth1/nimbus/sync/flare/worker/db.nim(177, 9) Hint: 'info' is declared but not used [XDeclaredButNotUsed]
```

nimbus/sync/flare/worker_desc.nim
mjfh: Sorry, thought you meant the `bnStr()` which was prominently displayed here. 

Jaceks pretty-printer has a different resolution for a long running process (w, d, h etc.). He also wants his function treated `inline`.

nimbus/sync/flare/worker_config.nim
mjfh: See intro comment

nimbus/sync/flare/worker_desc.nim
tersec: > He also wants his function treated `inline`.

He does? https://status-im.github.io/nim-style-guide/language.inline.html
> Avoid using explicit `{.inline.}` functions.

I believe that Jacek wrote that.

nimbus/sync/flare/worker_config.nim
tersec: I don't see how this explains it. The comment line
```
On `Geth`, responses to larger requests are all truncted to 1024 header
```
misspells the word "truncated" as "truncted".

nimbus/sync/flare/worker_desc.nim
mjfh: maybe, see introductory comment

tersec: If it's useful, in this PR, by merge time, alright. If not, then it shouldn't go in. It's evidently not used at all right now.

But it would be good to stop adding new and commented-out code to `nimbus-eth1`, just in case. We're approaching production readiness and at that point either code will be useful under well-defined circumstances, or it won't.

Basically same spirit as the trace++ logging level.

And, for things only useful with certain `when` conditional compilation settings:
1. this is a good reason not to have too much a proliferation of those, precisely because it makes testing for working compilation across all accessible settings literally exponentially harder. What if some specific combination of debug flags needs that variable which seems otherwise unused, etc?
2. then this needs some kind of CI coverage, or clear documentation under what circumstances it's useful. for example, if it's `when defined(foo) and not defined(bar)`, alright, but then that should be spelled out rather than just ad-hoc commented out

nimbus/sync/flare/worker/blocks_staged/debug.nim
```diff
@@ -0,0 +1,133 @@
...
+type
+  BlocksForImportQueueWalk = SortedSetWalkRef[BlockNumber,BlocksForImport]
+    ## Traversal descriptor (for `verifyStagedBlocksQueue()`)
+
+# ------------------------------------------------------------------------------
+# Public debugging helpers
+# ------------------------------------------------------------------------------
+
+proc rlpSize*(blk: ref BlocksForImport): int =
+  rlp.encode(blk[]).len
```
tersec, line 30: Is there a more efficient way of deriving this information than actually doing the encoding? That's a lot of work for an exercise in counting.

nimbus/sync/flare/worker/headers_staged/debug.nim
```diff
@@ -0,0 +1,111 @@
+# Nimbus
```
tersec, line 1: Is there a general approach to keeping these `debug` modules (this isn't the only one, but it's in this PR, so it's salient here) from bitrotting automatically by ensuring the CI knows how to test-compile them?

nimbus/sync/flare/worker/headers_staged/headers.nim
```diff
@@ -94,9 +70,8 @@ proc headersFetchReversed*(
...
     start = Moment.now()
 
-  when extraTraceMessages:
-    trace trEthSendSendingGetBlockHeaders & " reverse", peer, ivReq,
-      nReq=req.maxResults, useHash
+  trace trEthSendSendingGetBlockHeaders & " reverse", peer, ivReq,
+    nReq=req.maxResults, useHash, nRespErrors=buddy.only.nHdrRespErrors
 
   # Fetch headers from peer
   var resp: Option[blockHeadersObj]
```
tersec, line 77: New code should generally use `Opt` over `Option`

nimbus/sync/flare/worker/headers_staged/linked_hchain.nim
```diff
@@ -0,0 +1,73 @@
...
+  rev: seq[BlockHeader];
+  buddy: FlareBuddyRef;
+  topNumber: BlockNumber;
+  lhc: ref LinkedHChain; # update in place
+  info: static[string];
+    ): bool =
+  ## Returns sort of `lhc[] += rev[]` where `lhc[]` is updated in place.
+
+  # Verify top block number
+  assert 0 < rev.len # debugging only
```
tersec, line 33: Still should be `doAssert` unless it's some critical performance issue (but then probably shouldn't be asserting every iteration of some inner loop anyway). The goal is generally to have the release and debug builds vary as little as they can, and Nim `seq`s efficiently support `len`.

This is not some tine function, it does an `rlp.encode()`, various `seq` `setLen`, iterates over `seq`s, etc. Creating differences in the debug builds just means some behaviors might magically appear/disappear/be "solved" by running one or the other.

nimbus/sync/flare.nim
```diff
@@ -107,14 +107,12 @@ proc init*(
...
-    rng: ref HmacDrbgContext;
     maxPeers: int;
-    era1Dir: string;
+    chunkSize: int;
       ): T =
   new result
   result.initSync(ethNode, chain, maxPeers)
-  result.ctx.pool.rng = rng
-  result.ctx.pool.e1Dir = era1Dir
+  result.ctx.pool.nBodiesBatch = chunkSize
```
tersec, line 115: https://status-im.github.io/nim-style-guide/language.result.html

nimbus/sync/flare/README.md
```diff
@@ -127,3 +113,126 @@ If a *relatively complete* header chain is reached for the first time, the
...
+The headers corresponding to the half open interval `(T,B]` can be completed by
+fetching block bodies and then imported/executed.
+
+Running the sync process for *MainNet*
+--------------------------------------
+
+For syncing, a beacon node is needed that regularly informs via *RPC* of a
+recently finalised block header.
+
+The beacon node program used here is the *nimbus_beacon_node* binary from the
```
tersec, line 140: Indeed, `nimbus_beacon_node` is probably the best default to list, but there's also https://nimbus.guide/el-light-client.html

Is this compatible with flare?

nimbus/sync/flare/README.md
```diff
@@ -127,3 +113,126 @@ If a *relatively complete* header chain is reached for the first time, the
...
+
+### Syncing on a low memory machine
+
+On a system with memory with *8GiB* the following additional options proved
+useful for *nimbus* to reduce the memory footprint.
+
+For the *Era1* pre-load (if any) the following extra options apply to
+"*nimbus import*":
+
+       --chunk-size=1024
```
tersec, line 190: inconsistent indentation

nimbus/sync/flare/README.md
```diff
@@ -127,3 +113,126 @@ If a *relatively complete* header chain is reached for the first time, the
...
+       --debug-rocksdb-block-cache-size=1073741824
+       --debug-rdb-key-cache-size=67108864
+       --debug-rdb-vtx-cache-size=268435456
+
+Also, to reduce the backlog for *nimbus-eth2* stored on disk, the following
+changes might be considerd. For file
+*nimbus-eth2/vendor/mainnet/metadata/config.yaml* change setting constants:
+
+       MIN_EPOCHS_FOR_BLOCK_REQUESTS: 33024
+	   MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS: 4096
```
tersec, line 208: inconsistent indentation

nimbus/sync/flare/README.md
```diff
@@ -127,3 +113,126 @@ If a *relatively complete* header chain is reached for the first time, the
...
+to
+
+       MIN_EPOCHS_FOR_BLOCK_REQUESTS: 8
+	   MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS: 8
```
tersec, line 212: inconsistent indentation

nimbus/sync/flare/README.md
```diff
@@ -127,3 +113,126 @@ If a *relatively complete* header chain is reached for the first time, the
...
+
+       --debug-flare-chunk-size=384
+       --debug-rocksdb-max-open-files=384
+       --debug-rocksdb-write-buffer-size=50331648
+       --debug-rocksdb-block-cache-size=1073741824
+       --debug-rdb-key-cache-size=67108864
+       --debug-rdb-vtx-cache-size=268435456
+
+Also, to reduce the backlog for *nimbus-eth2* stored on disk, the following
+changes might be considerd. For file
```
tersec, line 204: "considered"

nimbus/sync/flare/worker/headers_staged/headers.nim
tersec: oops, it's an API issue

nimbus/sync/flare/worker/headers_unproc.nim
```diff
@@ -44,62 +44,83 @@ proc unprocFetch*(
...
+  ctx.lhc.unprocessed.clear()
+  ctx.lhc.borrowed = 0u
+
+proc headersUnprocSet*(ctx: FlareCtxRef; iv: BnRange) =
+  ## Set up new unprocessed range
+  ctx.headersUnprocSet()
+  discard ctx.lhc.unprocessed.merge(iv)
+
+proc headersUnprocSet*(ctx: FlareCtxRef; minPt, maxPt: BlockNumber) =
+  ## Ditto
```
tersec, line 120: [Documentation comments](https://nim-lang.org/docs/manual.html#lexical-analysis-comments) should generally be more self-contained than this, since they're intended to be extracted by automated tooling which might display them without any other context aside from the name of the function they're in, so the fact that 5 lines earlier the referent for "ditto" is defined doesn't help for them.

Can either flesh this out or make it a non-documentation comment

nimbus/sync/flare/worker/start_stop.nim
```diff
@@ -36,30 +28,33 @@ when enableTicker:
...
-    var count = 0
-  result = proc(h: BlockHeader) {.gcsafe, raises: [].} =
-    if ctx.lhc.beacon.header.number < h.number:
-      when extraTraceMessages:
-        if count mod 77 == 0: # reduce some noise
-          trace "updateBeaconHeaderCB", blockNumber=("#" & $h.number), count
-        count.inc
-      ctx.lhc.beacon.header = h
-      ctx.lhc.beacon.changed = true
+  result = proc(h: Hash256) {.gcsafe, raises: [].} =
```
tersec, line 54: https://status-im.github.io/nim-style-guide/language.result.html

nimbus/sync/flare/worker/start_stop.nim
```diff
@@ -36,30 +28,33 @@ when enableTicker:
...
+    # Rpc checks empty header against `Hash256()` rather than `EMPTY_ROOT_HASH`
+    if ctx.lhc.beacon.finalised == Hash256():
```
tersec, line 56: Using the `ZERO_HASH256` (which is indeed not `EMPTY_ROOT_HASH`) avoids constructing this at runtime, which Nim does, repeatedly and inefficiently

nimbus/sync/flare/worker/blocks_staged.nim
```diff
@@ -0,0 +1,277 @@
...
+proc fetchAndCheck(
+    buddy: FlareBuddyRef;
+    ivReq: BnRange;
+    blk: ref BlocksForImport; # update in place
+    info: static[string];
+      ): Future[bool] {.async.} =
+
+  let
+    ctx = buddy.ctx
+    offset = blk.blocks.len.uint
```
tersec, line 39: Every use of `offset` here is naturally/natively `int`: `seq` indexing and `setLen`. Sometimes it gets converted back to `int`, which is safe as far as it goes since it came from `int` from `len`.

But I'm not sure why it's ever converted to `uint` to begin with.

nimbus/sync/flare/worker/blocks_staged.nim
```diff
@@ -0,0 +1,277 @@
...
+  let bodies = block:
+    let rc = await buddy.bodiesFetch(blockHash, info)
+    if rc.isErr:
+      blk.blocks.setLen(offset)
+      return false
+    rc.value
+
+  # Append bodies, note that the bodies are not fully verified here but rather
+  # when they are imported and executed.
+  let nBodies = bodies.len.uint
```
tersec, line 67: Similar comment about `uint` as in https://github.com/status-im/nimbus-eth1/pull/2627/files#r1778587000

`n` is currently also `uint` due to the `for` loop bounds, and that actually does have some advantages, because Nim doesn't do some pointless range/index checks every loop iteration this way. But `nBodies` itself is still used in contexts where either it can be `int` or `uint` (logging), or it definitely should be `int` (`setLen`).

To retain the loop efficiency, maybe keep that as `nBodies.uint` if that seems useful, but in general, I'm not sure why `nBodies` overall is converted to `uint`.

nimbus/sync/flare/worker/blocks_staged.nim
```diff
@@ -0,0 +1,277 @@
...
+          n, nTxs=bodies[n].transactions.len, nBodies,
+          nRespErrors=buddy.only.nBdyRespErrors
+        break loop
+
+      blk.blocks[offset + n].transactions = bodies[n].transactions
+      blk.blocks[offset + n].uncles       = bodies[n].uncles
+      blk.blocks[offset + n].withdrawals  = bodies[n].withdrawals
+      blk.blocks[offset + n].requests     = bodies[n].requests
+
+  return offset < blk.blocks.len.uint
```
tersec, line 92: Without the `offset` `uint`, this just works naturally without the conversion

nimbus/sync/flare/worker/blocks_staged.nim
```diff
@@ -0,0 +1,277 @@
...
+    # Nothing to do
+    return false
+
+  let
+    ctx = buddy.ctx
+    peer = buddy.peer
+
+    # Fetch the full range of headers to be completed to blocks
+    iv = ctx.blocksUnprocFetch(
+      ctx.pool.nBodiesBatch.uint).expect "valid interval"
```
tersec, line 148: `blocksUnprocFetch` takes a `uint64`, which only coincidentally lines up with `uint` in 64-bit builds, but probably better to ensure they're explicitly the same.

nimbus/sync/flare/worker/blocks_unproc.nim
```diff
@@ -0,0 +1,115 @@
...
+proc blocksUnprocBottom*(ctx: FlareCtxRef): BlockNumber =
+  let iv = ctx.blk.unprocessed.ge().valueOr:
+    return high(BlockNumber)
+  iv.minPt
+
+
+proc blocksUnprocTotal*(ctx: FlareCtxRef): uint64 =
+  ctx.blk.unprocessed.total()
+
+proc blocksUnprocBorrowed*(ctx: FlareCtxRef): uint =
```
tersec, line 98: Depends/relies on a accidental `uint`/`uint64` conflation as `headersUnprocBorrowed`/`headersUnprocTotal` for same reasons: https://github.com/status-im/nimbus-eth1/pull/2627/files#r1778611383

nimbus/sync/flare/worker.nim
```diff
@@ -133,36 +184,78 @@ proc runMulti*(buddy: FlareBuddyRef) {.async.} =
...
-  # * store this range on the staged queue on the pool
-  if await buddy.stagedCollect info:
-    # * increase the top/right interval of the trused range `[L,F]`
-    # * save updated state and headers
-    discard buddy.ctx.stagedProcess info
-
-  else:
-    when extraTraceMessages:
-      debug info & ": nothing fetched, done", peer
+  let peer = buddy.peer
```
tersec, line 187: The comment block here says
```nim

  ## This peer worker is invoked if the `buddy.ctrl.multiOk` flag is set
  ## `true` which is typically done after finishing `runSingle()`. This
  ## instance can be simultaneously active for all peer workers.
```
but `runSingle()` unconditionally `raiseAssert`s?

PR #2629, opened 2024-09-13
Fix import stuck with era history behind 

advaita-saha:


nimbus/nimbus_import.nim
```diff
@@ -243,6 +243,11 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
         ).valueOr:
           importedSlot += 1
           continue
+        
+        # Checks if the Nimbus block number is ahead the era block number
+        if blockNumber >= blk.header.number:
+          error "Era history is behind client history", stateBlockNumber = blockNumber, eraBlockNumber = blk.header.number
```
tersec, line 249: wrap to 80 characters

nimbus/nimbus_import.nim
```diff
@@ -244,6 +244,12 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
           importedSlot += 1
           continue
 
+        # Checks if the Nimbus block number is ahead the era block number
+        if blockNumber >= blk.header.number:
+          error "Era history is behind client history",
```
arnetheduck, line 249: this is not an error - it's a successful result! basically, import can be restarted at any time and if you run out of files to process, it means the import terminated successfully since all files that exist were imported.

More useful would be to print a last "overview" notice log message that prints various information about the import so that the user can decide what to do next (get more era files, start the client as usual)

arnetheduck: there are also two messages about how it's making progress on finding a slot number - these are interesting for developers _maybe_ but they are certainly not that interesting for the end user - the logs are currently our "UX" so we have to balance what is interesting to developers (and for debugging) with providing the user a good overview of stuff relevant to them - the fact that it finds a pivot point is good to know at debug level maybe, but it certainly does not deserve two high-level logs.

nimbus/nimbus_import.nim
```diff
@@ -228,11 +228,40 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
...
+    debug "Finding slot number to resume import", startSlot, endSlot
+
+    while startSlot < endSlot:
+      let blk = getEthBlockFromEra(
+        era, historical_roots, historical_summaries, startSlot, clConfig.cfg
+      ).valueOr:
+        startSlot += 1
+        if startSlot == endSlot - 1:
+          error "No blocks found in the last era file"
+          quit QuitFailure
```
arnetheduck, line 249: broadly, exiting the program with quit is better avoided because it breaks the basic assumption that a functoin (like import) will return to the caller when it's done what it should.

Quitting should be the decision of the caller in _almost_ all cases (they might want to import and then continue doing something else).

nimbus/nimbus_import.nim
```diff
@@ -228,11 +228,40 @@ proc importBlocks*(conf: NimbusConf, com: CommonRef) =
...
+        else:
+          continue
+
+      startSlot += 1
+      if blk.header.number < blockNumber:
+        notice "Avaiable Era Files are already imported",
```
tersec, line 255:
```suggestion
        notice "Available Era Files are already imported",
```

nimbus/nimbus_import.nim
advaita-saha: Yes but for this particular case, when it iterates over a era file and finds no blocks this means there is something wrong with the era files, and definitely an error
Once thing that can be done instead of `quit` is `return` then the function goes back to it's caller. 

Then would you like all the `quit` to be replaced with `return` ?

arnetheduck: > over a era file and finds no blocks this means there is something wrong with the era files, and definitely an error

Not quite - it simply means we reached the end of available era files - just like when you're syncing the chain and you reach the head block and there are no more blocks.

Since the number of era files keeps growing, there's no particular era file that is the "last" - also, because of the aforementioned possibility to resume the import process, there's no particular correct "beginning" either - we simply run `import` to put everything we have in era files into the database and between runs, the number of available era files might change.

`return` is not quite right, though it's "better" - the function also allows for importing files from RLP dumps - these may come after the era files so probably it should continue with those.

One way to think about this function is to treat the era1/era/rlp files as opaque sources of blocks with the import procedure looking in each to find as many useful blocks as possible to import, which may be zero or lots.

PR #2640, opened 2024-09-19
Fluffy state network now enabled by default and improve status logs

web3-developer:
Changes in this PR:
- State network is now enabled by default
- Status log loop added to beacon and state networks to enable visibility of the respective routing table sizes.
- Status log loop added to portal node for metrics that are global and shared between each subnetwork.

### DIFF
--- a/fluffy/fluffy.nim
+++ b/fluffy/fluffy.nim
@@ -41,7 +41,7 @@ func optionToOpt[T](o: Option[T]): Opt[T] =
   else:
     Opt.none(T)
 
-proc run(config: PortalConf): PortalNode {.raises: [CatchableError].} =
+proc run(config: PortalConf) {.raises: [CatchableError].} =
   setupLogging(config.logLevel, config.logStdout, none(OutFile))
 
   notice "Launching Fluffy", version = fullVersionStr, cmdParams = commandLineParams()
@@ -166,6 +166,7 @@ proc run(config: PortalConf): PortalNode {.raises: [CatchableError].} =
       dataDir: string config.dataDir,
       storageCapacity: config.storageCapacityMB * 1_000_000,
     )
+
     node = PortalNode.new(
       config.network,
       portalNodeConfig,
@@ -251,7 +252,7 @@ proc run(config: PortalConf): PortalNode {.raises: [CatchableError].} =
 
     setupRpcServer(rpcWsServer)
 
-  return node
+  runForever()
 
 when isMainModule:
   {.pop.}
@@ -261,21 +262,6 @@ when isMainModule:
   )
   {.push raises: [].}
 
-  let node =
-    case config.cmd
-    of PortalCmd.noCommand:
-      run(config)
-
-  # Ctrl+C handling
-  proc controlCHandler() {.noconv.} =
-    when defined(windows):
-      # workaround for https://github.com/nim-lang/Nim/issues/4057
-      setupForeignThreadGc()
-
-    notice "Got interrupt, shutting down Fluffy..."
-    node.stop()
-    quit QuitSuccess
-
-  setControlCHook(controlCHandler)
-
-  runForever()
+  case config.cmd
+  of PortalCmd.noCommand:
+    run(config)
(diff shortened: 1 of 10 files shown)


fluffy/fluffy.nim
```diff
@@ -262,6 +261,21 @@ when isMainModule:
...
+      run(config)
+
+  ## Ctrl+C handling
+  proc controlCHandler() {.noconv.} =
+    when defined(windows):
+      # workaround for https://github.com/nim-lang/Nim/issues/4057
+      setupForeignThreadGc()
+
+    notice "Got interrupt, Fluffy shutting down..."
+    node.stop()
```
web3-developer, line 276: Currently this doesn't actually wait for every chronos task to complete because we are using cancelSoon in most places.

fluffy/network/beacon/beacon_light_client_manager.nim
```diff
@@ -320,8 +320,7 @@ proc start*(self: var LightClientManager) =
   doAssert self.loopFuture == nil
   self.loopFuture = self.loop()
 
-proc stop*(self: var LightClientManager) {.async: (raises: []).} =
```
kdeme, line 323: All the stop calls should remain async so that they can be awaited.

Else closing them down with awaiting in the ctrl-c handler will not actually ensure proper clean-up, defeating the point of having a graceful shutdown in the first place.

fluffy/fluffy.nim
```diff
@@ -262,6 +261,21 @@ when isMainModule:
...
+    quit QuitSuccess
+
+  setControlCHook(controlCHandler)
```
kdeme, line 279: I think a raw exception that might be thrown here needs to be catched?
At least that is what is being done in nimbus-eth2: https://github.com/status-im/nimbus-eth2/blob/bf4abf8b9e07c35442b00966f51cc9af5857af33/beacon_chain/nimbus_beacon_node.nim#L2264

fluffy/fluffy.nim
kdeme: Yes, any close call that does not actual wait for closing anything below should be deprecated and the awaited version should be used.

Like this one in discv5: https://github.com/status-im/nim-eth/blob/master/eth/p2p/discoveryv5/protocol.nim#L1120

fluffy/network/beacon/beacon_light_client_manager.nim
web3-developer: Yes, thanks. I'm working on fixing this now. I'm planning to make all the cancel calls use closeAndWait.

fluffy/fluffy.nim
kdeme: So this would be a bit of changes in portal wire and in the different networks. If you want you can split off those changes from this PR change to activate state network.

web3-developer: Thanks for sharing that example. I'll follow that pattern. Yes, it's turning out to be a bigger change than I expected so I'll do it in two PRs.

fluffy/fluffy.nim
web3-developer: Thanks, I'll look into that

web3-developer:
I've reverted the graceful shutdown changes which will now be part of my next PR.

fluffy/network/history/history_network.nim
```diff
@@ -704,17 +704,7 @@ proc processContentLoop(n: HistoryNetwork) {.async: (raises: []).} =
...
-      # drop a lot when using the logbase2 scale, namely `/ 2` per 1 logaritmic
-      # radius drop.
-      # TODO: Get some float precision calculus?
-      let radiusPercentage =
-        n.portalProtocol.dataRadius() div (UInt256.high() div u256(100))
-
       info "History network status",
-        radiusPercentage = radiusPercentage.toString(10) & "%",
-        radius = n.portalProtocol.dataRadius().toHex(),
-        dbSize = $(n.contentDB.size() div 1000) & "kb",
```
web3-developer, line 717: @kdeme Did you have any concerns about moving this log out into Portal node? I thought it would make sense to do so because these fields are based on the contentDb and are shared between all the portal sub networks.

PR #2645, opened 2024-09-20
Implement graceful shutdown in Fluffy

web3-developer:


fluffy/network/beacon/beacon_light_client_manager.nim
```diff
@@ -50,7 +50,7 @@ type
   GetBoolCallback* = proc(): bool {.gcsafe, raises: [].}
   GetSlotCallback* = proc(): Slot {.gcsafe, raises: [].}
 
-  LightClientManager* = object
+  LightClientManager* = ref object
```
web3-developer, line 53: Had to change this to a ref object so that the stop proc can become async

kdeme: And making `LightClientManager` in the stop call a `var` did not help?

fluffy/tests/beacon_network_tests/beacon_test_helpers.nim
```diff
@@ -45,7 +45,7 @@ proc start*(n: BeaconNode) =
   n.beaconNetwork.start()
 
 proc stop*(n: BeaconNode) {.async.} =
-  n.beaconNetwork.stop()
+  discard n.beaconNetwork.stop()
```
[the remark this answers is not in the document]
web3-developer: I tried that but it slows down and even breaks the tests in some cases. Probably because the tests stop and start the network multiple times. Adding the discard effectively allows the tests to behave as before.

fluffy/portal_node.nim
```diff
@@ -41,6 +49,9 @@ type
     stateNetwork*: Opt[StateNetwork]
     beaconLightClient*: Opt[LightClient]
     statusLogLoop: Future[void]
+    metricsServer*: Opt[MetricsHttpServerRef]
+    rpcHttpServer*: Opt[RpcHttpServer]
+    rpcWsServer*: Opt[RpcWebSocketServer]
```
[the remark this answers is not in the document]
web3-developer: I see your point. Yeah I wasn't sure about this part either. Okay, I'll move them out and clean them up separately for now. I'll keep the PortalClient idea for a separate PR.

fluffy/network/beacon/beacon_light_client_manager.nim
web3-developer: I tried that and it didn't work. This was the error: `Error: 'self' is of type <var LightClientManager> which cannot be captured as it would violate memory safety, declared here: /home/user/development/status-im/nimbus-eth1/fluffy/network/beacon/beacon_light_client_manager.nim(323, 12); using '-d:nimNoLentIterators' helps in some cases. Consider using a <ref var LightClientManager> which can be captured.`

Issue #2157, opened 2024-04-30
Fluffy shouldn't allow running multiple instances from the same data directory


web3-developer:
Fluffy doesn't fail or show any error message when starting a second Fluffy instance pointing to the same data directory. The `--data-dir` parameter is also not a required parameter and will use the default directory if not specified therefore it is rather easy for a user to not supply it and mistakenly run multiple instances pointing to the same location.

We should create a lock file on startup in the data directory so that if a second instance is started pointing to the same location it can read the lock file and shutdown with a error message to let the user know the root cause.

kdeme:
This got implemented in https://github.com/status-im/nimbus-eth1/pull/2647

PR #2647, opened 2024-09-23
Fix multiple instances running from same dataDir

bhaskar1001101:
Minimum working fix for #2157 

Tested locally with `./build/fluffy --rpc` on two instances.

* Do we want to write some specific information to the lock file?
* Do we want to implement OS level locks with `io2.lockFile` to prevent other processes from writing to the lock file?

### DIFF
--- a/fluffy/fluffy.nim
+++ b/fluffy/fluffy.nim
@@ -8,7 +8,7 @@
 {.push raises: [].}
 
 import
-  std/[os, enumutils],
+  std/[os, enumutils, exitprocs],
   confutils,
   confutils/std/net,
   chronicles,
@@ -61,6 +61,23 @@ proc run(
     fatal "Failed to create data directory",
       dataDir = config.dataDir, error = pathExists.error
     quit 1
+
+  # Make sure multiple instances to the same dataDir do not exist
+  let lockFilePath = config.dataDir.string / "fluffy.lock"
+  if isFile(lockFilePath):
+    fatal "Please ensure no other fluffy instances are running with the same data directory",
+     dataDir = config.dataDir
+    quit 1
+
+  if io2.writeFile(lockFilePath, "Fluffy is running").isErr():
+    fatal "Failed to create lock file", lock_file = lockFilePath
+    quit 1
+
+  addExitProc(proc() =
+    discard io2.removeFile(lockFilePath)
+  )
+
+  info "New lock file was created", lock_file = lockFilePath
 
   ## Network configuration
   let
--- a/fluffy/fluffy.nim
+++ b/fluffy/fluffy.nim
@@ -60,7 +60,7 @@ proc run(
   if pathExists.isErr():
     fatal "Failed to create data directory",
       dataDir = config.dataDir, error = pathExists.error
-    quit 1
+    quit QuitFailure
 
   # Make sure multiple instances to the same dataDir do not exist
   let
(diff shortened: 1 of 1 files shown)


kdeme (approved):
> * Do we want to write some specific information to the lock file?

I don't think we need any extra information for this functionality. What you write now is fine I think.

> 
> * Do we want to implement OS level locks with `io2.lockFile` to prevent other processes from writing to the lock file?

I'm fine with merging this version of the PR. It's better than having no lock file.

If you want, I'm also fine with adjusting to using the specific `lockfile` call (whether in this PR or an additional one). 
We currently use it already here if you want an example: https://github.com/status-im/nimbus-eth2/blob/bf4abf8b9e07c35442b00966f51cc9af5857af33/beacon_chain/filepath.nim#L27
But that is for more sensitive data (keystore management).

fluffy/fluffy.nim
```diff
@@ -59,6 +59,23 @@ proc run(
       dataDir = config.dataDir, error = pathExists.error
     quit 1
 
+  # Make sure multiple instances to the same dataDir do not exist
+  let lockFilePath = config.dataDir.string / "fluffy.lock"
+  if isFile(lockFilePath):
```
arnetheduck, line 64: this will cause the process to fail to restart if it crashes (and the file is not removed) which is poor behavior. the normal way set up create lock files is via `flock`: https://stackoverflow.com/questions/2964391/preventing-multiple-process-instances-on-linux - this works even if the application crashes (and doesn't remove the file) - this obviously needs to be translated to nim.

kdeme: Yes, that's the better version, where current code gets replaced with the io2.lockFile call, basically using calls like `openLockedFile` and `closeLockedFile` from the location I linked above, or similar. No?

bhaskar1001101: Added an OS managed exclusive lock.

fluffy/fluffy.nim
```diff
@@ -59,6 +59,28 @@ proc run(
...
     quit 1
 
+  # Make sure multiple instances to the same dataDir do not exist
+  let
+    lockFilePath = config.dataDir.string / "fluffy.lock"
+    lockFlags = {OpenFlags.Create, OpenFlags.Read, OpenFlags.Write}
+    lockFileHandleResult = openFile(lockFilePath, lockFlags)
+
+  if lockFileHandleResult.isErr():
+    error "Failed to open lock file", error = ioErrorMsg(lockFileHandleResult.error)
```
kdeme, line 69: Considering we are quitting here, this log message should also be a `fatal` log message.

fluffy/fluffy.nim
```diff
@@ -59,6 +59,28 @@ proc run(
...
+  let lockFileHandle = lockFileHandleResult.get()
+  if lockFile(lockFileHandle, LockType.Exclusive).isErr():
+    fatal "Please ensure no other fluffy instances are running with the same data directory",
+      dataDir = config.dataDir
+    quit 1
+
+  addExitProc(
+    proc() =
+      discard closeFile(lockFileHandle)
+      discard unlockFile(lockFileHandle, 0'i64, 0'i64)
```
kdeme, line 81: The order of these two calls needs to be reversed.

I'm pretty sure that `unlockFile` here will actually fail as the file is already closed, but you wouldn't notice because of the `discard`. You wouldn't actually notice this as an issue because I assume when the program terminates the locking gets removed anyhow, but it is cleaner to fix this order of operations.

kdeme: Also, you can use the `unlockFile` proc that takes only the lockFileHandle as parameters: https://github.com/status-im/nim-stew/blob/master/stew/io2.nim#L1675, but you will have to pass along the `result.value` of the `lockFile` call.

kdeme (changes requested):

fluffy/fluffy.nim
```diff
@@ -59,6 +59,28 @@ proc run(
...
+
+  if lockFileHandleResult.isErr():
+    error "Failed to open lock file", error = ioErrorMsg(lockFileHandleResult.error)
+    quit 1
+
+  let lockFileHandle = lockFileHandleResult.get()
+  if lockFile(lockFileHandle, LockType.Exclusive).isErr():
+    fatal "Please ensure no other fluffy instances are running with the same data directory",
+      dataDir = config.dataDir
+    quit 1
```
kdeme, line 76: nitpick but it's cleaner to use `QuitFailure`

```suggestion
    quit QuitFailure
```

fluffy/fluffy.nim
bhaskar1001101: Made the needed changes.

fluffy/fluffy.nim
bhaskar1001101: Done

kdeme (approved):

PR #2649, opened 2024-09-24
Fluffy Portal RPC Client

bhartnett:
This PortalRpcClient type can be used by Nimbus to fetch the block headers, bodies and receipts from the portal history network via the portal_history JSON-RPC endpoints. It provides convenience functions that encode/decode to/from the desired types. 

The nim-eth/common types are returned in the proc interfaces. This code intentionally doesn't validate the data returned because it is assumed that the portal node is trusted, and the portal node is already validating the content as it does normally when fetching data from the network.

Nimbus will accept a JSON-RPC endpoint on the command line, then create and connect an RPC client and use it to init the Portal RPC Client. Nimbus can then fetch headers, bodies and receipts via portal on demand.

Questions for discussion:
1. Is this a good location for this code for now? Or should it go somewhere else? If we are moving towards making this a library then perhaps it could go in a top level folder in the nimbus-eth1 project?
2. Should we implement retries at this layer or leave that to Nimbus?
3. How should we handle disconnects? For example if the rpc server goes down, then the rpc client will disconnect in the case of websockets but the RpcClient type has no methods to check if the connection is active or to reconnect. This likely needs to be fixed upstream in the nim-json-rpc library.
4. Side note: I found that the WebSocketRpcClient appears to only handle one concurrent request at a time per connection. Not sure if this is a limitation of websockets or if it can be fixed in nim-json-rpc.


kdeme (commented):
> . This code intentionally doesn't validate the data returned because it is assumed that the portal node is trusted, and the portal node is already validating the content as it does normally when fetching data from the network.

The Portal node is actually not validating portal rpc requests like `FindContent` or `RecursiveFindContent`. This has been a point of discussion before (and in our case would probably mean properly separating that portal rpc code per network).

Right now, you can however use the higher level calls in `history_network` code, like `getVerifiedBlockHeader` and `getBlockBody`, etc.

web3-developer:
> > . This code intentionally doesn't validate the data returned because it is assumed that the portal node is trusted, and the portal node is already validating the content as it does normally when fetching data from the network.
> 
> The Portal node is actually not validating portal rpc requests like `FindContent` or `RecursiveFindContent`. This has been a point of discussion before (and in our case would probably mean properly separating that portal rpc code per network).
> 
> Right now, you can however use the higher level calls in `history_network` code, like `getVerifiedBlockHeader` and `getBlockBody`, etc.

I missed this, thanks. I'll do the verification in that case re-using the history network code where possible.

fluffy/rpc/portal_rpc_client.nim
```diff
@@ -6,7 +6,144 @@
...
+  let
+    contentKeyBytes = blockHeaderContentKey(blockHash).encode().asSeq()
+    contentBytes = client.historyGetContent(contentKeyBytes)
+    headerWithProof = ?decodeSsz(contentBytes, BlockHeaderWithProof)
+
+  # TODO: How can we verify that the header is part of the canonical chain
+  # over JSON-RPC?
+  # The verifyHeader(n.accumulator, header, proof) call requires access to
+  # the accumulator which is not available here on the client side.
+  validateBlockHeaderBytes(headerWithProof.header.asSeq(), blockHash)
```
web3-developer, line 89: @kdeme Is there any way to verify that the header is canonical from the client side? In this case the client (Nimbus-Eth1) won't have history network running locally so it won't have access to the accumulator.

fluffy/rpc/portal_rpc_client.nim
```diff
@@ -6,7 +6,144 @@
...
+): Future[Result[BlockHeader, string]] {.async: (raises: []).} =
+  ## Fetches the block header for the given hash from the Portal History Network.
+  ## The data is first looked up in the node's local database before trying to
+  ## fetch it from the network.
+  ##
+  ## Note: This does not validate that the returned header is part of the canonical
+  ## chain, it only validates that the header matches the block hash. For example,
+  ## a malicious portal node could return a valid but non-canonical header such
+  ## as an uncle block that matches the block hash. For this reason the caller
+  ## needs to use another method to verify the header is part of the canonical chain.
```
web3-developer, line 78: @tersec The current design/spec of the portal JSON-RPC APIs don't do validation of the data on the RPC server side (inside Fluffy). For this reason, I need to do the validation on the client side but unfortunately, I can't easily verify if a requested header is canonical before returning it. So my question is: Will Nimbus need to rely on Fluffy only returning canonical blocks? Said another way, will Nimbus need to lookup blocks by number?

fluffy/rpc/portal_rpc_client.nim
web3-developer: Our options:
- Either we don't support this for now.
- Load the accumulator into Nimbus-Eth1 using the accumulator file.
- Update the portal JSON-RPC APIs to do full validation on the server side. This would make the API's more usable for usage in execution clients such as Nimbus.

kdeme: > * Update the portal JSON-RPC APIs to do full validation on the server side. This would make the API's more usable for usage in execution clients such as Nimbus.

Yes, I've been considering this also after looking at this PR.

Whether to do this or not is not explicitly specified. I will check with the other clients if they are doing this or not as I think it is a good idea as this is become the user face API for Portal. 

(It used to be only used for testing initially which at the start made it easier to use if no validation was done. But now, I think we should instead have `debug` api if we still need unverified calls for some reason).

fluffy/rpc/portal_rpc_client.nim
kdeme: For lookup by hash, the canonical validation is not really required, as it is implied that the client requesting it knows this (as it is providing the hash).

For lookup by number the case is different I would say, and it would still be useful to have the verification done (whether server side API call or as extra step in Nimbus, to be decided).

web3-developer: I removed the block number lookup for this reason as the verification is very weak. When looking up by hash we can verify the returned block header against the hash but there is a edge case where a malicious portal node could send a uncle header which isn't actually part of the portal network history but I see your point, the user of the API (Nimbus) will most likely know the context. If it has the block hash it probably knows if it is canonical.

fluffy/rpc/portal_rpc_client.nim
web3-developer: Thanks, I'm in favor of doing the validation in Fluffy in most case and make the behavior more intuitive for users of the API.

arnetheduck:
> This code intentionally doesn't validate the data returned 

it doesn't have to - ie this is effectively a replacement for the eth/68 protocol for fetching blocks so nimbus will be doing validation.

kdeme:
> it doesn't have to - ie this is effectively a replacement for the eth/68 protocol for fetching blocks so nimbus will be doing validation.

I don't think this is the right or the only way to look at it.

While it could be a replacement for fetching blocks when for example doing a "full sync" and in which case it would do it's own validation, afterwards it is mostly meant as a replacement for the current storage of blocks in local db. This latter case typically requires content to be verified (it gets verified before dropping in the db).

web3-developer:
> > This code intentionally doesn't validate the data returned
> 
> it doesn't have to - ie this is effectively a replacement for the eth/68 protocol for fetching blocks so nimbus will be doing validation.

I see what you mean, thanks. In that case, I'll make the validation optional so it can be disabled when required.

kdeme:
> as far as I can tell it doesn't store the content in the db after fetching.

This is not what I meant.

What I mean, is that the action of retrieving a block from portal replaces the action of retrieving a block from the local nimbus database.

The data in the local nimbus database is verified, thus the data it retrieves from portal also needs to be verified for that use case.

If that is not a use case we want to apply in nimbus, well then sure, verification is not required.

bhartnett:
> > as far as I can tell it doesn't store the content in the db after fetching.
> 
> This is not what I meant.
> 
> What I mean, is that the action of retrieving a block from portal replaces the action of retrieving a block from the local nimbus database.
> 
> The data in the local nimbus database is verified, thus the data it retrieves from portal also needs to be verified for that use case.
> 
> If that is not a use case we want to apply in nimbus, well then sure, verification is not required.

Sure, I see what you mean. It sounds like we should support both with and without verification depending on the use case.

PR #2651, opened 2024-09-24
FCU should consider ForkedChainRef when calculating valid ancestor

### LINKED ISSUE #2588: Impossible to produce blocks with `forkchoiceUpdated`
`master` and https://github.com/status-im/nimbus-eth1/pull/2423 have the same issue

If it already knows about the block:
https://github.com/status-im/nimbus-eth1/blob/0723464928517741bf75440d3bdc6c1644849703/nimbus/beacon/api_handler/api_forkchoice.nim#L154-L163

then it returns and never processes the payload attributes to trigger payload generation later on in that function:
https://github.com/status-im/nimbus-eth1/blob/0723464928517741bf75440d3bdc6c1644849703/nimbus/beacon/api_handler/api_forkchoice.nim#L189-L210

### LINKED ISSUE #2641: FCU ignored after payload is processed
When a payload is sent to eth1, it processes and validates fine
but then `fcu` is called, and says it is `updating to the old head` and thus ignoring the request

Shouldn't behave like that, but it is because of the payload processing
In the `new_payload`, we are also persisting the header to the db 
https://github.com/status-im/nimbus-eth1/blob/ea74e035ab019619fc00b46ab97496b502a188dd/nimbus/core/chain/forked_chain.nim#L88-L95

Then when the fcu call takes place, it checks if the header is there in the db, and it is thus ignoring the fcu call 
https://github.com/status-im/nimbus-eth1/blob/ea74e035ab019619fc00b46ab97496b502a188dd/nimbus/beacon/api_handler/api_forkchoice.nim#L158-L163

A similar issue arising from the same persistence of header #2588

jangko:
This PR should fix #2641 and fix #2588.

Test will come later because it's a bit complicated to create test for engine API.

### DIFF
--- a/nimbus/beacon/api_handler/api_forkchoice.nim
+++ b/nimbus/beacon/api_handler/api_forkchoice.nim
@@ -155,8 +155,7 @@ proc forkchoiceUpdated*(ben: BeaconEngineRef,
   # probably resyncing. Ignore the update.
   # See point 2 of fCUV1 specification
   # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/paris.md#specification-1
-  var canonHash: common.Hash256
-  if db.getBlockHash(header.number, canonHash) and canonHash == blockHash:
+  if ben.chain.isCanonicalAncestor(header.number, blockHash):
     notice "Ignoring beacon update to old head",
       blockHash=blockHash.short,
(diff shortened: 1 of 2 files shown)


tersec:
Sure, there are as you noted elsewhere inefficiencies in this, but it's basically fine and important for ongoing testing

PR #2657, opened 2024-09-25
Fluffy: Configure RPC APIs via CLI parameters

web3-developer:


kdeme (approved):

web3-developer:
Here is the hive PR to update the Fluffy startup script so that the discovery API is still available in the portal hive tests: https://github.com/ethereum/hive/pull/1168

PR #2664, opened 2024-09-27
Backend agnostic read-only DB wrapper framework

mjfh:
Proof-of-concept, up for discussion.

Code might raise to a wish-list what is really needed, useful, etc.

### DIFF
--- a/nimbus/db/edge_db/db_desc.nim
+++ b/nimbus/db/edge_db/db_desc.nim
@@ -1,0 +1,119 @@
+# Nimbus
+# Copyright (c) 2023-2024 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or
+#    http://www.apache.org/licenses/LICENSE-2.0)
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT) or
+#    http://opensource.org/licenses/MIT)
+# at your option. This file may not be copied, modified, or distributed except
+# according to those terms.
+
+{.push raises: [].}
+
+import
+  pkg/eth/common,
+  pkg/results
+
+type
+  EdgeDbError* = enum
+    ## Allows more granulated failure information.
+    NothingSerious = 0
+    EdgeKeyNotFound
+    EdgeColUnsupported
+    EdgeKeyTypeUnsupported
+
+  EdgeDbColumn* = enum
+    ## Specify object type to query for
+    Oops = 0
+    EthBlockData
+    EthHeaderData
+    EthBodyData
+
+  EdgeDbGetRef* = ref object of RootObj
+    ## Descriptor common to a set of `getFn()` implementations. This basic type
+    ## will be interited by sprcific implementations.
+
+
+  EdgeDbUintGetFn* =
+    proc(dsc: EdgeDbGetRef;
+         col: EdgeDbColumn;
+         key: uint64;
+        ): Result[Blob,EdgeDbError]
+        {.gcsafe, raises: [].}
+      ## Particular `getFn()` instance. Will return a RLP encoded serialised
+      ## data result.
+
+  EdgeDbBlobGetFn* =
+    proc(dsc: EdgeDbGetRef;
+         col: EdgeDbColumn;
+         key: openArray[byte];
+        ): Result[Blob,EdgeDbError]
+        {.gcsafe, raises: [].}
+      ## Ditto for `Blob` like key
+
+
+  EdgeDbUintGetPolicyFn* =
+    proc(edg: EdgeDbRef;
+         col: EdgeDbColumn;
+         key: uint64;
+        ): Result[Blob,EdgeDbError]
+        {.gcsafe, raises: [].}
+      ## Implemenation of `get()`. This function employs a set of `getFn()`
+      ## instances (in some order) for finding a result.
+
+  EdgeDbBlobGetPolicyFn* =
+    proc(edg: EdgeDbRef;
+         col: EdgeDbColumn;
+         key: openArray[byte];
+        ): Result[Blob,EdgeDbError]
+        {.gcsafe, raises: [].}
+      ##  Ditto for `Blob` like key
+
+
+  EdgeDbRef* = ref object
+    ## Visible database wrapper.
+    getDesc*: EdgeDbGetRef
+    uintGetFns*: seq[EdgeDbUintGetFn]
+    blobGetFns*: seq[EdgeDbBlobGetFn]
+    uintGetPolFn*: EdgeDbUintGetPolicyFn
+    blobGetPolFn*: EdgeDbBlobGetPolicyFn
+
+# ------------------------------------------------------------------------------
+# Public helpers, sequentially trying a list of `getFn()` instances
+# ------------------------------------------------------------------------------
+
+proc uintGetSeqentiallyUntilFound*(
+  edg: EdgeDbRef;
+  col: EdgeDbColumn;
+  key: uint64;
+    ): Result[Blob,EdgeDbError] =
+  ## Simple linear get policy.
+  var error = EdgeColUnsupported
+
+  for fn in edg.uintGetFns:
+    let err = edg.getDesc.fn(col,key).errorOr:
+      return ok(value)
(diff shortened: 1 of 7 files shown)


tests/test_edgedb/test_helpers.nim
```diff
@@ -0,0 +1,145 @@
...
+              if path.fileExists:
+                return ok(path)
+  echo "*** File not found \"", file, "\"."
+  err()
+
+# -----------------
+
+proc posixPrngRand(state: var uint32): byte =
+  ## POSIX.1-2001 example of a rand() implementation, see manual page rand(3).
+  state = state * 1103515245 + 12345;
```
tersec, line 54: It is useful to have an intentionally reproducible PRNG, but this particular one isn't very good, and conveniently due to exactly this inclusion in POSIX standards has been widely studied to the point of being an example of what not to do in [undergraduate CS lectures](https://inst.eecs.berkeley.edu/~cs161/fa08/Notes/random.pdf):
> The output of this generator isn’t very random. If we use the sample generator listed above, then the sequence of 16 key bytes will be highly non-random. For instance, it turns out that the low bit of each successive output of rand() will alternate (e.g., 0, 1, 0, 1, 0, 1, . . . ). Do you see why? The low bit of x * 1103515245 is the same as the low bit of x, and then adding 12345 just flips the low bit. Thus the low bit alternates.

It's true though that LCGs are simple and easy, and that's useful here too. For a while Mersenne Prime was typical, but that's a bit more involved.

[Xorshift](https://en.wikipedia.org/wiki/Xorshift#Example_implementation) might be one approach, more popular in the last few years. https://prng.di.unimi.it/ suggests it's fairly good quality as well.

https://nim-lang.org/docs/random.html uses this, even, and maybe using `initRand` with specific values and `std/random` suffices for this purpose.

tests/test_edgedb/test_helpers.nim
```diff
@@ -0,0 +1,145 @@
...
+    if w < top:
+      return w
+  raiseAssert "Not here (!)"
+
+# -----------------
+
+proc lastBlockNumber(e1db: Era1DbRef): BlockNumber =
+  var
+    minNum = BlockNumber(1)
+    maxNum = BlockNumber(4700013) # MainNet
```
tersec, line 77: This seems a bit specific? it might be enough practically to have this `maxNum` be efectively the max of mainnet/Holesky/Sepolia, since there presumably will never again be a network with era1 files since they'll launch merged, but then that should be noted, rather than treating the tests as only applicable to mainnnet (or if they are mainnet-specific by necessity, they probably shouldn't be framed as a general-purpose test utility).

tests/test_edgedb/test_helpers.nim
```diff
@@ -0,0 +1,145 @@
...
+    middle = (maxNum + minNum) div 2
```
tersec, line 78: Well, (in)famously, this was one of the longstanding binary search bugs because it can overflow, even if the relevant block numbers are unlikely to overflow a `BlockNumber = uint64`.

Practically, it probably can't, but

tests/test_edgedb/test_helpers.nim
```diff
@@ -0,0 +1,145 @@
...
+  for bn in first .. last:
+    result.add e1db.getEthBlock(bn).expect "valid eth block"
+
+# -------------------
+
+proc newCoreDbInstance*(): CoreDbRef =
+  AristoDbMemory.newCoreDbRef()
+
+proc newCommonInstance*(): CommonRef =
+  const networkId = MainNet
```
tersec, line 137: This is another case of, this is probably a bit too specific for a "backend agnostic", or it should at least be better-documented/enforced that all of this applies to mainnet only.

But there's a practical tradeoff here of, something like a Sepolia or Holesky era1 test can be done to completion much more quickly than mainnet, so constraining this to mainnet constraints tests to being only on small slices.

tests/test_edgedb/test_helpers.nim
```diff
@@ -0,0 +1,145 @@
...
+    middle = (maxNum + minNum) div 2
+    delta = maxNum - minNum
+  minNum
+
+# ------------------------------------------------------------------------------
+# Public functions
+# ------------------------------------------------------------------------------
+
+proc setTraceLevel* =
+  discard
```
tersec, line 94: None of these `setFooLevel` should need the `discard` statements, even when the `when` condition is false. These aren't `template`s, for which that would be necessary.

tests/test_edgedb/test_helpers.nim
```diff
@@ -0,0 +1,145 @@
...
+# Private helpers
+# ------------------------------------------------------------------------------
+
+proc findFilePathHelper(
+    file: string;
+    baseDir: openArray[string] = baseDir;
+    mainDir: openArray[string] = mainDir;
+    subDir: openArray[string] = subDir;
+      ): Result[string,void] =
+  for dir in baseDir:
```
tersec, line 38: Why this particular search pattern, apparently including searching the root directory (`$DirSep`)?

Also there's an odd mix here between this very generic search pattern and otherwise specialized to mainnet.

tests/test_edgedb.nim
```diff
@@ -0,0 +1,22 @@
...
+
+import
+  ./test_edgedb/[
+    test_era1_coredb,
+  ]
+
+proc edgeDbMain*() =
+  testEra1CoreDbMain()
+
+when isMainModule:
```
tersec, line 19: When imported from `all_tests` elsewhere, this `when` ensures it won't run any tests.

nimbus/db/edge_db/db_desc.nim
```diff
@@ -0,0 +1,119 @@
...
+# Public helpers, sequentially trying a list of `getFn()` instances
+# ------------------------------------------------------------------------------
+
+proc uintGetSeqentiallyUntilFound*(
+  edg: EdgeDbRef;
+  col: EdgeDbColumn;
+  key: uint64;
+    ): Result[Blob,EdgeDbError] =
+  ## Simple linear get policy.
+  var error = EdgeColUnsupported
```
tersec, line 91: The `error` symbol has a couple of other typical uses:
- `chronicles` logging `error`
- `results` magical variable `error`

Those already often conflict, e.g., https://github.com/arnetheduck/nim-results/issues/34 which triggered the workarounds in https://github.com/arnetheduck/nim-results/pull/35 and while those two are somewhat baked in, less chance of random future blowups to avoid reusing the `error` symbol beyond that unless necessary.

nimbus/db/edge_db/db_desc.nim
```diff
@@ -0,0 +1,119 @@
...
+
+  err(error)
+
+proc blobGetSeqentiallyUntilFound*(
+  edg: EdgeDbRef;
+  col: EdgeDbColumn;
+  key: openArray[byte];
+    ): Result[Blob,EdgeDbError] =
+  ## Ditto for `Blob` like key
+  var error = EdgeColUnsupported
```
tersec, line 107: `error` symbol overloading, see https://github.com/status-im/nimbus-eth1/pull/2664/files#r1779353613

tests/test_edgedb/test_era1_coredb.nim
```diff
@@ -0,0 +1,89 @@
...
+          blkData = edb.get(EthBlockData, bn).expect "valid block data"
+          hdrData = edb.get(EthHeaderData, bn).expect "valid header data"
+          bdyData = edb.get(EthBodyData, bn).expect "valid body data"
+          e1Blk = e1db.getEthBlock(bn).expect "valid eth block"
+          e1Tpl = e1db.getBlockTuple(bn).expect "valid block tuple"
+        check blkData == rlp.encode(e1Blk)
+        check hdrData == rlp.encode(e1Tpl.header)
+        check bdyData == rlp.encode(e1Tpl.body)
+
+when isMainModule:
```
tersec, line 84: Is this meant to be run standalone or imported? Nothing else in this PR references this module.

arnetheduck:
What is this wrapper for? ie for the moment, we should not need any additional wrappers or interfaces (in fact, we have too many already) - we should not be abstracting or interfacing anything prior to there being a non-abstract version working as intended.

mjfh:
Mothballed up until the need for such a wrapper will pop up ..

PR #2674, opened 2024-10-01
update web3 types to common eth

arnetheduck:


tersec:
Rebase after https://github.com/status-im/nimbus-eth1/pull/2678 might help Kurtosis

PR #2675, opened 2024-10-01
avoid threadvar in nim-serialization

arnetheduck:


tersec:
Rebase after https://github.com/status-im/nimbus-eth1/pull/2678 might help

PR #2681, opened 2024-10-02
rpc fixes and enable kurtosis

### LINKED ISSUE #2679: `eth.getTransactionReceipt` not registered rpc method
Here for all the cases, `addHttpService` and `addEngineApiServices` is adding the same serverAPI endpoints 
https://github.com/status-im/nimbus-eth1/blob/e19790614b1aba129bfc53bf12202a24956efd08/nimbus/rpc.nim#L267-L302

Thus resulting the eth endpoints being available mentioned in #2677 , but not all of them
For example `eth.getTransactionReceipt()` when called 
```bash
> eth.getTransactionReceipt("")
Error: 'eth_getTransactionReceipt' is not a registered RPC method
	at web3.js:6387:9(39)
	at send (web3.js:5116:62(29))
	at <eval>:1:26(3)
```

advaita-saha:
Once the blocks get's persisted to the db, the rpc can't serve it through the `ForkedChainRef`
Resulting in older blocks returning `null`

Screenshot_2024-10-02_at_3 26 09_PM

Also fixes #2679

### DIFF
--- a/nimbus/core/chain/forked_chain.nim
+++ b/nimbus/core/chain/forked_chain.nim
@@ -28,9 +28,9 @@ type
     forkJunction: BlockNumber
     hash: Hash256
 
-  BlockDesc = object
-    blk: EthBlock
-    receipts: seq[Receipt]
+  BlockDesc* = object
+    blk*: EthBlock
+    receipts*: seq[Receipt]
 
   BaseDesc = object
     hash: Hash256
@@ -582,9 +582,6 @@ func txRecords*(c: ForkedChainRef): Table[Hash256, Hash256] =
 
 func memoryBlocks*(c: ForkedChainRef): Table[Hash256, BlockDesc] =
   c.blocks
-
-func receipts*(b: BlockDesc): seq[Receipt] =
-  b.receipts
 
 func blockTransactions*(b: BlockDesc): seq[Transaction] =
   b.blk.transactions
--- a/nimbus/core/chain/forked_chain.nim
+++ b/nimbus/core/chain/forked_chain.nim
@@ -633,7 +633,19 @@ proc blockByHash*(c: ForkedChainRef, blockHash: Hash256): Opt[EthBlock] =
   do:
     return Opt.none(EthBlock)
 
-func blockByNumber*(c: ForkedChainRef, number: BlockNumber): Result[EthBlock, string] =
+proc blockByNumber*(c: ForkedChainRef, number: BlockNumber): Result[EthBlock, string] =
+  if number > c.cursorHeader.number:
+    return err("Requested block number not exists: " & $number)
+
+  if number < c.baseHeader.number:
+    var 
+      header: Header
+      body: BlockBody
+    if c.db.getBlockHeader(number, header) and c.db.getBlockBody(header, body):
+      return ok(EthBlock.init(move(header), move(body)))
+    else:
+      return err("Failed to get block with number: " & $number)
+
   shouldNotKeyError:
     var prevHash = c.cursorHash
     while prevHash != c.baseHash:
--- a/nimbus/core/chain/forked_chain.nim
+++ b/nimbus/core/chain/forked_chain.nim
@@ -45,7 +45,7 @@ type
     db: CoreDbRef
     com: CommonRef
     blocks: Table[Hash256, BlockDesc]
-    txRecords: Table[Hash256, Hash256]
+    txRecords: Table[Hash256, (Hash256, uint64)]
     baseHash: Hash256
     baseHeader: BlockHeader
     cursorHash: Hash256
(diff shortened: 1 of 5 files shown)


nimbus/rpc/server_api.nim
```diff
@@ -50,7 +50,12 @@ proc headerFromTag(api: ServerAPIRef, blockTag: BlockTag): Result[common.BlockHe
       return err("Unsupported block tag " & tag)
   else:
     let blockNum = common.BlockNumber blockTag.number
-    return api.chain.headerByNumber(blockNum)
+    let blkhdr = api.chain.headerByNumber(blockNum).valueOr:
+      try:
+        api.chain.db.getBlockHeader(blockNum)
```
jangko, line 55: `getBlockHeader` already in `headerByNumber` https://github.com/status-im/nimbus-eth1/blob/e19790614b1aba129bfc53bf12202a24956efd08/nimbus/core/chain/forked_chain.nim#L601

tersec:
This seems ok as a stop-gap, to get things running. But already needing two separate implementations of this, and probably more over time, each of which has to duplicate the logic, seems suboptimal.

nimbus/rpc/server_api.nim
```diff
@@ -74,7 +79,12 @@ proc blockFromTag(api: ServerAPIRef, blockTag: BlockTag): Result[EthBlock, strin
       return err("Unsupported block tag " & tag)
   else:
     let blockNum = common.BlockNumber blockTag.number
-    return api.chain.blockByNumber(blockNum)
+    let blk = api.chain.blockByNumber(blockNum).valueOr:
+      try:
+        api.chain.db.getEthBlock(blockNum)
```
jangko, line 84: perhaps move this into `ForkedChainRef.blockByNumber`?

advaita-saha: Yeah makes more sense

advaita-saha:
With this PR
All kurtosis tests pass. in a network of nimbus-eth1 and nimbus-eth2
TxPool works only when transactions are very low
All op code tests pass and chain stability tests pass

After the merge of this PR, kurtosis CI will help up not to regress from this point

nimbus/rpc/server_api.nim
```diff
@@ -258,3 +259,40 @@ proc setupServerAPI*(api: ServerAPIRef, server: RpcServer) =
...
+    ##
+    ## data: Hash of a transaction.
+    ## Returns ReceiptObject or nil when no receipt was found.
+    var
+      idx = 0'u64
+      prevGasUsed = GasInt(0)
+    
+    let txHash = data.ethHash()
+    if api.chain.txRecords.hasKey(txHash):
+      let blockhash = api.chain.txRecords[txHash]
```
jangko, line 274: Probably we can put both block hash and tx index into `txRecords` to optimize this a bit. `rlpHash` is a bit expensive. integer compare is much faster.

nimbus/rpc/server_api.nim
```diff
@@ -258,3 +259,40 @@ proc setupServerAPI*(api: ServerAPIRef, server: RpcServer) =
...
+      let blkdesc = api.chain.memoryBlocks[blockhash]
+      
+      for i, receipt in blkdesc.receipts:
+        let gasUsed = receipt.cumulativeGasUsed - prevGasUsed
+        prevGasUsed = receipt.cumulativeGasUsed
+        if txHash == blkdesc.blockTransactions[i].rlpHash:
```
jangko, line 280: and then we can compare i with tx index here

nimbus/core/chain/forked_chain.nim
```diff
@@ -570,6 +577,15 @@ func latestHash*(c: ForkedChainRef): Hash256 =
 func baseNumber*(c: ForkedChainRef): BlockNumber =
   c.baseHeader.number
 
+func txRecords*(c: ForkedChainRef): Table[Hash256, Hash256] =
```
jangko, line 580: hmm, looks like returning a table is expensive. Why not return only what we need? e.g. transactions or receipts?

nimbus/core/chain/forked_chain.nim
```diff
@@ -570,6 +577,15 @@ func latestHash*(c: ForkedChainRef): Hash256 =
...
+  c.txRecords
+
+func memoryBlocks*(c: ForkedChainRef): Table[Hash256, BlockDesc] =
```
jangko, line 583: ditto

nimbus/core/chain/forked_chain.nim
```diff
@@ -570,6 +577,15 @@ func latestHash*(c: ForkedChainRef): Hash256 =
...
+  c.blocks
+
+func blockTransactions*(b: BlockDesc): seq[Transaction] =
```
jangko, line 586: ditto

jangko (approved):

PR #2688, opened 2024-10-03
t8n: use common transaction reader

arnetheduck:


tools/t8n/helpers.nim
```diff
@@ -306,10 +288,11 @@ proc parseTxs*(ctx: TransContext, chainId: ChainId): seq[Result[Transaction, str
     result = newSeqOfCap[Result[Transaction, string]](ctx.txs.r.listLen)
     var rlp = ctx.txs.r
     for item in rlp:
-      if item.isList:
-        result.add parseTxLegacy(item)
-      else:
-        result.add parseTxTyped(item)
+      result.add try:
```
jangko, line 291: The above code will not works,

This snippet from nim-eth will works:
```Nim
for item in rlp:
    var tx: Transaction
    if item.isList:
      item.readTxLegacy(tx)
    else:
      var rr = rlpFromBytes(rlp.read(seq[byte])) # here is the problem
      rr.readTxTyped(tx)
    result.add tx
```

arnetheduck: ok - is there a case where the transactions have invalid rlp? it looks like the code is written to ignore such transactions

jangko: not ignoring, but it will raise RlpError exception

arnetheduck: also .. it seems like the code should just call `rlp.read(seq[Transaction])` in that case

arnetheduck: ok, there's a different code path that rejects each malformed transaction: https://github.com/status-im/nimbus-eth1/blob/master/tools/t8n/transition.nim#L251

arnetheduck: so anyway, `item.read(Transaction)` _is_ the snippet from nim-eth - that's what it does

arnetheduck: ah, no, it isn't . what a mess.

arnetheduck: ok, pushed a version that maybe does the right thing, without using those "private" functions

jangko: it works now

jangko (approved):

PR #2698, opened 2024-10-04
Fix eth/common & web3 related deprecation warnings for fluffy

kdeme:
This commit uses the new types in the new eth/common/ structure to remove deprecation warnings.

It is however more than just a mass replace as also all places where eth/common or eth/common/eth_types or eth/common/eth_types_rlp got imported have been revised and adjusted to a better per submodule based import.

There are still a bunch of toMDigest deprecation warnings but that convertor is not needed for fluffy code anymore so in theory it should not be used (bug?). It seems to still get imported via export leaks ffrom imported nimbus code I think.

### DIFF
--- a/fluffy/common/common_types.nim
+++ b/fluffy/common/common_types.nim
@@ -7,9 +7,9 @@
 
 {.push raises: [].}
 
-import results, ssz_serialization, stew/byteutils, eth/rlp, nimcrypto/hash
+import results, ssz_serialization, stew/byteutils, eth/rlp, eth/common/hashes
 
-export hash, ssz_serialization
+export hashes, ssz_serialization
 
 type
   Bytes2* = ByteVector[2]
@@ -17,7 +17,16 @@ type
 
   ContentId* = UInt256
   ContentKeyByteList* = ByteList[2048] # The encoded content key
-  BlockHash* = MDigest[32 * 8] # Bytes32
+  BlockHash* = Hash32
+
+func fromSszBytes*(T: type Hash32, data: openArray[byte]): T {.raises: [SszError].} =
+  if data.len != sizeof(result):
+    raiseIncorrectSize T
+
+  T.copyFrom(data)
+
+template toSszType*(v: Hash32): array[32, byte] =
+  v.data
 
 func `$`*(x: ByteList[2048]): string =
   x.asSeq.to0xHex()
--- a/fluffy/network/state/content/content_keys.nim
+++ b/fluffy/network/state/content/content_keys.nim
@@ -14,16 +14,16 @@ import
   nimcrypto/[hash, sha2, keccak],
   results,
   stint,
-  eth/common/eth_types,
+  eth/common/hashes,
   ssz_serialization,
   ./nibbles
 
 export ssz_serialization, common_types, hash, results
 
 type
-  NodeHash* = KeccakHash
-  CodeHash* = KeccakHash
-  AddressHash* = KeccakHash
+  NodeHash* = Hash32
+  CodeHash* = Hash32
+  AddressHash* = Hash32
 
   ContentType* = enum
     # Note: Need to add this unused value as a case object with an enum without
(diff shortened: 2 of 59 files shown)


fluffy/tools/portal_bridge/state_bridge/world_state_helper.nim
```diff
@@ -89,15 +89,15 @@ proc applyBlockRewards*(
...
+      uncleMinerAddress = Address(uncleMinerData.miner)
       uncleReward =
         (u256(8 + uncleMinerData.blockNumber - minerData.blockNumber) * baseReward) shr 3
     var accState = worldState.getAccount(uncleMinerAddress)
     accState.addBalance(uncleReward)
     worldState.setAccount(uncleMinerAddress, accState)
 
 const
-  DAORefundContract: EthAddress = address"0xbf4ed7b27f1d666546e30d74d50d173d20bca754"
+  DAORefundContract: Address = address"0xbf4ed7b27f1d666546e30d74d50d173d20bca754"
```
arnetheduck, line 100: this dao list is duplicated from nimbus - now that the types are the same, could reuse it

fluffy/tools/portal_bridge/state_bridge/world_state_helper.nim
```diff
@@ -71,15 +71,15 @@ proc applyStateDiff*(worldState: WorldStateRef, txDiff: TransactionDiff) =
...
+    minerData: tuple[miner: Address, blockNumber: uint64],
+    uncleMinersData: openArray[tuple[miner: Address, blockNumber: uint64]],
 ) =
   const baseReward = u256(5) * pow(u256(10), 18)
 
   block:
     # calculate block miner reward
     let
-      minerAddress = EthAddress(minerData.miner)
+      minerAddress = Address(minerData.miner)
```
arnetheduck, line 82: doesn't it hint unnecessary conversion here?

fluffy/tools/portal_bridge/state_bridge/world_state_helper.nim
kdeme: Ah thanks, I was not aware that came from nimbus. I'll reuse that one.

fluffy/tools/portal_bridge/state_bridge/world_state_helper.nim
kdeme: It does, and another one at line 92. Missed those, will fix.

kdeme: https://github.com/status-im/nimbus-eth1/pull/2698/commits/bad18bc64a29e0a7cf2a529766602ba59d6d28a4

fluffy/tools/portal_bridge/state_bridge/world_state_helper.nim
kdeme: https://github.com/status-im/nimbus-eth1/pull/2698/commits/bad18bc64a29e0a7cf2a529766602ba59d6d28a4

fluffy/common/common_types.nim
```diff
@@ -7,17 +7,26 @@
...
+export hashes, ssz_serialization
 
 type
   Bytes2* = ByteVector[2]
   Bytes32* = ByteVector[32]
 
   ContentId* = UInt256
   ContentKeyByteList* = ByteList[2048] # The encoded content key
-  BlockHash* = MDigest[32 * 8] # Bytes32
+  BlockHash* = Hash32
```
kdeme, line 20: I'm thinking about also removing this `BlockHash` alias. I will let it sit for now, but perhaps a follow-up PR will come :).
There is not much use to it, and it might be confusing into letting people think it is any different from Hash32.

If the variable names are always set to something good like `blockHash` or `block_hash`, and not just `hash` or `data`, then it is cleaner with just `Hash32` usage I believe.

fluffy/network/state/content/content_keys.nim
```diff
@@ -14,16 +14,16 @@ import
...
 
 export ssz_serialization, common_types, hash, results
 
 type
-  NodeHash* = KeccakHash
-  CodeHash* = KeccakHash
-  AddressHash* = KeccakHash
+  NodeHash* = Hash32
+  CodeHash* = Hash32
+  AddressHash* = Hash32
```
kdeme, line 26: Comment on `BlockHash` applies on these too.

fluffy/common/common_types.nim
arnetheduck: that's probably a good idea - unless it appears in a spec, I'd go for `Hash32`

arnetheduck:
> There are still a bunch of toMDigest deprecation warnings but that convertor is not needed 

https://github.com/nim-lang/Nim/issues/24241

a bit annoying :/ the trick is to not import eth_hash at all

kdeme:
> [nim-lang/Nim#24241](https://github.com/nim-lang/Nim/issues/24241)
> 
> a bit annoying :/ the trick is to not import eth_hash at all

Yeah, a lot of them went away when I adjusted to the more granular importing in the fluffy code base. But some `eth_hash` is likely still leaking in through imported nimbus code that did not yet go through the same exercise. 

Thanks for reproducing & filing that issue, that was on my todo list :)

PR #2704, opened 2024-10-07
replace deprecated types

chirag-parmar:
### Changelog

- rename `EthAddress` to `Address` to 'possibly' reduce imports.
- rename deprecated types `Hash256`, `BloomFilter`, `BlockHash`, `Blob`, `BlockHeader`, `VersionedHashes` and more.
- rename `EthBlock` to `Block` to 'possibly' reduce imports.
- remove usage `w3Addr`, `w3Hash`, `ethHash`, `ethAddr`  and other conversions wherever possible.
- use specific imports of types wherever possible


arnetheduck:
Rather than a straight search and replace, the plan here is to also update imports at the same time so as to move away from importing eth/common and instead import more specific types - in particular, most places that use `Hash256` actually only need to import `eth/common/hashes`.

The other part of this change is to remove unnecessary conversions, such as `w3hash` and friends since the types in nim-eth and nim-web3 now match. The warnings are a useful marker for code that hasn't yet been reviewed.

nimbus/common/chain_config.nim
```diff
@@ -11,7 +11,8 @@
 
 import
   std/[tables, strutils, times, macros],
-  eth/[common, rlp, p2p], eth/common/eth_types_json_serialization,
+  eth/[rlp, p2p], eth/common/eth_types_json_serialization,
+  eth/common/[base, hashes, addresses],
```
jangko, line 15: Looks like you need to import `eth/common/eth_types_rlp`  too to solve the failure in CI.

chirag-parmar: Thanks @jangko

chirag-parmar:
@kdeme @arnetheduck this is ready for review.

nimbus/beacon/web3_eth_conv.nim
```diff
@@ -25,14 +25,14 @@ export
 
 type
   Web3FixedBytes*[N: static int] = web3Types.FixedBytes[N]
```
jangko, line 27: No need to keep `Web3FixedBytes`, now it is the same type with nim-eth `FixedBytes`.

Ditto for `Web3Hash`, `Web3Address`, `Web3Bloom`, and `Web3PrevRandao`.

chirag-parmar: Had removed it earlier but reverted the changes. I'm not sure if ditto types have any use in Nim but I reverted the changes because they didn't cause any deprecation warnings. Will remove them again.

advaita-saha:
> #2733 might address the Kurtosis CI failure

Yes I rebase with master will solve the kurtosis issue

jangko:
tips: apart from fluffy, this kind of types and identifiers job usually need to build/compile:
- nimbus/nimbus
- tests/all_tests
- tests/test_tools_build
- nrpc/nrpc

if these four can be built using `nim c -c ...` usually there will be no problem in the CI(excluding logical bugs).

hive_integration/nodocker/engine/engine/invalid_payload.nim
```diff
@@ -166,7 +166,7 @@ method execute(cs: InvalidPayloadTestCase, env: TestEnv): bool =
 
       var attr = env.clMock.latestPayloadAttributes
       attr.timestamp = w3Qty(shadow.alteredPayload.timestamp, 1)
-      attr.prevRandao = w3Hash()
+      attr.prevRandao = Hash32()
```
jangko, line 169: same as above

hive_integration/nodocker/engine/engine/payload_execution.nim
```diff
@@ -349,7 +349,7 @@ method execute(cs: NewPayloadOnSyncingClientTest, env: TestEnv): bool =
 
       # Send the forkchoiceUpdated with a reference to the valid payload on the SYNCING client.
       var
-        random                = w3Hash()
+        random                = Hash32()
```
jangko, line 352: same as above

hive_integration/nodocker/rpc/client.nim
```diff
@@ -66,7 +66,7 @@ proc txReceipt*(client: RpcClient, txHash: eth_types.Hash32): Future[Option[Rece
     receiptType: LegacyReceipt,
     isHash     : rc.root.isSome,
     status     : rc.status.isSome,
-    hash       : rc.root.get(w3Hash()),
+    hash       : rc.root.get(),
```
jangko, line 69: should be `hash       : rc.root.get(default(Hash32))`

hive_integration/nodocker/engine/withdrawals/wd_base_spec.nim
```diff
@@ -160,12 +160,12 @@ proc verifyContractsStorage(ws: WDBaseSpec, env: TestEnv): Result[void, string]
 
   if latestPayloadNumber >= ws.forkHeight.uint64:
     # Shanghai
-    r.expectStorageEqual(WARM_COINBASE_ADDRESS, 100.u256.w3FixedBytes)    # WARM_STORAGE_READ_COST
-    p.expectStorageEqual(PUSH0_ADDRESS, latestPayloadNumber.u256.w3FixedBytes) # tx succeeded
+    r.expectStorageEqual(WARM_COINBASE_ADDRESS, FixedBytes[32](100.u256.toBytesBE))    # WARM_STORAGE_READ_COST
```
jangko, line 163: nim-eth have something like `100.u256.to(Bytes32)` for this kind of conversion.

chirag-parmar: very handy!! will change this. Thanks for the suggestion.

hive_integration/nodocker/rpc/client.nim
chirag-parmar: Hmm don't know how I missed it. Thanks

hive_integration/nodocker/engine/types.nim
```diff
@@ -103,11 +103,11 @@ template testCond*(expr, body: untyped) =
     body
     return false
 
-proc `==`*(a: Opt[BlockHash], b: Opt[common.Hash256]): bool =
+proc `==`*(a: Opt[Hash32], b: Opt[Hash32]): bool =
```
arnetheduck, line 106: this function can be removed entirely

nimbus/beacon/beacon_engine.nim
```diff
@@ -85,8 +84,8 @@ template wrapException(body: untyped): auto =
 
 # setInvalidAncestor is a callback for the downloader to notify us if a bad block
 # is encountered during the async sync.
-func setInvalidAncestor(ben: BeaconEngineRef,
-                         invalid, origin: common.BlockHeader) =
+proc setInvalidAncestor(ben: BeaconEngineRef,
```
arnetheduck, line 87: we tend to prefer `func` when that works, why the change here?

hive_integration/nodocker/engine/types.nim
chirag-parmar: makes sense. missed it.

nimbus/beacon/beacon_engine.nim
chirag-parmar: good catch. I think I missed the change [here](https://github.com/status-im/nimbus-eth1/commit/266b72698deeab5c7742f144af38dfe2738c466c) while rebasing. Should I open another PR?

arnetheduck: yes

PR #2705, opened 2024-10-07
Fluffy: Create separate portal rpc handlers for each sub-network

bhartnett:
This PR creates a separate rpc handler for each portal sub-network. 

I've removed the 'network' string parameter and basically copied the code into separate files and renamed a few files and procs. The validation that was added to state store no longer requires the `network == 'state'` check and that code is removed from history and beacon handlers. 

There now is a bit of code duplication which is to be expected and I'm thinking that we can refactor this out in future PRs when we add the network specific validation. Perhaps the duplicated code can go into functions in another file that is shared where it makes sense.

### DIFF
--- a/fluffy/rpc/rpc_portal_beacon_api.nim
+++ b/fluffy/rpc/rpc_portal_beacon_api.nim
@@ -51,87 +51,6 @@ proc installPortalBeaconApiHandlers*(rpcServer: RpcServer, p: PortalProtocol) =
     invalidValueErr =
       (ref errors.InvalidRequest)(code: -32602, msg: "Invalid content value")
 
-  rpcServer.rpc("portal_beaconNodeInfo") do() -> NodeInfo:
-    return p.routingTable.getNodeInfo()
-
-  rpcServer.rpc("portal_beaconRoutingTableInfo") do() -> RoutingTableInfo:
-    return getRoutingTableInfo(p.routingTable)
-
-  rpcServer.rpc("portal_beaconAddEnr") do(enr: Record) -> bool:
-    let node = Node.fromRecord(enr)
-    let addResult = p.addNode(node)
-    if addResult == Added:
-      p.routingTable.setJustSeen(node)
-    return addResult == Added
-
-  rpcServer.rpc("portal_beaconAddEnrs") do(enrs: seq[Record]) -> bool:
-    # Note: unspecified RPC, but useful for our local testnet test
-    for enr in enrs:
-      let node = Node.fromRecord(enr)
-      if p.addNode(node) == Added:
-        p.routingTable.setJustSeen(node)
-
-    return true
-
-  rpcServer.rpc("portal_beaconGetEnr") do(nodeId: NodeId) -> Record:
-    if p.localNode.id == nodeId:
-      return p.localNode.record
-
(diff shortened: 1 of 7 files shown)


kdeme (commented):
> There now is a bit of code duplication which is to be expected and I'm thinking that we can refactor this out in future PRs when we add the network specific validation. Perhaps the duplicated code can go into functions in another file that is shared where it makes sense.

Yeah, some code could go in functions in a rpc_portal_common or something like that.

However, as I look at this PR now, there are still quite a lot of calls that will never be different between the different subprotocols. E.g. all the ENR related calls. Perhaps we should still keep a generic version around for a part of the calls?

bhartnett:
> > There now is a bit of code duplication which is to be expected and I'm thinking that we can refactor this out in future PRs when we add the network specific validation. Perhaps the duplicated code can go into functions in another file that is shared where it makes sense.
> 
> Yeah, some code could go in functions in a rpc_portal_common or something like that.
> 
> However, as I look at this PR now, there are still quite a lot of calls that will never be different between the different subprotocols. E.g. all the ENR related calls. Perhaps we should still keep a generic version around for a part of the calls?

Yes good point and I guess when adding a new sub-protocol adding the enr related calls would be easier than having to remember to define all the endpoints each time. Okay, I'll make a portal common rpc handler.

fluffy/rpc/rpc_portal_beacon_api.nim
```diff
@@ -0,0 +1,270 @@
...
+    let
+      key = hexToSeqByte(contentKey)
+      content = hexToSeqByte(contentValue)
+      contentKeys = ContentKeysList(@[ContentKeyByteList.init(key)])
+      numberOfPeers =
+        await p.neighborhoodGossip(Opt.none(NodeId), contentKeys, @[content])
+
+    return numberOfPeers
+
+  rpcServer.rpc("portal_beaconRandomGossip") do(
```
bhartnett, line 261: @kdeme This endpoint doesn't appear to be in the spec. Should it be moved to a portal_debug namespace or even removed?

fluffy/rpc/rpc_portal_debug_history_api.nim
```diff
@@ -9,20 +9,19 @@
...
 import
   json_rpc/rpcserver,
-  stew/byteutils,
   ../network/wire/portal_protocol,
   ../eth_data/history_data_seeding,
   ../database/content_db
 
 export rpcserver
 
+# TODO: perhaps these endpoints should be named differently staring with "portal_debug_"?
```
bhartnett, line 18: Any thoughts on this?

bhartnett:
This PR conflicts with https://github.com/status-im/nimbus-eth1/pull/2689 so I'll wait for that to be merged in before pulling in that code to resolve the conflicts.

fluffy/rpc/rpc_portal_beacon_api.nim
bhartnett: I see, in that case I'll leave the name and just remove it from the other handlers. Probably we can rename to `portal_beaconGossip` after the spec is updated.

kdeme (approved):

PR #2710, opened 2024-10-08
Rename Fluffy debug rpc methods

bhartnett:


fluffy/docs/the_fluffy_book/docs/history-content-bridging.md
```diff
@@ -101,6 +101,7 @@ the assigned `--data-dir`.
 #### Step 2: Seed the epoch accumulators into the Portal network
 Run Fluffy and trigger the propagation of data with the
 `portal_history_propagateEpochRecords` JSON-RPC API call:
+TODO: Was this rpc endpoint removed? Should we remove this part from the book?
```
bhartnett, line 104: @kdeme This rpc method doesn't exist. Should we remove this part from the Fluffy book?

fluffy/rpc/rpc_calls/rpc_portal_debug_calls.nim
```diff
@@ -15,10 +15,13 @@ Opt[string].useDefaultSerializationIn JrpcConv
...
+  proc portal_debug_historyGossipHeaders(
+    era1File: string, epochRecordFile: Opt[string]
+  ): bool
+
+  proc portal_debug_historyGossipHeaders(era1File: string): bool
+  proc portal_debug_historyGossipBlockContent(era1File: string): bool
+  proc portal_debug_history_storeContent(dataFile: string): bool
+  proc portal_debug_history_propagate(dataFile: string): bool
+  proc portal_debug_history_propagateHeaders(dataFile: string): bool
+  proc portal_debug_history_propagateBlock(dataFile: string, blockHash: string): bool
```
bhartnett, line 27: It appears some of these methods are not used anywhere in the code. Do you think any of these should be removed?

kdeme: I use the first 2 sometimes still, and the storeContent from a file has some debugging/testing use. I'll have a look at the others, but they can probably go. I'll revise this in another PR, you can leave them in for now.

kdeme (approved):

PR #2712, opened 2024-10-08
Ordered trie

arnetheduck:
Speed up trie computations and remove redundant ways of performing this operation.

PR #2720, opened 2024-10-09
Bump nimbus-eth2 + add minilru to exclude list

kdeme:


kdeme:
nvm, got added already in https://github.com/status-im/nimbus-eth1/pull/2712

PR #2723, opened 2024-10-10
Fluffy: Add validation and local storage of content in remaining state portal rpc methods

bhartnett:
Changes in this PR:
- Refactor and cleanup some state network code.
- Added new validation functions which are used by the state portal rpc methods.
- Added validation of content keys to all state rpc methods.
- Updated stateFindContent, stateGossip and stateOffer to validate and store content locally before sending or returning.
- Updated stateRecursiveFindContent and stateTraceRecursiveFindContent to first lookup content in the local db, if it doesn't exist then do the recursive lookup, then validate and store fetched content locally before returning.

### DIFF
--- a/fluffy/network/state/state_network.nim
+++ b/fluffy/network/state/state_network.nim
@@ -48,23 +48,22 @@ proc new*(
     historyNetwork = Opt.none(HistoryNetwork),
     validateStateIsCanonical = true,
 ): T =
-  let cq = newAsyncQueue[(Opt[NodeId], ContentKeysList, seq[seq[byte]])](50)
+  let
+    cq = newAsyncQueue[(Opt[NodeId], ContentKeysList, seq[seq[byte]])](50)
+    s = streamManager.registerNewStream(cq)
+    portalProtocol = PortalProtocol.new(
+      baseProtocol,
+      getProtocolId(portalNetwork, PortalSubnetwork.state),
+      toContentIdHandler,
+      createGetHandler(contentDB),
+      createStoreHandler(contentDB, portalConfig.radiusConfig),
+      createRadiusHandler(contentDB),
+      s,
+      bootstrapRecords,
+      config = portalConfig,
+    )
 
-  let s = streamManager.registerNewStream(cq)
-
-  let portalProtocol = PortalProtocol.new(
-    baseProtocol,
-    getProtocolId(portalNetwork, PortalSubnetwork.state),
-    toContentIdHandler,
-    createGetHandler(contentDB),
-    createStoreHandler(contentDB, portalConfig.radiusConfig),
-    createRadiusHandler(contentDB),
-    s,
-    bootstrapRecords,
-    config = portalConfig,
-  )
-
-  return StateNetwork(
+  StateNetwork(
     portalProtocol: portalProtocol,
     contentDB: contentDB,
     contentQueue: cq,
@@ -109,7 +108,7 @@ proc getContent(
 
   n.portalProtocol.storeContent(contentKeyBytes, contentId, contentValueBytes)
 
-  return Opt.some(contentValue)
+  Opt.some(contentValue)
 
 proc getAccountTrieNode*(
     n: StateNetwork, key: AccountTrieNodeKey
--- a/fluffy/network/state/state_network.nim
+++ b/fluffy/network/state/state_network.nim
@@ -142,19 +142,6 @@ proc getStateRootByBlockNumOrHash*(
 
   Opt.some(header.stateRoot)
 
-proc getStateRootForValidation(
-    n: StateNetwork, offer: ContentOfferType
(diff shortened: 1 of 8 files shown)


fluffy/network/state/state_network.nim
```diff
@@ -132,16 +132,29 @@ proc getContractCode*(
...
+    n: StateNetwork, offer: ContentOfferType
+): Future[Result[Opt[Hash32], string]] {.async: (raises: [CancelledError]).} =
+  let maybeStateRoot =
+    if n.validateStateIsCanonical:
+      let stateRoot = (await n.getStateRootByBlockNumOrHash(offer.blockHash)).valueOr:
+        return err("Failed to get state root by block hash")
+      Opt.some(stateRoot)
+    else:
+      # Skip state root validation
+      Opt.none(Hash32)
```
kdeme, line 155: I find this quite confusing. I don't think `getStateRootForValidation` should be returning the `Opt.none(Hash32)`  when state canonicalness (is that a word?) is disabled. This should be checked a level up imo.

I think that you did that so it would fit more elegantly in the code lower down, but it's not very clear to a user reading the code.

fluffy/network/state/state_validation.nim
```diff
@@ -181,3 +181,50 @@ proc validateOffer*(
...
 
   validateRetrieval(key, offer.toRetrievalValue())
+
+func validateGetContentKey*(
+    keyBytes: ContentKeyByteList
+): Result[(ContentKey, ContentId), string] =
+  let key = ContentKey.decode(keyBytes).valueOr:
+    return err("Unable to decode content key")
+  if key.contentType == unused:
+    return err("ContentKey contentType: unused")
```
kdeme, line 191: I think we should add this failure in the `ContentKey.decode` call itself.

fluffy/network/state/state_validation.nim
```diff
@@ -181,3 +181,50 @@ proc validateOffer*(
     return err("hash of bytecode doesn't match the code hash in the account proof")
 
   validateRetrieval(key, offer.toRetrievalValue())
+
+func validateGetContentKey*(
+    keyBytes: ContentKeyByteList
+): Result[(ContentKey, ContentId), string] =
+  let key = ContentKey.decode(keyBytes).valueOr:
+    return err("Unable to decode content key")
```
kdeme, line 189: You might want to pass along the decoding failure (which will typically be some SSZ error)

fluffy/rpc/rpc_portal_state_api.nim
```diff
@@ -36,42 +36,56 @@ proc installPortalStateApiHandlers*(rpcServer: RpcServer, p: PortalProtocol) =
...
+        contentBytes = hexToSeqByte(contentItem[1])
+        contentValue = validateOfferGetValue(Opt.none(Hash32), key, contentBytes).valueOr:
+          raise invalidValueErr()
+        contentKV = ContentKV(contentKey: keyBytes, content: contentBytes)
+
+      contentItemsToStore.add((keyBytes, contentId, contentValue))
       contentItemsToOffer.add(contentKV)
 
+    for (keyBytes, contentId, contentValue) in contentItemsToStore:
+      p.storeContent(keyBytes, contentId, contentValue)
```
kdeme, line 87: Here I'm not sure what to do best, store or not.

fluffy/network/state/state_network.nim
bhartnett: Sure, I see what you mean, it's less readable perhaps. I'll revert it back to they way I did it before.

fluffy/network/state/state_validation.nim
bhartnett: Yep, will do

fluffy/network/state/state_validation.nim
bhartnett: Yes good point

fluffy/rpc/rpc_portal_state_api.nim
bhartnett: Yeah I wasn't sure about the methods that send or receive to/from a specific enr. We should at least aim to be consistent so portal_stateFindContent and portal_stateOffer should probably be similar in that they either store in db or they dont.

bhartnett: I'm thinking that perhaps both `portal_stateFindContent` and `portal_stateOffer` should not store in the db. 

Here's my reasoning:
- We have 2 'low level' methods that just store content locally in db, stateLocalContent and stateStore.
- We have 3 'high level' methods that store in the db and fetch over the network when required, stateGossip, stateRecursiveFindContent, stateTraceRecursiveFindContent.

So perhaps we should keep `portal_stateFindContent` and `portal_stateOffer` to act as low level endpoints that only go over the network to specific enrs and don't store in the db.

Issue #2258, opened 2024-05-31
Rename main binary to `nimbus_execution_client`


arnetheduck:
Given that we have `nimbus_beacon_node` and `nimbus_validator_client`, it makes sense to rename the execution to a similar name.

This frees up the name `nimbus` to become the integrated el-cl-vc binary and avoids compromising this UX with execution-only stuff.

jangko:
fixed by #2728

PR #2728, opened 2024-10-10
Rename nimbus binary to nimbus_execution_client

pedromiguelmiranda:
Renaming nimbus binary to nimbus_execution_client in order to free the "nimbus" binary name. addresses issue #2258 

Directory renaming will be addressed in a different issue (#2738)


advaita-saha:
Kurtosis check is breaking, as it keeps waiting for the CL to sync
Issue happened after https://github.com/ethpandaops/ethereum-package/pull/804

I am working with the ethpandaops to fix this, till then not to worry about the kurtosis check

Makefile
```diff
@@ -211,7 +211,9 @@ $(TOOLS): | build deps rocksdb
 		$(ENV_SCRIPT) nim c $(NIM_PARAMS) -d:chronicles_log_level=TRACE -o:build/$@ "$${TOOL_DIR}/$@.nim"
 
 # a phony target, because teaching `make` how to do conditional recompilation of Nim projects is too complicated
-nimbus: | build deps rocksdb
+nimbus: | build deps nimbus_execution_client
```
tersec, line 214: Still a `nimbus` target?

There is a compatibility argument for this, but not sure why the `build deps` and `rocksdb` parts get split here. It makes the `nimbus` target distinct from `nimbus_execution_client`.

advaita-saha: NOTE:- removing the nimbus target, will break kurtosis again, as it most likely needs changes in the launcher code in ethpandaops codebase of `ethereum-package`

tersec: Ok, we can do this in 2 stages, then, this PR, then switch Kurtosis and drop the `nimbus` target.

We don't have real compatibility guarantees for this kind of thing yet -- there's no release. But once we do, it'll be more involved.

So better to do it sooner in general. Breaking Kurtosis and being able to fix it in that one place is better than dealing with the fallout of a change once there's end-user support.

tersec: Or, maybe have some other temporary bridge support.

pedromiguelmiranda: nice catch. Also _nimbus_execution_client_ becomes a public target. will apply a different strategy

pedromiguelmiranda:
This pr only addresses the binary name as identified in #2258.

We could go further and rename the source directory from _nimbus_ to _nimbus_execution_client_ , however **all files history** under nimbus directory won't be easily accessible via _**git log**_ command, only via _**git log --follow**_.

Would like to get some opinions regarding if we are confortable with the directory renaming, or I can open a follow up issue to think/discuss about it about it

kdeme:
Fluffy failure is a flaky test due to a recent change. Can be ignored here.

tersec:
> This pr only addresses the binary name as identified in #2258.
> 
> We could go further and rename the source directory from _nimbus_ to _nimbus_execution_client_ , however **all files history** under nimbus directory won't be easily accessible via _**git log**_ command, only via _**git log --follow**_.
> 
> Would like to get some opinions regarding if we are confortable with the directory renaming, or I can open a follow up issue to think/discuss about it about it

Seems fine to leave the directory as-is for now. Can be discussed in a separate PR.

PR #2739, opened 2024-10-15
Fluffy: Enable content cache for state network

bhartnett:


fluffy/network/wire/portal_protocol.nim
```diff
@@ -1590,7 +1594,12 @@ proc storeContent*(
     contentKey: ContentKeyByteList,
     contentId: ContentId,
     content: seq[byte],
+    cacheContent = false,
 ): bool {.discardable.} =
+  if cacheContent:
+    # We cache content regardless of whether it is in our radius or not
+    p.contentCache.put(contentId, content)
```
kdeme, line 1601: Curious to hear what you think about putting it in the cache only in the case it does not get stored locally?

That would avoid double (even though temporary) storage and give more space to the cache.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -1600,6 +1609,22 @@ proc storeContent*(
   else:
     false
 
+proc getLocalContent*(
+    p: PortalProtocol, contentKey: ContentKeyByteList, contentId: ContentId
+): Opt[seq[byte]] =
+  # The cache can contain content that is not in our radius
+  let maybeContent = p.contentCache.get(contentId)
+  if maybeContent.isSome():
```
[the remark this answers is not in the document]
bhartnett: Sure, I'll add a metric for this.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -1600,6 +1609,22 @@ proc storeContent*(
...
+    return Opt.some(maybeContent.get())
```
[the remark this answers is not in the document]
bhartnett: Yes good point, I'll update.

fluffy/network/wire/portal_protocol.nim
bhartnett: That is an interesting idea. The benefit of caching everything is of course improved performance so when the database gets larger look ups from memory will be much faster. It also feels a bit messy and unconventional to do something like that in my opinion. Makes it harder to reason about whether it should exist in the cache or not. 

If the content gets removed from the db due to a change in the radius or more data coming in then the data is gone but not in the cache. Not sure if this scenario might cause issues. I'd lean towards caching all requests. Note that I'm not caching data received from offers, only data fetched during requests.

bhartnett:
@kdeme Any thoughts on what should be the default content cache size?

fluffy/network/wire/portal_protocol.nim
kdeme: Yes, I agree it feels messy which is why I asked about it. It didn't really appeal to me either as it brings caching and storage intermingled. But the idea behind is was because of how both the caching and storage are many times faster than the network request.

Good point also on the content that might get removed during radius adjustments. Feels like there would be some side effects like that and in general not a clean separation, so lets leave it as it is done now.

kdeme:
> @kdeme Any thoughts on what should be the default content cache size?

Not really, I think the current setting of 256 seems reasonable to start with.

bhartnett:
> > @kdeme Any thoughts on what should be the default content cache size?
> 
> Not really, I think the current setting of 256 seems reasonable to start with.

Actually I just reduced it to 100 for now to reduce the risk of it effecting the memory usage of the nodes in our fleet. I've also made it configurable so we can increase the cache size or even disable it if needed from the command line.

kdeme (approved):

PR #2750, opened 2024-10-18
Fluffy: Implement poke in state network

bhartnett:
Poke in the state network is feasible by passing in the parent offer when looking up content. The fetched state content is then combined with the parent offer to build the offer which is used in the poke. 

I'm going to suggest that poke is added to the state network specs because I can't see any reason for it to be disabled after implementing this change.

I've also added poke to the stateGetContent JSON-RPC method by supporting an optional parentContent parameter which should contain the parent offer which is required in order to enable poke.


bhartnett:
Leaving this as a draft PR until the specs are updated

fluffy/rpc/rpc_portal_state_api.nim
```diff
@@ -83,7 +83,9 @@ proc installPortalStateApiHandlers*(rpcServer: RpcServer, p: PortalProtocol) =
 
     SSZ.encode(offerResult).to0xHex()
 
-  rpcServer.rpc("portal_stateGetContent") do(contentKey: string) -> ContentInfo:
+  rpcServer.rpc("portal_stateGetContent") do(
+    contentKey: string, parentContent: Opt[string]
+  ) -> ContentInfo:
```
kdeme, line 88: For a user this API does complicate things. Feels like the "health" of the network should not be pushed towards the UX of user API. But perhaps its ok as an `Opt`...

kdeme (commented):
lgtm but would wait with merging until accepted in specs first.

fluffy/rpc/rpc_portal_state_api.nim
bhartnett: Actually I agree and for this reason I didn't propose changing the JSON-RPC spec to add this in my other PR. I'm planning to revert this part but I'll wait to see what happens with the spec changes first.

PR #2763, opened 2024-10-21
Implementation of EIP-7742 for Prague

### LINKED ISSUE #2753: Implement EIP-7742 for Pectra
https://eips.ethereum.org/EIPS/eip-7742

Was agreed to be included yesterday.

jangko:
TODOs:
- [x] How `maximumBlobCount` should be used in block creation?
- [x] Where `targetBlobCount` come from when converting to `Header`. From `ExecutionPayloadV4` or from `engine_newPayloadV4`'s param?
- [x] What about `engine_getPayloadV4`? should the response return `ExecutionPayloadV4` or additional `targetBlobCount` field?
- [x] Difference between EIP-7742 spec `targetBlobCount` of block header, and consensus spec `targetBlobsPerBlock` of `PayloadAttributesV4`
- [ ] emvstate: add targetBlobsPerBlock
- [ ] Upgrade on nimbus-eth2 side

fix #2753

### DIFF
--- a/hive_integration/nodocker/engine/cancun/customizer.nim
+++ b/hive_integration/nodocker/engine/cancun/customizer.nim
@@ -337,7 +337,7 @@ func getTimestamp*(cust: CustomPayloadData, basePayload: ExecutionPayload): uint
 # blockHash is calculated automatically.
 proc customizePayload*(cust: CustomPayloadData, data: ExecutableData): ExecutableData {.gcsafe.} =
   let requestsHash = calcRequestsHash(data.executionRequests)
-  var customHeader = blockHeader(data.basePayload, data.beaconRoot, requestsHash)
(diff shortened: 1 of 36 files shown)


tersec:
> TODOs:
> 
>     * [ ]  How `maximumBlobCount` should be used in block creation?

My understanding is that it's exactly the same as the current max blob count today (6), but not a constant across blocks.

>     * [ ]  Where `targetBlobCount` come from when converting to `Header`. From `ExecutionPayloadV4` or from `engine_newPayloadV4`'s param?

This is being [actively discussed](https://discord.com/channels/595666850260713488/1296496723702710393/1298013291053973505) in the EIP-7742 - EL and CL candidates thread in the #allcoredevs channel of the Eth R&D Discord server.

> 
>     * [ ]  What about `engine_getPayloadV4`? should the response return `ExecutionPayloadV4` or additional `targetBlobCount` field?

As far as I can tell, undetermined as of yet.

jangko:
removed from prague, closing

Issue #1769, opened 2023-09-22
Concurrent requests in a Portal protocol lookup get canceled before data is validated


kdeme:

bhartnett:
@kdeme Is the goal for this issue to update portal protocol to return the first valid content and then cancel all other pending requests? In other words do we want to add validation into the `contentLookup`? I'm assuming that cancelling the pending request isn't a problem as long as we know that at least a single valid content response is returned.

kdeme:
Yes, this is about not yet canceling the request until we are certain that the current one that we received has been validated and is correct. Practically that would probably mean adding a validation handler in wire protocol as you have already mentioned in the other PR.

I think that would be beneficial over retrying getContent attempts in the specific networks. But I am also not 100% sure, because in the happy case (which might/will occur more often), it means that we might be starting validation for no reason on a second or a third piece of data. Perhaps the actual validation should get queued? I haven't really thought it out that well yet.

bhartnett:
> Yes, this is about not yet canceling the request until we are certain that the current one that we received has been validated and is correct. Practically that would probably mean adding a validation handler in wire protocol as you have already mentioned in the other PR.

Note that the validation handler will need to be passed in to each call to contentLookup because the handler will be different for each type of content being validated and in some cases some context data needs to be available such as the block header when validating a block body. For this reason, using a single handler defined per sub-network won't work in this case.

> I think that would be beneficial over retrying getContent attempts in the specific networks. But I am also not 100% sure, because in the happy case (which might/will occur more often), it means that we might be starting validation for no reason on a second or a third piece of data. Perhaps the actual validation should get queued? I haven't really thought it out that well yet.

Since chronos is single threaded if we fetch the content, validate and then cancel remaining pending queries with no await calls then I believe the validation won't run on a second query. I'm not completely sure about this but I can try it out to see.

bhartnett:
> > Note that the validation handler will need to be passed in to each call to contentLookup because the handler will be different for each type of content being validated and in some cases some context data needs to be available such as the block header when validating a block body.
> 
> Well, you could handle each type of content in the handler as you do know the different content keys. A bit like is done for incoming offers? Regarding the block header, I think the data would have to be requested, which adds to the issue that I was mentioned of double/triple validation.

[...]

PR #2770, opened 2024-10-23
Fluffy: Make content request retries configurable

bhartnett:
Changes in this PR:
- Add a hidden CLI parameter to enable configuring the number of content request retries.
- Add retries to state network get content lookup.
- Minor cleanup in history network get content code, fix requestRetries to correctly reflect the count.
- Use new CLI parameter in both history network and state network get content lookups.

### DIFF
--- a/fluffy/conf.nim
+++ b/fluffy/conf.nim
@@ -305,9 +305,9 @@ type
     contentRequestRetries* {.
       hidden,
       desc: "Max number of retries when requesting content over the network.",
-      defaultValue: 3,
-      name: "content-request-retries"
-    .}: int
+      defaultValue: 1,
+      name: "debug-content-request-retries"
+    .}: uint
 
     contentCacheSize* {.
       hidden,
--- a/fluffy/conf.nim
+++ b/fluffy/conf.nim
@@ -302,6 +302,13 @@ type
       name: "force-prune"
     .}: bool
 
+    contentRequestRetries* {.
+      hidden,
+      desc: "Max number of retries when requesting content over the network.",
+      defaultValue: 3,
+      name: "content-request-retries"
+    .}: int
+
     contentCacheSize* {.
       hidden,
       desc:
--- a/fluffy/fluffy.nim
+++ b/fluffy/fluffy.nim
@@ -196,7 +196,7 @@ proc run(
       portalConfig: portalProtocolConfig,
       dataDir: string config.dataDir,
       storageCapacity: config.storageCapacityMB * 1_000_000,
-      contentRequestRetries: config.contentRequestRetries,
+      contentRequestRetries: config.contentRequestRetries.int,
     )
 
     node = PortalNode.new(
--- a/fluffy/fluffy.nim
+++ b/fluffy/fluffy.nim
@@ -196,6 +196,7 @@ proc run(
       portalConfig: portalProtocolConfig,
       dataDir: string config.dataDir,
       storageCapacity: config.storageCapacityMB * 1_000_000,
+      contentRequestRetries: config.contentRequestRetries,
     )
 
     node = PortalNode.new(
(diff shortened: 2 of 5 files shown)


fluffy/conf.nim
```diff
@@ -302,6 +302,13 @@ type
       name: "force-prune"
     .}: bool
 
+    contentRequestRetries* {.
+      hidden,
+      desc: "Max number of retries when requesting content over the network.",
+      defaultValue: 3,
+      name: "content-request-retries"
```
tersec, line 309: Nimbus in general, and I don't know if this should apply to Fluffy [has a convention](https://nimbus.guide/options.html) that if it's a `--debug-foo` flag, it's explicitly documented as not part of the compatibility surface.

But whether that's appropriate here depends on the intent, i.e. is this meant as hidden but more or less supported, or actually-debug/testing.

kdeme: Yeah, it is meant as debugging/testing in this case, and we have a few others of these.

I was not aware of the convention but it does sound like a good idea to apply this to our hidden debug/testing flags.

kdeme (commented):
The retries were originally added as a "temporary quick-fix" when we had most of the fluffy nodes providing data, but the Trin nodes not yet, hence it was always meant (at least in my mind) to be removed again. (I actually have them removed in my refactor PR).

Have you noticed/tested that these are still very much required and improve the content finding?

Additionally, I think if it is necessary to keep these retries (and add them to the other subnetworks), it would perhaps be cleaner to have them a layer up? Of course then it would be per application, and thus have some code duplication.

fluffy/conf.nim
bhartnett: In Fluffy up to this point we have been using hidden parameters for developer / non-user facing parameters. In this case I made it configurable because I'm not sure what the correct value should be and we will likely want to try out different values, perhaps also to debug issues. Not sure about hidden supported vs debug. 

@kdeme Did you have any thoughts on this?

kdeme: I'm not sure but I think what @tersec meant was just that we prefix these cli flags with `debug`.

For the link provided above I read: `Any debug-prefixed flags are considered ephemeral and subject to removal without notice.`

I'd be fine with doing that adjustment.

bhartnett:
> The retries were originally added as a "temporary quick-fix" when we had most of the fluffy nodes providing data, but the Trin nodes not yet, hence it was always meant (at least in my mind) to be removed again. (I actually have them removed in my refactor PR).
> 
> Have you noticed/tested that these are still very much required and improve the content finding?
> 
> Additionally, I think if it is necessary to keep these retries (and add them to the other subnetworks), it would perhaps be cleaner to have them a layer up? Of course then it would be per application, and thus have some code duplication.

I believe the retries are a good idea and likely need to be at the sub-network layer because validation is different for each sub-network. From my reading of the code when we do a content lookup we fetch the content using portal protocol but the content is not validated yet. Inside history network/state network/etc we validate and if validation fails then we should retry so that we can try getting the content from another node. It's possible for a malicious node to send us bad data and we should handle that scenario.

I guess we could remove the retries and make the user retry on failure but I would suggest at least a single retry would be preferred.

kdeme:
> No actually, the current implementation won't improve the content finding, it just handles the bad data scenario.

Yes, exactly. I realized that but it was originally not really added for that reason.

But as it is true that it helps with that, we can for now leave it in there like it is done now.

It is also the reason why we do this shuffle: https://github.com/status-im/nimbus-eth1/blob/master/fluffy/network/wire/portal_protocol.nim#L1141

But it still doesn't feel like the cleanest solution. And there is also this related issue: https://github.com/status-im/nimbus-eth1/issues/1769

Anyway, I'm fine with keeping it as is done in this PR for now.

PR #2777, opened 2024-10-25
Unify tx validation

### LINKED ISSUE #2695: Port transaction validation tests to new function
https://github.com/status-im/nimbus-eth1/blob/d6eb8c36f5ae655bb4a0159efa41bff6b113b527/nimbus/transaction.nim#L157 contains a transaction validation function that has fallen out of use.

The old function however has a set of checks and validations that the new function does not have - it also has test coverage which is missing from the new function - both of these issues need to be address such that the old tests and validations are checked.

jangko:
fix #2695 

overlapped checks are not ported. only checks not appear in the new function are ported. Also merge `intrinsicGas` for both `Transaction` and `CallParams`

### DIFF
--- a/nimbus/core/validate.nim
+++ b/nimbus/core/validate.nim
@@ -13,8 +13,10 @@
 import
   std/[sequtils, sets, strformat],
   ../db/ledger,
-  ".."/[transaction, common/common],
-  ".."/[errors],
+  ../common/common,
+  ../transaction/call_types,
+  ../errors,
+  ../transaction,
   ../utils/utils,
   "."/[dao, eip4844, gaslimit, withdrawals],
   ./pow/[difficulty, header],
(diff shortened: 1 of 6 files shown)


nimbus/core/validate.nim
```diff
@@ -182,6 +184,35 @@ proc validateUncles(com: CommonRef; header: Header;
 # Public function, extracted from executor
 # ------------------------------------------------------------------------------
 
+proc validateLegacySignature(tx: Transaction, fork: EVMFork): bool =
```
arnetheduck, line 187: not for this pr maybe, but in eth2 we use with `Result[void, cstring]` for checks like this which later allows code like:

```
  if tx.txType == TxLegacy:
    ?validateLegacySignature(tx, fork)
```

also, `validateLegacySignatureForm` or something like to to make clear that we're not checking the signature against the data but rather its form.

PR #2781, opened 2024-10-25
better logging for running client

advaita-saha:


nimbus/core/chain/forked_chain.nim
```diff
@@ -224,6 +226,11 @@ proc writeBaggage(c: ForkedChainRef, target: Hash32) =
       for tx in blk.blk.transactions:
         c.txRecords.del(rlpHash(tx))
       prevHash = header.parentHash
+      count.inc
+
+    notice "Finalized blocks persisted to Database",
+      numberOfBlocks = count,
+      last=target.short
```
arnetheduck, line 233: should log the new base (with blocknumber and hash) here so that a chronology of base updates can be reconstructed from the logs - let's make sure to use `base` as well so as to keep it distinct from `finalized`

arnetheduck: or hold on - do we write all finalized blocks to the database? also the ones newer than base? then my previous comment doesn't make sense..

nimbus/core/chain/forked_chain.nim
```diff
@@ -224,6 +226,11 @@ proc writeBaggage(c: ForkedChainRef, target: Hash32) =
       for tx in blk.blk.transactions:
         c.txRecords.del(rlpHash(tx))
       prevHash = header.parentHash
+      count.inc
+
+    notice "Finalized blocks persisted to Database",
```
arnetheduck, line 231:
```suggestion
    notice "Finalized blocks persisted",
```

nimbus/beacon/api_handler/api_newpayload.nim
```diff
@@ -195,4 +195,12 @@ proc newPayload*(ben: BeaconEngineRef,
     let blockHash = latestValidHash(db, parent, ttd)
     return invalidStatus(blockHash, vres.error())
 
+  info "New payload received and validated",
+    number = header.number, 
+    hash = blockHash.short, 
+    parent = header.parentHash.short,
+    txs = blk.transactions.len,
+    gasUsed = header.gasUsed,
+    blobGas = header.blobGasUsed.get()
```
arnetheduck, line 204: this field might not be present

nimbus/beacon/api_handler/api_forkchoice.nim
```diff
@@ -192,12 +192,20 @@ proc forkchoiceUpdated*(ben: BeaconEngineRef,
...
-    info "Created payload for sealing",
-      id = id.toHex,
-      hash = bundle.executionPayload.blockHash.short,
+    info "Created payload for block proposal",
       number = bundle.executionPayload.blockNumber,
-      attr = attrs
+      hash = bundle.executionPayload.blockHash.short,
+      txs = bundle.executionPayload.transactions.len,
+      gasUsed = bundle.executionPayload.gasUsed,
+      blobGasUsed = bundle.executionPayload.blobGasUsed.get(),
```
arnetheduck, line 200: ditto - `get` crashes the application when not set

arnetheduck, line 200:
```suggestion
      blobGasUsed = bundle.executionPayload.blobGasUsed.get(0),
```

arnetheduck: adding a default value avoids the panic

advaita-saha: Should crash logically, but how nimbus eth1 is written it is never set to `Opt.none()`
will `tryGet()` also crash, or better to use `get(0)` ?

nimbus/core/chain/forked_chain.nim
advaita-saha: base is updated independently, but can be added to the logs

nimbus/beacon/api_handler/api_forkchoice.nim
arnetheduck: better use `get(0)` - it's not obvious that it's "never" none when looking at this code locally.

nimbus/core/chain/forked_chain.nim
```diff
@@ -224,6 +226,13 @@ proc writeBaggage(c: ForkedChainRef, target: Hash32) =
       for tx in blk.blk.transactions:
         c.txRecords.del(rlpHash(tx))
       prevHash = header.parentHash
+      count.inc
+
+    notice "Finalized blocks persisted to",
```
tersec, line 231: While this trailing "to" works reasonably well for text-formatted logs, because Chronicles ensures the message is put in a specific position, followed directly by the fields, JSON logs don't necessarily follow this convenient order, e.g.,
```
{"lvl":"INF","ts":"2024-10-27 11:35:22.996+00:00","msg":"Slot start","topics":"beacnde","head":"a2c3ebc0:10267074","delay":"189us727ns","finalized":"320844:01d47808","peers":159,"slot":10267075,"sync":"synced","epoch":320846}
```
from `nimbus-eth2`.

Therefore, the message strings in these logs generally would be better served by being a standalone phrase, e.g., "Finalized blocks persisted", rather than something designed to operate as a more individually nongrammatical fragment ("Finalized blocks persisted to", which needs an object of that preposition to be grammatical).

nimbus/beacon/api_handler/api_forkchoice.nim
```diff
@@ -192,12 +192,22 @@ proc forkchoiceUpdated*(ben: BeaconEngineRef,
...
+      hash = bundle.executionPayload.blockHash.short,
+      txs = bundle.executionPayload.transactions.len,
+      gasUsed = bundle.executionPayload.gasUsed,
+      blobGasUsed = bundle.executionPayload.blobGasUsed.get(Quantity(0)),
+      id = id.toHex,
+      attrs = attrs
 
     return validFCU(Opt.some(id), blockHash)
 
+  info "Forkchoice updated to new head",
```
tersec, line 206: See discussions in https://github.com/status-im/nimbus-eth1/pull/2789#discussion_r1818054791; this is generally better rendered as either
- fork choice updated; or
- forkchoiceUpdated.

For logging purposes, the former is human-friendlier; it treats the engine API call name as an implementation detail, which is probably correct.

PR #2782, opened 2024-10-25
Beacon sync activation control update

mjfh:


nimbus/sync/sync_sched.nim
```diff
@@ -310,18 +399,17 @@ proc onPeerConnected[S,W](dsc: RunnerSyncRef[S,W]; peer: Peer) =
...
 
   # Add peer entry
-  discard dsc.buddies.lruAppend(peer.key, buddy, dsc.ctx.buddiesMax)
-
-  trace "Running peer worker", peer, nPeers,
-    nWorkers=dsc.buddies.len, maxWorkers
+  discard dsc.buddies.lruAppend(peer.key, buddy, dsc.buddiesMax)
 
+  # trace "Running peer worker", peer, nPeers,
+  #   nWorkers=dsc.buddies.len, maxWorkers
```
tersec, line 405: Commented-out code, absent pretty compelling reason, either shouldn't be in the codebase or should be not-commented-out. Otherwise, for example, it bitrots.

nimbus/sync/sync_sched.nim
```diff
@@ -343,8 +431,8 @@ proc onPeerDisconnected[S,W](dsc: RunnerSyncRef[S,W], peer: Peer) =
   else:
     rc.value.worker.ctrl.stopped = true # in case it is hanging somewhere
     dsc.buddies.del peer.key
-    trace "Disconnected buddy", peer, nPeers,
-      nWorkers=dsc.buddies.len, maxWorkers
+    # trace "Disconnected buddy", peer, nPeers,
```
tersec, line 434: ...

nimbus/sync/beacon/worker/db.nim
```diff
@@ -106,12 +119,19 @@ proc dbLoadSyncStateLayout*(ctx: BeaconCtxRef) =
       couplerHash:    latestHash,
       dangling:       latest,
       danglingParent: latestParent,
+      # There is no need to record a separate finalised head `F` as its only
+      # use is to serve as second argument in `forkChoice()` when committing
+      # a batch of imported blocks. Currently, there are no blocks to fetch
```
tersec, line 124: The sync should not be committing batches of anything -- one block at a time to forked chain, just like newpayload/fcu do.

nimbus/config.nim
```diff
@@ -370,6 +370,14 @@ type
       defaultValueDesc: $ProtocolFlag.Eth
       name: "protocols" .}: seq[string]
 
+    beaconSyncReset* {.
```
tersec, line 373: Why is this even an option? If the client restarts, the sync stops. The EL can hold some state in the meantime, sure, but sync is strictly driven by fcU.

In terms of avoiding sometimes inevitable client restarts from hurting sync too badly, the basic approach is that as base is updated, the next sync round does less, because it only has to go back to base. So it is useful to save the partial progress derived from sending blocks to `ForkedChain`, but (a) that's within ForkedChain's domain, not the syncer's, and (b) its not so useful to remember whether it was syncing before.

Also, generally speaking, the EL to function has to be connected to a CL, and the CL will within (typically) within 12 seconds send another fcU. No need for the EL to try to remember whether it was syncing. If it was, and it matters, the CL will ask again very soon.

nimbus/sync/sync_sched.nim
mjfh: ok :)

nimbus/sync/beacon/worker/db.nim
mjfh: This is how it works. Nevertheless blocks are fetched in batches

nimbus/config.nim
mjfh: This means that an interrupted sync cannot be resumed without RPC support?

tersec: Indeed.

mjfh: makes it simpler, then :)

nimbus/sync/beacon/worker.nim
```diff
@@ -150,31 +153,25 @@ proc runPool*(buddy: BeaconBuddyRef; last: bool; laps: int): bool =
...
   true # stop
 
 
-proc runPeer*(buddy: BeaconBuddyRef) {.async.} =
+proc runPeer*(buddy: BeaconBuddyRef; info: static[string]) {.async.} =
   ## This peer worker method is repeatedly invoked (exactly one per peer) while
   ## the `buddy.ctrl.poolMode` flag is set `false`.
   ##
-  const info = "RUNPEER"
   let peer = buddy.peer
```
tersec, line 164: `nimbus-eth1/nimbus/sync/beacon/worker.nim(164, 7) Hint: 'peer' is declared but not used [XDeclaredButNotUsed]`

tersec:
```
nimbus-eth1/nimbus/sync/beacon/worker.nim(18, 28) Warning: imported and not used: 'db' [UnusedImport]
```

nimbus/sync/beacon/worker.nim
mjfh: odd -- was removed in https://github.com/status-im/nimbus-eth1/pull/2782/commits/c32aa499e89dee52a541b54007861825ea5bfe01

PR #2783, opened 2024-10-25
Update networking defaults

arnetheduck:
* prefer IPv6 dual stack address, if available
* use `8551` as default engine api port

### DIFF
--- a/nimbus/config.nim
+++ b/nimbus/config.nim
@@ -18,6 +18,7 @@ import
     net
   ],
   pkg/[
+    chronos/transports/common,
     chronicles,
     confutils,
     confutils/defs,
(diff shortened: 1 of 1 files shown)


nimbus/config.nim
```diff
@@ -85,13 +86,16 @@ const
   defaultPort              = 30303
   defaultMetricsServerPort = 9093
   defaultHttpPort          = 8545
-  defaultEngineApiPort     = 8550
-  defaultListenAddress      = (static parseIpAddress("0.0.0.0"))
+  # https://github.com/ethereum/execution-apis/blob/main/src/engine/authentication.md#jwt-specifications
```
tersec, line 89: Not a stable link and might get reorganized in the repo; one reason to use stable links such as https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/authentication.md#jwt-specifications

PR #2789, opened 2024-10-27
`external_sync` rename to `sync`

advaita-saha:
Also changes the fcu call after each payload to end of epoch

### DIFF
--- a/nrpc/nrpc.nim
+++ b/nrpc/nrpc.nim
@@ -198,13 +198,17 @@ proc syncToEngineApi(conf: NRpcConf) {.async.} =
             Opt.none(PayloadAttributesV1)
           elif consensusFork == ConsensusFork.Capella:
             Opt.none(PayloadAttributesV2)
+          elif consensusFork == ConsensusFork.Deneb or
+            consensusFork == ConsensusFork.Electra:
+            Opt.none(PayloadAttributesV3)
           else:
-            Opt.none(PayloadAttributesV3) # For Deneb
+            doAssert(false, "Unsupported consensus fork")
+            Opt.none(PayloadAttributesV3)
 
       # Make the forkchoiceUpdated call based, after loading attributes based on the consensus fork
       let fcuResponse = await rpcClient.forkchoiceUpdated(state, payloadAttributes)
-      debug "Forkchoice Updated", state = state, response = fcuResponse
-      info "Forkchoice Update Sent", response = fcuResponse.payloadStatus.status
+      debug "forkchoiceUpdated", state = state, response = fcuResponse
+      info "forkchoiceUpdated Request sent", response = fcuResponse.payloadStatus.status
 
   while running and currentBlockNumber < headBlck.header.number:
     var isAvailable = false
@@ -230,7 +234,7 @@ proc syncToEngineApi(conf: NRpcConf) {.async.} =
         when consensusFork <= ConsensusFork.Capella:
           payloadResponse = await rpcClient.newPayload(payload)
           debug "Payload status", response = payloadResponse, payload = payload
-        elif consensusFork >= ConsensusFork.Deneb:
+        elif consensusFork == ConsensusFork.Deneb:
           # Calculate the versioned hashes from the kzg commitments
           let versioned_hashes = mapIt(
             forkyBlck.message.body.blob_kzg_commitments,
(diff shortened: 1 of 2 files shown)


nrpc/nrpc.nim
```diff
@@ -185,6 +185,27 @@ proc syncToEngineApi(conf: NRpcConf) {.async.} =
...
+          safeBlockHash: finalizedHash.asBlockHash,
+          finalizedBlockHash: finalizedHash.asBlockHash,
+        )
+        payloadAttributes =
+          when consensusFork == ConsensusFork.Bellatrix:
+            Opt.none(PayloadAttributesV1)
+          elif consensusFork == ConsensusFork.Capella:
+            Opt.none(PayloadAttributesV2)
+          else:
+            Opt.none(PayloadAttributesV3) # For Deneb
```
tersec, line 202: Useful here to check for specific known forks (Deneb and Electra) which use `PayloadAttributesV3` as an `elif`, and have some compile-time failure (e.g., `static: doAssert false`) on the `else` branch to automatically catch new forks, e.g., Fulu which will be coming in in the next few months to the codebase.

This will continue to compile and run, but start failing without notice, if Fulu uses a `PayloadAttributesV4`.

nrpc/nrpc.nim
```diff
@@ -185,6 +185,27 @@ proc syncToEngineApi(conf: NRpcConf) {.async.} =
...
+
+      # Make the forkchoiceUpdated call based, after loading attributes based on the consensus fork
+      let fcuResponse = await rpcClient.forkchoiceUpdated(state, payloadAttributes)
+      debug "Forkchoice Updated", state = state, response = fcuResponse
```
tersec, line 206: There are probably two main reasonable ways to render fcU as in words:
- using (no spaces) `forkchoiceUpdated` ([`forkchoiceUpdatedV1`](https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/paris.md#engine_forkchoiceupdatedv1), [`forkchoiceUpdatedV2`](https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/shanghai.md#engine_forkchoiceupdatedv2), [`forkchoiceUpdatedV3`](https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/cancun.md#engine_forkchoiceupdatedv3)) and mirroring the engine API's name and capitalization for this call; or
- "fork choice updated", i.e. framing it more colloquially as updating the fork choice system's state, using the same words in the same sequence as the engine API call to make it clear that's the mechanism by which it's happening.

But there's no "forkchoice" which is updated, so the middle ground (no matter the capitalization) of "forkchoice updated" isn't consistent with canonical usage of `forkchoiceUpdated`. When the consensus spec discusses "fork choice", it does it by using it that way, as a non-proper noun without capitalization and with two words:
- https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.8/fork_choice/safe-block.md
- https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.8/specs/phase0/fork-choice.md
- https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.8/specs/bellatrix/fork-choice.md
- https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.8/specs/capella/fork-choice.md
- https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.8/specs/deneb/fork-choice.md

What `nrpc` currently does isn't consistent with either.

nrpc/nrpc.nim
```diff
@@ -263,6 +270,9 @@ proc syncToEngineApi(conf: NRpcConf) {.async.} =
     currentBlockNumber = elBlockNumber()
     (headBlck, _) =
       client.getELBlockFromBeaconChain(BlockIdent.init(BlockIdentType.Head), clConfig)
+      
+  # FCU call for the last remaining payloads
```
tersec, line 274: Typically, the abbreviated version follows the capitalization in the engine API, i.e., **f**ork**c**hoice**U**pdated, or `fcU`.

nrpc/nrpc.nim
```diff
@@ -218,39 +239,25 @@ proc syncToEngineApi(conf: NRpcConf) {.async.} =
...
+        
         # Update the finalized hash
         # This is updated after the fcu call is made
         # So that head - head mod 32 is maintained
         # i.e finalized have to be mod slots per epoch == 0
         let blknum = forkyBlck.message.body.execution_payload.block_number
         if blknum < finalizedBlck.header.number and blknum mod 32 == 0:
           finalizedHash = headHash
+          # Make the forkchoicestate based on the the last
+          # `new_payload` call and the state received from the EL rest api
```
tersec, line 258: "REST" is a quasi-acronym and capitalized as such: https://en.wikipedia.org/wiki/REST

tersec: Also, the EL doesn't have a REST API, the CL does. The EL has a JSON-RPC API.

nrpc/nrpc.nim
tersec: I don't understand the intent here, more broadly: if `blknum >= finalizedBlck.header.number`, `nrpc sync` does not send fcU except once at the end, when it quits. It doesn't seem like operating in the post-finalization part of the chain has to be that much of a special case. That is, why is whether-to-send-fcU a function of `blknum < finalizedBlck.header.number and blknum mod 32 == 0`?

nrpc/nrpc.nim
advaita-saha: Also then should update the newPayload with Electra `execution_requests` 
Nim web3 is not yet updated for that

tersec: Indeed.

`nim-web3` has working Pectra engine API support:
- https://github.com/status-im/nim-web3/blob/6f441a9c1d7a90adca0019d01adec31f6fa35caa/web3/engine_api.nim#L30
- https://github.com/status-im/nim-web3/blob/6f441a9c1d7a90adca0019d01adec31f6fa35caa/web3/engine_api.nim#L55-L58

`nimbus-eth2` uses it today for `devnet-4`.

advaita-saha: `newPayload()` update is missing though 
ends at deneb 
https://github.com/status-im/nim-web3/blob/6f441a9c1d7a90adca0019d01adec31f6fa35caa/web3/engine_api.nim#L120-L126

I have added the electra support in `nrpc`, but having this api set complete in `nim-web3` would be nice

advaita-saha: Can be done later in another PR though

tersec: I've never been that sure of the value of those "convenience APIs", because while `newPayloadV1` and `newPayloadV2` at least share the argument count, so it makes sense to treat them as overloadable functions based on whether `payload` is `ExecutionPayloadV1` or `ExecutionPayloadV2`, `newPayloadV3` adds a parameter so can't be called the same way anyway:
```nim
template newPayload*(
    rpcClient: RpcClient,
    payload: ExecutionPayloadV3,
    versionedHashes: seq[VersionedHash],
    parentBeaconBlockRoot: Hash32): Future[PayloadStatusV1] =
  engine_newPayloadV3(
    rpcClient, payload, versionedHashes, parentBeaconBlockRoot)
```
and a hypothetical version for `engine_newPayloadV4` would have to add yet another parameter. These aren't straightforwardly generic functions that one just swaps in different execution payload types to, so at least my updates to `nim-web3` don't touch those convenience APIs, and I have no intention of adding the V4 version, because see above.

If someone else wants to, that's fine, but I still don't see the value of it.

tersec: > Can be done later in another PR though

This is the important part -- none of this should hold up this PR.

nrpc/nrpc.nim
```diff
@@ -185,6 +185,31 @@ proc syncToEngineApi(conf: NRpcConf) {.async.} =
...
+        payloadAttributes =
+          when consensusFork == ConsensusFork.Bellatrix:
+            Opt.none(PayloadAttributesV1)
+          elif consensusFork == ConsensusFork.Capella:
+            Opt.none(PayloadAttributesV2)
+          elif consensusFork == ConsensusFork.Deneb or
+            consensusFork == ConsensusFork.Electra:
+            Opt.none(PayloadAttributesV3)
+          else:
+            doAssert(false, "Unsupported consensus fork")
```
tersec, line 205: The `static:` part is important here, because it makes the error message compile-time.

PR #2793, opened 2024-10-28
rm --protocol CLI flag

tersec:
Without `snap`, the only options are `Eth` being present or absent, which controls two things:
- the RPC having the basic `eth_` methods; and
- Nimbus being intialized with a txpool or not.

For the former, that means
https://github.com/status-im/nimbus-eth1/blob/ba1cbed14fea101e1faea37ea5d312fa2c7c81fb/nimbus/rpc/common.nim#L34-L74
and it doesn't seem worth having this option to have a JSON-RPC server which can do some SHA3 hashing on demand. If one wants to disable the RPC servers entirely, there are already options for that.

For the latter the txpool doesn't really do anything unless requested specifically, so it also seems dubiously worthwhile.

### DIFF
--- a/nimbus/config.nim
+++ b/nimbus/config.nim
@@ -117,11 +117,6 @@ type
     noCommand
     `import`
 
-  ProtocolFlag* {.pure.} = enum
-    ## Protocol flags
-    Eth                           ## enable eth subprotocol
-    #Snap                          ## enable snap sub-protocol
-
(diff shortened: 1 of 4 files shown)


arnetheduck:
mmyeah, sgtm, the ux of enabling and disabling features probably shouldn't hingen on specific eth protocols that enable them

arnetheduck (approved):

PR #2802, opened 2024-10-30
Fluffy: Improve logging and add offer metrics to status logs.

bhartnett:
Set offer processing logs to debug level and add offer counts to state network.

### DIFF
--- a/fluffy/network/state/state_network.nim
+++ b/fluffy/network/state/state_network.nim
@@ -33,6 +33,9 @@ type StateNetwork* = ref object
   historyNetwork: Opt[HistoryNetwork]
   validateStateIsCanonical: bool
   contentRequestRetries: int
+  offersTotalCount: int
+  offersSuccessCount: int
+  offersFailedCount: int
 
 func toContentIdHandler(contentKey: ContentKeyByteList): results.Opt[ContentId] =
   ok(toContentId(contentKey))
@@ -70,6 +73,9 @@ proc new*(
     historyNetwork: historyNetwork,
     validateStateIsCanonical: validateStateIsCanonical,
     contentRequestRetries: contentRequestRetries,
+    offersTotalCount: 0,
+    offersSuccessCount: 0,
+    offersFailedCount: 0,
   )
 
 proc getContent(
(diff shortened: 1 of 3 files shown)


fluffy/network/state/state_network.nim
```diff
@@ -33,6 +33,9 @@ type StateNetwork* = ref object
   historyNetwork: Opt[HistoryNetwork]
   validateStateIsCanonical: bool
   contentRequestRetries: int
+  offersTotalCount: int
+  offersSuccessCount: int
+  offersFailedCount: int
```
bhartnett, line 38: The goal here is to count the number of offers processed, successes and failures. Should I be using `declareCounter` for this instead? Or is that only for the metrics server. I need to be able to return the counts in the logs.

kdeme: I believe this to be information mostly (only?) useful to developers?  I don't think we can expect a default user to understand the health of the node from this information.

For counters useful to developers we should be using metrics indeed. They are available then on the metrics server. You could still log them also (see https://github.com/status-im/nim-metrics?tab=readme-ov-file#logging), although I'm not sure if that is really needed.

By the way, we already have related offer metrics in the portal_protocol. The difference is that those are not aware of the validation failures/success that occurs here.

bhartnett: > I believe this to be information mostly (only?) useful to developers? I don't think we can expect a default user to understand the health of the node from this information.

I'll remove this from the info logs in that case.

> For counters useful to developers we should be using metrics indeed. They are available then on the metrics server. You could still log them also (see https://github.com/status-im/nim-metrics?tab=readme-ov-file#logging), although I'm not sure if that is really needed.

Good to know, thanks.

> By the way, we already have related offer metrics in the portal_protocol. The difference is that those are not aware of the validation failures/success that occurs here.

Yes I'm aware of the portal_protocol metrics but I wanted to see the count of offers failed/passed validation. This was driven mainly by seeing something similar in the Trin logs. 

It might be useful to log the more important metrics for each subnetwork at debug level. If one doesn't have the metrics infrastructure setup it provides a quick way to get an overview of what is happenning.

PR #2804, opened 2024-10-30
cl+el poc[WiP]

pedromiguelmiranda:
This is a work in progress for a cl+el single application.

**Some notes:**

**current state:**
beacon node metrics strategy ? issue with threads.
fetchgenesis mismatch error: 

~~comments with:` # NOTE, #TODO,` are related to improvements or refactorings that need  to be done
comments with: `# adapted from nimbus-eth2`, things that were required to be overridden or not used (or we think that is not used)
Probably the best course of action will be to create issues regarding it.
**current state:** able to synch (or it appears so) with two minor issues (check latest commit message)~~

image


nimbus_unified/consensus/adapted_block_chain_dag.nim
tersec, line 1: Better to refactor `block_chain_dag` in `nimbus-eth2` so it can be useful for both than copy/paste 2900 lines of code.

nimbus_unified/run_nimbus_unified.sh
tersec, line 1: It's not clear these wrapper shells scripts are useful. `nimbus-eth2` retains them for the sake of compatibility with existing installations because https://nimbus.guide/ refers to them, but potentially not worth introducing them in a context which has never had/needed/required them.

nimbus_unified/consensus/adapted_block_chain_dag.nim
arnetheduck: I'd expect a stream of small improvements in all projects before we land this, so as long as this is just a wip/pr, it's not really an issue

pedromiguelmiranda: You are both right. 

For now, the code has a lot of `Todo's`, `Note's` and `Adapted from eth2`, which can be translated to: Can and will need refactor. `Blockchain_dag` is one of them, but `nimbus_beacon_node` will require a lot of removals and improvements (first reason: declares gauges that metrics library does not like when it is declared on a thread).

Personally, I would tackle these improvements when we have a more "advanced" running software.

nimbus_unified/run_nimbus_unified.sh
pedromiguelmiranda: I agree, and due to your comment I now know why they are kept on nimbus-eth2.
For now it is useful for development purposes (anyone can quickly spin the application) but don't plan on merging them

nimbus_unified/run_nimbus_unified.sh
```diff
@@ -0,0 +1,44 @@
...
+--data-dir="build/data/shared_holesky_0" \
+--tcp-port=9000 \
+--udp-port=9000 \
+--rest \
+--rest-port=5052 \
+--metrics \
+"$@"
+# --el=http://127.0.0.1:8551 \
+# --jwt-secret="/tmp/jwtsecret" \
+# --web3-url=http://127.0.0.1:8551
```
tersec, line 27: remove this file

.gitmodules
```diff
@@ -241,3 +241,23 @@
 	path = vendor/nim-minilru
 	url = https://github.com/status-im/nim-minilru.git
 	branch = master
+[submodule "vendor/mainnet"]
```
tersec, line 244: split out submodule bumps

nimbus/nimbus_execution_client.nim
```diff
@@ -173,35 +173,33 @@ proc preventLoadingDataDirForTheWrongNetwork(db: CoreDbRef; conf: NimbusConf) =
...
-proc run(nimbus: NimbusNode, conf: NimbusConf) =
-  ## logging
-  setLogLevel(conf.logLevel)
-  if conf.logFile.isSome:
-    let logFile = string conf.logFile.get()
-    defaultChroniclesStream.output.outFile = nil # to avoid closing stdout
-    discard defaultChroniclesStream.output.open(logFile, fmAppend)
+proc run*(nimbus: NimbusNode, conf: NimbusConf) =
+  # ## logging
+  # setLogLevel(conf.logLevel)
```
tersec, line 178: don't randomly remove working EL code

nimbus_unified/version.nim
```diff
@@ -0,0 +1,62 @@
...
+    if line.startsWith(gitPrefix) and line.len > 8 + gitPrefix.len:
+      result = line[gitPrefix.len ..< gitPrefix.len + 8]
+
+# TODO: Currently prefixing these metric names as the non prefixed names give
+# a collector already registered conflict at runtime. This is due to the same
+# names in nimbus-eth2 nimbus_binary_common.nim even though there are no direct
+# imports of that file.
+
+declareGauge versionGauge,
+  "Nimbus version info (as metric labels)",
```
tersec, line 55: disables metrics elsewhere?

PR #2805, opened 2024-10-30
Add custom json-rpc method to initialize the trusted block root

kdeme:


fluffy/network/beacon/beacon_network.nim
```diff
@@ -357,7 +357,7 @@ proc validateContent(
 
       debug "Received offered content validated successfully", contentKey
     else:
-      error "Received offered content failed validation",
+      debug "Received offered content failed validation",
```
bhartnett, line 360: @kdeme Was this a mistake? I would have thought that failed validations should be logged at error log level.

kdeme: No, it was intentional. 

I can understand that these are still useful to have around at `error` or `warning` for devs for showing incompatibilities, but in the end they are not that useful for users. More so as we get to a stage where the different clients are/should be compatible, which means these fails would be cause due to malicious nodes or some old not updated lingering nodes.

In this specific case of the beacon network however the `error` is even worse, as currently this will also fail validation when an already applied (but valid) optimistic/finality update is received.

bhartnett: Ok, good to know. 

Yes I think the ban list of some sort would be a good idea. Perhaps we can look at something similar to what other EL clients already use in devp2p.

kdeme: We do this already in nimbus-eth2, it is nothing special, a table that tracks which nodes have been seen and a timer on how long they end up in there based on a reason.

See: https://github.com/status-im/nimbus-eth2/blob/stable/beacon_chain/networking/eth2_network.nim#L76 and https://github.com/status-im/nimbus-eth2/blob/stable/beacon_chain/networking/eth2_network.nim#L235-L251

PR #2808, opened 2024-10-30
Fix flaky test_portal_testnet CI test with retries

kdeme:


kdeme:
5 re-runs without issues on the portal testnet tests, thanks to added retry.

test_history_network can still fail though so removed that and will do a more proper fix.

PR #2819, opened 2024-11-01
holesky: fix timestamp

arnetheduck:


arnetheduck:
Originally I thought our timestamp was off but it looks correct after all as the new logging helps discover - turns out the holesky [readme](https://github.com/eth-clients/holesky/issues/115) is wrong instead.

Issue #1086, opened 2022-05-12
Split up content databases (kvstores) per network


kdeme:
Currently the quickest, simplest approach is taken and all is stored in one table / kvstore. However, this will not be scalable once we are dealing with lots of data.
This issue is about how to split this storage, basically this comment: https://github.com/status-im/nimbus-eth1/blob/master/fluffy/content_db.nim#L24

I think approach 1. mentioned there is probably the most straightforward path to take, but some investigation to better understand the implications of (the) other approaches is allowed ;-).

KonradStaniec:
I thought about it a bit and I like option `3` the most tbh. My reasons for it:
- having db per network make it possible to delegate handling of radius of each network to its database code which makes code a bit nicer.
- having db per network discourages sharing data thorough database, and encourages sharing data between networks through well defined interfaces.
- having one database make pruning code a bit more tricky, as then we need to how much data each table is taking which probably force us to keep more metadata in db. 

Disadvantages are:
- we have more files on disk (for 5 networks it will be 5 files, so imo it is not a big deal)
- to known how much data we store we need to call `db.size` few times and sum the values. Although it is rather not a big problem, as db limits should be defined per network basis.

kdeme:
FYI, similar (albeit not the same) technical question: https://github.com/status-im/nimbus-eth2/blob/039bece9175104b5c87a8c2ff6b1eafae731b05e/beacon_chain/validators/slashing_protection_v2.nim#L119

KonradStaniec:
> Not sure if I'd opt for different Radius handling per network (At least not if it makes things much more complex), as I'm not sure there is a need for this?

It maybe more complex to have  one more global radius, as different networks have different sizes so to adjust global radius we would need to take it into account somehow not to monopolize node storage by one type of data. Where in radius per network we keep the same size proportional logic everywhere.

> Anyway, these are exactly the things that need to be sketched out better I think before we make too many changes.

Having db per network incurs probably least changes for now (as we have one working network) it is just the question of initiating db in history network constructor instead of fluffy main. Where with multiple kvstores we would also need to update queries and calculating sizes of db. Configs would need to be updated in both approaches as in both of them user should configure different sizes for different networks. (at least I think so) 

I wonder, maybe we shoould delay making the decision until having another network, and some endpoint which get data from both of them, then implement some proof of concept for both approaches and then see which one we like more  ?

kdeme:
[...]

PR #2827, opened 2024-11-04
Fluffy: Implement offer cache

bhartnett:
I did some quick performance testing and confirmed that looking up data from a cache is always faster than even using contains on a small sqlite db (with the current default settings). 

This PR introduces a cache that stores the most recently accepted offers. This will be useful to speed up checking for existing offers which often happens during the gossip process as each node may receive offers for the same content from multiple peers. It will also help reduce the load on the database and increase lookup performance in general and will help speed up the state bridge gossip performance.

We may not need to split up the content databases as is suggested in this [issue](https://github.com/status-im/nimbus-eth1/issues/1086) if we use caching as an alternative.


fluffy/network/wire/portal_protocol.nim
```diff
@@ -161,8 +167,16 @@ type
...
 
+  # Caches content fetched from the network during lookups.
+  # Content outside our radius is also cached in order to improve performance
+  # of queries which may lookup data outside our radius.
   ContentCache = LruCache[ContentId, seq[byte]]
 
+  # Caches the most recently received content/offers.
+  # Content is only stored in this cache if it falls within our radius and similarly
+  # the cache is only checked if the content id is within our radius.
+  OfferCache = LruCache[ContentId, seq[byte]]
```
kdeme, line 178: I initially thought of this as a cache that only holds the content ids or the content keys (and thus could store more of them for the same data size). The cache would only be there to avoid hammering the database on a spam of offers.

After checking the PR I understand that it is similar as the `ContentCache`, in the sense that it is also used to check for data on requests. But with the difference that it only gets stored via the offer/accept flow (and json-rpc store method). This was not immediately clear to me.

In general the `ContentCache` is more useful because it foremost avoids re-doing recent network requests. This is something that `OfferCache` does not avoid.

bhartnett: Yes I remember we discussed only storing the content keys/ids and mapping to a bool in order to save space. I'm open to doing it this way which does provide some benefit for rejecting existing offers quickly but I believe that caching the recently received offers provides some other benefits because the cached offers are reusable in the other flows. 

The Sqlite in memory caching is limited and having some form of in memory caching per sub-network that is not specifically tied to recent queries will be useful. For example once we have the network fully synced and following the latest blocks, the most recently received offered content is likely to be looked up more frequently during normal usage of portal network by applications.

fluffy/network/beacon/beacon_network.nim
```diff
@@ -354,7 +354,9 @@ proc validateContent(
         return false
 
       let contentId = contentIdOpt.get()
-      n.portalProtocol.storeContent(contentKey, contentId, contentItem)
+      n.portalProtocol.storeContent(
+        contentKey, contentId, contentItem, cacheOffer = true
+      )
```
bhartnett, line 359: I wasn't sure about whether to cache in the store call. It may not be necessary and might be better to only cache offers received over the network.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -420,25 +484,21 @@ proc handleFindContent(
...
-  if p.inRange(contentId):
-    let contentResult = p.dbGet(fc.contentKey, contentId)
-    if contentResult.isOk():
-      let content = contentResult.get()
-      if content.len <= maxPayloadSize:
-        return encodeMessage(
-          ContentMessage(
-            contentMessageType: contentType, content: ByteList[2048](content)
-          )
+  let contentResult = p.getLocalContent(fc.contentKey, contentId)
```
bhartnett, line 487: When handling findContent messages we should return the content from the cache if available. When following the latest blocks, we likely will see more of these queries looking up the latest received offers.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -955,7 +1020,7 @@ proc offer(
           if contentIdResult.isOk():
             let
               contentId = contentIdResult.get()
-              contentResult = p.dbGet(contentKey, contentId)
+              contentResult = p.getLocalContent(contentKey, contentId)
```
bhartnett, line 1023: In the case when offering content via the database, the cache should be checked. The only downside I can think of is that it would influence the order of the content cache which may not be desired. It might be better to just check the offer cache and not the content cache, not sure about this yet.

fluffy/network/wire/portal_protocol.nim
bhartnett: Compare the in memory caching of Sqlite to RocksDb:
- Sqlite only uses the os page cache
- RocksDb uses the os page cache, the block cache (cache for recent reads) and MemTable (cache of recent writes)

For RocksDb the block cache improves the performance of recent reads and the MemTable improves the performance of reading recently written values (this is not the only purpose of the MemTable but one of its benefits). For the caching in Fluffy I was thinking we should have something similar in purpose to the RocksDb MemTable where we cache recent writes (we can keep the default size small to avoid increased memory usage). 

For Fluffy we have a single database/table shared across multiple logically isolated sub-networks for which the data doesn't normally need to be queried together which ideally should be stored in separate tables (but that leads to compilations with pruning). With the offer cache we at least have a separate write cache per subnetwork to alleviate load on the shared database.

My motivation for this change was partly because I noticed that the performance of returning a value from a large fully populated LRU cache is dramatically faster than even doing a contains check on a small Sqlite db.

PR #2829, opened 2024-11-04
Remove two simple sleeps from history network test

kdeme:
Replace simple sleeps in test_history_network with retries + sleeps. Current simple sleep setting would occasionally still fail on CI.


kdeme:
Have had 4x2 successful runs on CI with this version now.

PR #2863, opened 2024-11-22
Add missing fields to RPC object conversion

jangko:


advaita-saha:
I think it is better to keep 1 helper file instead of 2. Let's keep a general utils `rpc_utils` for now and remove `server_api_helpers` ( shift the functions inside to `rpc_utils` ) as we have dropped support for old p2p endpoints and only `server_api` exists now

PR #2865, opened 2024-11-22
Inline gas cost/instruction fetching

arnetheduck:
These make up 5:ish % of EVM execution time - even though they're trivial they end up not being inlined - this little change gives a practically free perf boost ;)

Also unify the style of creating the output to `setLen`..

### DIFF
--- a/nimbus/evm/code_stream.nim
+++ b/nimbus/evm/code_stream.nim
@@ -53,14 +53,19 @@ func readVmWord*(c: var CodeStream, n: static int): UInt256 =
 func len*(c: CodeStream): int =
   len(c.code)
 
-func next*(c: var CodeStream): Op {.inline.} =
-  # The extra >= 0 check helps eliminate `IndexDefect` from the optimized code
-  # which keeps this hotspot in the EVM small, code-size-wise
-  if c.pc >= 0 and c.pc < c.code.len:
-    result = Op(c.code.bytes[c.pc])
-    inc c.pc
+template next*(c: var CodeStream): Op =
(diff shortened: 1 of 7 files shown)


jangko:
need to fix test vector `tools/t8n/testdata/00-519/exp.txt`

```
diff --git a/tools/t8n/testdata/00-519/exp.txt b/tools/t8n/testdata/00-519/exp.txt
index d9509a336..253d4518b 100644
--- a/tools/t8n/testdata/00-519/exp.txt
+++ b/tools/t8n/testdata/00-519/exp.txt
@@ -1,2 +1,2 @@
 {"pc":0,"op":0,"gas":"0x0","gasCost":"0xfffffffffffecb68","memSize":0,"stack":[],"depth":1,"refund":0,"opName":"STOP","error":"PrcInvalidParam"}
-{"output":"","gasUsed":"0x13498","error":"PrcInvalidParam"}
+{"output":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","gasUsed":"0x13498","error":"PrcInvalidParam"}
```

jangko:
? doesn't it then make more sense to exclude the output from the "correctness" comparison entirely?

No, the tracer not only produce output from precompiles, but from the EVM. The above case happen because the last thing executed by the EVM is the precompile with invalid params. It is correct to check the output of the EVM. Whay I try to say is the EVM still correct even though the blake2f precompile behavior now a bit different.

arnetheduck:
https://github.com/status-im/nimbus-eth1/pull/2865/commits/abf3d736dbf5fc2b58296d8d024562b950f37f95

PR #2873, opened 2024-11-26
Fix multiple crashes due to doassert

advaita-saha:
Multiple crashes faced in holesky due to this assertion mainly when the node falls out of sync


advaita-saha:
> Which `doAssert`?

https://github.com/status-im/nimbus-eth1/blob/529921c93cf89a995ae35f5622bf675c0a0f6dae/nimbus/rpc/filters.nim#L20-L24

Issue #2874, opened 2024-11-26
Limit concurrent offers that can be received from each peer in Fluffy


bhartnett:
In order to reduce the impact of certain DOS attacks we should put limits in place to reduce the impact on Fluffy nodes.

We should enforce the following:
- For each content key/content id we can only receive one concurrent offer from each peer/enr for that content
- Limit the total number of offers that can be received concurrently from each peer/enr. This limit will be configurable as a debug parameter.

Nodes that hit the total number of offers limit will be added to the temporary ban list.

kdeme:
Yes, I agree that these would be two good limits to have enabled.

bhartnett:
@kdeme Do you think that content lookups should be included in the limits? 

In other words, when a node asks a peer for content to download it should be limited to one concurrent download/upload per contentId and limited to n concurrent downloads/uploads in total. When I say download I mean requesting content from a peer and when I say upload I mean offering content to a peer.

bhartnett:
Should we also limit the number of concurrent pings and findNodes requests which we respond to?

kdeme:
> Do you think that content lookups should be included in the limits?

Not fully sure on this one. It seems more like the responsibility of the application not to do such things. I mean, it's in the hands of the local node and not its peers for this to occur. But maybe there are situations where this can anyhow happen, although probably rare.

bhartnett:
> Not fully sure on this one. It seems more like the responsibility of the application not to do such things. I mean, it's in the hands of the local node and not its peers for this to occur. But maybe there are situations where this can anyhow happen, although probably rare.

What I meant is limiting the number of content lookups which we handle/respond to. My reasoning for applying the limit here as well is that we need to open a utp stream and so it seems reasonable to prevent a single peer requesting too much content. I've implemented this already but happy to revert if we don't want to go this way. See here: https://github.com/status-im/nimbus-eth1/pull/2885/commits/a96a134ea316b2aa833ce43cf7d3ab5315a4668e

kdeme:
Oh, yeah. I misunderstood as I already noticed that feature in your PR.

So yes, that definitely sounds good as a limit protection to have considering it is incoming from other peers.

bhartnett:
The changes for this task were implemented in this PR: https://github.com/status-im/nimbus-eth1/pull/2885

PR #2885, opened 2024-11-28
Fluffy: Limit concurrent offers that can be received from each peer

bhartnett:
Related to this issue: https://github.com/status-im/nimbus-eth1/issues/2874

We currently just limit transfers of offers per stream/subnetwork. This means transfers on one subnetwork won't effect the other subnetworks.

The limit is per offer so sending 2 batches of 64 offers will use up 128 of the quota for a peer.

I'm thinking to also limit the content lookup responses which will also share the pending transfers quota.

### DIFF
--- a/fluffy/network/wire/portal_protocol.nim
+++ b/fluffy/network/wire/portal_protocol.nim
@@ -446,8 +446,10 @@ proc handleOffer(p: PortalProtocol, o: OfferMessage, srcId: NodeId): seq[byte] =
       )
     )
 
-  var contentKeysBitList = ContentKeysBitList.init(o.contentKeys.len)
-  var contentKeys = ContentKeysList.init(@[])
+  var
+    contentKeysBitList = ContentKeysBitList.init(o.contentKeys.len)
+    contentKeys = ContentKeysList.init(@[])
+    contentIds = newSeq[ContentId]()
   # TODO: Do we need some protection against a peer offering lots (64x) of
   # content that fits our Radius but is actually bogus?
   # Additional TODO, but more of a specification clarification: What if we don't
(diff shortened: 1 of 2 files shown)


fluffy/network/wire/portal_protocol.nim
```diff
@@ -463,17 +465,19 @@ proc handleOffer(p: PortalProtocol, o: OfferMessage, srcId: NodeId): seq[byte] =
...
-      if p.inRange(contentId):
-        if not p.dbContains(contentKey, contentId):
-          contentKeysBitList.setBit(i)
-          discard contentKeys.add(contentKey)
+      if p.inRange(contentId) and p.stream.canAddPendingTransfer(srcId, contentId) and
+          not p.dbContains(contentKey, contentId):
+        p.stream.addPendingTransfer(srcId, contentId)
+        contentKeysBitList.setBit(i)
+        discard contentKeys.add(contentKey)
+        contentIds.add(contentId)
```
bhartnett, line 473: With this implementation we can accept a subset of the offers up to the pending transfers limit. An alternative way to implement it would be to check if the full batch of offers will fit and either accept or reject all. Not sure if this is preferred.

fluffy/network/wire/portal_stream.nim
```diff
@@ -35,6 +35,7 @@ const
   talkReqOverhead = getTalkReqOverhead(utpProtocolId)
   utpHeaderOverhead = 20
   maxUtpPayloadSize = maxDiscv5PacketSize - talkReqOverhead - utpHeaderOverhead
+  maxPendingTransfersPerPeer = 128
```
bhartnett, line 38: This means we can accept 128 concurrent offers (or 2 batches of 64) from each peer for each subnetwork.

fluffy/network/wire/portal_protocol.nim
kdeme: I think this version is fine

fluffy/network/wire/portal_stream.nim
kdeme: Probably ok as a value. 

It is also not possible to get 64  content keys for most of the content key types as the `len` of the total discv5 message would be to large. I think it is typically maxed at 32 values or so.

fluffy/network/wire/portal_stream.nim
```diff
@@ -69,28 +72,94 @@ type
...
+    transfers: TableRef[NodeId, HashSet[ContentId]],
+    nodeId: NodeId,
+    contentId: ContentId,
+    limit: int,
+): bool =
+  if not transfers.contains(nodeId):
+    return true
+
+  try:
+    (transfers[nodeId].len() < limit) and not transfers[nodeId].contains(contentId)
```
kdeme, line 93: Having 3x a lookup in the table here is not that great, but I guess it is fine as the table itself should never become large.

That is, for incoming offers a `contentQueue.full` check is done (so that's max 50), but this did make me notice that for incoming findContent we do not have a total limit currently.

kdeme (approved):
I think this is fine in general.

I do think we should  write somewhere down all the limits / DoS measures taken in combination, and see how that all fits together. We might have to revise some of the work after that.

bhartnett:
> I think this is fine in general.
> 
> I do think we should write somewhere down all the limits / DoS measures taken in combination, and see how that all fits together. We might have to revise some of the work after that.

Where would you suggest we put this write up? In an issue with a todo list?

fluffy/network/wire/portal_stream.nim
bhartnett: I'll improve this by storing the result of `transfers[nodeId]` in a temporary variable.

PR #2916, opened 2024-12-06
Fix race condition on syncer termination

mjfh:


nimbus/sync/sync_sched.nim
```diff
@@ -174,14 +174,13 @@ proc terminate[S,W](dsc: RunnerSyncRef[S,W]) =
...
-          try:
-            waitFor sleepAsync termWaitPollingTime
-          except CancelledError:
-            trace "Shutdown: peer timeout was cancelled",
-              peer=w.data.worker.peer, nWorkers=dsc.buddies.len
         else:
           dsc.buddies.del w.key # this is OK to delete
+      # Activate async jobs so they can finish
+      try:
+        waitFor sleepAsync termWaitPollingTime
```
arnetheduck, line 181: in general, you should not ever call `waitFor` outside of the top-level main function - it is not reentrant and this will cause bugs and issues - the solution is to turn this function into an `async` function.

PR #2917, opened 2024-12-06
Fluffy: Fix broken portal hive tests

bhartnett:
Fix bug in portal stream where connection id was not correctly generated when handling requests from peers.

### DIFF
--- a/fluffy/network/wire/portal_stream.nim
+++ b/fluffy/network/wire/portal_stream.nim
@@ -190,7 +190,8 @@ proc addContentOffer*(
   var id = ConnectionId.fromBytesBE(connectionId)
 
   # Generate a new id if already existing to avoid using a duplicate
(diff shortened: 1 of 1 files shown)


fluffy/network/wire/portal_stream.nim
```diff
@@ -212,12 +213,14 @@ proc addContentRequest*(
   # TODO: Should we check if `NodeId` & `connectionId` combo already exists?
   # What happens if we get duplicates?
   var connectionId: Bytes2
+  stream.rng[].generate(connectionId)
```
bhartnett, line 216: FYI @kdeme this was removed by mistake in my last PR. It caused the connectionId to default to 0 when handling requests. I find it surprising that most of the tests still passed. By the way, is zero a valid connectionId value?

PR #2919, opened 2024-12-09
Fix Blake2b precompile regression

jangko:
Introduced by #2865
Detected on Holesky block 2.406.802 tx no 11
And on MainNet block 19.633.393


arnetheduck:
Interesting find - though it raises the question: should the output always be cleared for all precompiles? is this not part of the spec? why are the differences allowed?

jangko:
should the output always be cleared for all precompiles? no
is this not part of the spec? not part of the spec
why are the differences allowed? the most probable explanation is the behavior came from the reference implementation.

PR #2925, opened 2024-12-11
Enable nimbus execution client to build on RISC-V boards

haurog:
To build the nimbus execution client on RISC-V boards, the ISA instruction set supported needs to be specified in the build instructions. A minimal version to build is to define the 'gc' extensions. In the future we might have to add the 'v' extension as well, but many boards do not support this extension yet. These changes are the same that were necessary for the nimbus consensus client: https://github.com/status-im/nimbus-eth2/pull/6439

It builds on my LicheePi 3a using these settings. The binary does start and seems to be working fine. I have not tested to sync an actual testnet or even mainnet with it yet. Will try to do so later.

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -102,6 +102,11 @@ elif defined(macosx) and defined(arm64):
   # Apple's Clang can't handle "-march=native" on M1: https://github.com/status-im/nimbus-eth2/issues/2758
   switch("passC", "-mcpu=apple-a14")
   switch("passL", "-mcpu=apple-a14")
+elif defined(riscv64):
+  # riscv64 needs specification of ISA with extensions. 'gc' is widely supported 
+  # and seems to be the minimum extensions needed to build. 
+  switch("passC", "-march=rv64gc")
+  switch("passL", "-march=rv64gc")
 else:
   switch("passC", "-march=native")
   switch("passL", "-march=native")


cmd0s:
Hi,

I compiled and successfully ran the Nimbus execution client on a Banana Pi F3. I paired it with the Nimbus consensus client running on a Raspberry Pi 5.

They are currently in the synchronization process, and everything looks promising so far. The logs on the console look good, and I can see the disk usage steadily increasing, which indicates that the data is being downloaded correctly. I'll let you know once the synchronization is complete.

Here’s the command I used to start Nimbus Eth1:
`nimbus_execution_client --network=mainnet --data-dir=/mnt/storage/nimbus_eth1 --jwt-secret=/root/jwt.hex --http-port=8545 --http-address=0.0.0.0 --rpc=true --ws=true --allowed-origins=* --tcp-port=7771 --udp-port=7771 --engine-api --engine-api=true --engine-api-address=0.0.0.0`

Nimbus_eth1_first_run

image

cmd0s:
During synchronization, I noticed some interesting log entries containing "TODO". 
I understand that this is software still in development :)
image

Below, I’m also attaching a screenshot from the Grafana dashboard used in the Web3 Pi project, where you can see CPU load, RAM usage, and disk usage during the first hour of operation.

image

cmd0s:
Unfortunately, the synchronization is not going as I expected.

The Nimbus consensus client on the RPi5 is synchronized now, but the Nimbus execution client cannot sync. I can see strange cycles in disk usage.

Perhaps the Nimbus dev team has some helpful insights?

image

I think I will first run it on the Raspberry Pi 5, and if it works there, I will move to RISC-V.

arnetheduck:
Really nice to see this in the wild!

>  I understand that this is software still in development :)

Indeed!

> I can see strange cycles in disk usage.

This is normal, due to how rocksdb works where data is "recompacted" when the changes add up to a large-enough batch.

Sync is indeed one of the areas where there are still frequent improvements landing - it should get more stable over time.

In the meantime, https://nimbus.guide/execution-client.html#top-up-blocks-from-the-consensus-node can help keep the node in sync.

I'll assume you did the initial import via era files or maybe a copy of the database from somewhere?

cmd0s:
I didn't do an initial import. My disks are already occupied from the start because other clients were previously running on them. However, they are now turned off, and their data is stored in separate directories - they just take up space :)

The only thing I did was start the Nimbus consensus client in TrustedNodeSync mode.

PR #2926, opened 2024-12-11
Move EVM code initialization outside of newComputation

### LINKED ISSUE #2883: Resolve EVM code outside of newComputation
From Discord chat:

https://github.com/status-im/nimbus-eth1/pull/2881 reduces seq overhead in EVM - that said, @jangko the design where newComputation sometimes does database lookups feels pretty hairy - can't that be moved somewhere else? it feels like unnecessary maintenance overhead to do the same thing in two places
[20.23]
basically, for precompiles in particular, we don't execution evm opcodes so correspondingly we don't need a stack/memory - the natural way to solve that would be that the code is always resolved "outside" of newComputation and if code is nil, we skip stack/mem allocation instead of that ugly flag..

this also would help with  the logic of reducing database lookups for precompiles - the database check is unnecessary (and expensive) - the PR introduces an ugly hack to fix this but ...

jangko:
fix #2883


nimbus/transaction/call_common.nim
```diff
@@ -165,56 +163,28 @@ proc setupHost(call: CallParams, keepStack: bool): TransactionHost =
...
-        host.msg.input_data = host.input[0].addr
-
-    let
-      cMsg = hostToComputationMessage(host.msg)
-    host.computation = newComputation(
-      vmState, call.sysCall, cMsg, code, isPrecompile = isPrecompile, keepStack = keepStack)
-    host.code = code
 
+  var code: CodeBytesRef
+  if call.isCreate:
```
arnetheduck, line 168: `let code = if ...` is generally nicer in situations like this

arnetheduck (approved):

PR #2937, opened 2024-12-13
Stream blocks during import

arnetheduck:
When running the import, currently blocks are loaded in batches into a `seq` then passed to the importer as such.

In reality, blocks are still processed one by one, so the batching does not offer any performance advantage. It does however require that the client wastes memory, up to several GB, on the block sequence while they're waiting to be processed.

This PR introduces a persister that accepts these potentially large blocks one by one and at the same time removes a number of redundant / unnecessary copies, assignments and resets that were slowing down the import process in general.

### DIFF
--- a/nimbus/core/chain/persist_blocks.nim
+++ b/nimbus/core/chain/persist_blocks.nim
@@ -11,6 +11,7 @@
 {.push raises: [].}
 
 import
+  stew/assign2,
   results,
   ../../evm/state,
   ../../evm/types,
@@ -29,8 +30,8 @@ export results
 
 type
   PersistBlockFlag* = enum
-    NoValidation # Validate the batch instead of validating each block in it
-    NoFullValidation # Validate the batch instead of validating each block in it
+    NoValidation # Disable chunk state root validation
+    NoFullValidation # Disable per-block validation
     NoPersistHeader
     NoPersistTransactions
     NoPersistUncles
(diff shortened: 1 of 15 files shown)


nimbus/nimbus_import.nim
```diff
@@ -33,6 +33,60 @@ declareCounter nec_imported_gas, "Gas processed during import"
...
+  except IOError as exc:
+    fatal "Could not open statistics output file", file = name, err = exc.msg
+    quit(QuitFailure)
+
+proc getMetadata(networkId: NetworkId): auto =
+  # Network Specific Configurations
+  # TODO: the merge block number could be fetched from the era1 file instead,
+  #       specially if the accumulator is added to the chain metadata
+  case networkId
+  of MainNet:
```
tersec, line 52: This entire structure of hardcoding mainnet/sepolia/holesky around the codebase already has limits with the devnets, because some functionality ends up gated behind only working for those specific networks for no very fundamental reason.

nimbus/core/chain/persist_blocks.nim
```diff
@@ -40,158 +41,144 @@ type
...
+
+  # Full validation means validating the state root at every block and
+  # performing the more expensive hash computations on the block itself, ie
+  # verifying that the transaction and receipts roots are valid - when not
+  # doing full validation, we skip these expensive checks relying instead
+  # on the source of the data to have performed them previously or because
+  # the cost of failure is low.
+  # TODO Figure out the right balance for header fields - in particular, if
+  #      we receive instruction from the CL while syncing that a block is
+  #      CL-valid, do we skip validation while "far from head"? probably yes.
```
tersec, line 128: The main thing which needs to be checked is likely the basic link back to the parent blocks each time matching the block hash, i.e. the exact set of conditions that the engine API requires even an unsynced EL to perform. Beyond that isn't critical to security, and can be optimized to taste.

arnetheduck: yeah, I guess a rule the EL can follow is that if the block is earlier than finalized (according to the cl), we do just parent-hash checking - if it's more recent than finalized, we do full checking with state root and everything - that *should* be pretty safe given that all the validators approved it already.

Anyway, this comment is just copy-pasted from the earlier code - it's slightly out of date because forkedchain no longer uses this code from what I remember

nimbus/nimbus_import.nim
arnetheduck: I don't disagree but fixing it is non-trivial and therefore out of scope for this PR  - as a first step, moving the existing logic to a function makes it easy to replace in a follow-up

tersec (approved):

Issue #2909, opened 2024-12-04
Revise logging in fluffy/portal code


kdeme:
Originally we had quite some logging that was happening on history/state/beacon network level at `info` log level. This was very useful for developers, but not really for regular users. Hence we moved some/most of that logging to `debug`. However, `debug` log level is way to verbose now, and we should be moving some of the `debug` to `trace` (especially a lot of the `utp` logs) to still have a similar experience as developer.

Additionally, there are quite some logs which are redundant considering the same or similar error message is also transferred a layer up via a `Result`.

RazorClient:
Hey can i take up this issue

PR #2940, opened 2024-12-15
Cleaning up debug level log

RazorClient:
pr to issue 
https://github.com/status-im/nimbus-eth1/issues/2909


kdeme (commented):
This is a bit more involved than just replacing the `debug` traces with `trace` where you did just now. In fact most of those should stay `debug`.

RazorClient:
Oh ..
Then I’ll be redoing the same with by tonight!

PR #2955, opened 2024-12-19
Only hashing immutable part of network configuration for dataDir ID

jangko:
@mjfh; @agnxsh ; @arnetheduck you can use hidden config `--rewrite-datadir-id` to reconfigure your existing database if you encounter "Data dir already initialized with other network configuration"

Future hardfork time or in the case of holesky and mainnet, got new `depositContractAddress` make the network configuration can change when new features added.

Because we need something stable to identify dataDir with network configuration, the new dataDir ID is calculated using only immutable parts of network configuration.

### DIFF
--- a/nimbus/config.nim
+++ b/nimbus/config.nim
@@ -417,6 +417,11 @@ type
       desc: "Print RDB statistics at exit"
       name: "debug-rdb-print-stats".}: bool
 
+    rewriteDatadirId* {.
+      hidden
+      desc: "Rewrite selected network config hash to database"
+      name: "rewrite-datadir-id".}: bool
+
     case cmd* {.
       command
       defaultValue: NimbusCmd.noCommand }: NimbusCmd
--- a/nimbus/config.nim
+++ b/nimbus/config.nim
@@ -420,7 +420,7 @@ type
     rewriteDatadirId* {.
       hidden
       desc: "Rewrite selected network config hash to database"
-      name: "rewrite-datadir-id".}: bool
+      name: "debug-rewrite-datadir-id".}: bool
 
     case cmd* {.
       command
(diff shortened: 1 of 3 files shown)


nimbus/config.nim
```diff
@@ -417,6 +417,11 @@ type
       desc: "Print RDB statistics at exit"
       name: "debug-rdb-print-stats".}: bool
 
+    rewriteDatadirId* {.
+      hidden
+      desc: "Rewrite selected network config hash to database"
+      name: "rewrite-datadir-id".}: bool
```
tersec, line 423: Might be better to use `--debug-rewrite-datadir-id`

https://nimbus.guide/options.html
> Any `debug`-prefixed flags are considered ephemeral and subject to removal without notice.

Right now, there aren't releases and nothing to support, but it would be useful not to get accidentally caught out with basically irrelevant early-development-only era flags when that happens or have to do a big clean up then to avoid it.

This is basically a hack/kludge around what amounts to a bug regarding how networks are currently identified.

jangko: https://github.com/status-im/nimbus-eth1/pull/2955/commits/d5da2e3eea2be0d5c4f145d2a4b94df7cd287308

nimbus/nimbus_execution_client.nim
```diff
@@ -158,17 +158,24 @@ proc setupMetrics(nimbus: NimbusNode, conf: NimbusConf) =
...
-      kvt.put(dataDirIdKey().toOpenArray, calculatedId.data).isOkOr:
-        fatal "Cannot write data dir ID", ID=calculatedId
-        quit(QuitFailure)
+      writeDataDirId(kvt, calculatedId)
       return
 
+  if conf.rewriteDatadirId:
+    writeDataDirId(kvt, calculatedId)
+    return
+
```
advaita-saha, line 178: Wouldn't this allow it to write an existing holesky database with sepolia config when the flag is passed?

jangko: and how to prevent that?

advaita-saha: Maybe storing networkId and networkParams separately 🤔, instead of hashing them together

PR #2960, opened 2024-12-19
aristo: fork support via layers/txframes

### LINKED ISSUE #2949: Need to fix some misconception in ForkedChain
- [x] No more cursordesc, cursorheads, pivotarc, arc. Just heads and branches like what the community call them.
- [x] updateFinalized happen before updateBase.
- [x] updateFinalized will responsible for pruning branches that is no longer can reach finalizedHead.
- [ ] updateBase will move base forward up to allowed distance, or to the nearest junction of a fork.
- [x] ~Replay to target block using a list built during traverse backward from target block to a known block.~ #2950

see also:
https://github.com/status-im/nimbus-eth2/blob/412ea71115ce6dded2841f18320c91feb86bf372/beacon_chain/consensus_object_pools/blockchain_dag.nim#L1963

### LINKED ISSUE #2950: Next step after fixing misconception in ForkedChain
after we fix #2949 

the next step after that is going to be to add to `BlockDesc` a `layer` or `blockT` field that holds the state associated with having applied that block - any time a fork is created, instead of replaying via `stagingTx`, this field will be used instead as a starting point - I've mentioned a few other, more efficient designs in the past, but this is the most simple one that gets us over the finish line

arnetheduck:
This change reorganises how the database is accessed: instead holding a "current frame" in the database object, a dag of frames is created based on the "base frame" held in `AristoDbRef` and all database access happens through this frame, which can be thought of as a consistent point-in-time snapshot of the database based on a particular fork of the chain.

In the code, "frame", "transaction" and "layer" is used to denote more or less the same thing: a dag of stacked changes backed by the on-disk database.

Although this is not a requirement, in practice each frame holds the change set of a single block - as such, the frame and its ancestors leading up to the on-disk state represents the state of the database after that block has been applied.

"committing" means merging the changes to its parent frame so that the difference between them is lost and only the cumulative changes remain - this facility enables frames to be combined arbitrarily wherever they are in the dag.

In particular, it becomes possible to consolidate a set of changes near the base of the dag and commit those to disk without having to re-do the in-memory frames built on top of them - this is useful for "flattening" a set of changes during a base update and sending those to storage without having to perform a block replay on top.

Looking at abstractions, a side effect of this change is that the KVT and Aristo are brought closer together by considering them to be part of the "same" atomic transaction set - the way the code gets organised, applying a block and saving it to the kvt happens in the same "logical" frame - therefore, discarding the frame discards both the aristo and kvt changes at the same time - likewise, they are persisted to disk together
- this makes reasoning about the database somewhat easier but has the downside of increased memory usage, something that perhaps will need addressing in the future.

Because the code reasons more strictly about frames and the state of the persisted database, it also makes it more visible where ForkedChain should be used and where it is still missing - in particular, frames represent a single branch of history while forkedchain manages multiple parallel forks - user-facing services such as the RPC should use the latter, ie until it has been finalized, a getBlock request should consider all forks and not just the blocks in the canonical head branch.

Another advantage of this approach is that `AristoDbRef` conceptually becomes more simple - removing its tracking of the "current" transaction stack simplifies reasoning about what can go wrong since this state now has to be passed around in the form of `AristoTxRef` - as such, many of the tests and facilities in the code that were dealing with "stack inconsistency" are now structurally prevented from happening. The test suite will need significant refactoring after this change.

Once this change has been merged, there are several follow-ups to do:

* there's no mechanism for keeping frames up to date as they get committed or rolled back - TODO
* naming is confused - many names for the same thing for legacy reason
* forkedchain support is still missing in lots of code
* clean up redundant logic based on previous designs - in particular the debug and introspection code no longer makes sense
* the way change sets are stored will probably need revisiting - because it's a stack of changes where each frame must be interrogated to find an on-disk value, with a base distance of 128 we'll at minimum have to perform 128 frame lookups for *every* database interaction - regardless, the "dag-like" nature will stay
* dispose and commit are poorly defined and perhaps redundant - in theory, one could simply let the GC collect abandoned frames etc, though it's likely an explicit mechanism will remain useful, so they stay for now

More about the changes:

* `AristoDbRef` gains a `txRef` field (todo: rename) that "more or less" corresponds to the old `balancer` field
* `AristoDbRef.stack` is gone - instead, there's a chain of `AristoTxRef` objects that hold their respective "layer" which has the actual changes
* No more reasoning about "top" and "stack" - instead, each `AristoTxRef` can be a "head" that "more or less" corresponds to the old single-history `top` notion and its stack
* `level` still represents "distance to base" - it's computed from the parent chain instead of being stored
* one has to be careful not to use frames where forkedchain was intended
- layers are only for a single branch of history!

TODO items for this PR:
* [x] update forkedchain to use the new mechanism, solving #2949 in the process
* [x] update / cleanup tests

fix #2949
fix #2950


jangko:
The regression in hive already fixed, now the FAIL/OK = 42/129.
I know there is still 4 regression compared to master branch. It is related to `getScore` in `newPayload` and `fCU`.
But that can be fixed in separate PR.

advaita-saha:
> The regression in hive already fixed, now the FAIL/OK = 42/129. I know there is still 4 regression compared to master branch. It is related to `getScore` in `newPayload` and `fCU`. But that can be fixed in separate PR. Because hive's `mapper.jq` for `nimbus-el` also need to be fixed, there is no direct effect if I fixed it here.

Yeah for sure majority issues are fixed now, but we need to also then fix `mapper.jq`, please create a issue regarding the changes required or you can also give a PR to hive

advaita-saha (approved):

nimbus/db/aristo/aristo_desc.nim
```diff
@@ -43,18 +43,14 @@ type
...
     ## Three tier database object supporting distributed instances.
-    top*: LayerRef                    ## Database working layer, mutable
-    stack*: seq[LayerRef]             ## Stashed immutable parent layers
-    balancer*: LayerRef               ## Balance out concurrent backend access
     backend*: BackendRef              ## Backend database (may well be `nil`)
 
-    txRef*: AristoTxRef               ## Latest active transaction
-    txUidGen*: uint                   ## Tx-relative unique number generator
+    txRef*: AristoTxRef               ## Bottom-most in-memory frame
+    # txUidGen*: uint                   ## Tx-relative unique number generator
```
tersec, line 53: Is this still necessary, i.e. is it a temporary or permanent comment?

arnetheduck: "probably" permanent but I think it may have been used in the debug code obsoleted by this approach - a follow-up PR would either rewrite or remove the debug code (though in reality, broadly I think the time would better be invested in solid test cases that verify different scenarios instead of just printing them.

nimbus/db/aristo/aristo_layers.nim
```diff
@@ -11,101 +11,61 @@
...
-  db.stack.mapIt(it.sTab.len).foldl(a + b, db.top.sTab.len)
-
-func nLayersKey*(db: AristoDbRef): int =
-  ## Number of vertex ID/key entries on the cache layers. This is an upper
-  ## bound for the number of effective vertex ID mappingss held on the cache
-  ## layers as there might be duplicate entries for the same vertex ID on
-  ## different layers.
-  ##
-  db.stack.mapIt(it.kMap.len).foldl(a + b, db.top.kMap.len)
+# func dup(sTab: Table[RootedVertexID,VertexRef]): Table[RootedVertexID,VertexRef] =
```
tersec, line 24: Seems better to remove this commented-out `dup` function than keep it commented - it doesn't appear to have any interesting archaeological content.

nimbus/db/core_db/base_iterators.nim
```diff
@@ -34,50 +34,40 @@ when CoreDbEnableApiTracking:
...
   else:
     fn(args)
 
-template call(
-    acc: CoreDbAccRef;
-    fn: untyped;
-    args: varargs[untyped];
-      ): untyped =
-  acc.distinctBase.parent.ariApi.call(fn, args)
+#template call(
```
tersec, line 56: Remove rather than comment out? Also seems to be a kind of straightforward Nim utility which doesn't have much archaeological value (along with the other `template call(kvt: CoreDbKvtRef; fn: untyped; args: varargs[untyped]): untyped =`, and the commented-out imports `std/typetraits` and `./kvt as use_kvt`).

nimbus/db/aristo/aristo_desc.nim
tersec: In general, the PR seems to remove tx UID generation/tracking though? Many places it is just removed entirely.

tersec (approved):

nimbus/core/chain/forked_chain/chain_kvt.nim
```diff
@@ -26,28 +26,29 @@ import
 
 proc fcKvtAvailable*(c: ForkedChainRef): bool =
   ## Returns `true` if `kvt` data can be saved persistently.
-  c.db.txFrameLevel() == 0
+  true
```
tersec, line 29:
```
$ rg -C3 -i fcKvtAvailable
nimbus/core/chain/forked_chain/chain_kvt.nim
24-  ../../../db/core_db,
25-  ./chain_desc
26-
27:proc fcKvtAvailable*(c: ForkedChainRef): bool =
28-  ## Returns `true` if `kvt` data can be saved persistently.
29-  true
30-
--
33-  ## that it saves all cached db data including `Aristo` data (although there
34-  ## should not be any.)
35-  ##
36:  if c.fcKvtAvailable():
37-    let db = c.com.db
38-    db.persistent(c.baseTxFrame.getSavedStateBlockNumber()).isOkOr:
39-      raiseAssert "fcKvtPersistent: persistent() failed: " & $$error
--
56-  ## Cache data on the `kvt` table marked for saving persistently. If the `kvt`
57-  ## table is unavailable, this function does nothing and returns `false`.
58-  ##
59:  if c.fcKvtAvailable():
60-    c.baseTxFrame.put(key, data).isOkOr:
61-      raiseAssert "fcKvtPut: put() failed: " & $$error
62-    return true
--
65-  ## Cache key for deletion on the  `kvt` table.  If the `kvt` table is
66-  ## unavailable, this function does nothing and returns `false`.
67-  ##
68:  if c.fcKvtAvailable():
69-    c.baseTxFrame.del(key).isOkOr:
70-      raiseAssert "fcKvtDel: del() failed: " & $$error
71-    return true

nimbus/sync/beacon/worker/db.nim
47-      ) =
48-  ## Delete stale headers and state
49-  let c = ctx.pool.chain
50:  if not c.fcKvtAvailable():
51-    trace kvtNotAvailable(info)
52-    return
53-
--
76-proc dbStoreSyncStateLayout*(ctx: BeaconCtxRef; info: static[string]) =
77-  ## Save chain layout to persistent db
78-  let c = ctx.pool.chain
79:  if c.fcKvtAvailable():
80-    discard c.fcKvtPut(LhcStateKey.toOpenArray, rlp.encode(ctx.layout))
81-    discard c.fcKvtPersistent()
82-  else:
--
162-  let
163-    c = ctx.pool.chain
164-    last = first + revBlobs.len.uint64 - 1
165:  if not c.fcKvtAvailable():
166-    # Need to cache it because FCU has blocked writing through to disk.
167-    for n,data in revBlobs:
168-      ctx.stash[last - n.uint64] = data
```

Seems like if this is truly unconditionally true, `fcKvtAvailable` can be removed and a bunch of the `if not c.fcKvtAvailable` codepaths removed, and the `if c.fcKvtAvailable` codepaths moved out of the `if` (or put in `block:` if appropriate).

nimbus/db/aristo/aristo_layers.nim
mjfh: I guess most of the debugging stuff has expired its shelf life. It was repeatedly updated but never re-designed. So I would suggest to remove everything here (checks and pretty printers) and possibly re-implement the few parts that are needed in the unit tests.

nimbus/core/chain/forked_chain/chain_kvt.nim
tersec: https://github.com/status-im/nimbus-eth1/pull/2960/commits/861f89b08a1e9c540819539107910a66dd3c0485

nimbus/db/aristo/aristo_layers.nim
tersec: https://github.com/status-im/nimbus-eth1/pull/2960/commits/861f89b08a1e9c540819539107910a66dd3c0485

nimbus/core/chain/forked_chain/chain_branch.nim
```diff
@@ -0,0 +1,138 @@
...
+    txFrame* : CoreDbTxRef
+    receipts*: seq[Receipt]
+    hash*    : Hash32
+
+  BlockPos* = object
+    branch*: BranchRef
+    index* : int
+
+  BranchRef* = ref object
+    blocks*: seq[BlockDesc]
```
arnetheduck, line 28: what's the content of `blocks`? is it all blocks all the way to the `base`?

jangko: Only one branch directly contains a base block. All other branches only contains blocks up to the fork block. And they can access the ancestor branch through `parent` branch.

arnetheduck: ok - my gut feeling is that the code would be simplified by removing the additional "branchref" layer and simply have a "BlockDescRef". this would be something to consider for future pr:s..

In particular, when a fork/pruning happens somewhere in the "middle" of the list of `blocks` here, the logic to split and recombine the branchref into a parent and two mini-branches is susceptible to subtle bugs. 

it's certainly not impossible to get this logic right, but even if it "saves" on ref instances by grouping ranges of branch blocks under a single entity, the "range of blocks" doesn't really appear in any first-principle descriptions of how the chain works and therefore, one always has to "reason" about this additional abstraction when relating it to the spec.

jangko:
image
The hive test shows the new FC module is stabilizing. From the 50 failing test case, 1 related to known issue of `ChainId` testing against `UInt256` while nimbus still using 64 bit `ChainId`. The rest of 49 cases related to BLS precompiles(EIP 2537) unable to handle infinity point correctly.

note: The hive test requires the new `mapper.jq` for nimbus-el https://github.com/ethereum/hive/pull/1239

advaita-saha:
> image The hive test shows the new FC module is stabilizing. From the 50 failing test case, 1 related to known issue of `ChainId` testing against `UInt256` while nimbus still using 64 bit `ChainId`. The rest of 49 cases related to BLS precompiles(EIP 2537) unable to handle infinity point correctly.
> 
> note: The hive test requires the new `mapper.jq` for nimbus-el [ethereum/hive#1239](https://github.com/ethereum/hive/pull/1239)

Doesn't work as you can see here https://hivenimbus.advaita.work
Your machine might have cached the master docker image, run with `--docker.nocache hive/clients/nimbus-el` and select the branch with `--client nimbus-el_forked-layers`

For verification you can check the commit hash of the runs visible in https://hivenimbus.advaita.work, one is master and other forked-layers
Also it's run with hive with your mapper.jq , which will be visible in the nimbus-el log genesis file, as deposit contract is added

arnetheduck:
how long does it take for hive to run? could we have a jenkins job for that?

advaita-saha:
> how long does it take for hive to run? could we have a jenkins job for that?

Around 45mins on a good machine, slow machines take a lot of time, can be parallelized based on resources. Major time is during compiling nimbus, after that it's around 20mins

I would not recommend having it in CI, because of constant changing of test cases by EF during development, but had a chat with Dustin to have a process of triggering a hive run in our servers

jangko:
Your mapper.jq is missing `mergeNetsplitBlock`. Although I cannot correlate between the error message and that missing field. 
image

advaita-saha:
Indeed the missing field is not related
I just used your branch from hive to build the tests, might be a problem there

Will check

jangko:
Rerun forked-layers branch and the result is same. Only one 7702 failing case, and the rest is 2537.

image

jangko:
@advaita-saha you can try to verify forked-layers branch using `test_blockchain_json` and `test_generalstate_json`. Run it locally on your machine. It should produce similar result to running hive against master branch. If the problem still persist then It is very weird. If both bc and gst test runs ok, then something wrong with your hive setup.

advaita-saha:
> @advaita-saha you can try to verify forked-layers branch using `test_blockchain_json` and `test_generalstate_json`. Run it locally on your machine. It should produce similar result to running hive against master branch. If the problem still persist then It is very weird. If both bc and gst test runs ok, then something wrong with your hive setup.

Indeed this is weird, let me try

nimbus/core/chain/forked_chain.nim
```diff
@@ -360,107 +220,238 @@ func calculateNewBase(
...
+  # during `beacon sync` or `nrpc sync`
+  if count > 1:
+    notice "Finalized blocks persisted",
+      numberOfBlocks = count,
+      baseNumber = c.baseBranch.tailNumber,
+      baseHash = c.baseBranch.tailHash.short
+  else:
+    debug "Finalized blocks persisted",
+      numberOfBlocks = count,
+      target = newBase.hash.short,
```
advaita-saha, line 446: After hours of debugging for the reason of failure in hive tests on my machine, 
the reason was, my nimbus instance was running in `debug` mode
And this line was causing the crash with `Error: unhandled exception: index 1 not in 0 .. 0 [IndexDefect]`

For now this PR has been dragged a lot, let's comment this line and add a TODO, which we can look into later
```suggestion
      # target = newBase.hash.short, TODO: fix index out of bound error
```

jangko: https://github.com/status-im/nimbus-eth1/pull/2960/commits/8dcd09b9a848150598ffc68a66e7ac125b543a2f

Fixed. Thanks for investigating this @advaita-saha . I know using `BlockPos` is a bit dangerous, maybe we can use `BlockDescRef` in the future and eliminate `BlockPos` altogether.

advaita-saha (approved):

nimbus/beacon/api_handler/api_forkchoice.nim
```diff
@@ -162,21 +162,22 @@ proc forkchoiceUpdated*(ben: BeaconEngineRef,
 
   # If the beacon client also advertised a finalized block, mark the local
   # chain final and completely in PoS mode.
+  let baseTxFrame = ben.chain.baseTxFrame
```
etan-status, line 165: Could there be multiple long branches and the new finalizedBlockHash is on a non-canonical branch? In that case, wouldn't we first have to switch to the new head branch (line 182)?

tersec: https://github.com/status-im/nimbus-eth1/issues/3083

PR #2963, opened 2024-12-20
devnet-5: Update EIP-7685: exclude empty requests data in commitment

jangko:


nimbus/beacon/api_handler/api_newpayload.nim
```diff
@@ -83,12 +83,35 @@ template validatePayload(apiVersion, payloadVersion, payload) =
...
         "excessBlobGas is expected from execution payload")
 
+func validateExecutionRequest(requests: openArray[seq[byte]]): Result[void, string] {.raises:[].} =
+  var previousRequestType = -1.byte
+  for request in requests:
+    if request.len == 0:
+      return err("Execution request data must not be empty")
+
+    let requestType = request[0]
+    if requestType <= previousRequestType:
```
tersec, line 93: This doesn't work because `byte` is unsigned:
```nim
const
  DEPOSIT_REQUEST_TYPE       = 0x00.byte
  WITHDRAWAL_REQUEST_TYPE    = 0x01.byte
  CONSOLIDATION_REQUEST_TYPE = 0x02.byte

var request_type = -1.byte
echo request_type
doAssert request_type == 255
doAssert request_type > DEPOSIT_REQUEST_TYPE
doAssert request_type > WITHDRAWAL_REQUEST_TYPE
doAssert request_type > CONSOLIDATION_REQUEST_TYPE
```

nimbus/beacon/api_handler/api_newpayload.nim
```diff
@@ -83,12 +83,35 @@ template validatePayload(apiVersion, payloadVersion, payload) =
       raise invalidParams("newPayload" & $apiVersion &
         "excessBlobGas is expected from execution payload")
 
+func validateExecutionRequest(requests: openArray[seq[byte]]): Result[void, string] {.raises:[].} =
+  var previousRequestType = -1.byte
+  for request in requests:
+    if request.len == 0:
+      return err("Execution request data must not be empty")
```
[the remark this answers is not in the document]
tersec: oops -- this is for the individual request

tersec:
[`getPayloadV4`](https://github.com/ethereum/execution-apis/blob/9707339bc8222f6d43b3bf0a7a91623f7ce52213/src/engine/prague.md#engine_getpayloadv4) appears to be missing?
> 2. The call **MUST** return `executionRequests` list representing execution layer triggered requests. Each list element is a `requests` byte array as defined by [EIP-7685](https://eips.ethereum.org/EIPS/eip-7685). The first byte of each element is the `request_type` and the remaining bytes are the `request_data`. Elements of the list **MUST** be ordered by `request_type` in ascending order. Elements with empty `request_data` **MUST** be excluded from the list.

nimbus/beacon/api_handler/api_newpayload.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/2963/commits/18c0a2b2840a23d7e11abbb4a072c5b7e44c56c1

jangko:
> [getPayloadV4](https://github.com/ethereum/execution-apis/blob/9707339bc8222f6d43b3bf0a7a91623f7ce52213/src/engine/prague.md#engine_getpayloadv4) appears to be missing?

https://github.com/status-im/nimbus-eth1/blob/036dd23e9b0bba425ad87dbbb0d938f379a26e8e/nimbus/beacon/api_handler/api_getpayload.nim#L69

https://github.com/status-im/nimbus-eth1/blob/036dd23e9b0bba425ad87dbbb0d938f379a26e8e/nimbus/rpc/engine_api.nim#L76C2-L78C1

tersec:
The `ExecutionBundle`'s representation for the `executionRequests` is fine (since right now, internally, there are exactly 3 sources):
https://github.com/status-im/nimbus-eth1/blob/036dd23e9b0bba425ad87dbbb0d938f379a26e8e/nimbus/beacon/payload_queue.nim#L34-L39

But when `getPayloadV4` (which, I wasn't very clear I guess, obviously it exists, but the EIP-7685 update does not appear to exist for it): gets this, it simply relays the bundle of `executionRequests out from the JSON-RPC endpoint:
https://github.com/status-im/nimbus-eth1/blob/036dd23e9b0bba425ad87dbbb0d938f379a26e8e/nimbus/beacon/api_handler/api_getpayload.nim#L82-L95

to be serialized implicitly via the default rules.

But two things are different now:
1. they need to each be prefixed with the request type byte; and
2. the empty one(s) need to be skipped

exactly the same/symmetrically as in `newPayloadV4`, i.e. `newPayloadV4` needs to be able to consume what `getPayloadV4` produces.

jangko:
It start from here in the txpool. I know it's a bit confusing because we don't have dedicated block builder. But until then, we have to use what we have.

https://github.com/status-im/nimbus-eth1/blob/03bef35c404c38942673aa8dc692dc136f453d8d/nimbus/core/tx_pool/tx_packer.nim#L366-L374

tersec:
> It start from here in the txpool. I know it's a bit confusing because we don't have dedicated block builder. But until then, we have to use what we have.
> 
> https://github.com/status-im/nimbus-eth1/blob/03bef35c404c38942673aa8dc692dc136f453d8d/nimbus/core/tx_pool/tx_packer.nim#L366-L374

Oops, indeed

PR #2984, opened 2025-01-05
Logging Fixes for INFO and DEBUG level

advaita-saha:
- [x] Log the event of adding a transaction to the `txpool`, and show warnings if _rejected_ with the cause of rejection
- [x] When the client is following chain, and _1 block_ is added every 12 seconds 1 block is persisted to the persistence storage, there it doesn't make sense to log every block. So a summarised persistence _INFO_ log is there when more blocks are persisted together ( case for beacon syncing / nrpc syncing ), else the logs will be shifted to _DEBUG_
- [x] Few logs had missing topics making it difficult find and isolate during devnet runs, so topics needed to be added to ( e.g `beacon engine`, `txpool` )
- [x] None of the syncer logs were visible at the _INFO_ level making it difficult for the user to understand what is happening in the background, all these logs were visible in the TRACE/DEBUG level, shifting of few logs from there to the _INFO_ level ( e.g start and suspending of syncer, successful staging of headers, import of blocks, etc )


nimbus/sync/beacon/worker/update.nim
```diff
@@ -121,7 +121,7 @@ proc startHibernating(ctx: BeaconCtxRef; info: static[string]) =
 
   ctx.hibernate = true
 
-  debug info & ": suspending syncer", L=ctx.chain.latestNumber.bnStr
+  info info & ": suspending syncer", L=ctx.chain.latestNumber.bnStr
```
tersec, line 124: Agree all these syncer state changes are worth making more visible.

nimbus/core/chain/forked_chain.nim
```diff
@@ -224,11 +227,18 @@ proc writeBaggage(c: ForkedChainRef, target: Hash32) =
       prevHash = header.parentHash
       count.inc
 
-    notice "Finalized blocks persisted",
-      numberOfBlocks = count,
-      last = target.short,
-      baseNumber = c.baseHeader.number,
-      baseHash = c.baseHash.short
+    if count > 1:
```
tersec, line 230: The intent here is to distinguish batch-syncing-persisted from persisting blocks in a more granular way once a node is synced?

advaita-saha: Kinda yes,
When my holesky was following the chain the INFO logs were like
```
INF new payload
INF fork choice updated
NTC 1 block persisted
```
Well in this case the user doesn't need to know the persistence data IMO and can lie in the DEBUG level, while when the syncer is syncing, the persistence data helps tracking the movement of the base for the user ( they know ok my node is syncing )

nimbus/core/tx_pool/tx_desc.nim
```diff
@@ -150,15 +154,22 @@ proc getNonce(xp: TxPoolRef; account: Address): AccountNonce =
 
 proc classifyValid(xp: TxPoolRef; tx: Transaction, sender: Address): bool =
   if tx.tip(xp.baseFee) <= 0.GasInt:
+    warn "Invalid Transaction: No tip"
```
tersec, line 157: How visible these should be depend on how much this is accessible by transactions the node might gather from other eth68/eth69 pools, and/or how much one considers the blocks it gets to be trusted.

In particular, it shouldn't allow eth68 peers to effectively spam (without cost, really, because they're not yet associated with a block yet) logs with invalid transactions. This `classifyValid` function is called from `addTx`, which is called to handle `eth_sendTransaction` and `eth_sendRawTransaction`, and for those purposes it's fine -- it's a user-triggered interaction and it's useful to report issues in some way regarding that.

But `addTx` is ultimately part of `tx_pool`'s public API and appears to be the canonical way for anything to add a transaction, and it's just kind of temporary happenstance that currently the only tx sources are basically trusted. Even if `nimbus-eth1` never ends up using eth68/eth69 to get transactions, it might use Portal or some other mechanism and they have basically the same issue.

It's useful for sure to have these checks visibly in some circumstances, just not quite this broadly.

nimbus/core/chain/forked_chain.nim
tersec: Yeah, agree on that point, that the
```
NTC 1 block persisted
```
isn't necessary.

But it'd be good to at least comment in the code that 1 isn't a per-se tunable parameter, there's something intrinsically special about it. Anything not-1 means it's potentially not in a steady-state synced configuration.

At some level this points to a broader logging issue -- blocks being persisted is kind of not the optimal abstraction level to log, probably, because it's an artifact of how `ForkedChain` works.

In particular, the blocks-persisted logs don't necessarily happen at the same granularity or with the same frequency as would be useful for a user to see, and that they happen, more or less, an epoch at a time, is from a user perspective kind of a random artifact.

Ideally it's syncing fast enough, indeed, that logging every 32 blocks is spammy -- but that doesn't mean that forkedchain should change, but rather that it's probably better for the user-level sync logging to be more decoupled from the mechanics of `ForkedChain`.

nimbus/core/tx_pool/tx_desc.nim
advaita-saha: Yeah ok makes sense to log the transactions submitted via `eth_` endpoints at the _INF_ and _WRN_ level

nimbus/sync/beacon/worker/blocks_staged.nim
```diff
@@ -290,7 +290,7 @@ proc blocksStagedImport*(
     nBlocks = qItem.data.blocks.len
     iv = BnRange.new(qItem.key, qItem.key + nBlocks.uint64 - 1)
 
-  debug info & ": import blocks ..", iv, nBlocks,
+  info info & ": import blocks ..", iv, nBlocks,
```
arnetheduck, line 293: Compared to the others, these logs randomly add a prefix - all "user"-level logs should look and feel the same (capitalization, style, voice etc). in particular, the `info` prefix here is not how chronicles is meant to be used: instead, topics should be used for any common static tags.

also, the data here doesn't make sense _for a user_ .. what's `B`? only a  developer intimately familiar with the code will know that, and likely not even them 6 months down the line.

arnetheduck: the aim here is to tell a story, much like a progress bar .. ie "staged blocks" - what does that mean and how is it different from "import blocks"? Syncing already has established terms for many of these concepts, there's no need to invent new words - compare with other clients, what they log when they sync.

Typically, what _should_ be logged is that we're first downloading headers and then blocks - is that the story that is immediately obvious when you read the logs as they pass by?

Also, one log should lead to the other, ie it should be possible to track how the database is gradually being built up by correlating numbers and hashes in logs - is that possible when reading these logs? ie do they form an identifiable logical sequence?

advaita-saha: Surely agreed on the 1st point, I wanted to remove `info` but was waiting for a bit discussion with jordan. But surely I agree that `info` should be removed 

Addressing the naming `B, L, C, D`, yes it doesn't makes sense to a random user, it will be changed to a more meaningful user friendly

arnetheduck:
> Logs about amount of peers are in the DEBUG, needs to be shifted to INFO level

In general, it's good if info logs strive to summarize additional data in a single line so as not to waste vertical space - if there are several related information points that make sense together, a single log line is better than multiple.

advaita-saha:
> > Logs about amount of peers are in the DEBUG, needs to be shifted to INFO level
> 
> In general, it's good if info logs strive to summarize additional data in a single line so as not to waste vertical space - if there are several related information points that make sense together, a single log line is better than multiple.

Yes had a discussion would with @tersec regarding the same, I am thinking of coupling the info about amount of peers with other logs ( similar to how it's done in `nimbus-eth2` )

Should we include the peer data with engineAPI logs ( as a running client it is always supposed to have a CL attached ) ? 🤔

PR #2991, opened 2025-01-10
Portal history JSON-RPC: Add validation, where reasonable

kdeme:
Add content validation + local storing to Portal history JSON-RPC methods:
- portal_historyGetContent
- portal_historyTraceGetContent

This means that for bodies and receipts being requested an additional network lookup (the header) is done. This is likely acceptable as they are considered high level calls.

Not adding the validation however for portal_historyPutContent method as this would highly limit the bridge gossip efficiency.

In general the Portal JSON-RPC API specifications are not very clear on which type of validation checks should be done in these methods.

### DIFF
--- a/fluffy/rpc/rpc_portal_history_api.nim
+++ b/fluffy/rpc/rpc_portal_history_api.nim
@@ -13,6 +13,7 @@ import
   json_serialization/std/tables,
   stew/byteutils,
   ../network/wire/portal_protocol,
+  ../network/history/history_network,
   ./rpc_types
 
 {.warning[UnusedImport]: off.}
@@ -29,7 +30,10 @@ TraceObject.useDefaultSerializationIn JrpcConv
 NodeMetadata.useDefaultSerializationIn JrpcConv
 TraceResponse.useDefaultSerializationIn JrpcConv
 
-proc installPortalHistoryApiHandlers*(rpcServer: RpcServer, p: PortalProtocol) =
+proc installPortalHistoryApiHandlers*(rpcServer: RpcServer, n: HistoryNetwork) =
+  template p(): PortalProtocol =
+    n.portalProtocol
+
   rpcServer.rpc("portal_historyFindContent") do(
     enr: Record, contentKey: string
   ) -> JsonString:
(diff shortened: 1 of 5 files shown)


kdeme:
Verified locally that the Hive tests keep working for all hive history network simulators.

kdeme:
Ah, this still fails in the Portal RPC client tests.

There is no easy fix for this as it doesn't use mainnet data and thus the added canonical verification fails.

fluffy/rpc/rpc_portal_history_api.nim
```diff
@@ -133,7 +150,7 @@ proc installPortalHistoryApiHandlers*(rpcServer: RpcServer, p: PortalProtocol) =
       contentId = p.toContentId(keyBytes).valueOr:
         raise invalidKeyErr()
 
-      valueBytes = p.dbGet(keyBytes, contentId).valueOr:
+      valueBytes = p.getLocalContent(keyBytes, contentId).valueOr:
```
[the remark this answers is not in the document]
bhartnett: If that does ever become a problem, we could disable the content cache in the tests.

fluffy/rpc/rpc_portal_history_api.nim
```diff
@@ -147,9 +164,11 @@ proc installPortalHistoryApiHandlers*(rpcServer: RpcServer, p: PortalProtocol) =
         raise invalidKeyErr()
       offerValueBytes = hexToSeqByte(contentValue)
 
-      # TODO: Do we need to convert the received offer to a value without proofs before storing?
       # TODO: validate and store content locally
-      # storedLocally = p.storeContent(keyBytes, contentId, valueBytes)
+      # Note: Not validating content as this would have a high impact on bridge
+      # gossip performance. Specification is quite ambiguous on what kind of
+      # validation should be done here anyhow.
+      # As no validation is done here, the content is not stored locally.
```
bhartnett, line 171: Maybe we should at least do some basic validation. Anything that doesn't require a network call such as decoding the types and checking the structure.

bhartnett:
> Ah, this still fails in the Portal RPC client tests.
> 
> There is no easy fix for this as it doesn't use mainnet data and thus the added canonical verification fails.

Yeah the tests use some fake data but I guess we can update them to use mainnet block data. Perhaps the data in portal-spec-tests.

bhartnett:
Overall I think having validation on by default for both `portal_historyGetContent` and `portal_historyPutContent` is a good idea so that other users which might use these endpoints for other use cases don't run into issues. 

The only reason to not add the validation is for performance reasons so perhaps we should implement a debug parameter that can be used to disable the validation when the extra performance is required. That way we can correctly implement the spec and be more consistent. For the state network I have the `--disable-state-root-validation` parameter for similar reasons which I use when running the state bridge with a Fluffy node locally. Without this parameter the validation would do an extra network lookup to fetch the block header in order to get the state root. 

The existing content cache should also help improve performance because we store the fetched block header in `getVerifiedBlockHeader` which means if we get the header, then the body and receipt for the same block we should only do one network call for the body and one for the receipt instead of two each because we should be getting the block header from the content cache.

fluffy/rpc/rpc_portal_history_api.nim
bhartnett: Actually for this I think we should do the full validation by default and have a debug parameter to disable it when required

kdeme:
> Overall I think having validation on by default for both `portal_historyGetContent` and `portal_historyPutContent` is a good idea so that other users which might use these endpoints for other use cases don't run into issues.

Yes, I've been going a bit back and forth with this, adding validation or not, but in the end it comes down to simply different use cases, so I think we just need different JSON-RPC methods doing one or the other. 

I do not really like an extra fluffy cli flag for this, as this is not very visible for the user and would require a restart. I rather just have different JSON-RPC methods or perhaps an optional parameter in the existing method. 

Regarding the current implementation:
- For the `GetContent`:
For block headers you can do basic block hash validation instead of the canonical validation, which might be sufficient as the user is requesting for a specific block hash anyhow. Although you could say that you might want to indicate an error to the user if there is a block hash request for a hash that does not exist on this chain (e.g. maybe the user started the node for the wrong chain/network).
Then for bodies and receipts, it gets again more complicated as the headers needs to be requested also (potentially doubling the  RTT).
So there is the added RTT, but I don't think the call is used so excessively in bridges that it matters that much?
Conclusion: I just went with the general verification call  that verifies everything to be consistent.

- For the `PutContent`: Because this is used by our bridges, I did not want to add the full verification, that extra RTT would really kill performance. I could add just the de-serialization of content to verify at least that, but I deemed that not *that*` useful.
But of course, as this is considered a "higher level" call, one could argue that it should have this validation. But in consider it not that harmful as receiving nodes should drop this data anyhow when invalid. The biggest issue would be with the fact that the local node might store it unvalidated in its own database, which is why I did not add that functionality (which does go against the spec).

kdeme:
Closing, some changes covered in https://github.com/status-im/nimbus-eth1/pull/3000
Will move addition of validation to a PR that adds also optional `validation` parameters.

Issue #2796, opened 2024-10-28
`nrpc` keeps sending blocks when el responds INVALID


arnetheduck:
due to a corrupted database, my nimbus-eth1 was responding `INVALID` to all requests.

`nrpc` however just happily kept trying to build on them - instead, it should stop:

```
INF 2024-10-28 19:01:20.241+01:00 newPayload Request sent                    blockNumber=2571310 response=INVALID
INF 2024-10-28 19:01:20.300+01:00 newPayload Request sent                    blockNumber=2571311 response=INVALID
INF 2024-10-28 19:01:20.350+01:00 newPayload Request sent                    blockNumber=2571312 response=INVALID
INF 2024-10-28 19:01:20.379+01:00 newPayload Request sent                    blockNumber=2571313 response=INVALID
INF 2024-10-28 19:01:20.404+01:00 newPayload Request sent                    blockNumber=2571314 response=INVALID
INF 2024-10-28 19:01:20.428+01:00 newPayload Request sent                    blockNumber=2571315 response=INVALID
INF 2024-10-28 19:01:20.450+01:00 newPayload Request sent                    blockNumber=2571316 response=INVALID
INF 2024-10-28 19:01:20.480+01:00 newPayload Request sent                    blockNumber=2571317 response=INVALID
INF 2024-10-28 19:01:20.509+01:00 newPayload Request sent                    blockNumber=2571318 response=INVALID
INF 2024-10-28 19:01:20.535+01:00 newPayload Request sent                    blockNumber=2571319 response=INVALID
INF 2024-10-28 19:01:20.554+01:00 newPayload Request sent                    blockNumber=2571320 response=INVALID
INF 2024-10-28 19:01:20.573+01:00 newPayload Request sent                    blockNumber=2571321 response=INVALID
INF 2024-10-28 19:01:20.596+01:00 newPayload Request sent                    blockNumber=2571322 response=INVALID
INF 2024-10-28 19:01:20.622+01:00 newPayload Request sent                    blockNumber=2571323 response=INVALID
INF 2024-10-28 19:01:20.640+01:00 newPayload Request sent                    blockNumber=2571324 response=INVALID
INF 2024-10-28 19:01:20.660+01:00 newPayload Request sent                    blockNumber=2571325 response=INVALID
INF 2024-10-28 19:01:20.677+01:00 newPayload Request sent                    blockNumber=2571326 response=INVALID
[...]

PR #2994, opened 2025-01-13
Nrpc re-org handling

advaita-saha:
Adds re-org handling capacity to `nrpc` along with fixes for #2796

### DIFF
--- a/nrpc/nrpc.nim
+++ b/nrpc/nrpc.nim
@@ -22,6 +22,7 @@ import
   beacon_chain/el/el_manager,
   beacon_chain/el/engine_api_conversions,
   beacon_chain/spec/[forks, state_transition_block],
+  beacon_chain/spec/datatypes/bellatrix,
   beacon_chain/spec/eth2_apis/[rest_types, rest_beacon_calls],
   beacon_chain/networking/network_metadata,
   eth/async_utils
@@ -45,7 +46,6 @@ template getCLBlockFromBeaconChain(
   var blck: ForkedSignedBeaconBlock
   if clBlock.isSome():
     let blck = clBlock.get()[]
-
     (blck, true)
   else:
     (blck, false)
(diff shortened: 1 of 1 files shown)


nrpc/nrpc.nim
```diff
@@ -249,18 +249,19 @@ proc syncToEngineApi(conf: NRpcConf) {.async.} =
...
             forkyBlck.message.body.blob_kzg_commitments,
             engine_api.VersionedHash(kzg_commitment_to_versioned_hash(it)),
           )
           # Execution Requests for Electra
-          let execution_requests = @[
-            SSZ.encode(forkyBlck.message.body.execution_requests.deposits),
-            SSZ.encode(forkyBlck.message.body.execution_requests.withdrawals),
-            SSZ.encode(forkyBlck.message.body.execution_requests.consolidations),
-          ]
+          let execution_requests =
```
tersec, line 259: This format has changed, see
- https://github.com/status-im/nimbus-eth2/pull/6766
- https://github.com/status-im/nimbus-eth2/pull/6822

nrpc/nrpc.nim
```diff
@@ -274,11 +285,17 @@ proc syncToEngineApi(conf: NRpcConf) {.async.} =
...
             executionRequests = execution_requests
         else:
-          static: doAssert(false, "Unsupported consensus fork")
+          static:
+            doAssert(false, "Unsupported consensus fork")
 
         info "newPayload Request sent",
           blockNumber = int(payload.blockNumber), response = payloadResponse.status
 
+        if payloadResponse.status != PayloadExecutionStatus.valid:
```
tersec, line 294: This is too strict -- syncing and accepted are fine too.

This should trigger on what https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.0/sync/optimistic.md#helpers calls `INVALIDATED`: `INVALID` or `INVALID_BLOCK_HASH`.

advaita-saha: But `syncing and accepted` shouldn't be expected the behaviour if `nrpc sync` is running and syncing properly

tersec: If the EL starts out as not synced, then in general an EL which does snap sync will go some hours before it starts responding with anything but syncing or accepted.

https://github.com/status-im/nimbus-eth1/blob/cee5fe9456205827f661a049b04df027ef49fa2a/nrpc/nrpc.nim#L212-L217
has a similar issue here.

PR #2995, opened 2025-01-13
Beacon sync revisting fc automove base2

mjfh:


nimbus/sync/beacon/worker/update.nim
```diff
@@ -282,17 +280,16 @@ proc updateSyncState*(ctx: BeaconCtxRef; info: static[string]) =
...
-       ctx.target.final != 0:            # .. ditto
+       ctx.target.changed:               # and there is a new target from CL
       ctx.setupCollectingHeaders info    # set up new header sync
     return
     # Notreached
 
   info "Sync state changed", prevState, thisState,
     head=ctx.chain.latestNumber.bnStr,
-    oldBase=(if ctx.layout.coupler == ctx.layout.dangling: "downloaded"
+    coupler=(if ctx.layout.coupler == ctx.layout.dangling: "dangling"
```
advaita-saha, line 290: As discussed with @tersec , at the `info` level logs keeping terms like `coupler`and `dangling`  would confuse the user. It would be better to go with terms similar to other clients

nimbus/sync/beacon/worker/blocks_staged.nim
```diff
@@ -367,7 +342,7 @@ proc blocksStagedImport*(
   ctx.updateMetrics()
 
   info "Import done", iv, nBlocks, base=ctx.chain.baseNumber.bnStr,
-    head=ctx.chain.latestNumber.bnStr, target=ctx.layout.final.bnStr
+    head=ctx.chain.latestNumber.bnStr
```
advaita-saha, line 345: A overall sync target is nice to have to look at wrt to understanding how far I am 
( without scrolling logs, POV of users )

advaita-saha (changes requested):

nimbus/sync/beacon/worker/update.nim
mjfh: might be, but base is completely wrong (see README), I don't really care and can revert that though

advaita-saha: What would be a better term for the base then ?

nimbus/sync/beacon/worker/blocks_staged.nim
mjfh: the target was wrong, `final` was never a target and does not exist anymore (see FC logic), maybe you take your pick from the [README](https://github.com/status-im/nimbus-eth1/tree/master/nimbus/sync/beacon#metrics)

nimbus/sync/beacon/worker/update.nim
mjfh: There was a reason why I used letter symbols.

mjfh: This should be left out completely then if it is not understood by a casual reader.

nimbus/sync/beacon/worker/blocks_staged.nim
advaita-saha: Ahhh right, also with this PR it gets removed, 
Then ig `target=ctx.layout.head.bnStr` would be the target

nimbus/core/chain/forked_chain.nim
```diff
@@ -462,6 +462,50 @@ proc updateHeadIfNecessary(c: ForkedChainRef, pvarc: PivotArc) =
...
+  # from draft PR #2845.
+  #
+  let
+    distanceFromBase = c.cursorHeader.number - c.baseHeader.number
+    hysteresis = max(1'u64, min(c.baseDistance div 4'u64, 32'u64))
+  # Finalizer threshold is baseDistance + 25% of baseDistancce capped at 32.
+  if distanceFromBase < c.baseDistance + hysteresis:
+    return ok()
+
+  # Move the base forward and stay away `baseDistance` blocks from
```
tersec, line 480: Why is this necessarily valid? For example, a chain might never have finalized.

nimbus/sync/beacon/worker/blocks_staged.nim
```diff
@@ -312,7 +311,7 @@ proc blocksStagedImport*(
           B=ctx.chain.baseNumber.bnStr, L=ctx.chain.latestNumber.bnStr,
           nthBn=nBn.bnStr, nthHash=qItem.data.getNthHash(n).short
         continue
-      ctx.pool.chain.importBlock(qItem.data.blocks[n]).isOkOr:
+      ctx.pool.chain.importBlock(qItem.data.blocks[n], autoRebase=true).isOkOr:
```
tersec, line 314: Does this loop know if the chain it's on is actually finalized?

jangko:
Note: This PR is postponed until FC module is ready to handle non-finalized branch or a very long branch yet to be finalized.

nimbus/core/chain/forked_chain.nim
```diff
@@ -462,6 +462,50 @@ proc updateHeadIfNecessary(c: ForkedChainRef, pvarc: PivotArc) =
 
   c.setHead(pvarc)
 
+proc autoUpdateBase(c: ForkedChainRef): Result[void, string] =
```
arnetheduck, line 465: btw, this function is a step in a good direction: the responsibility for updating the base should indeed rest firmly with FC.

However, there's nothing "special" with this flow at all, like this documentation makes it seem - instead, it should be considered a normal part of the importBlock flow _and_ of the fCU flow.

There are two things that can cause the base to be updated: 

* we received new blocks (either from RPC or the syncer)
* we move the finalization point (via fCU)

Both of these should update base _potentially_ and both flows should call the same function to do so, also so that the hysteresis computation below applies to both "sources" of base updates equally. This means that this same function should be called for *all* base updates that happen in FC.

Long-range syncing, ie syncing blocks that are older than finality, is a special case of the above logic because FC already knows that they form a linear history, but for the updateBase logic it does not matter: the execution head is what determines base in this case, that and the execution head in the case of long-range syncing is simply a block _on the DAG branch_ that leads to finality and eventually the consensus head.

mjfh:
Completely out of date -- needs to be re-developed from cratch

PR #3001, opened 2025-01-14
Improved logging

### LINKED ISSUE #2544: Implement logging options from eth2

pedromiguelmiranda:
Improved eth1 logging:
- using nimbus-eth2 options
- log file support removal
- auto detection of tty, colour support and related.

fix #2544

### DIFF
--- a/nimbus/sync/protocol/eth68.nim
+++ b/nimbus/sync/protocol/eth68.nim
@@ -1,6 +1,6 @@
 # Nimbus - Ethereum Wire Protocol
 #
-# Copyright (c) 2018-2024 Status Research & Development GmbH
+# Copyright (c) 2018-2025 Status Research & Development GmbH
 # Licensed under either of
 #  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or
 #    http://www.apache.org/licenses/LICENSE-2.0)
@@ -21,7 +21,8 @@ import
   stew/byteutils,
   ./trace_config,
   ./eth/eth_types,
-  ../../utils/utils
+  ../../utils/utils,
+  ../../common/logging
 
 export
   eth_types
--- a/nimbus/sync/sync_sched.nim
+++ b/nimbus/sync/sync_sched.nim
@@ -86,7 +86,7 @@
 import
   std/hashes,
   chronos,
-  eth/[p2p, p2p/peer_pool, common/eth_types_json_serialization],
+  eth/[p2p, p2p/peer_pool],
   stew/keyed_queue,
   ./sync_desc
 
@@ -430,7 +430,7 @@ proc onPeerConnected[S,W](dsc: RunnerSyncRef[S,W]; peer: Peer) =
       # This could happen if there are idle entries in the table, i.e.
       # somehow hanging runners.
       if dsc.ctx.noisyLog: trace "Peer table full! Dequeuing least used entry",
-        oldest.peer, oldest.only, nPeers, nWorkers=dsc.buddies.len, maxWorkers
+        oldestPeer=oldest.peer, oldestOnly=oldest.only, nPeers=nPeers, nWorkers=dsc.buddies.len, maxWorkers
       # Setting to `zombie` will trigger the worker to terminate (if any.)
       oldest.ctrl.zombie = true
 
--- a/nimbus/sync/sync_sched.nim
+++ b/nimbus/sync/sync_sched.nim
@@ -86,7 +86,7 @@
 import
   std/hashes,
   chronos,
-  eth/[p2p, p2p/peer_pool],
+  eth/[p2p, p2p/peer_pool, common/eth_types_json_serialization],
   stew/keyed_queue,
   ./sync_desc
 
@@ -430,7 +430,7 @@ proc onPeerConnected[S,W](dsc: RunnerSyncRef[S,W]; peer: Peer) =
       # This could happen if there are idle entries in the table, i.e.
       # somehow hanging runners.
       if dsc.ctx.noisyLog: trace "Peer table full! Dequeuing least used entry",
-        oldest, nPeers, nWorkers=dsc.buddies.len, maxWorkers
+        oldest.peer, oldest.only, nPeers, nWorkers=dsc.buddies.len, maxWorkers
       # Setting to `zombie` will trigger the worker to terminate (if any.)
       oldest.ctrl.zombie = true
 
--- a/nimbus/common/common.nim
+++ b/nimbus/common/common.nim
@@ -11,6 +11,7 @@
 
 import
   chronicles,
+  logging,
   ../db/[core_db, ledger, storage_types],
   ../utils/[utils],
   ".."/[constants, errors, version],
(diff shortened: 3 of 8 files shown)


tersec:
Yeah, this is arguably a `nim-json-serialization` bug:
```
/home/runner/work/nimbus-eth1/nimbus-eth1/nimbus/common/common.nim(128, 5) template/generic instantiation of `info` from here
/home/runner/work/nimbus-eth1/nimbus-eth1/vendor/nim-chronicles/chronicles.nim(380, 10) template/generic instantiation of `log` from here
/home/runner/work/nimbus-eth1/nimbus-eth1/nimbus/common/common.nim(134, 7) template/generic instantiation of `expandItIMPL` from here
/home/runner/work/nimbus-eth1/nimbus-eth1/vendor/nim-chronicles/chronicles/log_output.nim(835, 39) template/generic instantiation of `setProperty` from here
/home/runner/work/nimbus-eth1/nimbus-eth1/vendor/nim-chronicles/chronicles/log_output.nim(772, 10) template/generic instantiation of `[]=` from here
/home/runner/work/nimbus-eth1/nimbus-eth1/vendor/nim-chronicles/chronicles/log_output.nim(745, 17) template/generic instantiation of `writeField` from here
/home/runner/work/nimbus-eth1/nimbus-eth1/vendor/nim-json-serialization/json_serialization/writer.nim(94, 8) template/generic instantiation of `writeValue` from here
/home/runner/work/nimbus-eth1/nimbus-eth1/vendor/nim-json-serialization/json_serialization/writer.nim(387, 23) template/generic instantiation of `writeRecordValue` from here
/home/runner/work/nimbus-eth1/nimbus-eth1/vendor/nim-json-serialization/json_serialization/writer.nim(203, 8) template/generic instantiation of `enumInstanceSerializedFields` from here
/home/runner/work/nimbus-eth1/nimbus-eth1/vendor/nim-serialization/serialization/object_serialization.nim(48, 40) Error: type mismatch
Expression: fieldPairs(value)
  [1] value: uint64

Expected one of (first mismatch at [position]):
[1] iterator fieldPairs[S: tuple | object; T: tuple | object](x: S; y: T): tuple[
    key: string, a, b: RootObj]
[1] iterator fieldPairs[T: tuple | object](x: T): tuple[key: string, val: RootObj]
```
is because `nim-json-serialization`'s `writeValue` [only checks for `SomeInteger`, before fairly shortly assuming it has to be an object of some type](https://github.com/status-im/nim-json-serialization/blob/2b1c5eb11df3647a2cee107cd4cce3593cbb8bcf/json_serialization/writer.nim#L360-L387), a heuristic which fails in the case of a `distinct` of `SomeInteger`:
```nim
  elif value is SomeInteger:      # this case gets skipped
    w.stream.writeText value

  elif value is SomeFloat:
    # TODO Implement writeText for floats
    #      to avoid the allocation here:
    append $value

  elif value is (seq or array or openArray) or
      (value is distinct and distinctBase(value) is (seq or array or openArray)):
    when value is distinct:
      w.writeArray(distinctBase value)
    else:
      w.writeArray(value)

  elif value is (distinct or object or tuple):
    mixin flavorUsesAutomaticObjectSerialization

    type Flavor = JsonWriter.Flavor
    const isAutomatic =
      flavorUsesAutomaticObjectSerialization(Flavor)

    when not isAutomatic:
      const typeName = typetraits.name(type value)
      {.error: "Please override writeValue for the " & typeName & " type (or import the module where the override is provided)".}

    when value is distinct:
      # it tries to do this, but this is inappropriate for a `distinct SomeInteger`
      writeRecordValue(w, distinctBase(value, recursive = false))
```
This only trigers when `flavorUsesAutomaticObjectSerialization(Flavor)`, otherwise it will (more or less defensibly reasonable) report that it can't, in this case, serialize `EthTime`.

But when automatic serialization is enabled, this at least some kind of n-j-s issue.

nimbus/sync/sync_sched.nim
```diff
@@ -403,7 +403,7 @@ proc onPeerConnected[S,W](dsc: RunnerSyncRef[S,W]; peer: Peer) =
       # This could happen if there are idle entries in the table, i.e.
       # somehow hanging runners.
       if dsc.ctx.noisyLog: trace "Peer table full! Dequeuing least used entry",
-        oldest, nPeers, nWorkers=dsc.buddies.len, maxWorkers
+        oldest.peer, oldest.only, nPeers, nWorkers=dsc.buddies.len, maxWorkers
```
pedromiguelmiranda, line 406: @jangko , @mjfh 
 This trace of the var "oldest" is a little bit tricky with dyamic json logging, it requires to implement json object writers for all the type procedures involved, mostly on databases (rbtree, sorted sets and others).

 I don 't have a trace of this type (don't know also how to trigger it :/) to reach a conclusion of which types would be useful to log. I've extracted the most obvious ones. don t know which ones are missing. Would you mind sharing which ones would be useful?

 Alternative, we implement the handlers for the type procedures (example: GetVtxFn | GetKeyFn | GetTuvFn  and a lot others), but eventually, we will even have to implement json handlers for type procedures used in Chronos library.
 
 Personally speaking, I don't think logging these type procedures will add much information on a tracing.

tersec:
> Initially, I also agreed that it was a Nim-Json-Serializer bug. After some talks, I've changed my mind. njs is a public and generic library, and so any used defined types should be implemented by the user using the public available handlers (documented on njs README file).

Yeah, the bug I had in mind was mainly that it even tried to use the `fieldPairs` codepath with a non-`object`.

It's fine to group `  elif value is (distinct or object or tuple):` but then it doesn't make sense to assume that it must specifically be an `object`.

nimbus/sync/sync_sched.nim
mjfh: The `oldest` is used at other places as well. Are they all `tricky`? If no, why then here? The earlier line uses
 `oldest=("Node[" & $leastVal.key.address & "]")`,
would that do?

FYI: Currently there is not much of a problem solved with this variable although it is part of the rejection/zombie mechanism in order to avoid immediate reconnection loops. So it might be helpful to keep this entry nevertheless.

mjfh: Or use `oldest=oldest.peer` (should also work)

pedromiguelmiranda: _Tricky_ regarding the _trace_ log, not how the variable is used.
And this is the only place through the code base where we try to log via chronicles.

nimbus/sync/protocol/eth68.nim
```diff
@@ -21,7 +21,8 @@ import
   stew/byteutils,
   ./trace_config,
   ./eth/eth_types,
-  ../../utils/utils
+  ../../utils/utils,
+  ../../common/logging
```
tersec, line 25: Unexpected that `eth68.nim` needs to add any imports without other changes using said imports. Presumably, it does some logging which triggers the `writeValue`s, e.g., due to the `EthTime` thing?

pedromiguelmiranda: the only import added was the ```common/logging ``` and due to the fact that it has some ```trace``` log messages.

PR #3004, opened 2025-01-15
Re-name metrics variable to better fit into the current naming schemes

mjfh:
also:
  Force metrics update when all peers vanish. There might be not much
  activity in that case so that the set of metrics might become stale.


mjfh:
superseded by #3016

PR #3006, opened 2025-01-17
Fluffy: Cleanup unused code from PortalProtocol

bhartnett:
I noticed that these functions are never used so perhaps they should be removed.


bhartnett:
@kdeme Did you have any objections to me removing these functions? If they are not likely to be used any time soon, I'd suggest removing.

Issue #2809, opened 2024-10-31
Add peer ban list to Fluffy


kdeme:
Nodes that offer invalid data will currently be allowed to keep doing so wasting bandwidth and cpu of the receiving node.

Any offer that has a properly encoded content key and that falls in the data radius of the receiving node will be accepted.
Data will be send and once received it will be validated. If the data fails validation, the peer currently is not punished and thus allowed to keep offering this or other invalid data.

Similary a node can provide invalid data on a FindContent. Or send no data at all or very slow until uTP transfers time out.

We should add a temporary ban list for nodes that do this.

There might also be nodes that end up in the DHT of a specific subnetwork for some reason, but don't actually support the subnetwork. Might require a temporary (per network) ban list for these to avoid trying to ping over and over a node that is not supported.

kdeme:
We currently see a lot of following errors:
```
WRN 2024-11-06 11:11:29.853+01:00 Offer failed due to accept request failure  topics="portal_wire" error="No message data, peer might not support this talk protocol" contentKeys=@[0x00458ff92094a00fac6e393a435d58242c4971e9711adf1cd16215fc97856e4a1e] node=b7*406c26:164.92.247.243:5008
```

This is definitely a good candidate as reason for subnetwork specific temporary bans. As it really indicate a node that does not support a subnetwork.
In theory these should get filtered out of the subnetwork DHT, but if other clients don't do that properly or if the node does reply to ping/pongs on that subnetwork, this will not occur.

bhartnett:
@kdeme I remember you mentioned that there is an existing ban list implementation in nimbus-eth2 somewhere but I can't seem to find the link (if you shared it previously). Can share it here when you get a chance?

bhartnett:
How long should we ban nodes that are misbehaving?

So we should ban nodes for a period of time when they:
- Send offers that fail to validate (conteny key or value)
- Return responses to findContent that fails to validate
- Fails to respond or sends data too slow and hits timeout in response to pings, lookups and offers.
- Responds with empty or bad data to pings, lookups and offers.

Does this cover it or are there other scenarios that need to be covered?

kdeme:
> How long should we ban nodes that are misbehaving?

I think we might have different timeouts depending on the reason of the ban. But nodes that are simply incompatible should be banned for several hours if not a day. Nodes that have many timeouts perhaps less, as maybe the node/machine was just "busy" for a bit.

>Does this cover it or are there other scenarios that need to be covered?

[...]

PR #3007, opened 2025-01-17
Fluffy: Portal subnetwork peer ban list

bhartnett:
This PR relates to https://github.com/status-im/nimbus-eth1/issues/2809 and implements a per subnetwork peer ban list in the PortalProtocol type.

Changes completed:
- Banned peers table added to portal protocol used to record banned peers until a given timeout.
- Nodes can be banned for a given duration from the current time by nodeId.
- The ban list is checked before returning neighbours from the routing table, and banned nodes are filtered out so that we don't attempt to connect to banned nodes.
- The node lookup is updated to also filter out banned nodes when fetching nodes from peers.
- The contentLookup and traceContentLookup is updated to filter out banned nodes when fetching nodes from peers for the case when no content is found and we return the closest nodes to the content.
- When handling/receiving a portal message we check the ban list before processing it. If the sending peer is in the ban list then we don't process the message.
- Added helpers templates to ban a node when offer validation or content lookup validation fails.
- Ban nodes when offer and content lookup validations fail in the state network

Remaining work:
- Ban nodes when validations fail in the beacon and history networks.
- Ban misbehaving nodes for other scenarios and conditions in portal protocol.
- Review and decide on the best ban timeouts for each type of ban reason.


fluffy/network/wire/portal_protocol.nim
```diff
@@ -299,8 +336,11 @@ func getNode*(p: PortalProtocol, id: NodeId): Opt[Node] =
 func localNode*(p: PortalProtocol): Node =
   p.baseProtocol.localNode
 
-func neighbours*(p: PortalProtocol, id: NodeId, seenOnly = false): seq[Node] =
-  p.routingTable.neighbours(id = id, seenOnly = seenOnly)
+proc neighbours*(
+    p: PortalProtocol, id: NodeId, k: int = BUCKET_SIZE, seenOnly = false
+): seq[Node] =
+  # TODO: Maybe banned peers should be in the routing table and filtered out at that level
+  p.routingTable.neighbours(id, k, seenOnly).filterIt(not p.isBanned(it.id))
```
bhartnett, line 343: @kdeme Any thoughts on whether it would be a good idea to add the ban list into the routing table code in nim-eth? I believe the same routing table is used by the discv5 code and so it might be a good way to apply something similar at the discv5 level.

kdeme: I think it can be either in the routing table object or in PortalProtocol and DiscoveryProtocol. The latter would require a few more wrappers like you did now for `neighbours`.

What is more important I think is that banned nodes do not stay in the routing table. They should be removed from the routing table and not be allowed to re-added. That way the routing table gives space to valid nodes and a `neighbours` call gives better result.

I think achieving this might be slightly simpler when the ban list is part of routing table, not sure. 

But yes, the ban list would also be useful for discv5 I think.

fluffy/network/state/state_network.nim
```diff
@@ -109,6 +109,7 @@ proc getContent(
       continue
 
     validateRetrieval(key, contentValue).isOkOr:
+      n.portalProtocol.contentLookupFailedValidation(lookupRes.receivedFrom.id)
```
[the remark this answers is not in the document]
bhartnett: Good point. I'll remove those templates.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -126,6 +126,10 @@ const
   ## value in milliseconds
   initialLookups = 1 ## Amount of lookups done when populating the routing table
 
+  # Ban durations for the banned peers table
+  PeerBanDurationContentLookupFailedValidation = 30.minutes
+  PeerBanDurationOfferFailedValidation = 60.minutes
```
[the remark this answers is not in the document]
bhartnett: Yeah that sounds reasonable.

fluffy/network/wire/portal_protocol.nim
bhartnett: Thanks for the feedback. Good point about removing nodes from the routing table. I think considering that we want to keep the banned nodes out of the routing table and that this logic will likely get reused by discv5, it makes sense to move to the RoutingTable. Will reduce the amount of duplicated code and provide a cleaner solution

bhartnett:
@kdeme This is ready for another review.

bhartnett:
I've run the hive state tests locally and verified that they are all passing with these changes.

fluffy/network/beacon/beacon_network.nim
```diff
@@ -364,6 +364,8 @@ proc validateContent(
 
       debug "Received offered content validated successfully", srcNodeId, contentKey
     else:
+      if srcNodeId.isSome():
+        n.portalProtocol.banNode(srcNodeId.get(), NodeBanDurationOfferFailedValidation)
```
kdeme, line 368: This one is tricky, there are these cases which should not result in ban:
- https://github.com/status-im/nimbus-eth1/pull/3007/files#diff-41d12d3f36c29261d8d4bb41832f419a78f4a9920dc257374c43503f09321fdeR297-R300 -> These are rather failures because we are not synced and cannot verify the latest bootstraps
- https://github.com/status-im/nimbus-eth1/pull/3007/files#diff-41d12d3f36c29261d8d4bb41832f419a78f4a9920dc257374c43503f09321fdeR316 -> I have to double check, but from what I remember these will also fail on "duplicates", those that have already been verified once.
- And I think the same applies for optimistic and finality updates.

So for now I'd maybe remove the banNode here. We have to apply it more fine-grained I'm afraid.

fluffy/network/history/history_network.nim
```diff
@@ -147,6 +147,9 @@ proc getVerifiedBlockHeader*(
         return Opt.none(Header)
 
       header = validateCanonicalHeaderBytes(headerContent.content, id, n.accumulator).valueOr:
+        n.portalProtocol.banNode(
```
kdeme, line 150: While all these banNodes are fine in theory. They might cause issues with a occasional Trin bug where only partial data is send currently.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -661,26 +681,34 @@ proc reqResponse[Request: SomeMessage, Response: SomeMessage](
...
       proc(x: cstring): string =
         $x
     )
     .flatMap(
       proc(x: seq[byte]): Result[Message, string] =
-        decodeMessage(x)
+        let r = decodeMessage(x)
+        # Ban nodes that that send messages that fail to decode (includes empty messages)
+        if r.isErr():
+          p.banNode(dst.id, NodeBanDurationInvalidResponse)
```
kdeme, line 702: Was it not this ban that could cause a continuous "ban each other" due to empty talk-responses as you mentioned to me before?

fluffy/rpc/rpc_portal_state_api.nim
```diff
@@ -46,6 +46,7 @@ proc installPortalStateApiHandlers*(rpcServer: RpcServer, p: PortalProtocol) =
     of Content:
       let valueBytes = foundContentResult.content
       validateRetrieval(key, valueBytes).isOkOr:
+        p.banNode(node.id, NodeBanDurationContentLookupFailedValidation)
```
kdeme, line 49: No huge deal, but having these `banNode` calls also in the rpc api is a bit much. It would be nice if we could limit them to portal wire protocol and the specific networks.

kdeme: To be clear, the bans on this behaviour are correct, and should be done.

fluffy/network/beacon/beacon_network.nim
bhartnett: Sure, I'll remove this one in that case.

fluffy/network/history/history_network.nim
bhartnett: I think the longer term improvement for this would be to add counts of violations and then only ban nodes after a certain limit is reached. For this PR though I guess we could simply reduce the ban time if there are concerns with it not behaving as intended.

fluffy/network/wire/portal_protocol.nim
bhartnett: Well kind of. There was previously a much longer ban here which I have now removed. I guess it is still possible that nodes could keep banning each other because we send an empty response to banned nodes. At the moment banned nodes get the same response as nodes on the wrong sub-protocol. 

I thought that reducing the ban time would reduce the risk of this occurring. I'm happy to remove the ban here though if you think we should? Once we are able to ignore banned nodes without replying with an empty response then we could improve this in the future I guess.

fluffy/rpc/rpc_portal_state_api.nim
bhartnett: > No huge deal, but having these `banNode` calls also in the rpc api is a bit much. It would be nice if we could limit them to portal wire protocol and the specific networks.

Yeah I guess at some point I could refactor this to import the state network into the rpc api and then reuse the existing functions there which will already apply the bans as needed. Probably better to do that in a future PR though.

kdeme (approved):

PR #3010, opened 2025-01-17
Minimal implementation of Portal ping payload extensions spec

kdeme:


fluffy/network/wire/portal_protocol.nim
```diff
@@ -334,26 +335,63 @@ func truncateEnrs(
...
+    encodedCustomPayload.asSeq(), CustomPayloadExtensionsFormat
+  ).valueOr:
+    # invalid custom payload format, send back FailedToDecodePayload
+    return encodeErrorPayload(ErrorCode.FailedToDecodePayload)
+
+  if customPayload.`type` notin p.pingExtensionCapabilities:
+    return encodeErrorPayload(ErrorCode.ExtensionNotSupported)
+
+  case customPayload.`type`
+  of CapabilitiesType:
```
kdeme, line 351: The way currently decoding + actions taken + encoding is put all together is not very modular (and that also becomes visible in tests). 

I'll probably change this to have an object variant for this. I'd have to use a enum with different numbers however than the current numbers used for the types. But that is probably better anyhow so that there are no holes. (It does mean that full range cannot be covered, but that should be fine as realistically that range should not be hit)

fluffy/network/wire/portal_protocol.nim
```diff
@@ -703,25 +749,33 @@ proc recordsFromBytes*(
...
+    let customPayload = decodeSsz(
+      pong.customPayload.asSeq(), CustomPayloadExtensionsFormat
+    ).valueOr:
+      return err("Pong message contains invalid custom payload")
+
+    if customPayload.`type` != CapabilitiesType:
+      return err("Pong message contains invalid custom payload")
+
+    let payload = decodeSsz(customPayload.payload.asSeq(), CapabilitiesPayload).valueOr:
+      return err("Pong message contains invalid CapabilitiesPayload")
```
kdeme, line 772: Very minimal version and does not decode  anything else than CapabilitiesPayload, which is fine for now as that is the only one we send also.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -334,26 +335,63 @@ func truncateEnrs(
...
   enrs
 
+proc handlePingExtension(
+    p: PortalProtocol, encodedCustomPayload: ByteList[2048], srcId: NodeId
+): ByteList[2048] =
+  let customPayload = decodeSsz(
+    encodedCustomPayload.asSeq(), CustomPayloadExtensionsFormat
+  ).valueOr:
+    # invalid custom payload format, send back FailedToDecodePayload
+    return encodeErrorPayload(ErrorCode.FailedToDecodePayload)
```
kdeme, line 345: We might be able to get away with this part if the spec gets changed to have the type and the payload directly in the ping/pong message

PR #3014, opened 2025-01-21
Discard received data on uTP content stream timeout

kdeme:
Discard the received data on uTP content stream read timeout. Before the data was still added to the queue and being processed and should normally fail in validation. However as we know not all data got read it should not even move to the validation step.

Added however a FIN send after the timeout instead of the delayed socket clean-up which does not make much sense in that scenario either. Basically either be nice and still send a FIN or just destroy the socket immediatly.

### DIFF
--- a/fluffy/network/wire/portal_stream.nim
+++ b/fluffy/network/wire/portal_stream.nim
@@ -1,5 +1,5 @@
 # Nimbus
-# Copyright (c) 2022-2024 Status Research & Development GmbH
+# Copyright (c) 2022-2025 Status Research & Development GmbH
 # Licensed and distributed under either of
 #   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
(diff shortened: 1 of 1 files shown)


fluffy/network/wire/portal_stream.nim
```diff
@@ -327,11 +324,15 @@ proc readContentOffer(
...
+      # Read timed out, stop further reading and discard the data received so
+      # far as it will be incomplete.
       debug "Reading data from socket timed out, content offer failed",
         contentKeys = offer.contentKeys
-      break
+      # Still closing the socket (= sending FIN) but not waiting here for its
+      # ACK however, so no `closeWait`. Underneath the socket will still wait
+      # for the FIN-ACK (or timeout) before it destroys the socket.
+      socket.close()
+      return
```
bhartnett, line 335: Makes sense to discard the data instead of allowing a subset of it to get through, looks good.

Something to note is that `socket.connectionId` is used in `handleIncomingConnection` just after the call to `readContentOffer`. As long as the connectionId fields don't somehow get cleaned up then this won't be a problem.

bhartnett (approved):

fluffy/network/wire/portal_stream.nim
kdeme: I think that should be fine as there is no `await` in that code. That code should be handled immediately and the future gets passed along?

PR #3019, opened 2025-01-23
Remove SSZ Union usage in BlockHeaderWithProof type

kdeme:
Remove SSZ Union in BlockHeaderWithProof type by making the proof an SSZ encoded ByteList. The right type for the proof can be selected at the decoding step by first looking into the header for the timestamp and selecting the right type based on the hardfork the block is in.

Draft as this is for a proposal to spec change.


kdeme:
Proposal spec change here: https://github.com/ethereum/portal-network-specs/pull/362

kdeme:
Tests rely on https://github.com/ethereum/portal-spec-tests/pull/34

Issue #2738, opened 2024-10-15
Rename nimbus source directory


pedromiguelmiranda:
Now that nimbus binary is renamed to nimbus_execution_client (#2728 ), directory name could also be renamed to nimbus_execution_client (or similar suggestion) and free both the binary and the source directory.

However, there is a trade off regarding log history. All files history under nimbus directory won't be easily accessible via git log command, only via git log --follow (some git clients already implement this, not default with git command tool).

This tradeoff shoud be investigated before the change.

arnetheduck:
the case could be made for `execution_chain` or simply `execution` here, to complement `beacon_chain` that we use in eth2

arnetheduck:
Also, the makefile target name should be updated to `make nimbus_execution_client`

PR #3036, opened 2025-01-29
nimbus initial commit

pedromiguelmiranda:
project structure with:
- chronicles support
- config files 
- layers handlers.


tersec (approved):

Makefile
```diff
@@ -104,6 +104,7 @@ VERIF_PROXY_OUT_PATH ?= build/libverifproxy/
 	update \
 	nimbus \
 	nimbus_execution_client \
+	nimbus_unified \
```
arnetheduck, line 107: why not name it nimbus?

nimbus_unified/tests/test_nimbus_unified.nim
```diff
@@ -0,0 +1,127 @@
...
+    # Set up a simulated control-C hook
+    setControlCHook(localControlCHandler)
+
+    # Trigger the hook manually
+    localControlCHandler()
+
+    check isShutDownRequired.load() == true
+
+    #services running should be 0
+    for i in 0 .. cNimbusMaxServices - 1:
```
arnetheduck, line 126: in all of these cases, you can `for s in nimbus.serviceList: check s.isNone`  - no need for a constant

nimbus_unified/nimbus_unified.nim
```diff
@@ -0,0 +1,155 @@
...
+      let thread = nimbus.serviceList[i].get()
+      if thread.serviceHandler.running():
+        joinThread(thread.serviceHandler)
+        info "Exited service ", service = thread.name
+      nimbus.serviceList[i] = none(NimbusService)
+
+  notice "Exited all services"
+
+## Service monitoring
+proc monitor*(nimbus: Nimbus) =
```
arnetheduck, line 91: what's this for, ie what will it be doing?

nimbus_unified/version.nim
```diff
@@ -0,0 +1,62 @@
...
+    NimVersion
+
+  compileYear = CompileDate[0 ..< 4] # YYYY-MM-DD (UTC)
+  copyrightBanner* =
+    "Copyright (c) " & compileYear & " Status Research & Development GmbH"
+
+  # Short debugging identifier to be placed in the ENR
+  enrClientInfoShort* = toBytes("f")
+
+func getNimGitHash*(): string =
```
arnetheduck, line 40: we have a "nimbus_binary_common" file for hosting little helpers like this - the idea with that file is no matter what the binary contains, certain things work the same (logging options, version numbers etc) - it would be good to refactor the existing binaries to make use of such utilities, just like the logging.

nimbus_unified/tests/test_nimbus_unified.nim
```diff
@@ -0,0 +1,127 @@
...
+  ../nimbus_unified,
+  ../configs/nimbus_configs,
+  #eth1-configs
+  ../../nimbus/nimbus_desc
+
+# ----------------------------------------------------------------------------
+# Helper Functions
+# ----------------------------------------------------------------------------
+
+template fileExists(filename: string): bool =
```
arnetheduck, line 20: what's wrong with https://nim-lang.org/docs/files.html#fileExists%2CPath?

arnetheduck: as an alternative, if this function does not exist in `io2`, it's better to add a correctly implemented function there (ie an implementation that uses the right os utilities, similar to the above linked documentation but with different error handling

nimbus_unified/tests/test_nimbus_unified.nim
```diff
@@ -0,0 +1,127 @@
...
+  try:
+    discard readFile(filename)
+    true
+  except IOError:
+    false
+
+template removeFile(filename: string) =
+  try:
+    discard io2.removeFile(filename)
+  except IOError:
```
arnetheduck, line 30: `io2` does not raise exceptions, or at least it shouldn't - we use either `Result` or exceptions but not both in a single function (with a few exceptions that don't apply here).

Makefile
pedromiguelmiranda: the target `nimbus` is on rampdown:
https://github.com/status-im/nimbus-eth1/blob/dev/pedro/kikof/Makefile#L217
image

and we need to solve issue #2738 and update some README `make` commands [refs](https://github.com/status-im/nimbus-eth1/blob/master/README.md?plain=1#L59)

Given that this is a project in development yet to be released (https://nimbus.guide/execution-client.html), how much time should we wait for the rampdown period?

Agree that it would be nice to get the renaming before starting landing this.

nimbus_unified/nimbus_unified.nim
pedromiguelmiranda: Monitor if one of the services ended (gracefully) and an to ask remaining services to end gracefully.
(check one the service prototype wrappers for an example).

This is a small naive mechanism using polling, to be replaced by some wake up mechanism (maybe a condition?)

nimbus_unified/nimbus_unified.nim
```diff
@@ -0,0 +1,155 @@
...
+        raiseAssert exc.msg # shouldn't happen
+
+    notice "\tCtrl+C pressed. Shutting down services"
+    isShutDownRequired.store(true)
+    nimbus.exitServices()
+
+  setControlCHook(controlCHandler)
+
+  var
+    execution = LayerConfig(kind: Execution, executionConfig: NimbusConf())
```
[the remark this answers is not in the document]
pedromiguelmiranda: This is a nice hint. thank you
Wasn't aware of the configs..

nimbus_unified/version.nim
pedromiguelmiranda: Great. all projects are implementing similar code for the same purpous.
Let me try to land that before this.

nimbus_unified/tests/test_nimbus_unified.nim
pedromiguelmiranda: absolutely nothing wrong with those components. This was my half way implementing unit tests for the `createPidFile` which is not being used.
Which will be moved to binary common as well

nimbus_unified/tests/test_nimbus_unified.nim
pedromiguelmiranda: same as preivous comment

Makefile
pedromiguelmiranda: Fixed in 411a3cadfa3ac0be5eaf4b7d69b863c169d75957

pedromiguelmiranda:
latest updates:
- removed dead code.
- removed some structures and functionalitires not needed yet
- thread communication with shared data via channels: instead of creating each client configuration and pass itto each service thread, an array of configs will be passed instead. Each service thread well haev the responsability of creating its configuration and manage it.

nimbus/tests/nim.cfg
```diff
@@ -0,0 +1,14 @@
...
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+# Use only `secp256k1` public key cryptography as an identity in LibP2P.
+-d:"libp2p_pki_schemes=secp256k1"
+-d:"chronicles_runtime_filtering=on"
+
+--styleCheck:usages
+--styleCheck:hint
+--hint[Processing]:off
```
tersec, line 14: Why the apparently random mix of files ending without newlines and ones which do?

pedromiguelmiranda: great find. 
the project is using the nph formatter which formats all the nim files, and I forgot the non nim ones.

PR #3045, opened 2025-02-04
Fix tools helper and allow GST to parse new eest devnet-6 test vectors

jangko:
The new eest of devnet-6 test vector contains multiple sub fixtures in one json file. The old format(Frontier-Cancun), only contains one sub fixture.

### DIFF
--- a/tests/test_config.nim
+++ b/tests/test_config.nim
@@ -1,5 +1,5 @@
 # Nimbus
-# Copyright (c) 2019-2024 Status Research & Development GmbH
+# Copyright (c) 2019-2025 Status Research & Development GmbH
 # Licensed under either of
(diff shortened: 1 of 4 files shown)


advaita-saha:
https://github.com/status-im/nimbus-eth1/blob/97b17f8727e910f21aa1ea149651821059ff2b7c/tests/test_generalstate_json.nim#L228-L229

It would be better if we can pass single test vector files in debug mode, not from a fixed folder path

jangko:
Maybe you can try `tools/evmstate`. It has better defined/standardized cli and consume the same fixtures as gst. It also has better features. I will not make gst better, if it can correctly run test, it already fullfill it's purpose. 
evmstate tool is designed as debugging tool from beginning.

PR #3053, opened 2025-02-07
Headerwithproof migration and prune

kdeme:
Playing around with options regarding database changes that occur in https://github.com/status-im/nimbus-eth1/pull/3019

### DIFF
--- a/fluffy/database/content_db.nim
+++ b/fluffy/database/content_db.nim
@@ -14,7 +14,8 @@ import
   results,
   eth/db/kvstore,
   eth/db/kvstore_sqlite3,
-  ../network/state/state_content,
+  ../network/history/[history_content],
+  ../network/history/content/content_values_deprecated,
   ../network/wire/[portal_protocol, portal_protocol_config],
   ./content_db_custom_sql_functions
 
@@ -48,6 +49,8 @@ type
   RowInfo =
     tuple[contentId: array[32, byte], payloadLength: int64, distance: array[32, byte]]
 
+  ContentPair = tuple[contentKey: array[32, byte], contentItem: seq[byte]]
+
   ContentDB* = ref object
     backend: SqStoreRef
     kv: KvStoreRef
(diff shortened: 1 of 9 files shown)


kdeme:
Naive version is way too slow even for relative small databases.
It highly depends on how many actual headers with proof are stored, which when launching a new node might not be the same for some of our long running nodes. Especially as when testing I use the portal bridge to locally dump a lot of data in it.

But to give an example:
Database filled from the network: `1.1G Jan 14 17:25 fluffy.sqlite3`
```
NTC 2025-01-29 15:28:10.989+01:00 ContentDB migration: iterating over all content
NTC 2025-01-29 15:55:48.725+01:00 ContentDB migration done:                  contentAltered=46903 contentDeleted=223548
```
-> 27:38 mins
The important value is here `46903` updates (well rather INSERT OR REPLACE) and + `223548` deletes.

kdeme:
Delete version (https://github.com/status-im/nimbus-eth1/pull/3053/commits/8008aef243028c0c0b767ac83b2abd5173dd4aeb) is fine in terms of speed, but will not update the existing values that could be migrated.

On a big database with ` 953240` required deletions:
```
NTC 2025-02-07 16:31:32.686+01:00 ContentDB migration: deleting all headers without proof
NTC 2025-02-07 16:32:02.274+01:00 ContentDB migration done 
```

Might be able to do a similar update statement, not sure if supported with current API and what it would do in terms of memory.

Simpler option could be to just delete all, and re-inject all headers-only via our portal_bridge.

Secondary question is, whether to move this in to `fcli` as separate action to take or not?
In case of not, we probably need to add a version table or use sqlite `PRAGMA user_version`. The latter seems however already used for https://github.com/status-im/nim-eth/blob/master/eth/db/kvstore_sqlite3.nim#L595C32-L595C44

kdeme:
Closing this as https://github.com/status-im/nimbus-eth1/pull/3019 now has this functionality in the `fcli_db` tool

PR #3062, opened 2025-02-11
Speedup `eth_getLogs`

advaita-saha:
Currently in devnet-6-pectra noticed a very slow `eth_getLogs` calls from mainly lighthouse
And it was taking around 30seconds to complete a `eth_getLogs` call, which was also responsible for pushing `nimbus-eth1` off from synced state

This implements
- Fetches the transactionHashes from in-memory blocks using txRecords
- Removes the `.blockHash` call for every log despite of being from the same block

### DIFF
--- a/nimbus/core/chain/forked_chain.nim
+++ b/nimbus/core/chain/forked_chain.nim
@@ -640,9 +640,10 @@ func memoryTxHashesForBlock*(c: ForkedChainRef, blockHash: Hash32): seq[Hash32] 
     if blkHash == blockHash:
       cachedTxHashes.add((txHash, txIdx))
 
-  cachedTxHashes.sorted(proc(a, b: (Hash32, uint64)): int =
+  cachedTxHashes.sort(proc(a, b: (Hash32, uint64)): int =
       cmp(a[1], b[1])
-    ).mapIt(it[0])
+    )
+  cachedTxHashes.mapIt(it[0])
 
 proc latestBlock*(c: ForkedChainRef): Block =
   if c.activeBranch.headNumber == c.baseBranch.tailNumber:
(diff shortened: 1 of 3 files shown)


nimbus/core/chain/forked_chain.nim
```diff
@@ -632,6 +634,16 @@ func memoryTransaction*(c: ForkedChainRef, txHash: Hash32): Opt[(Transaction, Bl
     return Opt.some( (loc[].tx(index), loc[].number) )
   return Opt.none((Transaction, BlockNumber))
 
+func memoryTxHashesForBlock*(c: ForkedChainRef, blockHash: Hash32): seq[Hash32] =
+  var cachedTxHashes = newSeq[(Hash32, uint64)]()
+  for txHash, (blkHash, txIdx) in c.txRecords.pairs:
+    if blkHash == blockHash:
+      cachedTxHashes.add((txHash, txIdx))
+
+  cachedTxHashes.sorted(proc(a, b: (Hash32, uint64)): int =
```
tersec, line 643: The code is slightly simpler and probably faster (though to be fair, the benchmarks probably wouldn't show much difference -- mainly it's slightly simpler) if one swaps the order of the elements of the tuple in `cachedTxHashes`, then one can use the implied `system.cmp` function and save one call per invocation, the closure, etc.

So
```nim
  var cachedTxHashes = newSeq[(uint64, Hash32)]()
...
      cachedTxHashes.add((txIdx, txHash))
```

Also, three copies happen in the:
- adding to `cachedTxHashes` (can be avoided, probably not worth it)
- `sorted` rather than `sort` (no real reason to use `sorted` here -- it's not external data one has to preserve, `cachedTxHashes` is already a copy; and
- the `mapIt` (also copies)

Practically, the easiest one to remove is the `sorted` copy: use
```nim
  sort(cachedTxHashes)       # with the tuples in "natural" system.cmp sort order no custom comparison needed
  cachedTxHashes.mapIt(it[1])
```

PR #3065, opened 2025-02-13
Update Nimbus EVM code to use the latest nim-evmc which supports EVMC v12.1.0

bhartnett:
Relates to this PR: https://github.com/status-im/nim-evmc/pull/44

### DIFF
--- a/execution_chain/core/eip7691.nim
+++ b/execution_chain/core/eip7691.nim
@@ -25,10 +25,10 @@ func getTargetBlobGasPerBlock*(electra: bool): uint64 =
   else: TARGET_BLOB_GAS_PER_BLOCK.uint64
 
 const
-  EVMForkToFork: array[FkCancun..FkOsaka, HardFork] = [
+  EVMForkToFork: array[FkCancun..FkLatest, HardFork] = [
     Cancun,
     Prague,
-    Osaka
+    Osaka,
   ]
 
 func getMaxBlobsPerBlock*(com: CommonRef, fork: EVMFork): uint64 =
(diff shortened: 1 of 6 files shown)


jangko:
Usage of `EVMFork` and `..EVMFork.high` is to let the compiler produce error code at compile time when new fork is added. Because in those places, adding new fork definition need to add something too. If you replace it with `..FkOsaka`, then the compiler will silent, and we get runtime error instead which is not good.

bhartnett:
> Usage of `EVMFork` and `..EVMFork.high` is to let the compiler produce error code at compile time when new fork is added. Because in those places, adding new fork definition need to add something too. If you replace it with `..FkOsaka`, then the compiler will silent, and we get runtime error instead which is not good.

Hmm ok. The problem is that there has been a new `EVMC_EXPERIMENTAL` type added to `evmc_revision` and `EVMFork` is actually the same type as `evmc_revision`. I guess I can add `EVMC_EXPERIMENTAL` as another fork type to implement the change another way.

execution_chain/core/eip7691.nim
```diff
@@ -25,7 +25,7 @@ func getTargetBlobGasPerBlock*(electra: bool): uint64 =
   else: TARGET_BLOB_GAS_PER_BLOCK.uint64
 
 const
-  EVMForkToFork: array[FkCancun..EVMFork.high, HardFork] = [
```
bhartnett, line 28: @jangko The problem here is that `EVMFork.high` is equal to the `EVMC_EXPERIMENTAL` value which is not actually a valid fork. Do you have any suggestions on how I should fix this?

kdeme: assuming that `EVMC_EXPERIMENTAL` would always move one up from the latest fork, we can just run the loop till high - 1?

jangko: Add something like `FkLatest* = EVMFork.high-1`?

bhartnett: Sure, I'll go with that then.

bhartnett: @jangko I've updated the PR as suggested. Ready for final review.

PR #3081, opened 2025-02-17
update Ubuntu version: 20.04 is EOL

narimiran:


.github/workflows/ci.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         include:
           - target:
               os: linux
-            builder: ubuntu-20.04
+            builder: ubuntu-latest
```
tersec, line 57: In general, for the top-level user applications (Fluffy, the verifying proxy, and the EL, in this repo) it's still better to target a specific Ubuntu version than `ubuntu-latest`, which might switch to Ubuntu 24.04 before 22.04 is EOL.

For the libraries, sure, `ubuntu-latest` is the best way to to approach it, I agree, but for two repositories (`nimbus-eth1` and `nimbus-eth2` it's fine to control this more precisely.

narimiran: > `ubuntu-latest`, which might switch to Ubuntu 24.04 before 22.04 is EOL.

Based on [this](https://github.com/actions/runner-images?tab=readme-ov-file#available-images), `ubuntu-latest` is already 24.04.

> for two repositories (nimbus-eth1 and nimbus-eth2) it's fine to control this more precisely.

Ok, should the switch be made to 22.04 or 24.04?

tersec: 22.04

tersec: https://github.com/status-im/nimbus-eth1/pull/3081/commits/c7466b6850f250f66882a55cd5c931d2743febf5

PR #3082, opened 2025-02-17
wire in forkedchain to p2p

### LINKED ISSUE #2833: Update eth wire protocol handler to use ForkedChain
The wire protocol still access the database directly, likely will return incomplete data to peers. Although it will not pose much harm to other components, peers might have scoring algorithm and classify nimbus-eth1 as useless peer. If too much peers act like that, then nimbus EL will not have enough peers to sync to.

advaita-saha:
Screenshot 2025-02-17 at 4 39 46 PM

fix #2833

### DIFF
--- a/execution_chain/core/chain/forked_chain.nim
+++ b/execution_chain/core/chain/forked_chain.nim
@@ -694,8 +694,8 @@ proc receiptsByBlockHash*(c: ForkedChainRef, blockHash: Hash32): Result[seq[Rece
   c.hashToBlock.withValue(blockHash, loc):
     return ok(loc[].receipts)
   
-  let header = c.headerByHash(blockHash).valueOr:
-    return err("Block hash not found")
+  let header = c.baseTxFrame.getBlockHeader(blockHash).valueOr:
+    return err("Block header not found")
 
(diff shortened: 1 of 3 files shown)


execution_chain/core/chain/forked_chain.nim
```diff
@@ -675,6 +685,20 @@ proc blockByNumber*(c: ForkedChainRef, number: BlockNumber): Result[Block, strin
...
+proc blockHeader*(c: ForkedChainRef, blk: BlockHashOrNumber): Result[Header, string] =
+  if blk.isHash:
+    return c.headerByHash(blk.hash)
+  c.headerByNumber(blk.number)
+
+proc receiptsByBlockHash*(c: ForkedChainRef, blockHash: Hash32): Result[seq[Receipt], string] =
+  c.hashToBlock.withValue(blockHash, loc):
+    return ok(loc[].receipts)
+  
+  let header = c.headerByHash(blockHash).valueOr:
```
jangko, line 697: instead of `c.headerByHash`, why not `c.baseTxFrame.getBlockHeader`?

advaita-saha: Yeah makes sense to do that

PR #3087, opened 2025-02-18
Eip4444 - Portal Integration

advaita-saha:
- [x] rpc level integration
- [x] common block serve through forkedchain
- [x] p2p serving ( dependency #3082  )
- [x] EIP-4444 specific block limit in p2p
- [x] cleanup + comments


execution_chain/portal/portal.nim
```diff
@@ -0,0 +1,163 @@
...
+
+  T(
+    rpc: rpc,
+    limit: 0
+  )
+
+proc rpcProvider*(pc: PortalClientRef): RpcClient =
+  pc.rpc.provider
+
+proc toHeader(blkObj: BlockObject): Header =
```
kdeme, line 63: Note that you can use the native Portal JSON-RPC API that has the data in more convenient encoding than the EL JSON-RPC API.

We also have a helper JSON-RPC Nim API to access that easily here: https://github.com/status-im/nimbus-eth1/blob/master/fluffy/rpc/portal_rpc_client.nim

kdeme: I did just notice that in that portal_rpc_client API we have not yet added the block header by number call.

execution_chain/core/chain/forked_chain.nim
```diff
@@ -611,6 +612,12 @@ func txRecords*(c: ForkedChainRef, txHash: Hash32): (Hash32, uint64) =
 func isInMemory*(c: ForkedChainRef, blockHash: Hash32): bool =
   c.hashToBlock.hasKey(blockHash)
 
+func isHistoryExpiryActive*(c: ForkedChainRef): bool =
+  not c.portal.isNil
```
tersec, line 616: Is this still true for the testnets and other networks for which Portal has no coverage?

advaita-saha: yes

execution_chain/portal/portal.nim
```diff
@@ -0,0 +1,246 @@
...
+  
+  if conf.historyExpiry:
+    let 
+      rpc = conf.getPortalRpc()
+      portalEnabled =
+        if com.networkId == MainNet and rpc.isSome:
+          # Portal is only available for mainnet
+          true
+        else:
+          warn "Portal is only available for mainnet, history expiry won't be available"
```
tersec, line 67: This doesn't seem like it'd allow testing history expiry with e.g., Sepolia?

execution_chain/portal/portal.nim
```diff
@@ -0,0 +1,246 @@
...
+          false
+      limit = 
+        if conf.historyExpiryLimit.isSome:
+          conf.historyExpiryLimit.get()
+        else:
+          com.posBlock().get()
+
+    return T(
```
tersec, line 75: By omitting both the `return` here and in `return nil` a few lines later the whole thing can become an `if` expression and still work

execution_chain/portal/portal.nim
advaita-saha: yeah it does, maybe I should put a better warning message (current one is a bit mis-leading)

execution_chain/config.nim
```diff
@@ -445,6 +445,20 @@ type
         defaultValueDesc: $RpcFlag.Eth
         name: "ws-api" }: seq[string]
 
+      historyExpiry* {.
+        desc: "Enable the data from Portal Network"
+        defaultValue: false
+        name: "history-expiry" }: bool
+
+      historyExpiryLimit* {.
```
tersec, line 453: This is presumably set by spec? It's true that https://eips.ethereum.org/EIPS/eip-4444 is basically a placeholder, but for the pre-merge-only version of 4444 which is the first iteration, no fixed number of blocks would seem to be correct.

execution_chain/portal/portal.nim
```diff
@@ -0,0 +1,246 @@
...
+
+logScope:
+  topic = "portal"
+
+type
+  PortalRpc* = ref object
+    url*: string
+    provider: RpcClient
+
+type
```
tersec, line 29: Can omit this `type` and the whole thing works fine as one `type` section

execution_chain/portal/portal.nim
```diff
@@ -0,0 +1,246 @@
...
+  HistoryExpiryRef* = ref object
+    portalEnabled*: bool
+    rpc: Opt[PortalRpc]
+    limit*: base.BlockNumber # blockNumber limit till portal is activated, EIP specific
+
+proc init*(T: type PortalRpc, url: string): T =
+  let web3 = waitFor newWeb3(url)
```
tersec, line 36: Here, and for all of the other `waitFor`s in the PR, it's better to use `await`, as it integrates better with the existing Chronos event processing loop, results in cleaner stack traces, et cetera.

advaita-saha: True
But I am not sure if that's the best idea, as other parts of the `ForkedChain` is not async functions, so we do need to end there after-all with a `waitFor`, since these fetching functions are after-all used by FC module functions like `blockByHash`, `blockByNumber`, etc

Also making the whole `ForkedChain` async is IMO another discussion we should have, but out of scope of this PR

execution_chain/portal/portal.nim
```diff
@@ -0,0 +1,244 @@
...
+      else:
+        warn "Portal is only available for mainnet, skipping fetching data from portal"
+        false
+    limit = 
+      if conf.historyExpiryLimit.isSome:
+        conf.historyExpiryLimit.get()
+      else:
+        com.posBlock().get()
+
+  return T(
```
tersec, line 77: Don't need this `return`

execution_chain/sync/wire_protocol/handler.nim
```diff
@@ -110,13 +110,27 @@ proc getBlockBodies*(ctx: EthWireRef,
...
 
-    totalBytes += getEncodedLength(body)
-    list.add body
+    # EIP-4444 limit
+    if ctx.chain.isHistoryExpiryActive:
+      if blk.header.number > ctx.chain.portal.limit:
+        trace "handlers.getBlockBodies: blockBody older than expiry limit", blockHash
+        continue
+    
+    let blkBody = blk.body
```
tersec, line 131: This makes an unnecessary copy of the `blk.body`, when blocks can be sizable. Better to just refer to it by `blk.body` in the next two lines. Or, if one wants to give it a separate name/identifier, `template blkBody: untyped = blk.body` and then it just substitutes in at the `getEncodedLength`/`list.add` callsites, no extra copies.

execution_chain/core/chain/forked_chain.nim
```diff
@@ -773,12 +785,35 @@ proc blockBodyByHash*(c: ForkedChainRef, blockHash: Hash32): Result[BlockBody, s
...
+    var unavailable = false
+    let blk = c.baseTxFrame.getEthBlock(number)
+    # Txs not there in db - Happens during era1/era import, when we don't store txs and receipts
+    if blk.isErr or (blk.get.transactions.len == 0 and blk.get.header.transactionsRoot != emptyRoot):
+      unavailable = true
+    else:
+      return blk
+    
+    # Serves portal data if block not found in d
+    if unavailable and c.isPortalActive:
```
tersec, line 815: Either `unavailable` is true or `return blk` has already happened -- this extra variable doesn't change the meaning of this `if number <= c.baseBranch.tailNumber` body.

execution_chain/sync/wire_protocol/handler.nim
```diff
@@ -136,6 +149,16 @@ proc getBlockHeaders*(ctx: EthWireRef,
       return move(list)
     totalBytes = 0
 
+  # EIP-4444 limit
+  if chain.isHistoryExpiryActive:
+    case req.reverse:
```
tersec, line 154: `req.reverse` is a `bool`, so just `if` seems simpler here? `case` is a bit overkill

execution_chain/portal/portal.nim
```diff
@@ -0,0 +1,244 @@
...
+
+proc getBlockByNumber*(historyExpiry: HistoryExpiryRef, blockNumber: uint64, fullTxs: bool = true): Result[Block, string] =
+  debug "Fetching block from portal"
+  try:
+    let 
+      rpc = historyExpiry.rpcProvider.valueOr:
+        return err("Portal RPC is not available")
+      res = waitFor rpc.eth_getBlockByNumber(blockId(blockNumber), fullTxs)
+    if res.isNil:
+      return err("Block not found in portal")
```
tersec, line 184: Probably good to be consistent about capitalization of "Portal", at least in user-facing parts of the code (log messages and `err` strings ). Currently it sometimes uses "Portal" and sometimes "portal"; the former seems preferable, being that it is a proper noun.

tersec: Hm, ok, it is being consistent and treating it as a non-proper noun -- it's capitalized when and only when it's the first word. But taking cues from https://ethereum.org/en/developers/docs/networking-layer/portal-network/ that treats it as a proper noun.

execution_chain/core/chain/forked_chain.nim
```diff
@@ -693,6 +695,12 @@ func txRecords*(c: ForkedChainRef, txHash: Hash32): (Hash32, uint64) =
 func isInMemory*(c: ForkedChainRef, blockHash: Hash32): bool =
   c.hashToBlock.hasKey(blockHash)
 
+func isHistoryExpiryActive*(c: ForkedChainRef): bool =
+  not c.portal.isNil
+
+func isPortalActive*(c: ForkedChainRef): bool =
```
tersec, line 701: `isPortalActive` doesn't appear to be used outside `execution_chain/core/forked_chain`; if so it doesn't need to be exported

advaita-saha: This is kinda intentional for later use
but yeah best to not export it, unless it's needed

tersec: Yeah, the issue is that things which get marked that way never show up as `XDeclaredButNotUsed`, so it requires manual effort to find unused code. The exporting can be added as part of the PR which actually uses it.

Issue #3002, opened 2025-01-15
Need to upgrade FC module to handle very long branch or non finalized branch


jangko:
This feature is needed for #2995 

- [ ] Algorithm changes in FC module
- [ ] Possibly need to serialize aristo txFrame
- [ ] Requires #2960

jangko:
Copied from discord discussion:

basically, the first step that is needed is to keep track of the latest arrived fcu in forkedchain .. there are two cases:
we know the header -> we can store the "latest finalized block number" and use that in logic surrounding base updates and other things
we don't know the header -> we store the hash in a "pendingFcu" field - then, when we receive any header (either via newpayload or from the sync), can simply inform FC about it and FC can check if it's in the "pendingFCu" entry and promote it to to latestFinalizedNumber, then continue the above logic

in particular, there's no "long async queue" involved here.
once that's done, we can start reasoning about storing the "header chain" directly in the database instead of using the intermediate table that is used now 
also - cc @mjfh - we need to get rid of those out-of-band persist calls - the syncer must not interfere with the base flow - if need be, it can even have its own column family which is a lot easier now the backend has been cleaned of excessive abstraction

the other critical change needed to FC is to remove all writes to baseTxFrame essentially, such as here: https://github.com/status-im/nimbus-eth1/blob/9244cc9a300f4fd93db0fbe3f2c089ab542963ed/execution_chain/core/chain/forked_chain.nim#L333

that head update, it should not be written to the base txframe but rather to the txframe of the block being selected has head - same for finalized etc, they should always go into the same frame as the block that gets selected so that when we write the base to disk, we don't accidentally write a head hash that is not yet on disk

this "lastestFinalizedNumber" is suffcient to implement the correct "base update" logic after importBlock, ie block arrives, we check against latest known finalized number, and if the finalized number is far in the future, we can update the base (with some lag - ie the base "target" becomes roundDown32(min(max(head-distance, lastFinalizedNumber)) where head is the latest block we imported, lastestFinalizedNumber is the current finality point according to the CL (ie it can be both higher and lower than head depending on whether we're in sync or not) 
the syncer has the right idea here, ie it does something similar - it's just implemented in the wrong place, ie https://github.com/status-im/nimbus-eth1/blob/9244cc9a300f4fd93db0fbe3f2c089ab542963ed/execution_chain/core/chain/forked_chain/chain_header_cache.nim#L83

arnetheduck:
[...]

PR #3095, opened 2025-02-19
Add block quarantine to ForkedChain

### LINKED ISSUE #3056: Add quarantine to forkedchain
When syncing, the syncer works with stale information with respect to the latest blocks arriving from the CL - as such, when the syncer is about to finish, it's  likely that the chain has moved on to a new block or blocks that the EL has observed and discarded. 

Instead of using the syncer to top up this end-game gap, a quarantine in FC can hold up to N blocks that arrive from the API and/or elsewhere (doesn't matter where it comes from really) and apply these - it works the following way:

* when a block arrives whose parent is unknown, it is put in a bounded quarantine
* every time a block with a known parent is successfully added to FC, the quarantine is examined for blocks whose parent is now known (the block just added).
* These blocks are applied, removing them from the quarantine and repeating the process until no more blocks can be applied.

jangko:
fix #3056


execution_chain/core/chain/forked_chain/block_quarantine.nim
```diff
@@ -0,0 +1,51 @@
...
+
+type
+  Quarantine* = object
+    ## Keeps track of unvalidated blocks coming from the network
+    ## and that cannot yet be added to the chain
+    ##
+    ## This only stores blocks that cannot be linked to the
+    ## ForkedChain due to missing ancestor(s).
+
+    orphans: OrderedTable[Hash32, Block]
```
arnetheduck, line 27: `minilru` is probably the better choice here, which should simplify the addOrphan logic (and perform better)

execution_chain/core/chain/forked_chain/block_quarantine.nim
```diff
@@ -0,0 +1,51 @@
...
+# at your option. This file may not be copied, modified, or distributed except
+# according to those terms.
+
+import
+  std/tables,
+  results,
+  eth/common/blocks
+
+const
+  MaxOrphans = 128
```
arnetheduck, line 17: we will _probably_ need to introduce an async queue or a similar construct to process quarantine blocks - if we proces 128 blocks in one async time slice, that will likely disconnect everyone due to the long "downtime"..

in the meantime, I think it's enough that we keep 32:ish blocks here, that's already 6 minutes of block time.

advaita-saha: Yeah agreed, mostly the sync lags by maximum of 10 blocks
32 seems like a more reasonable number to keep

jangko: After I rethink about it, the block quarantine also overlap with `importHeader` functionality, especially when importing blocks to FC. They can share the same async worker. And somehow related to #3002

arnetheduck: `also overlap with` - not necessarily, ie it doesn't need headers to perform its work - when the normal syncer reaches a point here quarantine blocks exist, there's no need to separately download headers since the blocks in the quarantine already have headers too.

arnetheduck: ie I think the quarantine, even just as is, is a step forward towards stable sync

execution_chain/core/chain/forked_chain/block_quarantine.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/3095/commits/a079c86a9966fff93469b08bd820f77b0f164b89
requires https://github.com/status-im/nim-minilru/pull/3

advaita-saha:
Currently the block quarantine implementation has no effect on improving sync since,

> when a block arrives whose parent is unknown, it is put in a bounded quarantine

Here the arrival of the block means also means `payload` coming from `engineAPI`. Thus when a payload with unknown parent hash comes from the `engineAPI` call `newPayload` we should put it in the block quaratine, to be later used to complete the remaining ~10 blocks sync. _Not happening in the current impl_

So in the current implementation when a `payload` comes to us, with a unknow parent we are doing this. 
https://github.com/status-im/nimbus-eth1/blob/44db666c931ea2cb40acdbdf400b8422c724d429/execution_chain/beacon/api_handler/api_newpayload.nim#L174-L181

And currently we are only using the quarantine when we are trying to import a block into forked chain, which is fine. But in all the payloads we received with missing parent, are never added to the quarantine so in real life scenario never helps us
https://github.com/status-im/nimbus-eth1/blob/b3150eca6daad02d33d5260a808b396d7d8e628f/execution_chain/core/chain/forked_chain.nim#L523-L545

Ideally when the sync is happening we are getting a lot of `newPayload` and `fCU` calls from the CL. Now these payloads are _possible_ future parts of the chain when we are syncing older blocks. And we want to use these received payloads to complete the sync _if possible_ i.e validity checked.  This helps the syncer not lag behind by 7-10 blocks every sync cycle
That being said, it implies also using the block quarantine _before_ calling the syncer in `fCU` call and `newPayload` with unknow parent to check we can construct the chain

execution_chain/core/chain/forked_chain/block_quarantine.nim
```diff
@@ -0,0 +1,45 @@
...
+# according to those terms.
+
+import
+  std/tables,
+  results,
+  minilru,
+  eth/common/blocks
+
+const
+  MaxOrphans = 32
```
arnetheduck, line 18: let's reduce to 8 and merge as is - 8 is already >1minute of blocks which will help the syncing end game significantly - the async aspect should probably be dealt with outside of the orphan list, ie _likely_ with an `AsyncQueue` somewhere, and this somewhere should cover both syncing and quarantine blocks, but in the meantime, a small buffer like 8 here can be processed in one round.

arnetheduck:
>  But in all the payloads we received with missing parent, are never added to the quarantine so in real life scenario never helps us

indeed, it's important that newPayload blocks are added to quarantine in with a round-robin eviction strategy

jangko:
Perhaps we only need to harmonize between `newPayload` own queue and FC quarantine.
The simplest thing we can do is put block into `quarantine` when `newPayload` put the block into its own queue.

PR #3097, opened 2025-02-19
Adjust and activate post-merge proofs for block headers

kdeme:
Not yet fully activated for proofs based on historical summaries as the historical summaries do not get loaded yet.

This is rather a quick version of this to be able to test https://github.com/ethereum/portal-spec-tests/pull/39

Also, I'm not sure if I like the change of `Vector` instead of `List` for the `ExecutionBlockProof` for capella / deneb / etc. But it is currently what we decided in specs.

### DIFF
--- a/fluffy/network/history/validation/block_proof_historical_summaries.nim
+++ b/fluffy/network/history/validation/block_proof_historical_summaries.nim
@@ -1,5 +1,5 @@
 # Fluffy
-# Copyright (c) 2023-2024 Status Research & Development GmbH
+# Copyright (c) 2023-2025 Status Research & Development GmbH
 # Licensed and distributed under either of
 #   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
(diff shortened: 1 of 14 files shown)


fluffy/network/history/validation/block_proof_historical_summaries.nim
```diff
@@ -28,22 +28,63 @@ import
...
+    # - 16 as there are 14 fields
+    # - 12 as pos of field is 12
+    gIndex = GeneralizedIndex(gIndexMidLevel * 1 * 16 + 12)
+
+  var proof: seq[Digest] = newSeq[Digest](11)
+  ?beaconBlock.build_proof(gIndex, proof)
+
+  ok(ExecutionBlockProofList(proof))
+
+func verifyProof*(
```
kdeme, line 78: So with this proof in the specs currently being a List, we would have to also verify the List size per different fork. (or even if there would only be the current fork, as the List can just be too small)

kdeme:
Closing in favour for https://github.com/status-im/nimbus-eth1/pull/3118 which only enables the historical roots based proofs for now.

PR #3099, opened 2025-02-20
Beacon sync recalibrate blocks queue

mjfh:


execution_chain/sync/beacon/worker/blocks_staged.nim
```diff
@@ -147,8 +147,8 @@ func blocksStagedCanImportOk*(ctx: BeaconCtxRef): bool =
       if ctx.pool.nBuddies == 0:
         return true
 
-      # Start importing if the queue is long enough.
-      if ctx.pool.blocksStagedQuLenMax <= ctx.blk.staged.len:
+      # Start importing if the queue is filled up enough
```
arnetheduck, line 150: why wait here? 1 block should be enough to start importing, no?

mjfh: When importing while peers are still downloading you will most certainly run into timeouts so loosing the peers.

arnetheduck: how is this different when you have more than N blocks in the queue already? ie is it ok to download and process blocks in parallel when there are many blocks resting in the queue but not when there are few?

arnetheduck: also, when did you last test this? there were many bugs in the peer management itself as well as performance issues that have been fixed.

mjfh: I can give it a try to do it simultaneously, importing and downloading. We'll see (as you see these are mere flags to be set.)

PR #3100, opened 2025-02-21
Nimbus Verified Proxy

chirag-parmar:


bhartnett:
The approach you've taken here is interesting. 

Unfortunately, it will only work if the RPC provider returns the correct access list. We don't want to trust the RPC provider and so we can't really trust that they will provide a correct access list. They could for example provide you with an access list that will cause the eth_call result to be completely different, even though you are verifying the account and storage proofs of each account and slot there is no way to verify that you are looking up the correct pre-state before the transaction is executed.

I think using the access list purely as a performance optimization would be ok as long as the EVM itself still looks up the correct state during transaction execution. In this scenarios the worst that the RPC provider could do is a DOS by sending you a very large fake access list but since the RPC provider can already effectively DOS your application by not responding to requests this is likely acceptable.

bhartnett:
> Unfortunately, it will only work if the RPC provider returns the correct access list. We don't want to trust the RPC provider and so we can't really trust that they will provide a correct access list. They could for example provide you with an access list that will cause the eth_call result to be completely different, even though you are verifying the account and storage proofs of each account and slot there is no way to verify that you are looking up the correct pre-state before the transaction is executed.

You might be able to solve this problem by building up the access list yourself by executing the in memory EVM multiple times, each time collecting more of the access list keys and state until eventually executing the transaction against the full state which will give you the result for `eth_call`.

Here is the algorithm:
```
- Fetch code to be executed
- while the evm execution is not successful or the touched account,code,slots have changed since the last execution:
    - execute the code using in memory EVM
    - collect the touched account,code,slots from the evm state
    - fetch all touched account,code,slots concurrently and pass into a new evm state
```
The last execution will be successful and should be using the correct keys. 

You can use the `makeMultiKeys` function in ledger.nim to collect the touched account keys and storage slots from the EVM state after executing the transaction. You might also be able to use the accessList functions for this.

nimbus_verified_proxy/rpc/rpc_eth_api.nim
```diff
@@ -13,194 +13,259 @@ import
...
-    lcProxy.blockCache.getPayloadByHash(blockHash)
+#  lcProxy.proxy.rpc("eth_getBlockByNumber") do(
+#    blockTag: BlockTag, fullTransactions: bool
+#  ) -> Opt[BlockObject]:
+#    lcProxy.getBlockByTag(blockTag)
+#
+#  lcProxy.proxy.rpc("eth_getBlockByHash") do(
+#    blockHash: Hash32, fullTransactions: bool
+#  ) -> Opt[BlockObject]:
+#    lcProxy.blockCache.getPayloadByHash(blockHash)
```
kdeme, line 263: What is the plan here now with the removal of the block cash?

"Recent" blocks were coming in and being stored in that cache. Now that you only have the headers, is the idea to request blocks over the regular EL JSON-RPC API?

Also, with the headers only cache, you are already kind of storing those headers via the consensus light client, albeit not directly addressable by hash.

chirag-parmar: > "Recent" blocks were coming in and being stored in that cache

These were blocks were directly imported through the p2p layer. Their verification defeats the purpose of a light client. I am not sure why we were doing that before.

> Now that you only have the headers, is the idea to request blocks over the regular EL JSON-RPC API?

Yes, the idea is that exactly. The idea is to save on bandwidth, first by *not* downloading consensus block data and second by downloading the whole block on a need-only basis. Furthermore, I expect the state to be queried more often than the transactions.

> you are already kind of storing those headers via the consensus light client, albeit not directly addressable by hash

I looked into the eth2 code base and it didn't seem like the light client stored the update (`LightClientOptimisticUpdate` and `LightClientFinalityUpdate`) headers anywhere. But it is possible that I missed looking at some code. Could you please point me to it?

kdeme: > I am not sure why we were doing that before.

I think they were there because before capella fork the `LightClientHeader` did not hold the `ExecutionPayloadHeader` yet. And thus the only way to have that data (and all the roots to verify with) was by having the recent blocks.

Now (well, since capella) it indeed makes much less sense to get all that data as it probably was/is the main bandwidth consumer.





> I looked into the eth2 code base and it didn't seem like the light client stored the update (`LightClientOptimisticUpdate` and `LightClientFinalityUpdate`) headers anywhere. But it is possible that I missed looking at some code. Could you please point me to it?

No, you're right about this. I was mixing it up with something else.

kdeme:
I think it would be a good plan to already put this PR in a cleaned-up state that would be ready to review / ready to merge.

Your changes into using the json-rpc API instead of the libp2p enabled block cache right now is already a good improvement that can go into the code base. No need to wait for other improvements/features, those can go in other PRs (this avoids to-big-to-review PRs and brings quicker improvements to the web3 ~~verifier~~ purifier)

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -117,114 +146,55 @@ proc run*(
...
+    # light client is set to optimistic finalization mode
     lightClient = createLightClient(
       network, rng, lcConfig, cfg, forkDigests, getBeaconTime, genesis_validators_root,
       LightClientFinalizationMode.Optimistic,
     )
 
-  verifiedProxy.installEthApiHandlers()
-
-  info "Listening to incoming network requests"
+  # find out what this does
```
kdeme, line 155: This registers the basic protocol that allows for ping/getMetadata/goodbye for a node in libp2p. 
I think it should remain available for basic communication with nodes on the network.

Eventually when you move to add the REST API sync version, this can be either be removed if we are to only support REST. Or if we want to support both modes, this gets disabled in case of REST mode.

nimbus_verified_proxy/types.nim
```diff
@@ -0,0 +1,29 @@
+# nimbus_verified_proxy
+# Copyright (c) 2022-2025 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
```
kdeme, line 7: add `{.push raises: [].}`

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -117,114 +146,55 @@ proc run*(
...
+
+  # verify chain id that the proxy is connected to
   waitFor verifiedProxy.verifyChaindId()
 
   proc onFinalizedHeader(
       lightClient: LightClient, finalizedHeader: ForkedLightClientHeader
   ) =
     withForkyHeader(finalizedHeader):
-      when lcDataFork > LightClientDataFork.None:
+      when lcDataFork > LightClientDataFork.Altair:
```
kdeme, line 173: Curious to know why this had to be upped 1 fork?

nimbus_verified_proxy/header_store.nim
```diff
@@ -0,0 +1,141 @@
...
+  std/tables,
+  beacon_chain/spec/beaconstate,
+  beacon_chain/spec/datatypes/[phase0, altair, bellatrix],
+  beacon_chain/[light_client, nimbus_binary_common],
+  beacon_chain/el/engine_api_conversions,
+  minilru,
+  results
+
+type HeaderStore* = ref object
+  headers: LruCache[Hash32, Header]
```
kdeme, line 23: I guess it works the use an `LruCache` because of the way you have implemented it (using `peek` for the getters and scanning over the cache before add), but to me it does feel like a non-intuitive way to use an LruCache. 

It is basically being used/selected mostly for its "automatic" pruning of the oldest headers here, but not at all for keeping its most "touched" content, hence the use of only `peek` and the requirement to add only new data.

The required full scan on add also makes the efficient add of the lru cache undone, but with the set max value of `64` that's probably not really an issue currently.

This implementation is also easy to make mistakes against in future changes, so it should at least have a big warning on how to use (already has this warning on the add currently).

nimbus_verified_proxy/nimbus_verified_proxy.nim
chirag-parmar: I think I forgot the comment in there. Thank you for the explanation

bhartnett:
We have existing users of the Nimbus Verified Proxy and so we need to be careful not to break any of the functionality, so in my opinion this PR is too big to be merged, especially because we don't have a staging branch etc. I also noticed that there are no tests added.

I would suggest treating this as a draft PR to get feedback and then create smaller PRs with smaller subsets of the features and have some unit tests covering the changes in each PR. Just my thoughts, curious to hear what others in the team think. @kdeme @tersec

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -73,16 +68,40 @@ proc run*(
...
       notice "commandLineParams() exception"
 
+  # load constants and metadata for the selected chain
+  let metadata = loadEth2Network(config.eth2Network)
+
+  # initialze verified proxy
   let
-    metadata = loadEth2Network(config.eth2Network)
     chainId = getConfiguredChainId(metadata)
+    authHooks = @[httpCors(@[])] # TODO: for now we serve all cross origin requests
```
bhartnett, line 77: Why did you need to introduce this CORS change when it previously wasn't needed?

nimbus_verified_proxy/header_store.nim
```diff
@@ -0,0 +1,141 @@
...
+  hashes: Table[base.BlockNumber, Hash32]
```
bhartnett, line 24: It appears that the `hashes` table only grows and the entries never get deleted. Maybe you should use a LRUCache here as well.

nimbus_verified_proxy/header_store.nim
```diff
@@ -0,0 +1,141 @@
...
+
+proc earliest*(self: HeaderStore): Opt[Header] =
+  if self.headers.len() == 0:
+    return Opt.none(Header)
+
+  var hash: Hash32
+  for h in self.headers.keys:
+    hash = h
+
+  self.headers.peek(hash)
```
bhartnett, line 129: Why not just store the earliest and latest block number (or hash) and then use them to return the earliest and latest headers.

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -73,16 +68,40 @@ proc run*(
...
+    rpcProxy = RpcProxy.new(
+      [initTAddress(config.rpcAddress, config.rpcPort)], clientConfig, authHooks
+    )
+
+    headerStore = HeaderStore.new(64) # block cache contains blocks downloaded from p2p
 
+  let com = CommonRef.new(
+    DefaultDbMemory.newCoreDbRef(),
+    taskpool = nil,
+    config = chainConfigForNetwork(chainId),
```
bhartnett, line 90: Note that chainId is not always the same as networkId. This `chainConfigForNetwork` function expects a networkId so I'm not sure if there might be an issue here.

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,104 @@
...
+  eth/trie/[ordered_trie, trie_defs],
+  json_rpc/[rpcproxy, rpcserver, rpcclient],
+  web3/[primitives, eth_api_types, eth_api],
+  ../types,
+  ./blocks
+
+export results, stint, hashes_rlp, accounts_rlp, eth_api_types
+
+template rpcClient(vp: VerifiedRpcProxy): RpcClient =
+  vp.proxy.getClient()
```
bhartnett, line 28: This `rpcClient` template is defined in each file. Would be better to define this in just one place and then import.

nimbus_verified_proxy/rpc/accounts.nim
```diff
@@ -0,0 +1,173 @@
...
+  json_rpc/[rpcproxy, rpcserver, rpcclient],
+  web3/[primitives, eth_api_types, eth_api],
+  ../types
+
+export results, stint, hashes_rlp, accounts_rlp, eth_api_types
+
+template rpcClient(vp: VerifiedRpcProxy): RpcClient =
+  vp.proxy.getClient()
+
+proc getAccountFromProof(
```
bhartnett, line 29: It appears that this function is now defined in two places. The other one in validate_proof.nim

nimbus_verified_proxy/rpc/rpc_eth_api.nim
```diff
@@ -8,199 +8,305 @@
...
-  lcProxy.proxy.rpc("eth_getCode") do(
-    address: Address, quantityTag: BlockTag
+  vp.proxy.rpc("eth_call") do(
+    tx: TransactionArgs, blockTag: BlockTag, optimisticStateFetch: Opt[bool]
   ) -> seq[byte]:
+    if tx.to.isNone():
+      raise newException(ValueError, "to address is required")
+
+    if blockTag.kind == bidAlias:
+      raise newException(ValueError, "tag not yet implemented")
```
bhartnett, line 203: Why is tag not implemented? Looks like this was just copied from Fluffy.

nimbus_verified_proxy/rpc/rpc_eth_api.nim
```diff
@@ -8,199 +8,305 @@
...
+    return callResult.output
+
+  vp.proxy.rpc("eth_createAccessList") do(
+    tx: TransactionArgs, blockTag: BlockTag, optimisticStateFetch: Opt[bool]
+  ) -> AccessListResult:
+    if tx.to.isNone():
+      raise newException(ValueError, "to address is required")
+
+    if blockTag.kind == bidAlias:
+      raise newException(ValueError, "tag not yet implemented")
```
bhartnett, line 224: Same here. Tag should be supported.

nimbus_verified_proxy/rpc/evm.nim
```diff
@@ -0,0 +1,41 @@
...
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+import
+  ../../fluffy/evm/async_evm, ../../fluffy/evm/async_evm_backend, ./accounts, ../types
+
+export async_evm, async_evm_backend
+
```
bhartnett, line 12: In order to get reasonable performance you should create a cache for the most recently fetched accounts, slots and code. This should be populated after validation when fetching this data from the downstream RPC provider.

nimbus_verified_proxy/rpc/rpc_eth_api.nim
```diff
@@ -8,199 +8,305 @@
...
     let
-      blk = lcProxy.getBlockByTagOrThrow(quantityTag)
-      blockNumber = blk.number.uint64
+      header = (await vp.getHeaderByTag(blockTag)).valueOr:
+        raise newException(ValueError, error)
+
+      optimisticStateFetch = optimisticStateFetch.valueOr:
+        true
+
+    let callResult = (await vp.evm.call(header, tx, optimisticStateFetch)).valueOr:
```
bhartnett, line 212: As a performance optimization, before executing call on the evm, I would suggest calling `eth_createAccessList` on the downstream RPC provider and then using the returned keys to concurrently fetch the accounts and storage slots to populate a cache which should be used by the AsyncEvm backend.

nimbus_verified_proxy/header_store.nim
kdeme: FYI: I do realize that the original `BlockCache` was implemented in a similar way, and that you improved it somewhat due to using peek (peek was not part of the API at creation of BlockCache I believe). But as you are reworking this I mostly wanted to point out that I do not think it is something we necessarily need to stick with as solution for this cache.

Also, as @bhartnett pointed out, there is an issue with pruning the number to hashes mapping table, which will be difficult to solve cleanly as the LRU prunes behind the scenes (without hook into it).

tersec:
> We have existing users of the Nimbus Verified Proxy and so we need to be careful not to break any of the functionality, so in my opinion this PR is too big to be merged, especially because we don't have a staging branch etc. I also noticed that there are no tests added.
> 
> I would suggest treating this as a draft PR to get feedback and then create smaller PRs with smaller subsets of the features and have some unit tests covering the changes in each PR. Just my thoughts, curious to hear what others in the team think. @kdeme @tersec
> 
> Another option I guess would be to create a staging branch where the changes can be tested a bit longer before being merged into master but this would not be my preferred approach.

I'd echo this. Also, yes, it's possible to keep it in staging longer, unless it needs to be a larger PR, it would better to split out individual changes from this to review and merge, then this branch can be updated to reflect those being merged.

kdeme:
> Just my thoughts, curious to hear what others in the team think.

Verified proxy is definitely lacking a good set of tests which currently makes it difficult making changes without breaking things. In the past the only testing done was basically running it with a wallet (e.g. Metamask) and testing some end-to-end functionality there.

Now as going forward and moving this "PoC" into a properly maintained and developed product, and adding more functionality to it, we should definitely have more unit testing (and especially for this newly added functionality). To be fair, the current version does not really have much functionality, libp2p consensus light client, which is already tested, some basic proving and this block cache.
In terms of testing the JSON-RPC functionality (end-to-end tests), I think we can have tests with some static response data (mock the response basically) and load that in the tests just to verify the underlying logic of our JSON-RPC methods.


Regarding smaller PRs: Yes, I always prefer to have features split up (easier to review, to test, to revert, etc.). Practically for this PR I think the block cache changes could be a PR that gets merged (once comments are addressed), and the addition and usage of async EVM could be another (or even several depending on which API method starts to use them already). I noticed that there is currently an issue that causes a lot of the JSON-RPC code to be commented. Separating PRs would allow us to already get some of this functionality merged.

A staging branch I personally do not prefer, in general, but also in this case specifically because we do not  have a way to really verify/test that it works for users. Aside from using it ourselves for some time.
Note that this sort of also applies to the big vs smaller PRs comment, but with that difference that with smaller PRs we can potentially diagnose issues quicker (git bisect, etc.) and revert them cleaner. 
And in a way, our master is basically the staging branch, as we are not doing releases yet :)

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -0,0 +1,284 @@
...
+      if hLatest.isSome:
+        return ok(hLatest.get().number)
+      else:
+        return err("Couldn't get the latest block number from header store")
+    else:
+      return err("No support for block tag " & $blockTag)
+  else:
+    return ok(base.BlockNumber(distinctBase(blockTag.number)))
+
+proc convHeader(blk: eth_api_types.BlockObject): Header =
```
tersec, line 46: https://status-im.github.io/nim-style-guide/language.proc.html
> Prefer `func` - use `proc` when side effects cannot conveniently be avoided.

But there appear to be no side effects here to begin with.

nimbus_verified_proxy/header_store.nim
```diff
@@ -0,0 +1,141 @@
+# nimbus_verified_proxy
+# Copyright (c) 2022-2025 Status Research & Development GmbH
```
tersec, line 2: How is this file from 2022-2024? It pops into existence in this PR.

nimbus_verified_proxy/rpc/accounts.nim
```diff
@@ -0,0 +1,173 @@
+# nimbus_verified_proxy
+# Copyright (c) 2022-2025 Status Research & Development GmbH
```
tersec, line 2: This file appears to have no pre-2025 existence.

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -0,0 +1,284 @@
...
+    return err("hashed block header doesn't match with blk.hash(downloaded)")
+
+  if blockHash != blk.hash:
+    return err("the downloaded block hash doesn't match with the requested hash")
+
+  let earliestHeader = self.headerStore.earliest.valueOr:
+    return err("syncing")
+
+  # walk blocks backwards(time) from source to target
+  let isLinked = await self.walkBlocks(
```
kdeme, line 132: I think we should have some limit here of how much back in time you can request a block.

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,104 @@
...
+
+template toLog(lg: LogObject): Log =
+  Log(address: lg.address, topics: lg.topics, data: lg.data)
+
+proc toLogs(logs: openArray[LogObject]): seq[Log] =
```
tersec, line 33: `func` (etc)

also https://status-im.github.io/nim-style-guide/language.result.html
> Avoid using `result` for returning values.

nimbus_verified_proxy/rpc/rpc_eth_api.nim
```diff
@@ -8,199 +8,305 @@
...
+        raise newException(ValueError, error)
 
-    info "Forwarding eth_getTransactionCount", blockNumber
+    for r in rxs:
+      if r.transactionHash == txHash:
+        return r
+
+    raise newException(ValueError, "receipt couldn't be verified")
+
+  #  # eth_blockNumber - get latest tag from header store
```
tersec, line 145: More commented-out code. This is fine as an implementation tracking thing in a draft PR, but preferably not to merge.

nimbus_verified_proxy/nimbus_verified_proxy.nim
chirag-parmar: So I rearranged the boot code for the verified proxy to have a proper initialization flow. I did that to understand the verified proxy's workings and the interplay of the light client with it. 

The CORS change is just a rearrangement diff i.e it existed before.

cc: @kdeme

fluffy/evm/async_evm.nim
```diff
@@ -108,6 +108,9 @@ proc init*(
 
   AsyncEvm(com: com, backend: backend)
 
+proc init*(T: type AsyncEvm, backend: AsyncEvmStateBackend, com: CommonRef): T =
+  AsyncEvm(com: com, backend: backend)
```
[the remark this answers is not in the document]
chirag-parmar: I did not want to initialize two instances `CommonRef`. More so because the other instance is merely used for gas price APIs (which internally use `CommonRef` for fork calculation, I assume)

nimbus_verified_proxy/header_store.nim
chirag-parmar: In retrospect, that makes a lot of sense, i'll add it in. Initially, it did not hurt to let it grow because even a million blocks would amount to only ~32MB but since we are probably also aiming for resource constrained embedded devices and not just smartphone wallets it makes sense to add a limit. Regardless of the resource constraints, it still does make sense to add a limit and just set it to a high number

nimbus_verified_proxy/rpc/accounts.nim
```diff
@@ -0,0 +1,173 @@
...
+      codeHash: accountCodeHash,
+    )
+    accountEncoded = rlp.encode(acc)
+    accountKey = toSeq(keccak256((accountAddress.data)).data)
+
+  let proofResult = verifyMptProof(mptNodesBytes, stateRoot, accountKey, accountEncoded)
+
+  case proofResult.kind
+  of MissingKey:
+    return ok(EMPTY_ACCOUNT)
```
tersec, line 53: Don't need any of these three `return`s explicitly, that way it becomes a case expression rather than a case statement and Nim effectively statically checks that this is either the last thing in a block/function or is discarded.

chirag-parmar:
> Separating PRs would allow us to already get some of this functionality merged without that commented code holding it back.

@bhartnett @tersec @kdeme the reasoning is very sound. I'll create smaller PRs which addresses the reviews and also seperates out code. I'll link the new PRs in your reviews here. Thanks a ton for these reviews, learning a lot

nimbus_verified_proxy/rpc/rpc_eth_api.nim
chirag-parmar: Indeed most of the code for evm execution was taken directly from fluffy. I'm fairly sure I removed this bit of code, but I think I lost the diff while stashing some changes for the Pr on AsyncEVM. Will add it back in(i.e. remove the code)

nimbus_verified_proxy/header_store.nim
bhartnett: Yeah it can hold more items but it should still be bounded.

fluffy/evm/async_evm.nim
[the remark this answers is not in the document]
bhartnett: We probably don't need an instance of `CommonRef` in the NVP just for determining a fork. After looking at that area of code, it appears that you should be able to do something like this to get the fork without creating an instance of `CommonRef`:
```
let forkTransitionTable = config.toForkTransitionTable() # store this somewhere precomputed

ToEVMFork[toHardFork(forkTransitionTable, forkDeterminationInfo(header))]
```

I haven't tried this by it should be possible even with some small changes to Nimbus code if needed.

nimbus_verified_proxy/nimbus_verified_proxy.nim
chirag-parmar: the merge happened at bellatrix i.e. execution payload is available only starting from bellatrix. Using the light client for verified proxy pre-bellatrix doesnt make sense.

tersec: Ok, worth commenting as such in the code along these lines.

PR #3110, opened 2025-02-27
Collect keys of all read storage slots even if zero when building MultiKeys

bhartnett:
This change is needed for the Fluffy EVM which uses the Nimbus in memory EVM. 

In this case the EVM is looking up a slot that hasn't yet been set so it returns a zero value. It fails with InvalidJumpDest because the transaction doesn't yet have the correct state set.

In Fluffy and also Nimbus Verified Proxy we need to figure out what state is accessed when the transaction is executed so that we can fetch that state over the network, store it in the memory evm and then finally execute the transaction with the correct state. In this case we still want to know about the slot keys accessed that returned zero values so that we can look them up over the network to get the correct non zero values.

### DIFF
--- a/execution_chain/db/ledger.nim
+++ b/execution_chain/db/ledger.nim
@@ -790,7 +790,6 @@ proc update(wd: var WitnessData, acc: AccountRef) =
 
   if not acc.originalStorage.isNil:
     for k, v in acc.originalStorage:
-      if v.isZero: continue
       wd.storageKeys.incl k
 
   for k, v in acc.overlayStorage:
--- a/tests/test_ledger.nim
+++ b/tests/test_ledger.nim
@@ -713,6 +713,45 @@ proc runLedgerBasicOperationsTests() =
       check 2.u256 in vals
       check 3.u256 in vals
 
+    test "Test MultiKeys - Set storage":
+      var
+        ac = LedgerRef.init(memDB.baseTxFrame())
+        addr1 = initAddr(1)
+
+      ac.setStorage(addr1, 1.u256, 1.u256) # Non-zero value
+      ac.setStorage(addr1, 2.u256, 0.u256) # Zero value
+
+      ac.collectWitnessData()
+      let multikeys = ac.makeMultiKeys().keys
+
+      check:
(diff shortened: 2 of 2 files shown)


bhartnett:
FYI @chirag-parmar you will likely need this change when implementing `eth_call` in the Nimbus Verified Proxy.

bhartnett:
> this fix looks off though - the way I remember things, the way the evm is specified is that _all_ slots implicitly have zero as value unless they've been changed to something else - this implies that the fix should be in the code that attempts to read the slot and not here.

Actually I made a mistake in my description of the problem and reason for this change. In this case the EVM is looking up a slot that hasn't yet been set so it returns a zero value. It fails with InvalidJumpDest because the transaction doesn't yet have the correct state set. So the EVM is actually behaving correctly.

In Fluffy and also Nimbus Verified Proxy we need to figure out what state is accessed when the transaction is executed so that we can fetch that state over the network, store it in the memory evm and then finally execute the transaction with the correct state. In this case we still want to know about the slot keys accessed that returned zero values so that we can look them up over the network to get the correct non zero values.

Note that MultiKeys are not used anywhere else in the codebase and this minor change in behavior is required in order to support the above use case.

execution_chain/db/ledger.nim
```diff
@@ -790,7 +790,6 @@ proc update(wd: var WitnessData, acc: AccountRef) =
 
   if not acc.originalStorage.isNil:
     for k, v in acc.originalStorage:
```
arnetheduck, line 792:
```suggestion
    for k in acc.originalStorage.keys():
```

arnetheduck (approved):
Ok .. it feels like there should exist a test case covering this scenario, ie ideally we'd add a test that verifies this behavior so it doesn't get lost.

jangko:
| Note that MultiKeys are not used anywhere else in the codebase

That's correct, it was created for now obsolete block witness protocol. And if verified proxy requires all storage slot keys, then the fix is correct. The original use case is for replicating the state of a block, and zero value means the slot is removed. But I think verified proxy want different thing.

PR #3116, opened 2025-03-03
Tidying up wire protocol handler and add limit

### LINKED ISSUE #3102: Put limit to eth/68 methods: GetHeaders, GetBlockBodies, GetReceipts, and GetPooledTransactions
https://github.com/ethereum/devp2p/blob/master/caps/eth.md

The spec mentions about soft limit and implementation defined limits. And nimbus EL have no limit implemented on those methods.

Request timeout or other problems might occur because of this unbounded behavior.

jangko:
Unfortunately, because of rlp encoding, the rate limiter like in nimbus-eth2 cannot be applied to nimbus-eth1 wire protocol handler immediately. So we stick with what the spec says.

In nimbus-eth2 or consensus layer in general, the wire protocol is encoded in SSZ. So it is possible to send the message in chunk relatively easy.

While execution client is using RLP encoding. Even before sending the wire message, the full payload length prefix must be calculated, therefore a special RLP encoder is needed. And because the full payload must be held in memory, benefits of incremental read from database cannot be applied.

fix #3102


jangko:
I'll merge this. The issues with hive devp2p/eth wire protocol test will be fixed in other PR.

PR #3119, opened 2025-03-04
Fluffy: Portal EVM call

bhartnett:
This Portal EVM implementation uses the Nimbus in-memory EVM to execute transactions using the portal state network state data. Currently only call is supported.

Rather than wire in the portal state lookups into the EVM directly, the approach taken here is to optimistically execute the transaction multiple times with the goal of building the correct access list so that we can then lookup the required state from the portal network, store the state in the in-memory EVM and then finally execute the transaction using the correct state. The Portal EVM makes use of data in memory during the call and therefore each piece of state is never
fetched more than once. We know we have found the correct access list if it doesn't change after another execution of the transaction.

The assumption here is that network lookups for state data are generally much slower than the time it takes to execute a transaction in the EVM and therefore executing the transaction multiple times should not significantly slow down the
call given that we gain the ability to fetch the state concurrently.

There are multiple reasons for choosing this approach:
- Firstly updating the existing Nimbus EVM to support using a different state backend (portal state in this case) is difficult and would require making non-trivial changes to the EVM.
- This new approach allows us to look up the state concurrently in the event that multiple new state keys are discovered after executing the transaction. This should in theory result in improved performance for certain scenarios. The default approach where the state lookups are wired directly into the EVM gives the worst case performance because all state accesses inside the EVM are completely sequential.


execution_chain/common/common.nim
```diff
@@ -217,7 +218,8 @@ proc init(com         : CommonRef,
   # By default, history begins at genesis.
   com.startOfHistory = GENESIS_PARENT_HASH
 
-  com.initializeDb()
+  if initializeDb:
+    com.initializeDb()
```
bhartnett, line 222: Made this configurable so that it can be disabled in Fluffy which doesn't need it. Removes the logs also which are not relevant in the portal context.

execution_chain/stateless/multi_keys.nim
```diff
@@ -186,3 +186,59 @@ proc visitMatch*(m: var MultiKeysRef, mg: MatchGroup, depth: int): KeyData =
...
+      if kd1.storageKeys != kd2.storageKeys:
+        return false
+      else:
+        continue
+
+    # storageKeys not nil
+    if not equalsStorageMode(kd1.storageKeys, kd2.storageKeys):
+      return false
+
+  return true
```
bhartnett, line 244: This code is required to check if the keys have changed. Is there a better way to compare case object ref types?

fluffy/evm/portal_evm.nim
```diff
@@ -0,0 +1,223 @@
...
+              if slotIdx notin fetchedStorage:
+                debug "Fetching storage slot", address = k.address, slotKey
+                let storageFut = evm.stateNetwork.getStorageAtByStateRoot(
+                  header.stateRoot, k.address, slotKey
+                )
+                storageQueries.add(StorageQuery.init(k.address, slotKey, storageFut))
+
+      # Store fetched state in the in-memory EVM
+      for q in accountQueries:
+        let acc = (await q.accFut).valueOr:
```
bhartnett, line 199: When awaiting these futures here the internal future type raises a CatchableError when the async proc called actually raises a CancelledError. Not sure what is causing this and how I should handle it.

fluffy/portal_node.nim
```diff
@@ -190,6 +192,11 @@ proc new*(
     historyNetwork: historyNetwork,
     stateNetwork: stateNetwork,
     beaconLightClient: beaconLightClient,
+    portalEvm:
+      if historyNetwork.isSome() and stateNetwork.isSome():
+        Opt.some(PortalEvm.init(historyNetwork.get(), stateNetwork.get()))
+      else:
+        Opt.none(PortalEvm),
```
bhartnett, line 199: The PortalEvm is created only when both the history network and the state network are running. Should we start the portal evm using an alternative method such as having a separate cli argument?

fluffy/rpc/rpc_eth_api.nim
```diff
@@ -418,3 +425,36 @@ proc installEthApiHandlers*(
...
+
+  # TransactionArgs* = object
+  #   `from`*: Opt[Address]    # (optional) The address the transaction is sent from.
+  #   to*: Opt[Address]        # The address the transaction is directed to.
+  #   gas*: Opt[Quantity]      # (optional) Integer of the gas provided for the transaction execution. eth_call consumes zero gas, but this parameter may be needed by some executions.
+  #   gasPrice*: Opt[Quantity] # (optional) Integer of the gasPrice used for each paid gas.
+  #   maxFeePerGas*: Opt[Quantity]         # (optional) MaxFeePerGas is the maximum fee per gas offered, in wei.
+  #   maxPriorityFeePerGas*: Opt[Quantity] # (optional) MaxPriorityFeePerGas is the maximum miner tip per gas offered, in wei.
+  #   value*: Opt[UInt256]     # (optional) Integer of the value sent with this transaction.
+  #   nonce*: Opt[Quantity]    # (optional) integer of a nonce. This allows to overwrite your own pending transactions that use the same nonce
```
bhartnett, line 437: The eth_call implementation here is not the core focus of this PR. It is simply used for some basic manual testing. I will create another future PR which refines and further tests the various eth_call scenarios.

fluffy/network/state/state_network.nim
```diff
@@ -262,10 +262,10 @@ proc processContentLoop(n: StateNetwork) {.async: (raises: []).} =
 proc statusLogLoop(n: StateNetwork) {.async: (raises: []).} =
   try:
     while true:
+      await sleepAsync(60.seconds)
```
bhartnett, line 265: Moved the sleep so that during startup the logs look better/cleaner.

fluffy/portal_node.nim
kdeme: Conceptually I do see the PortalEvm as an optional part that is not required to run a pure portal node. That is a node that only does its duties on a Portal Network protocol level and does not provide higher level API such as the EL JSON API. (We could even rewrite the types to have `PortalClient` or `PortalLightClient` which has the subtypes of `PortalNode` and `PortalEvm`.)

In that sense it should be possible to run a portal node with history/state/beacon subnetworks enabled but the evm not running. This could be achieved with an extra cli argument.

Another way  to achieve this (with perhaps better UX?) could be to enable it based on enabling of linked functionality. I believe in this case that would just mean in case the EL (`eth`) JSON-RPC API gets enabled, which it currently default does.

This would avoid having to add additional "Not supported" errors for certain JSON-RPC endpoints in case you would use a separate extra cli argument.

If we would go that route, then I am uncertain whether we should then keep `eth` as default enabled, but I'm leaning towards yes.

bhartnett: > Conceptually I do see the PortalEvm as an optional part that is not required to run a pure portal node. That is a node that only does its duties on a Portal Network protocol level and does not provide higher level API such as the EL JSON API. (We could even rewrite the types to have `PortalClient` or `PortalLightClient` which has the subtypes of `PortalNode` and `PortalEvm`.)
> 
> In that sense it should be possible to run a portal node with history/state/beacon subnetworks enabled but the evm not running. This could be achieved with an extra cli argument.
> 
> Another way to achieve this (with perhaps better UX?) could be to enable it based on enabling of linked functionality. I believe in this case that would just mean in case the EL (`eth`) JSON-RPC API gets enabled, which it currently default does.
> 
> This would avoid having to add additional "Not supported" errors for certain JSON-RPC endpoints in case you would use a separate extra cli argument.
> 
> If we would go that route, then I am uncertain whether we should then keep `eth` as default enabled, but I'm leaning towards yes.

Sure, I'll remove it from `PortalNode` and only create the `PortalEvm` instance when `eth` rpc is enabled. I think this is the simplest way for now considering that it will only be used in the `eth` rpc at this point. In the future when we implement the transaction gossip sub-network the requirements may change and at that point it could be moved somewhere else if/when needed.

bhartnett:
Merging this now as the implementation is basically done. I'll add the unit tests in a future PR.

PR #3121, opened 2025-03-05
Tip is optional in EIP-1559

advaita-saha:


execution_chain/core/tx_pool/tx_desc.nim
```diff
@@ -226,8 +223,8 @@ proc classifyValid(xp: TxPoolRef; tx: Transaction, sender: Address): bool =
       return false
 
   if tx.txType >= TxEip1559:
-    if tx.tip(xp.baseFee) < 1.GasInt:
-      debug "Invalid transaction: EIP-1559 transaction with tip lower than 1"
+    if tx.tip(xp.baseFee) < 0.GasInt:
```
jangko, line 226: GasInt now is `uint64`, probably we need to fix `tip` function too when it produce negative result e.g. use Result[T] and return error.

advaita-saha: Not needed IMO
Even if it tries to move into negative it will overflow due to being `uint`. And the tip calculation has a minimization function `min()` thus nullifying the case of that value ever getting selected 

Rather I think it's safe to remove the negative check from here

advaita-saha: Yess will rework the validation, currently it's spead in different areas and confusing

PR #3129, opened 2025-03-07
Make db initialization configurable

bhartnett:
Changes in this PR:
- Add a flag to make the CommonRef initializeDb() proc configurable (enabled by default) so that it can be disabled when using the Nimbus in-memory evm which doesn't always require the same data to be in the database. For the portal use case, all data is loaded on demand from portal network.

### DIFF
--- a/execution_chain/db/ledger.nim
+++ b/execution_chain/db/ledger.nim
@@ -453,6 +453,10 @@ proc getCode*(ac: LedgerRef,
               returnHash: static[bool] = false): auto =
   let acc = ac.getAccount(address, false)
   if acc.isNil:
+    # We need to record that the code was read even if the account doesn't exist
+    # so that the returned multikeys correctly show that the code lookup occurred
+    ac.witnessCache[address] = WitnessData(codeTouched: true)
+
     when returnHash:
       return (EMPTY_CODE_HASH, CodeBytesRef())
     else:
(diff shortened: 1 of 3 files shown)


execution_chain/db/ledger.nim
```diff
@@ -453,6 +453,10 @@ proc getCode*(ac: LedgerRef,
               returnHash: static[bool] = false): auto =
   let acc = ac.getAccount(address, false)
   if acc.isNil:
+    # We need to record that the code was read even if the account doesn't exist
+    # so that the returned multikeys correctly show that the code lookup occurred
+    ac.witnessCache[address] = WitnessData(codeTouched: true)
```
jangko, line 458: For EVM execution without `witnessCache`, who will clear the table?

jangko: Probably the whole `witnessCache` code path should be put behind compile time flag.

bhartnett: > For EVM execution without `witnessCache`, who will clear the table?

It appears that the existing code does not clear the witnessCache. Perhaps the previous logic to clear the witnessCache was removed at some point.

bhartnett: > Probably the whole `witnessCache` code path should be put behind compile time flag.

Great point, I guess this has the risk of impacting performance without the compile time flag. Actually I was thinking about rewriting and improving the witnessCache code to make it better suited to what I'm doing and putting it behind a compile time flag as you suggested.

I'll remove the witness cache change from this PR and do it in a future PR.

PR #3133, opened 2025-03-07
Snapshot cap

arnetheduck:
When a lot of heads are created, the current strategy of 1-snapshot-per-head creates a lot of snapshots - such as on holesky - and these don't get cleared until finality which has been evasive.

This PR is a hotfix that caps the number of snapshots at 10 which is pretty arbitrary - when a snapshot is cleared, applying the first block to *that* branch is slow but subsequent blocks are fine _until_ the snapshot is cleared again.

The code is pretty crude - the snapshots will be cleared whenever 10 blocks have been applied and not when there are 10 snapshots - this is mostly fine though, if a branch doesn't get attention after 10 blocks, it's likely dead on arrival.


execution_chain/db/core_db/base.nim
```diff
@@ -451,11 +451,15 @@ proc txFrameBegin*(parent: CoreDbTxRef): CoreDbTxRef =
...
 
+proc clearSnapshot*(tx: CoreDbTxRef) =
+  tx.aTx.clearSnapshot()
+
 proc dispose*(tx: CoreDbTxRef) =
   tx.aTx.dispose()
   tx.kTx.dispose()
   tx[].reset()
 
+
```
tersec, line 462: intentional?

arnetheduck, line 462:
```suggestion
```

tersec: https://github.com/status-im/nimbus-eth1/pull/3133/commits/cda348e9a9d6b51ebf154d0b27b4879323c7c7bf

PR #3138, opened 2025-03-13
multi-arch docker builder 

advaita-saha:
The changes ensure that we dynamically tag our images based on the branch name and a short commit SHA while also addressing the push-by-digest error that occurred with plain tags. The workflow now builds multi-architecture images (`amd64` and `arm64`) and creates a manifest list that includes two variants for each build: one tagged with the branch name (e.g., `master`) and one with the branch name concatenated with the short SHA (e.g., `master-abcdef0`).


advaita-saha:
Once merged, a follow up PR would be to hive, for leveraging the extremely fast build times as per other client standards

PR #3139, opened 2025-03-13
Beacon sync stop syncing when progress deemed impossible

mjfh:


execution_chain/config.nim
```diff
@@ -356,11 +356,19 @@ type
       desc: "Number of worker threads (\"0\" = use as many threads as there are CPU cores available)"
       name: "num-threads" .}: int
 
-    beaconBlocksQueueHwm* {.
+    beaconSyncScrumFile* {.
+      hidden
+      desc: "Load a file containg an rlp-encoded object \"(Header,Hash32)\" " &
+            "to be used " &
+            "as the first scrum target before any other request from the CL " &
```
arnetheduck, line 363: what is "scrum"?

mjfh: like in rugby, common sprint -- any better idiom?

arnetheduck: I don't know rugby really, and I don't know what a common sprint is either - like "buddy" that we've discussed before for which `syncPeer` is a better term because it's synthesized for concepts already involved in the syncing process, a good name here would relate to what the algorithm does - failing that, it needs a comment that explains it.

PR #3145, opened 2025-03-15
Refactor wire protocol(eth68): separate request methods and response handlers

jangko:
Separation of request methods and response handlers into their own module will help us avoid cyclic import or employ convoluted abstractions for user handler.

So the dependency graph can be arranged like:
  response_handlers -> user_handlers -> request_methods

Now the p2pProtocol macro is not used for eth68 anymore, only for devP2P. Next PR may reintroduce a simpler but more versatile macro, if needed. Or we can use a more sophisticated template, or both.

With this refactoring, implementation of transactions broadcast, including blob transactions, will be much easier.

This PR also introduce some modifications to reduce copies when the packet received, intended to be rebroadcasted again. This optimization cannot be achieved if most of the code hide behind macros. Or the macros become too complex.
Future PR could address the simplification of of writing both request methods and response handlers.

### DIFF
--- a/execution_chain/sync/wire_protocol/types.nim
+++ b/execution_chain/sync/wire_protocol/types.nim
@@ -11,8 +11,9 @@
 {.push raises: [].}
 
 import
-  eth/common
-
+  eth/common,
+  ../../core/[chain, tx_pool]
+  
 type
   NewBlockHashesAnnounce* = object
     hash*: Hash32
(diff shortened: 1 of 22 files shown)


execution_chain/sync/wire_protocol/types.nim
```diff
@@ -33,7 +34,21 @@ type
...
+  PooledTransactionsRequest* =object
+    txHashes*: seq[Hash32]
+
+  ReceiptsRequest* =object
+    blockHashes*: seq[Hash32]
+
+  EthWireRef* = ref object of RootRef
+    chain* : ForkedChainRef
+    txPool*: TxPoolRef
+
```
tersec, line 54: not sure what's happening here -- GitHub shows this as an extra empty line but without its own line ending?

jangko: no idea too, nothing suspicious shows on my editor when I turn on show all symbols/unprintable chars.

PR #3148, opened 2025-03-17
custom network support in `nrpc`

advaita-saha:
This is needed while testing and debugging for `devnet-6`, seems like a nice to have add-on


tersec:
The UX on this is, I'm not sure -- it seems odd to have this situation where the `--network` parameter knows it can't find a built-in one and then one has to use this `--custom-network-file` (`as /path/to/metadata.yaml`, distinct from `--custom-network`, which is `as /path/to/genesis.json`) parameter.

They're in the same directories for all the EF-supported networks:
- https://github.com/eth-clients/mainnet/tree/main/metadata
- https://github.com/eth-clients/sepolia/tree/main/metadata
- https://github.com/eth-clients/holesky/tree/main/metadata
- https://github.com/eth-clients/hoodi/tree/main/metadata

So while it's fine to document that these files are different (`metadata.yaml` vs `genesis.json`) the actual path is the same and having two different command-line options for basically the same thing is odd.

There's a separate question of, should it be a `--custom-network` at all or a `nimbus-eth2`-style implied check in `--network`, but at least for this PR, it seems ok to at least avoid creation of similar-but-not-identical `--custom-network-file`.

nrpc/config.nim
```diff
@@ -187,8 +190,12 @@ proc makeConfig*(cmdLine = commandLineParams()): NRpcConf
 
   var networkId = result.getNetworkId()
 
-  if result.customNetwork.isSome:
-    result.networkParams = result.customNetwork.get()
+  if result.customNetworkFolder.len > 0:
+    var networkParams = NetworkParams()
+    if not loadNetworkParams(result.customNetworkFolder&"/genesis.json", networkParams):
```
tersec, line 195: This specific kind of concatenation is what [`/`](https://nim-lang.org/docs/ospaths2.html#/,string,string) (a shorthand for [`joinPath`](https://nim-lang.org/docs/ospaths2.html#joinPath,string,string)) is for.

nrpc/nrpc.nim
```diff
@@ -80,8 +80,14 @@ template loadNetworkConfig(conf: NRpcConf): (RuntimeConfig, uint64, uint64) =
   elif conf.networkId == HoodiNet:
     (getMetadataForNetwork("hoodi").cfg, 0'u64, 0'u64)
   else:
-    error "Unsupported network", network = conf.networkId
-    quit(QuitFailure)
+    notice "Loading custom network, assuming post-merge"
+    if conf.customNetworkFolder.len == 0:
+      error "Custom network file not provided"
+      quit(QuitFailure)
+    let (cfg, unloaded) = readRuntimeConfig(conf.customNetworkFolder&"/config.yaml")
```
tersec, line 87: See https://github.com/status-im/nimbus-eth1/pull/3148/files#r1999868058

PR #3149, opened 2025-03-17
Implement rlpx peer connections limit

mjfh:
why
  The number of incoming connections were not restricted in the `p2p`
  handler. Even worse, there was a nimbus option `--max-peers` which
  was used to set up the minimum(sic) number of peers to keep
  connected with.

  Now, the re-organised nimbus options mean:
  * `--min-peers` -- minimum number of connections to hold
  * `--max-peers` -- maximum number of all peer connections

caveat
  While the p2p module now rejects new incoming connections when
  the limit is reached it does allow more connection list entries
  for outgoing connection when explicitly using the `connectToNode()`
  function (as in `rpc/common.nim`, `sync/peers.nim` and probably
  others.) In practice, this lead to at most one more connections on
  the list.

### DIFF
--- a/execution_chain/networking/p2p.nim
+++ b/execution_chain/networking/p2p.nim
@@ -82,6 +82,7 @@ proc newEthereumNode*(
     clientId = "nim-eth-p2p",
     addAllCapabilities = true,
     minPeers = 10,
+    maxPeers = 25,
     bootstrapNodes: seq[ENode] = @[],
     bindUdpPort: Port,
     bindTcpPort: Port,
@@ -101,6 +102,7 @@ proc newEthereumNode*(
   result.connectionState = ConnectionState.None
   result.bindIp = bindIp
   result.bindPort = bindTcpPort
+  result.maxPeers = maxPeers
 
   result.discovery = newDiscoveryProtocol(
     keys.seckey, address, bootstrapNodes, bindUdpPort, bindIp, rng)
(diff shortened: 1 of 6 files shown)


execution_chain/networking/p2p.nim
```diff
@@ -121,6 +123,11 @@ proc processIncoming(server: StreamServer,
                      remote: StreamTransport): Future[void] {.async: (raises: []).} =
   try:
     var node = getUserData[EthereumNode](server)
+
+    if node.maxPeers <= node.peerPool.connectedNodes.len:
+      await node.rlpxReject(remote, TooManyPeers)
```
arnetheduck, line 128: instead of rejecting, we should simply not accept the connection from the socket until there's a free slot, similar to how nimbus-eth2 does it - this makes the OS handle connection spam better.

mjfh: I thought about that but looking at the P2P docs it suggests to respond to the `hello` message with a reason code.

So what is your take on that https://github.com/ethereum/devp2p/blob/master/rlpx.md#p2p-capability

mjfh: Ok, and you don't wish to use it. I will shut down the connection early then.

arnetheduck: > I will shut down the connection early

arnetheduck: Not shut down - what I mean is that you should not call `accept` at at all _until_ there's room for another connection.. 

Also, separately from this, what eth2 implements actually is a hard and a soft limit - the hard limit works like that: it doesn't call accept until there's a slot. 

The soft limit on the other hand is enforced with a delay - it's done this way to allow peers to connect and establish a reputation and then _the worst_ peer is disconnected after a while .. this is overall better because it keeps the number of connected peers in range over time while also allowing better peers to connect and establish themselves.

It's a good idea to study how eth2_network does this and reimplement the same strategy (but with a different scoring function for what consitutes a good peer, which in eth1 is different).

mjfh: will have a look

PR #3154, opened 2025-03-19
Beacon sync handle widely varying block sizes

mjfh:
Needed to run on panda/devnet-6

### DIFF
--- a/execution_chain/config.nim
+++ b/execution_chain/config.nim
@@ -339,13 +339,13 @@ type
       desc: "Number of worker threads (\"0\" = use as many threads as there are CPU cores available)"
       name: "num-threads" .}: int
 
-    beaconSyncScrumFile* {.
+    beaconSyncTargetFile* {.
       hidden
       desc: "Load a file containg an rlp-encoded object \"(Header,Hash32)\" " &
             "to be used " &
-            "as the first scrum target before any other request from the CL " &
-            " is accepted"
-      name: "debug-beacon-sync-scrum-file" .}: Option[InputFile]
+            "as the first target before any other request from the CL " &
+            "is accepted"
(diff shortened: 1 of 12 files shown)


execution_chain/config.nim
```diff
@@ -339,17 +339,17 @@ type
...
-            " is accepted"
-      name: "debug-beacon-sync-scrum-file" .}: Option[InputFile]
+            "as the first target before any other request from the CL " &
+            "is accepted"
+      name: "debug-beacon-sync-target-file" .}: Option[InputFile]
 
     beaconSyncBlocksQueueHwm* {.
       hidden
-      desc: "Limit number of blocks on staging queue for beacon sync"
+      desc: "Limit memory size of the staged block queue for beacon sync"
```
arnetheduck, line 352: per earlier discussions, this kind of limit is not needed on the CL when syncing - CL blocks are per definition larger than EL blocks - what is the difference that motivates this alternative strategy?

mjfh: Instead of a number of blocks on the queue the size becomes relevant. Otherwise it will suck on devnet6.

Here blocks vary from small (e.g. ~90KiB) to at least 4MiB. So counting blocks on the queue designed fo an av size of 90KiB with e.g. 3072 blocks (previous default) would consume 270MiB, but 15GiB in the case of 5MiB blocks.

arnetheduck: > Because we loose peers when importing

This is bug effectively, or rather an architectural problem in the syncer. We cannot work around this bug with this PR - instead, we need to address the bug itself - it is not normal for peers to be dropped during block processing and if they are, the root cause must be determined (too long processing times, invalid requests being sent, protocol violations etc have all been past causes and could still be present).

PR #3163, opened 2025-03-25
Fluffy: Simplify and optimize ContentDb pruning

kdeme:
Simplify by shrinking the data radius with 5% and just deleting all content that no longer falls within that range.

This gives roughly a 150x speed improvement on a 4GB database. The downside is that we cannot be certain that a single 5% radius drop will result in a similar storage drop.


kdeme:
Closing for updated version in https://github.com/status-im/nimbus-eth1/pull/3753

PR #3165, opened 2025-03-26
Fluffy: Make portal evm backend configurable

bhartnett:
Changes in this PR:
- Portal evm no longer depends on history network or state network.
- The functions used to fetch state are now configurable (getAccountProc, getStorageProc, getCodeProc). 
- The block header is now passed in as a parameter to the call function.

This allows the portal async evm to be re-used in the Nimbus Verified Proxy and also makes unit testing easier because the state and header data can be fetched from another location other than portal.

### DIFF
--- a/fluffy/evm/portal_evm.nim
+++ b/fluffy/evm/portal_evm.nim
@@ -18,14 +18,12 @@ import
   ../../execution_chain/db/ledger,
   ../../execution_chain/common/common,
   ../../execution_chain/transaction/call_evm,
-  ../../execution_chain/evm/[types, state, evm_errors],
-  ../network/history/history_network,
-  ../network/state/[state_endpoints, state_network]
+  ../../execution_chain/evm/[types, state, evm_errors]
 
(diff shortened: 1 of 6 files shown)


bhartnett:
@chirag-parmar You should now be able to use this code in the Nimbus Verified Proxy by passing in different state fetch functions (getAccountProc, getStorageProc, getCodeProc) which should fetch the state from the downstream RPC provider. Note that the values returned from these state procs should be validated using eth_getProof before being returned and passed into the evm. The evm doesn't do validation of state internally. 

The state backend should also cache the most recently fetched items (can use minilru for this) for performance reasons. You can also use the downstream RPC `eth_createAccessList` to prefetch and store the expected required state in the state LRU cache before executing the transaction. Note that this is fine purely as a performance optimization because the state fetch strategy inside the async evm will eventually execute the transaction with the correct access list even if a bad access list is provided. The only assumption is that the values returned from the state backend for each access list key need to be correct which is why the validation is required.

kdeme (approved):

PR #3167, opened 2025-03-26
eth: move NibblesBuf to nim-eth

arnetheduck:
Allows hexary trie in nim-eth to take advantage of the faster `NibblesBuf` code which, apart from lowering maintenance by not having two implementations, should give a small perf boost in portal / verif-proxy.


arnetheduck:
Requires https://github.com/status-im/nim-eth/pull/784 to be merged first!

PR #3180, opened 2025-04-01
FC header cache get its own column family

jangko:
This is the second piece of preparation for #3002

### DIFF
--- a/execution_chain/core/chain/forked_chain/chain_header_cache.nim
+++ b/execution_chain/core/chain/forked_chain/chain_header_cache.nim
@@ -68,6 +68,8 @@ import
   pkg/eth/[common, rlp],
   pkg/results,
   "../../.."/[common, db/core_db, db/storage_types],
+  ../../../db/kvt,
+  ../../../db/kvt/[kvt_utils, kvt_tx_frame],
   ../forked_chain,
   ./[chain_branch, chain_desc]
 
@@ -96,6 +98,7 @@ type
     state: FcHdrState          # state of header chain cache
     session: FcHdrSession      # additional session variables
     notify: FcNotifyCB         # client app notification
+    txFrame: KvtTxRef          # metadata and headers storage with it's own column family
 
 const
   HashCacheSize = 50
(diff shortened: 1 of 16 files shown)


arnetheduck:
This looks off - the separate CF should be only for the "temporary" headers that the syncer downloads backwards .. ordinary headers should still be stored in the same txframe as the state (since these reference / rely on each other)

jangko:
> the separate CF should be only for the "temporary" headers that the syncer downloads backwards .. ordinary headers should still be stored in the same txframe as the state

That is exactly what this PR doing. There is no interference/contamination between ordinary headers and temporary headers. It's just the FC haven't take over the "pending FCU". When that happen, The header cache will be simpler. And they are not even sharing the same `KvtDbRef`. Only the lowest rocksdb instance is shared.

arnetheduck:
> That is exactly what this PR doing.

ah ok, I misunderstood it as storing the ordinary headers separately. the code is a bit confusing with the weird counter for the imports - isn't there some .. cleaner way to do this?

jangko:
I wish there is a cleaner way. But for now we have to satisfied by this weird workaround. The EL client itself have no problem without this mechanism. But fluffy will fail to compile because it has no persistent backend.

arnetheduck:
> I wish there is a cleaner way. 

well, I mean that when creating rocksdb, can't we create the extra cf and give it to the syncer directly so that it doesn't have to go through fc/core/base stuff at all? the point would be to isolate this "temporary" storage entirely within the syncer

jangko:
Ah, ok. I will decouple them and remove the weird imports.

execution_chain/core/chain/forked_chain/chain_header_cache.nim
```diff
@@ -340,9 +343,10 @@ proc init*(T: type ForkedCacheRef; c: ForkedChainRef): T =
   ## informed when and how the API is fully functional.
   ##
   let
-    db = c.baseTxFrame
-    state = db.getState.valueOr: FcHdrState()
-    fc = T(chain: c, state: state)
+    be = c.db.kvtBackend()
+    kvt = be.kvtTemporaryHeaderStorage()
```
arnetheduck, line 347: let's maybe name it accordingly, ie `syncKvt` or something

arnetheduck (approved):

mjfh:
Looks OK to me

mjfh (approved):

PR #3183, opened 2025-04-01
Delay state root checks for known-finalized blocks (fixes #3096)

### LINKED ISSUE #3096: When syncing finalized blocks, have forkedchain perform bulk verification/persist
Consider the following scenario:

* according to consensus, the head block is 128 and the finalized block is 64
* we download headers for block `[1..128]` and verify they form a chain
* we start importing blocks one by one

Since blocks up to 64 have been finalized, we can be (fairly) certain that validators have already established their validity, so we can reasonably skip some checks for them - in particular, it makes sense to amortize the cost of state root validation and / or database persistence by updating the base only once per N blocks.

N should probably be kept fairly low - 32-128 perhaps - once `head.number >= roundUp(base.number+baseDistance, N)`, a single state root verification should be applied (which, as a side effect, computes the hash keys of all updated nodes) that covers the entire range of N blocks.

Of course, for blocks newer than the finality point, we must perform full validation for every block.

arnetheduck:
When a majority of validators (according to forkChoiceUpdated) have verified a block, it's reasonable to skip or at least delay some verification so that it can be done in bulk / faster.

Here, we delay the state root verification, doing it when it's time to update base instead of eagerly when processing the block - with a 32-block window, this leads to a speedup of live syncing of around 2-3x, with potential for more with a larger window.

We run the state root computation on a regular basis not only as a sanity check but also to populate the database and not end up with "too much" work at the end - there's lots of follow-up tuning that can be done here.

The current version of the code piggy-backs the savings on top of the batching that the syncer does - of course, the syncer _should not_ be batching like this - it should be feeding blocks to FC as soon as they arrive - once FC is fully responsible for updating the base as outlined in #3002, the next step is to improve the FC base update algorithm to include some form of hysteresis - either fixed by number of blocks or by looking at the change set size and updating base when there is "enough" but not too many changes queued up - a trivial fixed-batch-size version of that is to simply round the base target block number to some suitable multiple like 32.

fix #3096


jangko:
@arnetheduck is there anything you want to add to this PR or can we merge this?

arnetheduck:
we can merge it with a smaller batch size, ie something like 4 blocks - computing the keys of 32 blocks requires some more optimizations to not hog the async thread for too long.

tersec:
> we can merge it with a smaller batch size, ie something like 4 blocks - computing the keys of 32 blocks requires some more optimizations to not hog the async thread for too long.

Yeah, in general see
- https://github.com/status-im/nimbus-eth1/issues/3231

jangko:
Then we must elevate the #3231 priority over this PR.

PR #3193, opened 2025-04-04
Add Portal protocol versioning

kdeme:
Incomplete as it is currently not possible to always retrieve the ENR from a peer from which the node gets an incoming request. 

The best solution that I see for this right now is to add ENRs to the discv5 SessionCache and thus make this retrievable for each node that has an ongoing session.

edit: currently fully implemented with v1 functionality enabled also, relies on https://github.com/status-im/nim-eth/pull/790

### DIFF
--- a/fluffy/fluffy.nim
+++ b/fluffy/fluffy.nim
@@ -31,6 +31,7 @@ import
     rpc_portal_nimbus_beacon_api, rpc_portal_debug_history_api,
   ],
   ./database/content_db,
+  ./network/wire/portal_protocol_version,
   ./portal_node,
   ./version,
   ./logging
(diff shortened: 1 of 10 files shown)


bhartnett (approved):

kdeme:
> @kdeme Did you think we should ban nodes that have no compatible portal version?

Yes, good point. I hadn't considered this but that would be a better way than just blocking in the `addNode` each time.
I will leave the PR like this for now considering we do not have banning enabled yet.

I think following steps would be then to test & enable banning. Then we alter the code to ban on incompatible version (with a small banned time period).

bhartnett:
> Yes, good point. I hadn't considered this but that would be a better way than just blocking in the `addNode` each time. I will leave the PR like this for now considering we do not have banning enabled yet.
> 
> I think following steps would be then to test & enable banning. Then we alter the code to ban on incompatible version (with a small banned time period).

Sure, makes sense. Can be done in a separate PR.

Issue #1524, opened 2023-03-28
explore the possibility of unify process_transaction and tx_packer


jangko:
Both `process_transaction.nim` and `tx_packer.nim` share a lot of similarities: executing a transaction.
Although they now have differences, maybe we can somehow unify both of them similar to what we have done with various emv call in the past.

The new EIP-1153 and old one like EIP-158/161 proves having two transaction execution increase maintenance cost because we have to modify code in two places rather than only in one place.

`tx_packer` is less tested compared to `process_transaction` hence may lead to regression without we being aware of it.

PR #3195, opened 2025-04-04
Nimbus EVM: Fix gas calculation when setting up message for call

bhartnett:
I discovered that the intrinsicGas deduction was not correctly being applied when running my implementation of eth_call and it turned out that in that scenario I had set a very low gas limit which was less than the intrinsicGas but the message.gas field is an unsigned integer so in this case the value underflows to a very large number. 

I also found that in the evm call tests we use `testCallEvm` which sets up the call parameters using callParamsForTest which disables the intrinsicGas setting it to zero so this scenario isn't triggered.

### DIFF
--- a/execution_chain/transaction/call_common.nim
+++ b/execution_chain/transaction/call_common.nim
@@ -142,7 +142,7 @@ proc setupHost(call: CallParams, keepStack: bool): TransactionHost =
                          CallKind.Call,
       # flags: {},
       # depth: 0,
-      gas:             call.gasLimit - intrinsicGas,
+      gas:             if call.gasLimit < intrinsicGas: 0.GasInt else: call.gasLimit - intrinsicGas,
       contractAddress: call.to,
       codeAddress:     call.to,
       sender:          call.sender,


execution_chain/transaction/call_common.nim
```diff
@@ -142,7 +142,7 @@ proc setupHost(call: CallParams, keepStack: bool): TransactionHost =
                          CallKind.Call,
       # flags: {},
       # depth: 0,
-      gas:             call.gasLimit - intrinsicGas,
+      gas:             if call.gasLimit < intrinsicGas: 0.GasInt else: call.gasLimit - intrinsicGas,
```
jangko, line 145: The reason why `processTransaction` works, because before entering EVM, the transaction gone through a series of validation. And I think because of that `eth_call` should also do the same. If we only fix it here, it will fail in other places. And many of EIP rules also bypassed by `eth_call`. Better approach should redesign `processTransaction` similar to `callEvm` so it can be used in `processBlock`, `tx pool` and `rpc`. see #1524

bhartnett: Yeah I noticed that the code path for executing a transaction is different from the evm call code path and call needs to have the correct validations as you said. Unifying these into one common code path is a good idea.

bhartnett: I'm also thinking that it might be a good idea to put this fix in anyway considering that we have multiple code paths that can execute a computation in the EVM and the fact that an underflow can occur here (perhaps if some validation is missing) isn't very safe. At least with this change, the evm execution will fail with out of gas in the scenario where an underflow would have occurred. Unless there is something I'm missing, I think we should put in this change.

@jangko Considering my comments above, did you have any objections to me merging this one?

jangko: > Having said that, it looks like Geth doesn't use common validation on the Message

Geth `Transaction` object and `TransactionArgs` have minimal validation on their own. Most of the validation is done inside the Geth EVM. They do have common validation before EVM execution take place.

And nimbus `processTransaction` with it's `Transaction` validation is too focused on block execution during early stage of development. At that time the txpool is non existent. The RPC is also still very crude. And now it become a debt for us.

I know this is not easy. That's why I put `hard` label to the linked issue. But the correct way to fix it is unify them. I will merge this one if you add some comment that we still need to have a proper validation somewhere and this PR is only a temporary measure.

bhartnett: Well I guess my point is that we probably need both the improvement you are suggesting as well as this change because the underflow can occur under certain conditions and it should not ever be allowed to continue even if there is a missing validation or programmer bug in another location in the code base. In other words it should be a precondition that gasLimit is >= intrinsicGas right? I guess I could be missing something because I'm not sure exactly how you are planning to add this common validation.

Perhaps it would be better to make this part an `doAssert call.gasLimit >= intrinsicGas` to make it clear that the intent is to prevent the underflow.

jangko: > Perhaps it would be better to make this part an doAssert call.gasLimit >= intrinsicGas to make it clear that the intent is to prevent the underflow.

It's not only `gasLimit` need a precondition, `value` must me checked against `sender`'s account `balance`. `versionedHashes` cannot coexist with `authorizationList`, and they have to be checked against fork actvation, etc. So simple `doAssert` although useful does not cover entire `precondition`.

> I'm not sure exactly how you are planning to add this common validation

- `processTransaction` and `callEvm` both calculate intrinsicGas, they should reduced to one call.
- txPool.txPacker should share the similar `processTransaction` (a new version of `processTransaction`) with block execution.
- both `eth_call` and `eth_estimateGas`(or more) should share `TransactionArgs` validation and the `TransactionArgs` validation also shared with the above `processTransaction`(part of it, not the whole `processTransaction`).

> I also found that in the evm call tests we use testCallEvm which sets up the call parameters using callParamsForTest which disables the intrinsicGas

Tests should use regular `callEvm`, `testCallEvm` is good for transition from ancient-nimbus-evm to today-nimbus-evm. But we should phase out `testCallEvm`. There is no point testing `testCallEvm` while in production we use `callEvm`. Existing tests using `testCallEvm` must be fixed/refactored to use regular `callEvm`.

bhartnett: Thanks for the explanation. This sounds like a reasonable plan and I think you are probably the best person to make these changes considering your knowledge in this area. Having said that I still feel rather strongly that putting this small change in (as a short term solution perhaps) does no harm and only reduces the risk of problems caused by this underflow. I am not questioning the importance of those other validations that you mentioned but I don't see any strong argument against merging in this change as it is because it improves the codebase in the short term. 

I've added a comment as you suggested.

PR #3202, opened 2025-04-08
fluffy: add Portal wire accept codes

kdeme:
Add accept/decline codes for offer's accept response: https://github.com/ethereum/portal-network-specs/pull/370

Put in draft as this needs to become selective by version

### DIFF
--- a/fluffy/network/wire/messages.nim
+++ b/fluffy/network/wire/messages.nim
@@ -27,7 +27,7 @@ const
 
 type
   ContentKeysList* = List[ContentKeyByteList, contentKeysLimit]
-  ContentKeysBitList* = BitList[contentKeysLimit]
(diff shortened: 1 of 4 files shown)


bhartnett:
> Put in draft as this needs to become selective by version

Yes in order to support the old version we might need to keep the BitList type in addition to the new ByteList type and handle the message based on the peer's highest support protocol version.

kdeme:
> > Put in draft as this needs to become selective by version
> 
> Yes in order to support the old version we might need to keep the BitList type in addition to the new ByteList type and handle the message based on the peer's highest support protocol version.

Yes, but to make it somewhat clean I think I will actually copy the whole messages.nim file. As the decoding part of the portal wire messages will become quite messy else.

kdeme:
Closing as moved functionality into PR https://github.com/status-im/nimbus-eth1/pull/3193

PR #3204, opened 2025-04-09
fCU early notification of finalized hash to FC.

jangko:
Before fCU actually call FC.forkchoice, it will go through a series of validation, and the finalized hash may never arrive at FC.forkChoice at all although the finalized hash is valid. e.g. head, finalized, or safe block are being downloaded by the syncer.


execution_chain/beacon/api_handler/api_forkchoice.nim
```diff
@@ -109,6 +108,7 @@ proc forkchoiceUpdated*(ben: BeaconEngineRef,
       hash   = headHash.short
 
     # Inform the header cache (used by the syncer)
+    chain.notifyFinalizedHash(update.finalizedBlockHash)
```
arnetheduck, line 111: should also update `latestFinalizedBlockNumber` with `max(latestFinalizedBlockNumber, newNumber)` .. 

also, should set `latestFCU` even if we can't find the number

arnetheduck:
while this PR fixes the urgent base update issue, many of the checks and updates performed in `api_forkChoice` should probably move into `forkChoice` proper

execution_chain/beacon/api_handler/api_forkchoice.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/3204/commits/cbe96b892b66fc6a20e320ec9fa494e334778d9f

arnetheduck: not quite done here: need to look up the block number of `update.finalizedBlockHash` in the quarantine - the syncer will not do it if it's already syncing something else.

PR #3209, opened 2025-04-11
fix: spec deviation in `newPayloadV4`

advaita-saha:
Found using hive, reported by EF testing team 

> executionRequests: Array of DATA - List of execution layer triggered requests. Each list element is a requests byte array as defined by [EIP-7685](https://eips.ethereum.org/EIPS/eip-7685). The first byte of each element is the request_type and the remaining bytes are the request_data. Elements of the list MUST be ordered by request_type in ascending order. Elements with empty request_data MUST be excluded from the list. If any element is out of order, has a length of 1-byte or shorter, or more than one element has the same type byte, client software MUST return -32602: Invalid params error.

Ref -> https://github.com/ethereum/execution-apis/blob/main/src/engine/prague.md#request

This infers to if execution request type is invalid in length then we issue a invalid params error, other than that we must return INVALID as payloadStatus

### DIFF
--- a/execution_chain/beacon/api_handler/api_newpayload.nim
+++ b/execution_chain/beacon/api_handler/api_newpayload.nim
@@ -87,7 +87,7 @@ template validatePayload(apiVersion, payloadVersion, payload) =
       raise invalidParams("newPayload" & $apiVersion &
         "excessBlobGas is expected from execution payload")
 
-# https://github.com/ethereum/execution-apis/blob/main/src/engine/prague.md#engine_newpayloadv4
+# https://github.com/ethereum/execution-apis/blob/40088597b8b4f48c45184da002e27ffc3c37641f/src/engine/prague.md#request
 template validateExecutionRequest(requests: openArray[seq[byte]], apiVersion: Version) =
   var previousRequestType = -1
   for request in requests:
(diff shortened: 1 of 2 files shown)


execution_chain/beacon/api_handler/api_newpayload.nim
```diff
@@ -87,28 +87,31 @@ template validatePayload(apiVersion, payloadVersion, payload) =
       raise invalidParams("newPayload" & $apiVersion &
         "excessBlobGas is expected from execution payload")
 
-func validateExecutionRequest(requests: openArray[seq[byte]]): Result[void, string] {.raises:[].} =
+# https://github.com/ethereum/execution-apis/blob/main/src/engine/prague.md#engine_newpayloadv4
```
tersec, line 90: The `main` branch changes, links can become invalid, and it does not identify which exact specification was taken into account last in updating this implementation. Unfortunately, the `execution-api`s repository hasn't made a tagged release in a long time, but it's still preferable to use e.g., https://github.com/ethereum/execution-apis/blob/c43a2d05af83f4f245b7188c777a2cfbab43edf2/src/engine/prague.md#engine_getpayloadv4 which doesn't have these issues.

tests/test_engine_api.nim
```diff
@@ -269,6 +268,27 @@ proc newPayloadV4InvalidRequests(env: TestEnv): Result[void, string] =
...
+      params.payload,
+      params.expectedBlobVersionedHashes,
+      params.parentBeaconBlockRoot,
+      params.executionRequests)
+
+  if res.isErr:
+    return err("res should success")
+
+  if res.get.status != PayloadExecutionStatus.invalid:
+    return err("res.status should equals to PayloadExecutionStatus.invalid")
```
tersec, line 288: either "should equal" or "should be equal to" presumably?

PR #3210, opened 2025-04-11
Fluffy: Implement AsyncEvm estimateGas and eth_estimateGas RPC endpoint

bhartnett:
Changes in this PR:
- `AsyncEvm` cleanup and refactor.
- `estimateGas` function added to `AsyncEvm` so that it can be reused by the Nimbus Verified Proxy and also tested in isolation.
- Implemented `eth_estimateGas` RPC endpoint.
- Tests for `AsyncEvm` `estimateGas`.

### DIFF
--- a/execution_chain/transaction/call_evm.nim
+++ b/execution_chain/transaction/call_evm.nim
@@ -58,22 +58,9 @@ proc rpcCallEvm*(args: TransactionArgs,
 
 proc rpcEstimateGas*(args: TransactionArgs,
                      header: Header,
-                     headerHash: Hash32,
-                     com: CommonRef,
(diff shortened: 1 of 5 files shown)


execution_chain/transaction/call_evm.nim
```diff
@@ -58,22 +58,9 @@ proc rpcCallEvm*(args: TransactionArgs,
 
 proc rpcEstimateGas*(args: TransactionArgs,
```
bhartnett, line 59: No changes to the logic here in the `call_evm` module. Only a refactor to make the function support passing in the `BaseVMState` object.

PR #3214, opened 2025-04-14
Implement transactions broadcast

jangko:
Filter and accepts potentially good transactions. Removes bad transactions from circulation. Disconnect bad peer if they messing around with blob transactions with no or invalid sidecar.

## TODOs:
- [x] Filter blob transactions with different order compared to hashes when handling NewPooledTransactionHashes.
- [x] Disable broadcast/receive broadcast when syncing, including no addition to txpool.
- [ ] Broadcast to newly connected peers.
- [x] Peers still connected?
- [x] Broadcast `blockRangeUpdate` to connected peers. (req. [eth/69](https://github.com/status-im/nimbus-eth1/pull/3283))
- [x] handle `BlockRangeUpdate` from peers. (req. [eth/69](https://github.com/status-im/nimbus-eth1/pull/3283))
- [x] Remove txs regularly from txpool if blocks imported even though no blocks build


advaita-saha:
Superseded by #4487

PR #3221, opened 2025-04-16
Re factor header chain cache and some fixes

mjfh:


execution_chain/sync/beacon/worker/headers_staged.nim
```diff
@@ -140,18 +145,27 @@ proc headersStagedCollect*(
         # heap so that `async` can capture that properly.
         lhc = (ref LinkedHChain)(peerID: buddy.peerID)
 
-        # Fetch headers and fill up the headers list of `lhc`,
-        # get returned the last unprocessed block number
+        # Fetch headers and fill up the headers list of `lhc`. The function
+        # returnedsthe last unprocessed block number.
```
tersec, line 149: typo

Issue #3164, opened 2025-03-25
Use optimised rlp hash writer in `aristo_compute`


arnetheduck:
`computeKey` is one of the places where we compute _the most_ hashes that would benefit from avoiding memory allocations (since they're numerous and small), and yet the new HashWriter is not in use there: https://github.com/status-im/nimbus-eth1/blob/232e7560e066810b67a09a193cd00bc4a4f9788d/execution_chain/db/aristo/aristo_compute.nim#L146

arnetheduck:
More here: https://github.com/status-im/nim-eth/blob/37bf6db28b81e3cd45a101457bc6909d9081c4b4/eth/common/transactions_rlp.nim#L238

basically, need to grep the codebase for any rlp interactions that we can get rid of

PR #3226, opened 2025-04-18
Use hash writer

chirag-parmar:
https://github.com/status-im/nimbus-eth1/issues/3164


execution_chain/db/aristo/aristo_compute.nim
```diff
@@ -271,7 +311,7 @@ proc computeKeyImpl(
 
   if vtx.vType notin Leaves:
     ?db.putKeyAtLevel(rvid, vtx, key, level, batch)
-  ok (key, level)
+  return ok (key, level)
```
arnetheduck:
```suggestion
  ok (key, level)
```

PR #3233, opened 2025-04-23
Fluffy: Implement offer cache to hold content ids of recent offers

bhartnett:
Reduces load on the database during gossip process as it is very common to receive multiple copies of the same content from different peers as content is gossipped through the network.

Changes in this PR:
- Offer cache holds content id of recently stored offers
- Content cache is checked before checking the database during offer flow
- If the database is pruned then the content ids in the offer cache are invalidated so that we don't incorrectly indicate that some pruned content is already stored.


fluffy/network/wire/portal_protocol.nim
```diff
@@ -1797,8 +1811,16 @@ proc storeContent*(
...
   # content in range is stored.
   if p.inRange(contentId):
-    doAssert(p.dbPut != nil)
-    p.dbPut(contentKey, contentId, content)
+    let dbPruned = p.dbPut(contentKey, contentId, content)
+    if dbPruned:
+      # invalidate all cached content incase it was removed from the database
+      # during pruning
+      for k, v in p.offerCache.mpairs():
+        v = false
```
kdeme, line 1819: Probably faster to just reinitialize the cache? And that way you also don't need to boolean I think?

kdeme (commented):
I'm fine with this cache addition as offers do typically come in fairly close to eachother.

Two things however:
- Curious to see any actual data on this.
- I was also wondering how often this occurs versus the version where offers come in too close to each other that the actual content of the first offer is not stored yet. Of course for that scenario we cannot really add a cache as the content offered could be failed to send or invalid.

fluffy/network/wire/portal_protocol.nim
bhartnett: > Probably faster to just reinitialize the cache? And that way you also don't need to boolean I think?

I was looking though the minilru code and I didn't find a clear function so I went with this method. But yes, reinit is probably better. I'll update.

bhartnett:
> * Curious to see any actual data on this.
I guess I can add some cache hit cache miss metrics for this and gossip some data in a local testnet to see the results.

> * I was also wondering how often this occurs versus the version where offers come in too close to each other that the actual content of the first offer is not stored yet. Of course for that scenario we cannot really add a cache as the content offered could be failed to send or invalid.
This is something we could use metrics to get data on a well. To address this problem I think we should put a limit on the max number of concurrent offers per content id. The current limits are per content id and per peer but there is no limit on multiple peers sending the same content concurrently.

bhartnett:
@kdeme Here are some metrics showing the usage of the offer cache when running a local testnet with 16 nodes and gossipping content using 20 workers running for 10 mins or so. All content is sent to one of the fluffy instances (node 2) which gossips the content to its peers.

**Node 1:**
```
# HELP portal_offer_cache_hits Portal wire protocol local content lookups that hit the offer cache
# TYPE portal_offer_cache_hits counter
portal_offer_cache_hits_total{protocol_id="500a"} 47330.0
portal_offer_cache_hits_created{protocol_id="500a"} 1745418350.0

# HELP portal_offer_cache_misses Portal wire protocol local content lookups that don't hit the offer cache
# TYPE portal_offer_cache_misses counter
portal_offer_cache_misses_total{protocol_id="500a"} 9918.0
portal_offer_cache_misses_created{protocol_id="500a"} 1745418350.0
```

**Node 2 (the node which the portal bridge is connected to):**
```
# HELP portal_offer_cache_hits Portal wire protocol local content lookups that hit the offer cache
# TYPE portal_offer_cache_hits counter

# HELP portal_offer_cache_misses Portal wire protocol local content lookups that don't hit the offer cache
# TYPE portal_offer_cache_misses counter
portal_offer_cache_misses_total{protocol_id="500a"} 47465.0
portal_offer_cache_misses_created{protocol_id="500a"} 1745418349.0
```

**Node 3:**
```
# HELP portal_offer_cache_hits Portal wire protocol local content lookups that hit the offer cache
# TYPE portal_offer_cache_hits counter
portal_offer_cache_hits_total{protocol_id="500a"} 48479.0
portal_offer_cache_hits_created{protocol_id="500a"} 1745418350.0

# HELP portal_offer_cache_misses Portal wire protocol local content lookups that don't hit the offer cache
# TYPE portal_offer_cache_misses counter
portal_offer_cache_misses_total{protocol_id="500a"} 9175.0
portal_offer_cache_misses_created{protocol_id="500a"} 1745418350.0
```

**Node 4:**
```
# HELP portal_offer_cache_hits Portal wire protocol local content lookups that hit the offer cache
# TYPE portal_offer_cache_hits counter
portal_offer_cache_hits_total{protocol_id="500a"} 17025.0
portal_offer_cache_hits_created{protocol_id="500a"} 1745418350.0

# HELP portal_offer_cache_misses Portal wire protocol local content lookups that don't hit the offer cache
# TYPE portal_offer_cache_misses counter
portal_offer_cache_misses_total{protocol_id="500a"} 18976.0
portal_offer_cache_misses_created{protocol_id="500a"} 1745418350.0
```

kdeme (approved):

PR #3237, opened 2025-04-24
Addition to #3204: Try to resolve finalized block number from quarantine

jangko:


execution_chain/core/chain/header_chain_cache.nim
```diff
@@ -466,6 +488,11 @@ proc put*(
   if rev.len == 0:
     return ok()                                    # nothing to do
 
+  notice "HC updated",
```
arnetheduck, line 491:
```suggestion
  debug "HC updated",
```

execution_chain/core/chain/header_chain_cache.nim
```diff
@@ -516,18 +543,11 @@ proc put*(
...
-        #   TODO: syncer and also ForkedCache should not start session
-        #   using finalizedHash, as evident from hive test
-        #   the FCU head can have block number bigger than finalized block.
-        #   Syncer and ForkedCache should only deal with FCU headHash.
-        #   TODO: move notifyBlockHashAndNumber call to fcPutHeader
-        #   where one of the headers should also the finalized header.
-        hc.chain.notifyBlockHashAndNumber(hash, bn)
+      if hash == hc.chain.pendingFCU:
+        if hc.chain.tryUpdatePendingFCU(hash, hdr.number):
+          notice "PendingFCU resolved to block number",
```
arnetheduck, line 548:
```suggestion
          debug "PendingFCU resolved to block number",
```

arnetheduck (approved):

execution_chain/core/chain/header_chain_cache.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/3237/commits/e32b4de930f23129eec84fecf1391e26220d6d99

execution_chain/core/chain/header_chain_cache.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/3237/commits/e32b4de930f23129eec84fecf1391e26220d6d99

jangko:
Updating `pendingFCU` and resolve it into block number earlier only solve half of the problem. The racing between EL and CL still a problem. Because EL syncing progress faster than CL, the updated fin hash, although updated regularly, cannot match EL progress. And if the syncing session is very long, the memory consumption problem is still there. But this PR can be merged due to no more improvement can be achieved without big modification. And the modification will come in the next PR.

execution_chain/core/chain/forked_chain.nim
```diff
@@ -135,8 +135,7 @@ proc validateBlock(c: ForkedChainRef,
   # handled region(head - baseDistance)
   # e.g. live syncing with the tip very far from from our latest head
   if c.pendingFCU != zeroHash32 and
```
arnetheduck, line 137: `c.pendingFCU != zeroHash32` is not needed here

execution_chain/core/chain/header_chain_cache.nim
```diff
@@ -309,6 +323,25 @@ proc tryFcParent(hc: HeaderChainRef; hdr: Header): HeaderChainMode =
 # Private fork choice call back function
 # ------------------------------------------------------------------------------
 
+proc resolveFinHash(hc: HeaderChainRef; f: Hash32) =
```
arnetheduck, line 326: this function must set pendingFCU even if it doesn't find it right now .. ie when it's called from api_forkchoice, it might be unknown to all these sources

arnetheduck: ah, maybe I misread the code, it's not being called from there .. maybe ..

execution_chain/core/chain/forked_chain.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/3237/commits/74fe451c328d05eefeccf6f47c769797e7dfac4c

PR #3238, opened 2025-04-24
Fix syscall fail issue and missing code

advaita-saha:
EIP-7002 updated along with EIP-7215
https://github.com/ethereum/EIPs/pull/9508/files
https://github.com/ethereum/EIPs/pull/9582/files


advaita-saha (commented):
Not sure if this is the best approach to be done here
If the code doesn't exist, we don't need to process further over the EVM
https://github.com/status-im/nimbus-eth1/blob/a91406e4d074f45d1b049a3b7d50346e2917cb3e/execution_chain/core/executor/process_transaction.nim#L204-L206

jangko:
So it's a breaking changes I see. I think the t8n test vectors must be fixed too by adding predeployed contract code.
Of course you will see additional mismatch like stateroot mismatch too, but simply replace it with new value should solve the problem.

advaita-saha:
Screenshot 2025-04-24 at 5 00 09 PM

jangko (approved):
nice

PR #3240, opened 2025-04-25
Adjust tests for supporting SSZ vector Capella and Deneb HWP

kdeme:
- Adjust tests for supporting HeaderWithProof updated version with SSZ vector instead of list
- Add & adjust generation of test vectors for this
- Enable validation of this for Capella & Deneb onwards, however it will still fail due to no recent historical_summaries download
- Update portal-spec-tests test vectors


kdeme:
As per spec change: https://github.com/ethereum/portal-network-specs/pull/396

PR #3241, opened 2025-04-28
Fluffy: Add metrics to count offer accept codes

bhartnett:
Add metrics to count accept codes when both sending and receiving offers in Fluffy. These metrics will be useful to gain better visibility into the propagation of offers across the network.

After adding these metrics and then running the state bridge locally with a testnet of 8 nodes I collected the following metrics after running from genesis for around 10 minutes. Collected from Node 1 which is not the node connected to the bridge:
```
# HELP portal_offer_accept_codes Portal wire protocol accept codes received from peers after sending offers
# TYPE portal_offer_accept_codes counter
portal_offer_accept_codes_total{protocol_id="500a",accept_code="3"} 30242.0
portal_offer_accept_codes_created{protocol_id="500a",accept_code="3"} 1745822003.0
portal_offer_accept_codes_total{protocol_id="500a",accept_code="2"} 10765.0
portal_offer_accept_codes_created{protocol_id="500a",accept_code="2"} 1745822003.0
portal_offer_accept_codes_total{protocol_id="500a",accept_code="0"} 1016.0
portal_offer_accept_codes_created{protocol_id="500a",accept_code="0"} 1745822059.0
portal_offer_accept_codes_total{protocol_id="500a",accept_code="4"} 10089.0
portal_offer_accept_codes_created{protocol_id="500a",accept_code="4"} 1745822006.0

# HELP portal_handle_offer_accept_codes Portal wire protocol accept codes returned to peers when handing offers
# TYPE portal_handle_offer_accept_codes counter
portal_handle_offer_accept_codes_total{protocol_id="500a",accept_code="3"} 35864.0
portal_handle_offer_accept_codes_created{protocol_id="500a",accept_code="3"} 1745822003.0
portal_handle_offer_accept_codes_total{protocol_id="500a",accept_code="2"} 2030.0
portal_handle_offer_accept_codes_created{protocol_id="500a",accept_code="2"} 1745822003.0
portal_handle_offer_accept_codes_total{protocol_id="500a",accept_code="0"} 13078.0
portal_handle_offer_accept_codes_created{protocol_id="500a",accept_code="0"} 1745822003.0
portal_handle_offer_accept_codes_total{protocol_id="500a",accept_code="4"} 30197.0
portal_handle_offer_accept_codes_created{protocol_id="500a",accept_code="4"} 1745822010.0
```

Accept codes from the portal wire spec for reference:
```
  Accepted* = byte 0x0
  DeclinedGeneric* = byte 0x1
  DeclinedAlreadyStored* = byte 0x2
  DeclinedNotWithinRadius* = byte 0x3
  DeclinedRateLimited* = byte 0x4
  DeclinedInboundTransferInProgress* = byte 0x5
```

The number of `DeclinedNotWithinRadius` accept codes returned is relatively high and I find this surprising because ideally we shouldn't even send the content to nodes if it is outside their radius. Once this change is deployed to the network it will be interesting to see if portal mainnet shows similar numbers.

### DIFF
--- a/fluffy/network/wire/portal_protocol.nim
+++ b/fluffy/network/wire/portal_protocol.nim
@@ -42,6 +42,12 @@ declareCounter portal_message_requests_outgoing,
 declareCounter portal_message_response_incoming,
   "Portal wire protocol incoming message responses",
   labels = ["protocol_id", "message_type"]
+declareCounter portal_offer_accept_codes,
+  "Portal wire protocol accept codes received from peers after sending offers",
+  labels = ["protocol_id", "accept_code"]
+declareCounter portal_handle_offer_accept_codes,
+  "Portal wire protocol accept codes returned to peers when handing offers",
+  labels = ["protocol_id", "accept_code"]
 
 const requestBuckets = [1.0, 3.0, 5.0, 7.0, 9.0, Inf]
 declareHistogram portal_lookup_node_requests,
@@ -560,6 +566,9 @@ proc handleOffer(
   # of content to process and potentially gossip around. Don't accept more
   # data in this case.
   if p.stream.contentQueue.full():
+    portal_handle_offer_accept_codes.inc(
+      o.contentKeys.len, labelValues = [$p.protocolId, $DeclinedRateLimited]
+    )
     return ok(
       AcceptMessage(
         connectionId: Bytes2([byte 0x00, 0x00]),
@@ -603,6 +612,10 @@ proc handleOffer(
         discard contentKeys.add(contentKey)
         contentIds.add(contentId)
         contentAccepted = true
+
+      portal_handle_offer_accept_codes.inc(
+        labelValues = [$p.protocolId, $contentKeysAcceptList[i]]
+      )
     else:
       # Return empty response when content key validation fails
       return err("Invalid content key")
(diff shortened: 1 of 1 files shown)


kdeme (approved):

kdeme:
It is definitely not normal to have so many `DeclinedNotWithinRadius`.

But I think I know where this is coming from. In your local testnet you are only having 8 nodes, so it is likely that the NHgossip call never finds the `n` required nodes and then just does a `neighbors` lookup. And the result of that lookup will not check whether the nodes are within radius before offering content.

So I think it is specifically a side effect of the low node count test you are running. But it will be interesting to see if this also occurs on the mainnet, as if it does we definitely have a bug.

bhartnett:
> It is definitely not normal to have so many `DeclinedNotWithinRadius`.
> 
> But I think I know where this is coming from. In your local testnet you are only having 8 nodes, so it is likely that the NHgossip call never finds the `n` required nodes and then just does a "closest nodes" lookup on the network. And the result of that lookup will not check whether the nodes are within radius before offering content.
> 
> So I think it is specifically a side effect of the low node count test you are running. But it will be interesting to see if this also occurs on the mainnet, as if it does we definitely have a bug.

Yes, you are right. The problem is that `--debug-max-gossip-nodes=4` is set in the testnet script and when there are only 8 nodes then each node often won't have at least 4 peers in it's local routing table so it does the lookup then offer over the network without checking the radius. I set the --debug-max-gossip-nodes parameter to 1 and the problem is fixed. Do you think we should set this to 1 or 2 as a default for the testnet script?

PR #3254, opened 2025-05-03
`nrpc` logs with estimated completion time from bps

advaita-saha:


nrpc/nrpc.nim
```diff
@@ -8,8 +8,9 @@
 # those terms.
```
tersec, line 9: Would be good to add `{.push raises: [].}` -- it seems to compile with it fine.

advaita-saha: Ahh yess

PR #3255, opened 2025-05-03
Era files downloader tool

advaita-saha:
Trivial but very handy, specially after EF putting their history endpoints behind cloudflare

### DIFF
--- a/scripts/era_downloader.sh
+++ b/scripts/era_downloader.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# Copyright (c) 2025 Status Research & Development GmbH. Licensed under
+# either of:
+# - Apache License, version 2.0
+# - MIT license
+# at your option. This file may not be copied, modified, or distributed except
+# according to those terms.
+
 # Usage: ./download_era.sh <download_url> <download_path>
 set -e
 
--- a/scripts/era_downloader.sh
+++ b/scripts/era_downloader.sh
@@ -8,7 +8,7 @@
 # according to those terms.
 
 # Usage: ./download_era.sh <download_url> <download_path>
-set -e
+set -eo pipefail
 
 if [ $# -ne 2 ]; then
   echo "Usage: $0 <download_url> <download_path>"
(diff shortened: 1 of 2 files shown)


scripts/era_downloader.sh
```diff
@@ -0,0 +1,78 @@
...
+fi
+
+mkdir -p "$DOWNLOAD_DIR"
+cd "$DOWNLOAD_DIR" || exit 1
+
+curl -s "$DOWNLOAD_URL" | \
+grep -Eo 'href="[^"]+"' | \
+cut -d'"' -f2 | \
+grep -Ei '\.(era|era1|txt)$' | \
+sort -u > urls_raw.txt
```
tersec, line 33: For both `urls_raw.txt` and `urls.txt`, maybe `mktemp` or some similar approach to avoid accidentally overwriting possible files in the directory in which this script is run called either, because they're `rm`'d unconditionally at the end.

scripts/era_downloader.sh
```diff
@@ -0,0 +1,78 @@
...
+
+# Copyright (c) 2025 Status Research & Development GmbH. Licensed under
+# either of:
+# - Apache License, version 2.0
+# - MIT license
+# at your option. This file may not be copied, modified, or distributed except
+# according to those terms.
+
+# Usage: ./download_era.sh <download_url> <download_path>
+set -e
```
tersec, line 11: without `-o pipefail` this doesn't help with the pipes in this script.

scripts/era_downloader.sh
```diff
@@ -42,17 +47,17 @@ case "$DOWNLOAD_URL" in
   *)            BASE_URL="$DOWNLOAD_URL" ;;
 esac
 
-# 🛠️ Prepend base URL to each file
-awk -v url="$BASE_URL" '{ print url "/" $0 }' urls_raw.txt > urls.txt
+# Prepend full URL
+awk -v url="$BASE_URL" '{ print url "/" $0 }' "$URLS_RAW_FILE" > "$URLS_FILE"
 
-TOTAL_FILES=$(wc -l < urls.txt)
+TOTAL_FILES=$(wc -l < "$URLS_FILE")
```
tersec, line 53: why the input redirect? `wc` can handle filenames directly.

advaita-saha: Well to avoid handling the filename in output

Few other options than this would be 
```bash
TOTAL_FILES=$(<"$URLS_FILE" wc -l)  # POSIX-safe and fast
```
or
```bash
TOTAL_FILES=$(wc -l "$URLS_FILE" | awk '{print $1}')
```

PR #3259, opened 2025-05-05
Refactor block import logic to be async

### LINKED ISSUE #3231: Cooperative concurrency in block processing
The import loop should process blocks in a cooperative, event-loop-friendly way. This can be done by:

- Importing **one block per iteration** of an async loop.
- Explicitly yielding to the event loop (`await idleAsync()` or similar) between imports.

For inspiration, see `nimbus-eth2`:
- [`BlockProcessor.runQueueProcessingLoop`](https://github.com/status-im/nimbus-eth2/blob/4e440277cf8a3fed72f32eb2f01fc5e910ad6768/beacon_chain/gossip_processing/block_processor.nim#L914-L929)
- [`blockchain_dag.nim` pruning logic](https://github.com/status-im/nimbus-eth2/blob/4e440277cf8a3fed72f32eb2f01fc5e910ad6768/beacon_chain/consensus_object_pools/blockchain_dag.nim#L56-L60)

These approaches cap long-running CPU tasks to avoid networking stalls.

### Action Items

- [ ] Refactor block import logic to be async
- [ ] Ensure regular yielding to the event loop
- [ ] Limit per-loop block processing to a small, bounded number (e.g., 1 block per iteration)
- [ ] Monitor peer connectivity and responsiveness during sync

jangko:
fix #3231

### DIFF
--- a/execution_chain/core/block_import.nim
+++ b/execution_chain/core/block_import.nim
@@ -21,7 +21,7 @@ import
 proc importRlpBlocks*(blocksRlp:seq[byte],
                       chain: ForkedChainRef,
                       finalize: bool):
-                        Future[Result[void, string]] {.async: (raises: []).} =
+                        Future[Result[void, string]] {.async: (raises: [CancelledError]).} =
   var
     # the encoded rlp can contains one or more blocks
     rlp = rlpFromBytes(blocksRlp)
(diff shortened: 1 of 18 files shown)


execution_chain/core/block_import.nim
```diff
@@ -13,14 +13,15 @@ import
   chronicles,
   eth/rlp,
   stew/io2,
+  chronos,
   ./chain,
   ../config,
   ../utils/utils
 
-proc importRlpBlocks*(blocksRlp: openArray[byte],
+proc importRlpBlocks*(blocksRlp:seq[byte],
```
arnetheduck, line 21: rlp block import should probably not be async, ie there should probably be a sync codepath for this use case

jangko: Why not? It's only executed once and then the program will terminate.

arnetheduck: yeah, that's the point: async introduces a bit of performance overhead which is not justified here since the network is not running

PR #3264, opened 2025-05-07
Fluffy: Multiple workers for processing received content in state network

bhartnett:
Before this change each Fluffy instance only processes one content/offer from the content queue at a time. In theory we should be able to speed up the process of distributing content across the network by allowing concurrent processing so that each node can process and gossip multiple content offers at once.

It would be nice to have faster distribution across the network so that the Fluffy bridges don't need to be 'smart' and can simply connect to locally running Fluffy nodes which use neighborhoodGossip to send content across the network.

If the distribution across the network is too slow than we might need to build a 'smart' bridge that sends content directly to each node that requires it as this method doesn't rely on the gossip of content between peers as much.

### DIFF
--- a/fluffy/network/state/state_network.nim
+++ b/fluffy/network/state/state_network.nim
@@ -36,7 +36,7 @@ const pingExtensionCapabilities = {CapabilitiesType, BasicRadiusType}
 type StateNetwork* = ref object
   portalProtocol*: PortalProtocol
   contentQueue*: AsyncQueue[(Opt[NodeId], ContentKeysList, seq[seq[byte]])]
-  processContentLoop: Future[void]
+  processContentLoops: seq[Future[void]]
   statusLogLoop: Future[void]
   historyNetwork: Opt[HistoryNetwork]
   validateStateIsCanonical: bool
@@ -275,7 +275,9 @@ proc start*(n: StateNetwork) =
 
   n.portalProtocol.start()
 
-  n.processContentLoop = processContentLoop(n)
+  for i in 0..<25:
+    n.processContentLoops.add processContentLoop(n)
+
   n.statusLogLoop = statusLogLoop(n)
 
 proc stop*(n: StateNetwork) {.async: (raises: []).} =
@@ -284,12 +286,12 @@ proc stop*(n: StateNetwork) {.async: (raises: []).} =
   var futures: seq[Future[void]]
   futures.add(n.portalProtocol.stop())
 
-  if not n.processContentLoop.isNil():
-    futures.add(n.processContentLoop.cancelAndWait())
+  for loop in n.processContentLoops:
+    futures.add(loop.cancelAndWait())
   if not n.statusLogLoop.isNil():
     futures.add(n.statusLogLoop.cancelAndWait())
 
   await noCancel(allFutures(futures))
 
-  n.processContentLoop = nil
+  n.processContentLoops.setLen(0)
   n.statusLogLoop = nil
--- a/fluffy/network/state/state_network.nim
+++ b/fluffy/network/state/state_network.nim
@@ -41,6 +41,7 @@ type StateNetwork* = ref object
   historyNetwork: Opt[HistoryNetwork]
   validateStateIsCanonical: bool
   contentRequestRetries: int
+  contentQueueWorkers: int
 
 func toContentIdHandler(contentKey: ContentKeyByteList): results.Opt[ContentId] =
   ok(toContentId(contentKey))
@@ -56,7 +57,11 @@ proc new*(
     historyNetwork = Opt.none(HistoryNetwork),
     validateStateIsCanonical = true,
     contentRequestRetries = 1,
+    contentQueueWorkers = 25,
 ): T =
+  doAssert(contentRequestRetries >= 0)
+  doAssert(contentQueueWorkers >= 1)
+  
   let
     cq = newAsyncQueue[(Opt[NodeId], ContentKeysList, seq[seq[byte]])](50)
     s = streamManager.registerNewStream(cq)
@@ -80,6 +85,7 @@ proc new*(
     historyNetwork: historyNetwork,
     validateStateIsCanonical: validateStateIsCanonical,
     contentRequestRetries: contentRequestRetries,
+    contentQueueWorkers: contentQueueWorkers
   )
 
 proc getContent(
@@ -211,7 +217,7 @@ proc processOffer*(
 
   ok()
 
-proc processContentLoop(n: StateNetwork) {.async: (raises: []).} =
+proc contentQueueWorker(n: StateNetwork) {.async: (raises: []).} =
   try:
     while true:
       let (srcNodeId, contentKeys, contentValues) = await n.contentQueue.popFirst()
@@ -275,8 +281,8 @@ proc start*(n: StateNetwork) =
 
   n.portalProtocol.start()
 
-  for i in 0..<25:
-    n.processContentLoops.add processContentLoop(n)
+  for i in 0 ..< n.contentQueueWorkers:
+    n.processContentLoops.add(contentQueueWorker(n))
 
   n.statusLogLoop = statusLogLoop(n)
 
--- a/fluffy/network/state/state_network.nim
+++ b/fluffy/network/state/state_network.nim
@@ -57,11 +57,11 @@ proc new*(
     historyNetwork = Opt.none(HistoryNetwork),
     validateStateIsCanonical = true,
     contentRequestRetries = 1,
-    contentQueueWorkers = 25,
+    contentQueueWorkers = 8,
 ): T =
   doAssert(contentRequestRetries >= 0)
   doAssert(contentQueueWorkers >= 1)
-  
+
   let
     cq = newAsyncQueue[(Opt[NodeId], ContentKeysList, seq[seq[byte]])](50)
     s = streamManager.registerNewStream(cq)
@@ -85,7 +85,7 @@ proc new*(
     historyNetwork: historyNetwork,
     validateStateIsCanonical: validateStateIsCanonical,
     contentRequestRetries: contentRequestRetries,
-    contentQueueWorkers: contentQueueWorkers
+    contentQueueWorkers: contentQueueWorkers,
   )
 
 proc getContent(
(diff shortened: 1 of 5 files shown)


fluffy/network/state/state_network.nim
```diff
@@ -275,7 +275,9 @@ proc start*(n: StateNetwork) =
 
   n.portalProtocol.start()
 
-  n.processContentLoop = processContentLoop(n)
+  for i in 0..<25:
+    n.processContentLoops.add processContentLoop(n)
```
bhartnett, line 279: @kdeme Do you have any thoughts on whether we should add workers here to make the processContentLoop concurrent and speed up the throughput?

I'm just wondering if there are any downsides to doing this.

kdeme: I see no real reason why we could/should not do this.

The `processOffer` relies on a `async` step (getBlockHeaderByBlockNumOrHash), so I do think it makes sense and should improve content processing.

The improvement will perhaps be less than expected due to the content cache that we have and thus for the same state root (->block number requested), a synchronous path will be taken.  Nevertheless, for the cases where the cache is not hit, it would help.

Similar logic applies on history network, although there we have the `async` step only for 2 of the 4 content types.

kdeme: Actually I do see some concerns, or at least something to keep an eye on when adding this:
- To many workers might perhaps cause you to hit a lot of processing (verification) in a row. We need to make sure that this does not cause the async part (request for block) timeout or so. I don't think this will be an actual issue however due to the short time the verification/validation takes.
- More workers means also more neighborhood gossips started over time. I have not fully thought this through but I believe it does alter the flow of accepting -> processing -> gossiping data. Where now with extra workers the `contentQueue` will be processed quicker, but the neighborhood gossip will `await` at the step of adding to the `offerQueue`, meaning that more of those calls will be in rotation. While before the `contentQueue` would be processed slower, however we would also cut-off accepting more offers when this queue is full.

Not saying we should not go for more workers (>1) here, but we have to careful not to get the node overloaded.

bhartnett: > Actually I do see some concerns, or at least something to keep an eye on when adding this:
> 
> * To many workers might perhaps cause you to hit a lot of processing (verification) in a row. We need to make sure that this does not cause the async part (request for block) timeout or so. I don't think this will be an actual issue however due to the short time the verification/validation takes.

Thanks for the feedback. If I understand you correctly for this point, I guess you mean that running multiple block lookup requests concurrently might increase the risk of timeout. Yes, that could be a problem if setting the worker count too large if the node is not able to handle processing so many content lookup requests concurrently.

> * More workers means also more neighborhood gossips started over time. I have not fully thought this through but I believe it does alter the flow of accepting -> processing -> gossiping data. Where now with extra workers the `contentQueue` will be processed quicker, but the neighborhood gossip will `await` at the step of adding to the `offerQueue`, meaning that more of those calls will be in rotation. While before the `contentQueue` would be processed slower, however we would also cut-off accepting more offers when this queue is full.

Good point, this is something to think about and get right. I'm thinking that the number of state network process content workers which process the `contentQueue` should be less than the number of gossip workers which processes the `offerQueue`. We also need to keep in mind that each offer successfully validated by a content worker will be sent to approx 4 peers (because maxGossipNodes = 4) which creates 4 offer requests on the `offerQueue`. Currently the number of offer workers defaults to 50. Perhaps the number of process content workers should be set to around 1/4th of the offer workers.

bhartnett:
I did some basic performance testing today using a local test network:
- Using 64 nodes all running just on the state network.
- With state root checks (block header lookups) disabled.
- The state bridge ran with 10 concurrent workers
- For the test just the first 2000 blocks were gossipped into the network on each run
- Sending content to only a single node in the test (node 1)
- Metrics collected in each test run from node 1.
- Tested 3 scenarios where the number of workers was configured to by 1, 10 or 25.

Here are the results:

**Using 1 worker:**

```
# HELP portal_offer_accept_codes Portal wire protocol accept codes received from peers after sending offers
# TYPE portal_offer_accept_codes counter
portal_offer_accept_codes_total{protocol_id="500a",accept_code="2"} 22.0
portal_offer_accept_codes_created{protocol_id="500a",accept_code="2"} 1746596324.0
portal_offer_accept_codes_total{protocol_id="500a",accept_code="0"} 42614.0
portal_offer_accept_codes_created{protocol_id="500a",accept_code="0"} 1746596322.0
portal_offer_accept_codes_total{protocol_id="500a",accept_code="4"} 7281.0
portal_offer_accept_codes_created{protocol_id="500a",accept_code="4"} 1746596323.0

# HELP portal_handle_offer_accept_codes Portal wire protocol accept codes returned to peers when handing offers
# TYPE portal_handle_offer_accept_codes counter
portal_handle_offer_accept_codes_total{protocol_id="500a",accept_code="2"} 191120.0
portal_handle_offer_accept_codes_created{protocol_id="500a",accept_code="2"} 1746596322.0

# HELP portal_gossip_offers_successful Portal wire protocol successful content offers from neighborhood gossip
# TYPE portal_gossip_offers_successful counter
portal_gossip_offers_successful_total{protocol_id="500a"} 49906.0
portal_gossip_offers_successful_created{protocol_id="500a"} 1746596322.0

# HELP portal_gossip_offers_failed Portal wire protocol failed content offers from neighborhood gossip
# TYPE portal_gossip_offers_failed counter
portal_gossip_offers_failed_total{protocol_id="500a"} 83.0
portal_gossip_offers_failed_created{protocol_id="500a"} 1746596327.0

# HELP portal_gossip_with_lookup Portal wire protocol neighborhood gossip that required a node lookup
# TYPE portal_gossip_with_lookup counter

# HELP portal_gossip_without_lookup Portal wire protocol neighborhood gossip that did not require a node lookup
# TYPE portal_gossip_without_lookup counter
portal_gossip_without_lookup_total{protocol_id="500a"} 12498.0
portal_gossip_without_lookup_created{protocol_id="500a"} 1746596322.0
```

**Using 10 workers:**

```
# HELP portal_offer_accept_codes Portal wire protocol accept codes received from peers after sending offers
# TYPE portal_offer_accept_codes counter
portal_offer_accept_codes_total{protocol_id="500a",accept_code="2"} 51.0
portal_offer_accept_codes_created{protocol_id="500a",accept_code="2"} 1746596971.0
portal_offer_accept_codes_total{protocol_id="500a",accept_code="0"} 46862.0
portal_offer_accept_codes_created{protocol_id="500a",accept_code="0"} 1746596966.0
portal_offer_accept_codes_total{protocol_id="500a",accept_code="4"} 1773.0
portal_offer_accept_codes_created{protocol_id="500a",accept_code="4"} 1746596969.0

# HELP portal_handle_offer_accept_codes Portal wire protocol accept codes returned to peers when handing offers
# TYPE portal_handle_offer_accept_codes counter
portal_handle_offer_accept_codes_total{protocol_id="500a",accept_code="2"} 108441.0
portal_handle_offer_accept_codes_created{protocol_id="500a",accept_code="2"} 1746596967.0

# HELP portal_gossip_offers_successful Portal wire protocol successful content offers from neighborhood gossip
# TYPE portal_gossip_offers_successful counter
portal_gossip_offers_successful_total{protocol_id="500a"} 48683.0
portal_gossip_offers_successful_created{protocol_id="500a"} 1746596967.0

# HELP portal_gossip_offers_failed Portal wire protocol failed content offers from neighborhood gossip
# TYPE portal_gossip_offers_failed counter
portal_gossip_offers_failed_total{protocol_id="500a"} 41.0
portal_gossip_offers_failed_created{protocol_id="500a"} 1746596976.0

# HELP portal_gossip_with_lookup Portal wire protocol neighborhood gossip that required a node lookup
# TYPE portal_gossip_with_lookup counter

# HELP portal_gossip_without_lookup Portal wire protocol neighborhood gossip that did not require a node lookup
# TYPE portal_gossip_without_lookup counter
portal_gossip_without_lookup_total{protocol_id="500a"} 12181.0
portal_gossip_without_lookup_created{protocol_id="500a"} 1746596966.0
```

**Using 25 workers:**

```
# HELP portal_offer_accept_codes Portal wire protocol accept codes received from peers after sending offers
# TYPE portal_offer_accept_codes counter
portal_offer_accept_codes_total{protocol_id="500a",accept_code="2"} 61.0
portal_offer_accept_codes_created{protocol_id="500a",accept_code="2"} 1746597388.0
portal_offer_accept_codes_total{protocol_id="500a",accept_code="0"} 50010.0
portal_offer_accept_codes_created{protocol_id="500a",accept_code="0"} 1746597382.0
portal_offer_accept_codes_total{protocol_id="500a",accept_code="4"} 94.0
portal_offer_accept_codes_created{protocol_id="500a",accept_code="4"} 1746597384.0

# HELP portal_handle_offer_accept_codes Portal wire protocol accept codes returned to peers when handing offers
# TYPE portal_handle_offer_accept_codes counter
portal_handle_offer_accept_codes_total{protocol_id="500a",accept_code="2"} 117142.0
portal_handle_offer_accept_codes_created{protocol_id="500a",accept_code="2"} 1746597383.0

# HELP portal_gossip_offers_successful Portal wire protocol successful content offers from neighborhood gossip
# TYPE portal_gossip_offers_successful counter
portal_gossip_offers_successful_total{protocol_id="500a"} 50159.0
portal_gossip_offers_successful_created{protocol_id="500a"} 1746597383.0

# HELP portal_gossip_offers_failed Portal wire protocol failed content offers from neighborhood gossip
# TYPE portal_gossip_offers_failed counter
portal_gossip_offers_failed_total{protocol_id="500a"} 45.0
portal_gossip_offers_failed_created{protocol_id="500a"} 1746597387.0

# HELP portal_gossip_with_lookup Portal wire protocol neighborhood gossip that required a node lookup
# TYPE portal_gossip_with_lookup counter

# HELP portal_gossip_without_lookup Portal wire protocol neighborhood gossip that did not require a node lookup
# TYPE portal_gossip_without_lookup counter
portal_gossip_without_lookup_total{protocol_id="500a"} 12551.0
portal_gossip_without_lookup_created{protocol_id="500a"} 1746597382.0
```

bhartnett:
For now the content queue workers are only created in the state network. We can roll a similar change out to both beacon and history networks in a future PR.

I've set the worker count to 8 which is reasonably low. Using a conservative choice for now.

PR #3269, opened 2025-05-08
Beacon sync maintenance update and header download fix

mjfh:


execution_chain/core/chain/header_chain_cache.nim
```diff
@@ -191,31 +186,42 @@ proc delInfo(db: KvtTxRef) =
 
 
 proc putHeader(db: KvtTxRef; h: Header) =
-  ## Store rlp encoded header
+  ## Store the argument `header` indexed by block number, and the hash lookup
+  ## of the parent header.
```
arnetheduck, line 190: why the number of the parent header and not the current one?

mjfh: Seems to avoid computing a hash, cc: @jangko

PR #3270, opened 2025-05-09
fluffy: pass headers into backend procedures instead of stateroot

chirag-parmar:
Passing headers instead of state roots provides the flexibility of querying the state(accounts, storage slots, code) either using the state root(like for the use case of portal network) or using the block number/hash(like for the use case of nimbus verified proxy).

If the header is not passed, nimbus verified proxy will have to maintain a cache mapping state roots to block numbers. IMO this would be an inefficiency because all calls to the async evm are made against headers which already have all the relevant information required.

### DIFF
--- a/fluffy/evm/async_evm.nim
+++ b/fluffy/evm/async_evm.nim
@@ -215,8 +215,7 @@ proc callFetchingState(
           let slotIdx = (adr, v.storageSlot)
           if slotIdx notin fetchedStorage:
             debug "Fetching storage slot", address = adr, slotKey = v.storageSlot
-            let storageFut =
-              evm.backend.getStorage(header, adr, v.storageSlot)
+            let storageFut = evm.backend.getStorage(header, adr, v.storageSlot)
             if not stateFetchDone:
               storageQueries.add(StorageQuery.init(adr, v.storageSlot, storageFut))
               if not optimisticStateFetch:
--- a/fluffy/evm/async_evm.nim
+++ b/fluffy/evm/async_evm.nim
@@ -82,15 +82,15 @@ type
     address: Address
     codeFut: Future[Opt[seq[byte]]]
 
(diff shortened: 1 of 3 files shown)


kdeme (commented):
> If the header is not passed, nimbus verified proxy will have to maintain a cache mapping state roots to block numbers. IMO this would be an inefficiency because all calls to the async evm are made against headers which already have all the relevant information required.

I think it is fine to pass the header, because from the quick look I had, it seems that the calls build on top of those are already doing that. But I'll let @bhartnett double check if there are not issues with this.

So what you are saying is that in the purifier implementation if the `GetAccountProc`, `GetStorageProc` and `GetCodeProc` you need to have access to the right block number/hash to be able to request the proper data, right?

kdeme:
I see https://github.com/status-im/nimbus-eth1/pull/3271/files already has this adjustment.

bhartnett:
My initial concern with using block header here was that I thought nim would copy the header when passed into each function but it turns out that nim actually passes these value types by reference when they are larger than a certain size. In my testing anything larger than around 20-30 bytes is passed by reference (value types). 

So I approve of these changes but I've converted this to a draft PR because I've already pull this into my PR here: https://github.com/status-im/nimbus-eth1/pull/3271

chirag-parmar:
> So what you are saying is that in the purifier implementation if the GetAccountProc, GetStorageProc and GetCodeProc you need to have access to the right block number/hash to be able to request the proper data, right?

Yes

chirag-parmar:
> So I approve of these changes but I've converted this to a draft PR because I've already pull this into my PR here: https://github.com/status-im/nimbus-eth1/pull/3271

Shall I close this PR?

bhartnett:
> > So I approve of these changes but I've converted this to a draft PR because I've already pull this into my PR here: #3271
> 
> Shall I close this PR?

Yes, we can close it.

PR #3273, opened 2025-05-12
Fluffy: Use neighboursInRange in neighborhoodGossip and filter out src node in handleFindContent

bhartnett:
Changes in this PR:
- Created new helper function to return nodes which are in range of a content id from the routing table.
- Use the new `neighboursInRange` function in `neighborhoodGossip` which should result in a minor improvement where we should get slightly more in range nodes from the routing table which should reduce the number of node lookups overall.
- Fixed a bug in handleFindContent where we were not correctly filtering out the source node as defined in the portal wire spec.

This change depends on the nim-eth update in this PR: https://github.com/status-im/nim-eth/pull/792

### DIFF
--- a/fluffy/network/wire/portal_protocol.nim
+++ b/fluffy/network/wire/portal_protocol.nim
@@ -552,7 +552,7 @@ proc handleFindContent(
   # Node does not have the content, or content is not even in radius,
   # send closest neighbours to the requested content id.
   let
-    closestNodes = p.neighbours(NodeId(contentId), seenOnly = true)
+    closestNodes = p.neighboursInRange(contentId, seenOnly = true)
     enrs = truncateEnrs(closestNodes, maxPayloadSize, enrOverhead)
   portal_content_enrs_packed.observe(enrs.len().int64, labelValues = [$p.protocolId])
 
--- a/fluffy/network/wire/portal_protocol.nim
+++ b/fluffy/network/wire/portal_protocol.nim
@@ -360,11 +360,6 @@ func getNode*(p: PortalProtocol, id: NodeId): Opt[Node] =
 func localNode*(p: PortalProtocol): Node =
   p.baseProtocol.localNode
 
-template neighbours*(
-    p: PortalProtocol, id: NodeId, k: int = BUCKET_SIZE, seenOnly = false
-): seq[Node] =
-  p.routingTable.neighbours(id, k, seenOnly)
-
 func distance(p: PortalProtocol, a, b: NodeId): UInt256 =
   p.routingTable.distance(a, b)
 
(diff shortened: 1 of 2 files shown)


fluffy/network/wire/portal_protocol.nim
```diff
@@ -542,8 +558,11 @@ proc handleFindContent(
 
   # Node does not have the content, or content is not even in radius,
   # send closest neighbours to the requested content id.
+
   let
-    closestNodes = p.neighbours(NodeId(contentId), seenOnly = true)
```
bhartnett, line 546: I think we had a bug here where we weren't filtering out the src node and local node.

fluffy/network/wire/portal_protocol.nim
```diff
@@ -542,8 +558,11 @@ proc handleFindContent(
...
+    closestNodes = p.neighboursInRange(
+      contentId, seenOnly = true, excluding = toHashSet([p.localNode.id, srcId])
+    )
```
bhartnett, line 565: Actually does the routing table ever contain the local node? Not sure if that one needs to be filtered out.

kdeme: No, the local node cannot get added to the routing_table: https://github.com/status-im/nim-eth/blob/master/eth/p2p/discoveryv5/routing_table.nim#L394

fluffy/network/wire/portal_protocol.nim
```diff
@@ -380,6 +380,22 @@ func inRange(
...
+    id: ContentId,
+    k: int = BUCKET_SIZE,
+    seenOnly = false,
+    excluding = initHashSet[NodeId](),
+): seq[Node] =
+  func nodeInRange(nodeId: NodeId): bool =
+    if excluding.contains(nodeId):
+      return false
+    let radius = p.radiusCache.get(nodeId).valueOr:
+      return false
```
kdeme, line 394: Fine for the NHgossip, but for `FindContent` we are now not forwarding potential good node because we might not have their radius information.

This gets quiet difficult to understand what would be best; not forward them at all or do forward them and the receiving node might actually have the radius information needed. I would stick with the specification here for now, but you can always bring this up in the specs.

bhartnett: Good point. Ok I'll remove the change to FindContent for now.

bhartnett: I've rolled back the change to handleFindContent so that it returns all closest nodes ignoring the radius as before but I've kept the fix to filter out the source node which should be done according to the portal wire specs.

kdeme (approved):

PR #3275, opened 2025-05-12
Remove account leaf cache

arnetheduck:
With static vid lookup we can find cached accounts in `O(1)` in the regular caches - no need to waste time and effort on maintaining a separate path-based cache for it.

5% speed bump from just the removal - more to be had by reallocating the memory to other causes.


arnetheduck:
this one is not a db change, once the underlying static-vid branch is merged

Issue #3280, opened 2025-05-13
Error: unhandled exception: field 'number' is not accessible for type 'BlockHashOrNumber' using 'isHash = true' [FieldDefect]


yorickdowne:
Nimbus EL `v0.1.0-d3215efb/linux-amd64/Nim-2.2.4`
Lighthouse `v7.0.1-e42406d`

Nimbus synced on mainnet with history expiry

I see the above error message and Nimbus fails. The attached logs show no Portal client; this also happens when a Portal URL is provided.

Attached 1000 lines of log - let me know if you need DEBUG

[nimbus-el-fail.log](https://github.com/user-attachments/files/20185110/nimbus-el-fail.log)

kdeme:
Thanks for the report.

@advaita-saha 

This does seem to be related to the addition of Portal support:
Looks like the issue is the accessing of the number field (https://github.com/status-im/nimbus-eth1/blob/d3215efbe9783b05711409c257b1658862fcb348/execution_chain/sync/wire_protocol/handler.nim#L155) while it can also be a block hash (`BlockHashOrNumber` is used).

yorickdowne:
I connected a Fluffy to this and errors continue. On the Fluffy side I see:

```
portal-1  | INF 2025-05-13 10:27:54.320+00:00 Portal node status                         radiusPercentage=100% radius=ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff dbSize=1300094kb
portal-1  | WRN 2025-05-13 10:28:16.304+00:00 Failed to get block header from history    topics="portal_state" blockNumOrHash=0x8661fab6bf7b65f5ad6dcb05ae0aaaa938b3a592f11264cd93dca32bb732513c
portal-1  | ERR 2025-05-13 10:28:16.304+00:00 Received offered content failed validation topics="portal_state" srcNodeId=ok(d6f157e98c663d0ad3be093932ca7a0e7f5dbb1da72c327325fe7032ea310509) contentKeyBytes=0x20240000001c51b558b934ee5b3db6b26a09ddbcc4501e737501a09dd03fabcc01724bf2fb000cc6 error="Failed to get block header by hash"
portal-1  | WRN 2025-05-13 10:28:16.352+00:00 Failed to get block header from history    topics="portal_state" blockNumOrHash=0x8661fab6bf7b65f5ad6dcb05ae0aaaa938b3a592f11264cd93dca32bb732513c
[...]

PR #3282, opened 2025-05-13
Fix number field not accessible when block header by hash

kdeme:
Potential quickfix for https://github.com/status-im/nimbus-eth1/issues/3280

I have not tested it.


advaita-saha (approved):
LGTM
Should work, haven't tested though

PR #3283, opened 2025-05-13
Support multiple wire protocol version and implement eth/68 + eth/69

### LINKED ISSUE #3245: Support eth/69 wire protocol, optionally besides eth/68
If we want to support multiple protocol versions at runtime, we need to first remove existing `p2pProtocol` macro usage.
This has been done for `eth/68`, only `devp2p` protocol need refactoring.

Then from there we can modify `rlpx` interface to support multiple protocol version.

If we only want to support `eth/69`, just remove some fields and methods.

Possible inclusion in Osaka:
- https://github.com/ethereum/EIPs/blob/6ac7c8f8dbbc701f24aa3b69a910fab2bfdf7b23/EIPS/eip-7607.md#other-eips

Spec:
- https://github.com/ethereum/devp2p/blob/b0c213de97978053a0f62c3ea4d23c0a3d8784bc/caps/eth.md#eth69-eip-7642-april-2025

jangko:
fix #3245 

Turn out it is easier to support multiple eth/XX version than I thought before.
No need to modify the rlpx too much.
This PR keep the `eth/68` and add the new `eth/69`.


advaita-saha:
maybe issue with the encoding 
getting `rlp.nim(43)              raiseExpectedBlob` in eth69

Maybe try adding a roundtrip test to the append and read

jangko:
Looks like missing this:

```Nim
proc read*(rlp: var Rlp, T: type Receipt69): T {.raises: [RlpError].} =
  rlp.read(result)
```

advaita-saha:
Now you can bump nim-eth to master, the PR is merged

PR #3285, opened 2025-05-13
Bump nim-eth to 29ce002cbbf5e0d0b92f154ef37146a8d305cefe

bhartnett:


bhartnett:
> @chirag-parmar The CI failures here appear to be from your recent commit in nim-eth here: [status-im/nim-eth#787](https://github.com/status-im/nim-eth/pull/787)
> 
> Can you please have a look?

I've reverted the nim-eth change which caused the broken tests for now. This change will need to be tested and re-merged once ready.

PR #3288, opened 2025-05-15
proxy: swap block cache for header store

chirag-parmar:
This PR swaps out the block cache for the header cache. There are multiple reasons as to why this is more efficient.

1. `BlockObject`s were manually downloaded over the P2P layer and were cached without verification. Storing blocks without verification provides no benefit in terms of trust for the verified proxy. If verified, the proxy wouldn't be a light client (EL).
2. All light client updates are headers and have trust associated with them via sync committee and therefore it makes more sense to cache headers.
3. The main use case of the proxy, wallets, tend to query state data more often than block data. Block data is mostly required for tx verification which can be done without the entire block being downloaded (not fully true?)

Because of the swap, some RPC methods that depend on the `BlockData` are removed. Refer this draft PR which already has most of the modified code: https://github.com/status-im/nimbus-eth1/pull/3100

This PR also refactors the `run` method for better code readability.

- [x] refactor `run`
- [x] swap out blockcache for headerstore
- [x] tests for header store


nimbus_verified_proxy/header_store.nim
```diff
@@ -0,0 +1,147 @@
...
+        excessBlobGas = Opt.some(p.excess_blob_gas)
+        parentBeaconBlockRoot = Opt.some(forkyHeader.beacon.parent_root.asBlockHash)
+    else:
+      let
+        blobGasUsed = Opt.none(uint64)
+        excessBlobGas = Opt.none(uint64)
+        parentBeaconBlockRoot = Opt.none(Hash32)
+
+    when lcDataFork >= LightClientDataFork.Electra:
+      # TODO: there is no visibility of the execution requests hash in light client header
```
tersec, line 48: What observable effects does this have?

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -73,16 +66,32 @@ proc run*(
...
+  var verifiedProxy = VerifiedRpcProxy.new(rpcProxy, headerStore, chainId)
+
+  # add handlers that verify RPC calls /rpc/rpc_eth_api.nim
+  verifiedProxy.installEthApiHandlers()
+
+  # just for short hand convenience
   template cfg(): auto =
     metadata.cfg
 
+  # initiialize beacon node genesis data, beacon clock and forkDigests
```
tersec, line 94:
```suggestion
  # initialize beacon node genesis data, beacon clock and forkDigests
```

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -117,114 +136,55 @@ proc run*(
...
+    # light client is set to optimistic finalization mode
     lightClient = createLightClient(
       network, rng, lcConfig, cfg, forkDigests, getBeaconTime, genesis_validators_root,
       LightClientFinalizationMode.Optimistic,
     )
 
-  verifiedProxy.installEthApiHandlers()
-
-  info "Listening to incoming network requests"
+  # find out what this does
```
tersec, line 145: seems like a TODO still?

nimbus_verified_proxy/header_store.nim
```diff
@@ -0,0 +1,147 @@
...
+        withdrawalsRoot: withdrawalsRoot,
+        blobGasUsed: blobGasUsed,
+        excessBlobGas: excessBlobGas,
+        parentBeaconBlockRoot: parentBeaconBlockRoot,
+        requestsHash: requestsHash,
+      )
+    else:
+      # INFO: should never reach this point because running verified
+      # proxy for altair doesn't make sense
+      let h = Header()
```
tersec, line 80: If the verified proxy/purifier is going to be this nonfunctional pre-Bellatrix (completely reasonable), then it should be more explicit than this on the UX side. Maybe this isn't the PR for that, i.e. the `HeaderStore` implementation might not be where this should live.

nimbus_verified_proxy/header_store.nim
chirag-parmar: it doesn't affect the validation pipeline for execution block validation

nimbus_verified_proxy/header_store.nim
```diff
@@ -0,0 +1,147 @@
...
+        parentBeaconBlockRoot = Opt.some(forkyHeader.beacon.parent_root.asBlockHash)
+    else:
+      let
+        blobGasUsed = Opt.none(uint64)
+        excessBlobGas = Opt.none(uint64)
+        parentBeaconBlockRoot = Opt.none(Hash32)
+
+    when lcDataFork >= LightClientDataFork.Electra:
+      # TODO: there is no visibility of the execution requests hash in light client header 
+      let requestsHash = Opt.none(Hash32)
```
tersec, line 49: (unfortunately also) these can be both `const`

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -73,16 +66,32 @@ proc run*(
     except Exception:
       notice "commandLineParams() exception"
 
+  # load constants and metadata for the selected chain
+  let metadata = loadEth2Network(config.eth2Network)
+
+  # initialze verified proxy
```
tersec, line 72: typo

nimbus_verified_proxy/header_store.nim
```diff
@@ -0,0 +1,147 @@
...
+  web3/eth_api_types,
+  std/tables,
+  beacon_chain/spec/beaconstate,
+  beacon_chain/spec/datatypes/[phase0, altair, bellatrix],
+  beacon_chain/[light_client, nimbus_binary_common],
+  minilru,
+  results
+
+type HeaderStore* = ref object
+  headers: LruCache[Hash32, Header]
```
bhartnett, line 21: Just curious, why do you cache the `Header` instead of the `BlockObject` as was done in the previous `BlockCache`?

nimbus_verified_proxy/header_store.nim
```diff
@@ -0,0 +1,147 @@
...
+  hashes: LruCache[base.BlockNumber, Hash32]
+  capacity: int
```
bhartnett, line 23: It appears that this capacity field isn't used to so can be removed?

nimbus_verified_proxy/header_store.nim
```diff
@@ -0,0 +1,147 @@
...
+    capacity: max,
+  )
+
+func len*(self: HeaderStore): int =
+  len(self.headers)
+
+func isEmpty*(self: HeaderStore): bool =
+  len(self.headers) == 0
+
+proc add*(self: HeaderStore, header: ForkedLightClientHeader): Result[bool, string] =
```
bhartnett, line 94: This HeaderStore design requires that the headers always get added in the correct order, otherwise the ordering in the LRUCaches will be off. Can we be sure that these headers will always be added in the correct order?

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -73,16 +66,32 @@ proc run*(
...
+    # TODO: write a comment
+    clientConfig = config.web3url.asClientConfig()
 
-  for node in metadata.bootstrapNodes:
-    lcConfig.bootstrapNodes.add node
+    rpcProxy = RpcProxy.new(
+      [initTAddress(config.rpcAddress, config.rpcPort)], clientConfig, authHooks
+    )
+
+    headerStore = HeaderStore.new(64) # block cache contains blocks downloaded from p2p
```
bhartnett, line 83: The 64 value here should be configurable on the cli using a hidden parameter. FYI normally for parameters like this we use the `--debug` prefix to indicate a developer parameter. You can see similar examples in the Nimbus Portal config.

nimbus_verified_proxy/header_store.nim
```diff
@@ -0,0 +1,147 @@
...
+
+  Opt.none(Hash32)
+
+func earliest*(self: HeaderStore): Opt[Header] =
+  if self.headers.len() == 0:
+    return Opt.none(Header)
+
+  var hash: Hash32
+  for h in self.headers.keys:
+    hash = h
```
bhartnett, line 126: You can reuse the `earliestHash` function in this `earliest` function to avoid code duplication.

nimbus_verified_proxy/header_store.nim
```diff
@@ -0,0 +1,147 @@
...
+
+  self.headers.peek(hash)
+
+func earliestHash*(self: HeaderStore): Opt[Hash32] =
+  if self.headers.len() == 0:
+    return Opt.none(Hash32)
+
+  var hash: Hash32
+  for h in self.headers.keys:
+    hash = h
```
bhartnett, line 136: This is a rather inefficient because it needs to loop through the entire list to get the earliest key. Perhaps a better way would be to simple store the earliest block number as a separate field in the cache and increment it as each new header is stored. The block number can then be used to return the earliest hash on demand.

nimbus_verified_proxy/rpc/rpc_eth_api.nim
```diff
@@ -39,144 +39,13 @@ type
...
-
-proc getBlockByTagOrThrow(
-    proxy: VerifiedRpcProxy, quantityTag: BlockTag
-): BlockObject {.raises: [ValueError].} =
-  getBlockByTag(proxy, quantityTag).valueOr:
-    raise newException(ValueError, "No block stored for given tag " & $quantityTag)
-
 proc installEthApiHandlers*(lcProxy: VerifiedRpcProxy) =
   lcProxy.proxy.rpc("eth_chainId") do() -> UInt256:
     lcProxy.chainId
```
bhartnett, line 48: Why are all these implemented RPC endpoints being removed/deleted below? I didn't see them being added anywhere else. This PR should focus just on updating the `HeaderStore`

bhartnett:
@chirag-parmar The description of this PR should explain why the `BlockCache` is being replaced by the `HeaderStore`. In other words it would be helpful to know the high level reason for this change. 

Adding more documentation in the PR like this will help others quickly understand the PR without having to dig too much into the details.

nimbus_verified_proxy/rpc/rpc_eth_api.nim
```diff
@@ -185,22 +54,13 @@ proc installEthApiHandlers*(lcProxy: VerifiedRpcProxy) =
   lcProxy.proxy.registerProxyMethod("eth_sendRawTransaction")
   lcProxy.proxy.registerProxyMethod("eth_getTransactionReceipt")
```
bhartnett, line 56: More RPC endpoints being removed here. Not sure why.

nimbus.nimble
```diff
@@ -125,7 +125,7 @@ task nimbus_verified_proxy, "Build Nimbus verified proxy":
   buildBinary "nimbus_verified_proxy", "nimbus_verified_proxy/", "-d:chronicles_log_level=TRACE"
 
 task nimbus_verified_proxy_test, "Run Nimbus verified proxy tests":
-  test "nimbus_verified_proxy/tests", "test_proof_validation", "-d:chronicles_log_level=ERROR -d:nimbus_db_backend=sqlite"
```
bhartnett, line 128: You might also need to update the make file.

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -117,114 +136,63 @@ proc run*(
...
     )
 
-  verifiedProxy.installEthApiHandlers()
-
-  info "Listening to incoming network requests"
+  # registerbasic p2p protocols for maintaing peers ping/status/get_metadata/... etc.
   network.registerProtocol(
     PeerSync,
     PeerSync.NetworkState.init(cfg, forkDigests, genesisBlockRoot, getBeaconTime),
   )
```
bhartnett, line 149: Why was this code below removed?

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -117,114 +136,63 @@ proc run*(
       rng, lcConfig, netKeys, cfg, forkDigests, getBeaconTime, genesis_validators_root
     )
```
bhartnett, line 138: Why was the below code removed? Or was it just moved? In my opinion, the way this file has been refactored/re-organised makes it hard to tell what has actually changed and I'm not sure of the reason for the restructure here.  Generally, it would be better to put refactors that don't change any functionality in separate PRs as this doesn't look specifically related to swaping in the header store.

nimbus_verified_proxy/nimbus_verified_proxy.nim
chirag-parmar: great point

nimbus_verified_proxy/header_store.nim
chirag-parmar: true that

chirag-parmar: on second thought, use earliestHash will require unnecessary copying/initialization of `Opt`  and then a branching condition to resolve the option. 2 lines of redundancy is acceptable, IMHO.

nimbus_verified_proxy/header_store.nim
chirag-parmar: We can never be sure if the headers stored have a time gap, In other words, the proxy can go offline for sometime and then come back online to receive newer updates. In such situations, updating earliest might get difficult. Still can be done, but I think it would be better if we tackle this in another PR. What do you think? (if you agree, I'll create a follow-up issue so we don't forget about it)

chirag-parmar:
> The description of this PR should explain why the BlockCache is being replaced by the HeaderStore. In other words it would be helpful to know the high level reason for this change.

Agreed and it is a very valid point. Will add a description

nimbus_verified_proxy/header_store.nim
chirag-parmar: Added the reason in the description of the PR

nimbus_verified_proxy/header_store.nim
chirag-parmar: rectified

nimbus_verified_proxy/rpc/rpc_eth_api.nim
chirag-parmar: Added the reason in the above PR

nimbus_verified_proxy/header_store.nim
kdeme: I think the main idea is to only store the headers as they come in via the consensus client REST API and request the blocks "on demand". To keep the client more light weight.

See also discussion/explanation here: https://github.com/status-im/nimbus-eth1/pull/3100#discussion_r1985198204

nimbus_verified_proxy/header_store.nim
kdeme: I had a similar remark here: https://github.com/status-im/nimbus-eth1/pull/3100#discussion_r2086375132

nimbus_verified_proxy/nimbus_verified_proxy.nim
kdeme: Most of that code was there to follow libp2p block gossip (for storing new blocks in the BlockCache). Some seems to be code that was just moved.

nimbus_verified_proxy/rpc/rpc_eth_api.nim
kdeme: Yes, we should not be removing RPC endpoints. Same functionality must remain available.

If for some reason of change in implementation, an RPC endpoint can temporarily not be served, it should be proxied for now. But ideally we do not remove more "purified" functionality.

nimbus_verified_proxy/tests/test_header_store.nim
```diff
@@ -0,0 +1,130 @@
...
+    capellaData: capella.LightClientHeader(
+      beacon: default(capella.BeaconBlockHeader),
+      execution: capella.ExecutionPayloadHeader(
+        block_number: uint64(number), block_hash: Hash32(toBytesBE(u256(number)))
+      ),
+      execution_branch: default(capella.ExecutionBranch),
+    ),
+  )
+
+let store = HeaderStore.new(64)
```
kdeme, line 31: We typically write tests such that they can be run independently. It seems that here the tests are written in such way that they each depend on the previous test cases being run. This should be changed so that each test cases stands on its own.

chirag-parmar: fixed

nimbus_verified_proxy/rpc/rpc_eth_api.nim
chirag-parmar: added them back. They now use `HeaderStore`

nimbus_verified_proxy/header_store.nim
chirag-parmar: Addressed this concern in the latest commit

nimbus_verified_proxy/header_store.nim
chirag-parmar: So while addressing hte `HeaderStore` ordering issue I realized that `earliest` is used for walking blocks (see draft PR). In which case, it could also be the earliest finalized header seen by the proxy (regardless of pruned headers). So I added in support for storing finalized headers and using them to store earliest headers.

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
```diff
@@ -64,6 +64,14 @@ type VerifiedProxyConf* = object # Config
     name: "data-dir"
   .}: OutDir
 
+  # In-Memory Cache Size
+  cacheLen* {.
+    desc: "Length of the header cache maintained in memory",
+    defaultValue: 64,
+    defaultValueDesc: "64",
+    name: "cache-len"
+  .}: int
```
bhartnett, line 73: This should probably be a hidden parameter as it doesn't need to be user facing but will likely be useful for us (developers). The name should start with 'debug'. Here is an example from Nimbus Portal:
```
    contentCacheSize* {.
      hidden,
      desc:
        "Size of the in memory local content cache. This is the max number " &
        "of content values that can be stored in the cache.",
      defaultValue: defaultPortalProtocolConfig.contentCacheSize,
      name: "debug-content-cache-size"
    .}: int
```

chirag-parmar:
@kdeme and @tersec shall I merge this once the pipeline passes?

tersec (approved):

PR #3300, opened 2025-05-19
eth/69: Disconnect peer when receive invalid blockRangeUpdate

### LINKED ISSUE #3299: disconnect on invalid `blockRangeUpdate` in eth69
Validate received BlockRangeUpdate. 
Hive test sends an invalid announcement with earliest > latest and expects a disconnect, but Nimbus just keeps the connection open

Ref: https://hivetests2.ethdevops.io/suite.html?suiteid=1747568995-c5cb8c9886dce0c681fb730ccfe8c348.json&suitename=eth#test-29

jangko:
Hive test sends an invalid announcement with earliest > latest and expects a disconnect, but Nimbus just keeps the connection open.

fix #3299

### DIFF
--- a/execution_chain/sync/wire_protocol/responder.nim
+++ b/execution_chain/sync/wire_protocol/responder.nim
@@ -284,7 +284,7 @@ proc blockRangeUpdateUserHandler(peer: Peer; packet: BlockRangeUpdatePacket) {.
       latestHash = packet.latestHash.short
 
   if packet.earliest > packet.latest:
-    trace "Disconnecting peer because of protocol breach",
+    debug "Disconnecting peer because of protocol breach",
       remote = peer.remote, clientId = peer.clientId,
       msg = "blockRangeUpdate must have latest >= earliest"
     await peer.disconnect(BreachOfProtocol)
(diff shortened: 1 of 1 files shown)


execution_chain/sync/wire_protocol/responder.nim
```diff
@@ -283,6 +283,10 @@ proc blockRangeUpdateUserHandler(peer: Peer; packet: BlockRangeUpdatePacket) {.
       earliest = packet.earliest, latest = packet.latest,
       latestHash = packet.latestHash.short
 
+  if packet.earliest > packet.latest:
+    await peer.disconnect(BreachOfProtocol)
+    return
+
```
advaita-saha, line 289: We should add a trace log here, incase some client mentions breach of protocol from our side, we know why it is happening

kdeme: I would suggest a `debug` log, as it it something that really should not happen so you want to see it in debug logs to inform other clients for wrong behaviour, and I assume `trace` is rarely activated?

jangko: I agree we should use `debug`.

advaita-saha: Sure am fine with debug too

PR #3303, opened 2025-05-19
Portal Client: Remove offer workers and support sending rate limited offers

bhartnett:
The overall goal of this change was to improve the neighborhood gossip function which is used by the portal bridge so that we can actually see how many peers accepted the gossiped content in the put content JSON-RPC response, rather than just the attempted count which is what is currently returned. This is important so the bridge can know when to retry sending content if no peers actually accepted or if the offers requests failed.

Changes in this PR:
- Removed the offer workers and offer queue. These were both used to implement rate limited concurrent sending of offers.
- Added a token bucket to PortalProtocol used to rate limit offer sending.
- Updated neighborhood gossip to use the new rate limited offer functions which now check the accept code in the offer response.
- Updated random gossip to use the new rate limited offer functions as well.
- History network and beacon network no longer us asyncSpawn when calling neighborhood gossip.
- Updated poke to use the new rate limited offer functions. `triggerPoke` is called using asyncSpawn which should be safe because it only sends offers when the rate limit is not at capacity.
- Added content queue workers to history and beacon network so we can process the received content concurrently. This is important now due to removing the async queue and no longer using asyncSpawn.

### DIFF
--- a/portal/network/wire/portal_protocol.nim
+++ b/portal/network/wire/portal_protocol.nim
@@ -1745,30 +1745,14 @@ proc queryRandom*(
 proc offerBatchGetPeerCount*(
     p: PortalProtocol, offers: seq[OfferRequest]
 ): Future[int] {.async: (raises: [CancelledError]).} =
-  # Start up to offers.len() concurrent offers and collect the futures
-  var futs = newSeqOfCap[Future[PortalResult[ContentKeysAcceptList]]](offers.len())
-  for offer in offers:
-    futs.add(p.offerRateLimited(offer))
+  let futs = await allFinished(offers.mapIt(p.offerRateLimited(it)))
 
-  # Await each future and for each successful offer where at least one content item
-  # was accepted or has already been stored, we add to the peer count.
   var peerCount = 0
   for f in futs:
-    let acceptList =
-      try:
-        (await f).valueOr:
-          continue
-      except CancelledError as e:
-        raise e
-      except CatchableError as e:
-        raiseAssert(e.msg) # Shouldn't happen
+    if f.completed() and f.value().isOk():
+      inc peerCount # only count successful offers
 
-    for acceptCode in acceptList:
-      if acceptCode == Accepted or acceptCode == DeclinedAlreadyStored:
-        inc peerCount
-        break # only need to get one success to count the peer
-
-  return peerCount
+  peerCount
 
 proc neighborhoodGossip*(
     p: PortalProtocol,


portal/network/wire/portal_protocol.nim
```diff
@@ -1731,6 +1743,35 @@ proc queryRandom*(
...
+  for offer in offers:
+    futs.add(p.offerRateLimited(offer))
+
+  # Await each future and for each successful offer where at least one content item
+  # was accepted, add to the peer count
+  var peerAcceptedCount = 0
+  for f in futs:
+    let acceptList =
+      try:
+        (await f).valueOr:
```
bhartnett, line 1760: Unfortunately, the await on the future here can throw CatchableError. Not sure why it does but I suspect it may be a CancelledError. This is the reason for the except in this section in order to make the compiler happy.

portal/network/wire/portal_protocol.nim
```diff
@@ -1329,19 +1339,21 @@ proc triggerPoke*(
...
+      # tryConsume actually deducts tokens and there is currently
+      # no API to check the remaining capacity of the bucket so we just
+      # add the token back here
+      p.offerTokenBucket.replenish(1)
+
+      let
+        contentKV = ContentKV(contentKey: contentKey, content: content)
+        list = List[ContentKV, contentKeysLimit].init(@[contentKV])
+        req = OfferRequest(dst: node, kind: Direct, contentList: list)
+      discard await p.offerRateLimited(req)
```
bhartnett, line 1352: This could slow down the queries which call poke. Not sure if there is a better way to do this. Would it be safe to use asyncSpawn here?

bhartnett: Actually I think this will be ok because `triggerPoke` can safely be called using asyncSpawn.

portal/network/history/history_network.nim
```diff
@@ -161,7 +162,7 @@ proc getVerifiedBlockHeader*(
     n.portalProtocol.storeContent(
       contentKey, contentId, headerContent.content, cacheContent = true
     )
-    n.portalProtocol.triggerPoke(
+    asyncSpawn n.portalProtocol.triggerPoke(
```
bhartnett, line 165: I believe it should be safe to use asyncSpawn when calling `triggerPoke` because it won't be blocked by the offer token bucket filling up. If the offer bucket is full the poke implementation will simply skip sending each offer.

portal/network/wire/portal_protocol.nim
```diff
@@ -1731,6 +1743,35 @@ proc queryRandom*(
...
+        (await f).valueOr:
+          continue
+      except CatchableError as e:
+        if e of CancelledError:
+          raise cast[ref CancelledError](e)
+        else:
+          raiseAssert(e.msg) # Shouldn't happen
+
+    for acceptCode in acceptList:
+      if acceptCode == Accepted or acceptCode == DeclinedAlreadyStored:
```
bhartnett, line 1769: For now counting both Accepted and DeclinedAlreadyStored since we currently only have one peerCount fields to work with and it is useful to know how many peers in total have the content regardless of whether they just received it or already have previously stored it. In the future I may suggest a spec change on the put content JSON-RPC to support returning more metadata in the response if it becomes necessary to improve the bridges.

portal/network/wire/portal_protocol.nim
bhartnett: Sure, I'll do that. Much cleaner.

portal/network/wire/portal_protocol.nim
kdeme: Yes, I think it is fine to use asyncSpawn on the caller site considering it will not block on a full token bucket

portal/network/wire/portal_protocol.nim
```diff
@@ -1218,18 +1220,26 @@ proc offer*(
...
+    p: PortalProtocol, offer: OfferRequest
+): Future[PortalResult[ContentKeysAcceptList]] {.async: (raises: [CancelledError]).} =
+  try:
+    await p.offerTokenBucket.consume(1)
+  except CatchableError as e:
+    if e of CancelledError:
+      raise cast[ref CancelledError](e)
     else:
-      portal_gossip_offers_failed.inc(labelValues = [$p.protocolId])
+      raiseAssert(e.msg) # Shouldn't happen
```
kdeme, line 1232: same comment as lower down

portal/network/wire/portal_protocol.nim
```diff
@@ -1731,6 +1743,35 @@ proc queryRandom*(
...
+): Future[int] {.async: (raises: [CancelledError]).} =
+  # Start up to offers.len() concurrent offers and collect the futures
+  var futs = newSeqOfCap[Future[PortalResult[ContentKeysAcceptList]]](offers.len())
+  for offer in offers:
+    futs.add(p.offerRateLimited(offer))
+
+  # Await each future and for each successful offer where at least one content item
+  # was accepted we add to the peer count
+  var peerCount = 0
+  for f in futs:
```
kdeme, line 1757: Could also use `allFutures` from chronos

portal/network/wire/portal_protocol.nim
bhartnett: Actually I do find the spec for this endpoint to be a bit unclear. The description for the PutContentResult endpoint says this:
`Returns the number of peers that the content was gossiped to and a flag indicating whether the content was stored locally or not`. 

So the question is what does `gossiped to` mean? I thought it meant the number of peers to which the content was transferred/sent, not just the number of peers to which the content was offered.

kdeme: Yeah, I can see how it could get interpreted like that. I just remember from the past that it is meant to be interpreted like I said, as I've seen this discussion before.

But as stated, I would be pro adding more meta info from the accept codes.

kdeme (approved):

PR #3313, opened 2025-05-21
Portal Client: Add debug parameter to generate node id having a given prefix

bhartnett:


bhartnett:
When making this change, I've discovered a minor issue where when a private key is supplied on the command line then we don't write it out to the key file but the enr on disk does get updated. If the node previously was started without the the private key param then this leads to a problem where the key file and enr on disk are out of sync and then we get this error:
```
lientcolonObjectType_EEN25nimbus_portal_client_conf10PortalConfE
/home/user/development/status-im/nimbus-eth1/vendor/nim-eth/eth/p2p/discoveryv5/protocol.nim(1089) _ZN8protocol11newProtocolEN4keys10PrivateKeyE3OptIN3net9IpAddressEE3OptIN13nativesockets4PortEE3OptIN13nativesockets4PortEE9openArrayI5tupleI6string3seqI5uInt8EEE9openArrayIN3enr6RecordEE3OptIN3enr6RecordEEN13nativesockets4PortEN3net9IpAddressE4bool4boolN8protocol15DiscoveryConfigE3refIN12bearssl_rand15HmacDrbgContextEE.constprop.0
/home/user/development/status-im/nimbus-eth1/vendor/nim-results/results.nim(946) _ZN7results6expectE6ResultI4void7cstringE6string
/home/user/development/status-im/nimbus-eth1/vendor/nim-results/results.nim(433) _ZN7results17raiseResultDefectE6string7cstring
/home/user/development/status-im/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/excpt.nim(349) _ZN6system18rawWriteStackTraceE3varI3seqIN6system15StackTraceEntryEEE
/home/user/development/status-im/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/stacktraces.nim(62) _ZN11stacktraces30auxWriteStackTraceWithOverrideE3varI3seqIN6system15StackTraceEntryEEE
Error: unhandled exception: Record within size limits and correct key: Public key does not correspond with given private key [ResultDefect]
```

If we no longer write the ENR to disk then this problem will likely be solved. One more reason to do that change.

kdeme (commented):
Would it not make more sense to add this NodeId prefix option as something that runs only on first start for the usual `networkKeyFile` option? So basically, under the condition that the network key file does not exist yet, and else it doesn't do anything.

When using this to set up a few nodes for a bridge it would be better for this generation to happen only once?

bhartnett:
> Would it not make more sense to add this NodeId prefix option as something that runs only on first start for the usual `networkKeyFile` option? So basically, under the condition that the network key file does not exist yet, and else it doesn't do anything.
> 
> When using this to set up a few nodes for a bridge it would be better for this generation to happen only once?

Yes I agree. I'll change it to work this way.

PR #3314, opened 2025-05-22
Portal Client: Support sending only single offer in neighborhoodGossip and randomGossip

bhartnett:


kdeme (commented):
> After looking through the portal specs I also can't find any mention of neighborhood gossip or random gossip needing to support multiple content key/values.


Not specifically mentioned, but an offer can contain up to 64 (I think) content keys.
This is currently probably used by nobody, and as you mentioned, the JSON-RPC is not even specified to support that use case.

However, the original reason to have that was to make injection more efficient, e.g. group same zone of content ids and offer them together. In that scenario it does make sense to also keep NHGossip or randomGossip grouping them together instead of splitting them up as all different offers, which would be less efficient.

We have experience in the past with doing database lookups for close content and offering that. But we never added it in the "field" as it was not a priority to focus on this efficiency improvement. A place where this might also come in handy is the "welcome basket": https://github.com/ethereum/portal-network-specs/issues/350



> 2\. In the history and beacon network, if just one of the content key/values failed validation then all the remaining content keys were being dropped. This is now fixed by processing each key, value separately.


Yes, I considered this a feature. To avoid processing other data that is most likely invalid (as it comes from the same node).

> 1. The `neighborhoodGossip` only used the first content key to find the closest neighbors and therefore any other content keys were likely not being sent to the correct in range peers.


The assumption here is that the original node gossiping this would have properply grouped the content per content id range. Of course this could be improved, one could take the middle of the content range.
In case some node just offers content spread all over the content id range, most of it would not be accepted. But we could have an additional check for this (e.g. in case we are a big radius node and accept most of it, but the range is way to big to make sense).

bhartnett:
> Not specifically mentioned, but an offer can contain up to 64 (I think) content keys. This is currently probably used by nobody, and as you mentioned, the JSON-RPC is not even specified to support that use case.

Yes 64 is the max in the spec but the practical limit depends on how big the content is. For the state network we can't fit 64 offers because they are too big. Currently the only way we can receive the multiple offers like this if when we receive them from other clients and also from the Trin bridge which uses the offer RPC.

> However, the original reason to have that was to make injection more efficient, e.g. group same zone of content ids and offer them together. In that scenario it does make sense to also keep NHGossip or randomGossip grouping them together instead of splitting them up as all different offers, which would be less efficient.

I'm not sure this kind of grouping of content would work very well in practice because if a node has a large radius then it could receive content from many places in the key space even within a single offer request because it should be able to accept them but when it goes to offer the content to other nodes, it can't assume that each content key is close enough to its peers. It would need to actually check if each content id is within the radius of each of its closest peers. Now sure we could build a more complex neighborhood gossip implementation that handles this better and I'm open to doing this but I'm wondering if the extra complexity is worth it.

> Yes, I considered this a feature. To avoid processing other data that is most likely invalid (as it comes from the same node).

Ok interesting. The comment in the code lead me to believe that this was an existing problem, rather than a feature. I wonder how valid this assumption is that the other data is most likely invalid. Isn't it possible the some data gets truncated due to network packet loss or similar and in that case if only one content item fails validation we would still want to process all other content items?

For the scenario where the sending node has a bug or is sending bad content consistently, I think we should use peer scoring and node bans to prevent them continuing to do so.

> The assumption here is that the original node gossiping this would have properply grouped the content per content id range. Of course this could be improved, one could take the middle of the content range. In case some node just offers content spread all over the content id range, most of it would not be accepted. But we could have an additional check for this (e.g. in case we are a big radius node and accept most of it, but the range is way to big to make sense).

I don't think we can make this assumption. Peers can send content that falls anywhere within the nodes radius and with nodes of different sizes I don't see how this idea could work. 

If course it is possible to have some performance improvement by supporting multiple items in neighborhood gossip because we can send batches of the content where they do actually fall within the radius of each peer. But in order to do this more precisely we would need to lookup the closest n nodes for each of the content ids in the batch, and then for each list of n nodes, find the overlapping nodes and then combine the their content ids into a single offer before sending it as a batch but with this method we need to be careful that the batch created is not too large.

kdeme:
> Yes 64 is the max in the spec for offer but the practical limit depends on how big the content is.

It depends on how big the content keys are. Yes, typically there will be not more than 32 that fit.


> For neighborhood gossip it doesn't mention batches.

I do think this originally was the idea, but nobody ever deployed this. It does seem that the specification does not really mention this and can be interpreted as you did due to this specific sentence: `A DHT node is offered and receives a piece of content that it is interested in.`

Either way, our implementation will not differ on whether an offer is 1 or multiple, in order to do NHgossip. So if we interpret the spec as such (as opposed to suggest to clarify it), than we should drop NHgossip for offer len > 1.



> I'm not sure this kind of grouping of content would work very well in practice because if a node has a large radius then it could receive content from many places in the key space even within a single offer request because it should be able to accept them but when it goes to offer the content to other nodes

Yes, which is what I said above with my suggestion to check on id range spread. But getting this right depends on the network composition. Probably difficult to get right and might need adaptation, I guess that's why it didn't happen yet. 

Note also that the cost of adding a few content keys that do not get accepted is not high. Especially compared to sending them separate.

Anyhow, the discussion if this is feasible or not aside, the bigger issue I have with this PR is that it splits a single offer/accept which gets gossiped now,  into potentially way more more gossips. That's to much of an amplification factor imo. Sounds DoS'able. So either we keep sending them as one, or we drop it completely, or we do some smarter filtering of some of the content in there before we forward it.

Another important note: The situation is different for `randomGossip`. This is typically used for content where the content id / radius does not matter. And recently there is actually a use case for sending multiple content items: https://github.com/ethereum/portal-network-specs/blob/master/history/history-network.md#ephemeral-header-by-offer . So these we should definitely not split up.




> Isn't it possible the some data gets truncated due to network packet loss or similar and in that case if only one content item fails validation we would still want to process all other content items?

No, I specifically requested a spec change for that. The node will notice when the data received is incomplete due to the size prefix: https://github.com/status-im/nimbus-eth1/blob/master/portal/network/wire/portal_stream.nim#L347

bhartnett:
> Yes, which is what I said above with my suggestion to check on id range spread. But getting this right depends on the network composition. Probably difficult to get right and might need adaptation, I guess that's why it didn't happen yet.

I see, yeah this could be an interesting solution to explore at some point.

> Note also that the cost of adding a few content keys that do not get accepted is not high. Especially compared to sending them separate.

Yes it is true that it doesn't cost much in terms of performance but it isn't ideal to send the content to the wrong peers and perhaps also not distribute it to interested peers. I would guess that this does happen at least some of the time.

> Anyhow, the discussion if this is feasible or not aside, the bigger issue I have with this PR is that it splits a single offer/accept which gets gossiped now, into potentially way more more gossips. That's to much of an amplification factor imo. Sounds DoS'able. So either we keep sending them as one, or we drop it completely, or we do some smarter filtering of some of the content in there before we forward it.
> 
> Another important note: The situation is different for `randomGossip`. This is typically used for content where the content id / radius does not matter. And recently there is actually a use case for sending multiple content items: https://github.com/ethereum/portal-network-specs/blob/master/history/history-network.md#ephemeral-header-by-offer . So these we should definitely not split up.

Ok, given these points then I won't go ahead with this change but maybe I'll implement an improvement to the multiple neighborhood gossip instead.

> No, I specifically requested a spec change for that. The node will notice when the data received is incomplete due to the size prefix: https://github.com/status-im/nimbus-eth1/blob/master/portal/network/wire/portal_stream.nim#L347

Interesting, ok good to know.

PR #3320, opened 2025-05-22
load the jwt secret from data-dir if available

advaita-saha:
Previously, if `--jwt-secret` was not explicitly passed, the client would unconditionally generate a new JWT secret and write it to `${dataDir}/jwt.hex`. This behavior caused unnecessary secret churn across restarts, breaking authentication with connected CL unless the old key was manually preserved.

This PR introduces a safeguard: if the default `${dataDir}/jwt.hex` file already exists, the client will load and reuse the existing JWT secret instead of generating a new one. A new key is only created if: `--jwt-secret` is not provided, and no `${dataDir}/jwt.hex` file is present

### DIFF
--- a/execution_chain/rpc/jwt_auth.nim
+++ b/execution_chain/rpc/jwt_auth.nim
@@ -1,5 +1,5 @@
 # Nimbus
-# Copyright (c) 2022-2024 Status Research & Development GmbH
+# Copyright (c) 2022-2025 Status Research & Development GmbH
 # Licensed and distributed under either of
 #   * MIT license (license terms in the root directory or at
 #     https://opensource.org/licenses/MIT).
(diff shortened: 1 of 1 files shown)


execution_chain/rpc/jwt_auth.nim
```diff
@@ -209,7 +209,9 @@ proc jwtSharedSecret*(
   # startup, or show error and continue without exposing the authenticated
   # port.
   #
-  if config.jwtSecret.isNone:
+  var jwtSecretPath = config.dataDir.string & "/" & jwtSecretFile # default path
```
tersec, line 212: [`/`](https://nim-lang.org/docs/ospaths2.html#%2F%2Cstring%2Cstring)

PR #3322, opened 2025-05-23
Fusaka-devnet-0/EIP-7823: Set upper bounds for MODEXP

jangko:
fix #7823


execution_chain/evm/precompiles.nim
```diff
@@ -257,6 +257,11 @@ func modExp(c: Computation, fork: EVMFork = FkByzantium): EvmResultVoid =
     expLen  = expL.safeInt
     modLen  = modL.safeInt
 
+  if fork == FkOsaka:
```
advaita-saha, line 260: Shouldn't this be >= FkOsaka ? as we now have Bpo forks

Issue #3101, opened 2025-02-21
Nimbus Verified Proxy


chirag-parmar:
### Pricing Verification
- [x] `eth_blobBaseFee`
- [x] `eth_gasPrice`
- [x] `eth_maxPriorityFeePerGas`

### Execution Block Header Verification (+ Tx Trie Verification)

Notes:  we download execution payload from the beacon light client but they are not verified. Also to maintain memory usage if a block misses the cache we might need to re-download and re-verify the block 

- [x] `eth_getBlockByHash`
- [x] `eth_getBlockByNumber`
- [x] `eth_getUncleCountByBlockHash`
- [x] `eth_getUncleCountByBlockNumber`

### Transaction and Receipt Trie Verification
- [x] `eth_getBlockReceipts`
- [x] `eth_getBlockTransactionCountByHash`
- [x] `eth_getBlockTransactionCountByNumber`
- [x] `eth_getTransactionByBlockHashAndIndex`
- [x] `eth_getTransactionByBlockNumberAndIndex`
- [x] `eth_getTransactionByHash`
- [x] `eth_getTransactionReceipt`
- [x] `eth_feeHistory`

### EVM execution for verification - In Progress
- [x] `eth_call`
- [x] `eth_estimateGas`
- [x] `eth_sendRawTransaction`
- [x] `eth_createAccessList`

### Account Trie Verification - Complete
- [x] `eth_getBalance`
- [x] `eth_getCode`
- [x] `eth_getStorageAt`
- [x] `eth_getTransactionCount`

### Methods requiring a wallet 

Notes: requires an API so that control can be forwarded to user application

- [ ] `eth_accounts`
- [ ] `eth_coinbase`
- [ ] `eth_sign`
- [ ] `eth_signTransaction`
- [ ] `eth_sendTransaction`

### Logs Verification
- [x] `eth_getFilterChanges`
- [x] `eth_getFilterLogs`
- [x] `eth_getLogs`
- [x] `eth_newFilter`
- [x] `eth_uninstallFilter`

### stuff that doesn't need verification or is irrelevant
- [x] `eth_blockNumber`
- [x] `eth_chainId`
- [x] `eth_syncing`
- [ ] `eth_newBlockFilter`
- [ ] `eth_newPendingTransactionFilter`

## Other Improvements
- [x] ~use hardcoded bootstrapping light client peers for faster startup~
- [x] ~reduce code duplication between fluffy and nimbus verified proxy~
[...]

PR #3338, opened 2025-05-27
proxy: add blocks support

chirag-parmar:
This PR adds support for block based RPC endpoints by adding the following:

- [x] helper functions for downloading and verifying blocks and headers via JSON-RPC API under `rpc/blocks.nim`
- [x] helper functions to verify transaction trie roots and withdrawals root under `rpc/transactions.nim`
- [x] adds block based JSON RPC method handlers ([check this](https://github.com/status-im/nimbus-eth1/issues/3101))


nimbus_verified_proxy/header_store.nim
```diff
@@ -116,6 +116,18 @@ func finalized*(self: HeaderStore): Opt[Header] =
...
   self.finalizedHash
 
+func contains*(self: HeaderStore, hash: Hash32): bool =
+  if self.headers.contains(hash):
+    return true
+
+  return false
+
+func contains*(self: HeaderStore, number: base.BlockNumber): bool =
+  if self.hashes.contains(number):
```
tersec, line 126: Likewise here, can just return the value of `self.hashes.contains(number)`, no extra conditionals, etc

nimbus_verified_proxy/rpc/transactions.nim
```diff
@@ -0,0 +1,81 @@
...
+    if x.kind == tohTx:
+      convertedTxs.add toTransaction(x.tx)
+    else:
+      return err("cannot construct a transaction trie using only txhashes")
+
+  return ok(convertedTxs)
+
+proc checkTxHash*(txObj: TransactionObject, txHash: Hash32): bool =
+  let tx = toTransaction(txObj)
+  if tx.rlpHash != txHash:
```
tersec, line 65: This entire function can return the value of `toTransaction(txObj).rlpHash == txHash`, no extra conditionals

nimbus_verified_proxy/rpc/transactions.nim
```diff
@@ -0,0 +1,81 @@
...
+
+func vHashes(x: Opt[seq[Hash32]]): seq[VersionedHash] =
+  if x.isNone:
+    return
+  else:
+    x.get
+
+func authList(x: Opt[seq[Authorization]]): seq[Authorization] =
+  if x.isNone:
+    return
```
tersec, line 28: implicitly relies in `result`, better to explicitly return the empty sequence

nimbus_verified_proxy/rpc/transactions.nim
```diff
@@ -0,0 +1,81 @@
...
+import
+  stint,
+  results,
+  eth/common/eth_types_rlp,
+  eth/trie/[ordered_trie, trie_defs],
+  json_rpc/[rpcserver, rpcclient],
+  web3/[eth_api_types, eth_api],
+  ../../execution_chain/beacon/web3_eth_conv
+
+template calcWithdrawalsRoot*(withdrawals: openArray[Withdrawal]): Root =
```
tersec, line 17: does anything in this PR use this?

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -9,59 +9,252 @@
...
+    baseFeePerGas: blk.baseFeePerGas,
+    withdrawalsRoot: blk.withdrawalsRoot,
+    blobGasUsed: blk.blobGasUsed.u64,
+    excessBlobGas: blk.excessBlobGas.u64,
+    parentBeaconBlockRoot: blk.parentBeaconBlockRoot,
+    requestsHash: blk.requestsHash,
+  )
+
+proc walkBlocks(
+    self: VerifiedRpcProxy,
```
bhartnett, line 73: Minor issue: Just an FYI that using the term `self` is not very common in nim code. For example in this case you could use `vp` or `verifiedProxy`.

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -9,59 +9,252 @@
...
+    sourceNum: base.BlockNumber,
+    targetNum: base.BlockNumber,
+    sourceHash: Hash32,
+    targetHash: Hash32,
+): Future[bool] {.async: (raises: []).} =
+  var nextHash = sourceHash
+  info "starting block walk to verify", blockHash = targetHash
```
bhartnett, line 80:
```suggestion
  info "Starting block walk to verify requested block", blockHash = targetHash
```

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -9,59 +9,252 @@
...
+    if nextHeader.parentHash == targetHash:
+      return true
+
+    nextHash = nextHeader.parentHash
+
+  return false
+
+proc verifyHeader(
+    self: VerifiedRpcProxy, header: Header, hash: Hash32
+): Future[Result[bool, string]] {.async.} =
```
bhartnett, line 113: You can use `Result[void, string]` here. The boolean is redundant because Result.isOk() already indicates success.

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -9,59 +9,252 @@
...
-    proxy.headerStore.latest
-  of BlockNumber:
-    proxy.headerStore.get(base.BlockNumber(distinctBase(tag.blockNumber)))
-
-proc getHeaderByTagOrThrow*(
-    proxy: VerifiedRpcProxy, quantityTag: BlockTag
-): Header {.raises: [ValueError].} =
-  getHeaderByTag(proxy, quantityTag).valueOr:
-    raise newException(ValueError, "No block stored for given tag " & $quantityTag)
+    return ok(base.BlockNumber(distinctBase(blockTag.number)))
```
bhartnett, line 39: All the 'return' statements above can be removed so that all the code here becomes an expression which is safer. You could also use the early return pattern, for example:
```
  if blockTag.kind != bidAlias:
    return ok(base.BlockNumber(distinctBase(blockTag.number)))

  let tag = blockTag.alias.toLowerAscii()
  case tag
  of "latest":
    let hLatest = self.headerStore.latest().valueOr:
      return err("Couldn't get the latest block number from header store")
      
    ok(hLatest.number)
  else:
    err("No support for block tag " & $blockTag)
```

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -9,59 +9,252 @@
...
+proc resolveBlockTag*(
+    self: VerifiedRpcProxy, blockTag: BlockTag
+): Result[base.BlockNumber, string] =
   if blockTag.kind == bidAlias:
-    let tag = blockTag.alias.toLowerAscii
+    let tag = blockTag.alias.toLowerAscii()
     case tag
     of "latest":
-      return ok(QuantityTag(kind: LatestBlock))
+      let hLatest = self.headerStore.latest()
```
bhartnett, line 31: Can use valueOr here.

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -9,59 +9,252 @@
...
+  )
+
+  if not isLinked:
+    return err("the requested block is not part of the canonical chain")
+
+  return ok(true)
+
+proc verifyBlock(
+    self: VerifiedRpcProxy, blk: BlockObject, fullTransactions: bool
+): Future[Result[bool, string]] {.async.} =
```
bhartnett, line 133: Same here `Result[bool, string]` can be changed to `Result[void, string]`.

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -9,59 +9,252 @@
...
+  let header = convHeader(blk)
+
+  let status = await self.verifyHeader(header, blk.hash)
+
+  if status.isErr():
+    return err(status.error)
```
bhartnett, line 139: You should be able to use the `?` operator here (from the results library) to return the errer if `isErr` and unwrap the value otherwise. For example:
```
let status = ?(await self.verifyHeader(header, blk.hash))
```

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -9,59 +9,252 @@
...
+
+  # verify transactions
+  if fullTransactions:
+    let verified = verifyTransactions(header.transactionsRoot, blk.transactions).valueOr:
+      return err(error)
```
bhartnett, line 144: Can also use the `?` operator here and just discard the result. For example:
```
  if fullTransactions:
    discard ?verifyTransactions(header.transactionsRoot, blk.transactions)
```

nimbus_verified_proxy/rpc/transactions.nim
```diff
@@ -0,0 +1,81 @@
...
+  if x.isNone:
+    return
+  else:
+    x.get
+
+func authList(x: Opt[seq[Authorization]]): seq[Authorization] =
+  if x.isNone:
+    return
+  else:
+    x.get
```
bhartnett, line 30: I would suggest that the `vHashes` and `authList` functions are not really needed as all they do is unwrap an `Opt`.

bhartnett:
@chirag-parmar It would be good to get some basic unit tests on the code in the blocks.nim and transactions.nim files. I would aim for at least one test covering the success scenario of each of the public exported functions.

nimbus_verified_proxy/rpc/transactions.nim
```diff
@@ -0,0 +1,62 @@
...
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+import
+  stint,
+  results,
+  eth/common/eth_types_rlp,
+  eth/trie/[ordered_trie, trie_defs],
+  json_rpc/[rpcserver, rpcclient],
```
bhartnett, line 13: Do you need these RPC imports here?

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -9,59 +9,235 @@
...
+            return err(
+              "Couldn't get block " & $nextHash & " during the chain traversal: " & e.msg
+            )
+
+        trace "getting next block",
+          hash = nextHash,
+          number = blk.number,
+          remaining = distinctBase(blk.number) - targetNum
+
+        convHeader(blk)
```
kdeme: I don't think the block hash of the `nextHeader` gets verified here anywhere does it?
In case it gets requested through the JSON-RPC API I think it would be valuable to do that verification. Else a provider can let you back track a wrong chain and you would only error out after 1000 requests.

nimbus_verified_proxy/rpc/rpc_eth_api.nim
```diff
@@ -21,69 +21,124 @@ import
...
+  vp.proxy.rpc("eth_getUncleCountByBlockHash") do(blockHash: Hash32) -> Quantity:
+    let blk = (await vp.getBlock(blockHash, false)).valueOr:
+      raise newException(ValueError, error)
+
+    Quantity(blk.uncles.len())
+
+  vp.proxy.rpc("eth_getBlockTransactionCountByNumber") do(
+    blockTag: BlockTag
+  ) -> Quantity:
+    let blk = (await vp.getBlock(blockTag, true)).valueOr:
```
kdeme, line 102: Do you need the `fullTransactions` here on true?

nimbus_verified_proxy/rpc/rpc_eth_api.nim
```diff
@@ -21,69 +21,124 @@ import
...
+      raise newException(ValueError, error)
+
+    Quantity(blk.transactions.len)
+
+  vp.proxy.rpc("eth_getBlockTransactionCountByHash") do(blockHash: Hash32) -> Quantity:
+    let blk = (await vp.getBlock(blockHash, true)).valueOr:
```
kdeme, line 108: Do you need the `fullTransactions` here on true?

nimbus_verified_proxy/rpc/rpc_eth_api.nim
```diff
@@ -21,69 +21,124 @@ import
...
+    x.tx
+
   # TODO:
   # Following methods are forwarded directly to the web3 provider and therefore
   # are not validated in any way.
-  lcProxy.proxy.registerProxyMethod("net_version")
-  lcProxy.proxy.registerProxyMethod("eth_call")
-  lcProxy.proxy.registerProxyMethod("eth_sendRawTransaction")
-  lcProxy.proxy.registerProxyMethod("eth_getTransactionReceipt")
-  lcProxy.proxy.registerProxyMethod("eth_getBlockByNumber")
```
kdeme, line 85: I think the `eth_getBlockByNumber` did not actually get added? Even though the underlying functionality is there.

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -9,59 +9,235 @@
...
+
+    nextHash = nextHeader.parentHash
+
+  err("the requested block is not part of the canonical chain")
+
+proc verifyHeader(
+    vp: VerifiedRpcProxy, header: Header, hash: Hash32
+): Future[Result[void, string]] {.async.} =
+  # verify calculated hash with the requested hash
+  if header.rlpHash != hash:
```
kdeme: Deprecated, use `computeRlpHash` to make it clear that a computation happens here.

```suggestion
  if header.computeRlpHash != hash:
```

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -9,59 +9,235 @@
...
+    return err("hashed block header doesn't match with blk.hash(downloaded)")
+
+  let latestHeader = vp.headerStore.latest.valueOr:
+    return err("syncing")
+
+  # walk blocks backwards(time) from source to target
+  ?(
+    await vp.walkBlocks(
```
kdeme: Can we not immediately try to access the header of this block in our `HeaderStore`, and only if it is not there do the block walk?  I am assuming that anything in the `HeaderStore` is considered canonical (up until finalization), but perhaps that is not the case currently.

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -9,59 +9,235 @@
...
+
+proc verifyHeader(
+    vp: VerifiedRpcProxy, header: Header, hash: Hash32
+): Future[Result[void, string]] {.async.} =
+  # verify calculated hash with the requested hash
+  if header.rlpHash != hash:
+    return err("hashed block header doesn't match with blk.hash(downloaded)")
+
+  let latestHeader = vp.headerStore.latest.valueOr:
+    return err("syncing")
```
kdeme:
```suggestion
    return err("Cannot verify header due to still syncing")
```

Or something similar to make the error more clear?

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -9,59 +9,235 @@
...
+
+  ?(await vp.verifyHeader(header, blk.hash))
+
+  # verify transactions
+  if fullTransactions:
+    ?verifyTransactions(header.transactionsRoot, blk.transactions)
+
+  # verify withdrawals
+  if blk.withdrawals.isSome():
+    if blk.withdrawalsRoot.get() != orderedTrieRoot(blk.withdrawals.get()):
```
kdeme: If `withdrawalsRoot` is `None` here and `withdrawals` gets some invalid data  this could crash. Unlikely to occur as it would normally fail at `verifyHeader` for recent blocks, but for much older blocks it could be abused.

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -9,59 +9,235 @@
...
+      return err("withdrawals within the block do not yield the same withdrawals root")
+
```
kdeme, line 154: I think there is still a possibility now where provider gives a  `blk.withdrawals.isNone()`  but puts in the correct `withdrawalsRoot` so that the `verifyHeader` call does not fail.

And next we do not verify the withdrawals because it was set to `None`?

nimbus_verified_proxy/rpc/rpc_eth_api.nim
chirag-parmar: true. Changing it

nimbus_verified_proxy/rpc/rpc_eth_api.nim
chirag-parmar: because without asking for all the transactions I cannot be sure if every transaction hash returned(in the `false` case) is actually part of the block.

nimbus_verified_proxy/rpc/rpc_eth_api.nim
chirag-parmar: ditto

nimbus_verified_proxy/rpc/blocks.nim
chirag-parmar: good catch. changing it

nimbus_verified_proxy/rpc/transactions.nim
chirag-parmar: no. good catch. changing it

nimbus_verified_proxy/rpc/blocks.nim
chirag-parmar: good catch. changing this

nimbus_verified_proxy/rpc/blocks.nim
chirag-parmar: Good catch. I changed/inverted the logic. It now checks if the `withdrawalsRoot` is specified. if yes it checks the root. If not, it verifies that the withdrawals field is empty.

chirag-parmar:
@bhartnett added some basic tests

bhartnett:
> @bhartnett added some basic tests

I remember we discussed the idea of mocking out the RPC backend so that you can test more of the logic. Looks like that hasn't been completed yet.

chirag-parmar:
> > @bhartnett added some basic tests
> 
> I remember we discussed the idea of mocking out the RPC backend so that you can test more of the logic. Looks like that hasn't been completed yet.

I though I would make a seperate PR adding all the helios tests with the mock transports

PR #3342, opened 2025-05-29
Portal client: Add accept codes metadata to neighborhood gossip and PutContent JSON-RPC endpoints

bhartnett:


bhartnett:
After these changes the portal hive tests are still passing, I guess because the additional/new fields are ignored. So I don't see any reason why we can't preemptively merge this change considering that the spec change will also be merged soon.

bhartnett:
I didn't add the accept metadata to the random gossip JSON-RPC endpoint response because it doesn't currently return the PutContentResult type. 

@kdeme If you would like me to add the accept codes to the `portal_beaconRandomGossip` endpoint, please let me know. We could probably return the same `PutContentResult` type for this endpoint as well.

portal/network/wire/portal_protocol.nim
```diff
@@ -191,6 +191,15 @@ type
     of Database:
       contentKeys: ContentKeysList
 
+  OfferBatchMetadata* = object
```
kdeme, line 194: Perhaps we could go with one type here that immediatly fits also the json-rpc request.
And use `AcceptMetadata.useDefaultSerializationIn JrpcConv` on that one.

kdeme (commented):
> @kdeme If you would like me to add the accept codes to the `portal_beaconRandomGossip` endpoint, please let me know. We could probably return the same `PutContentResult` type for this endpoint as well.

The beacon version for put content should actually do the random gossip for now. As it should use the default gossip mechanism for that subnetwork. So this is wrongly implemented still.

bhartnett:
> > @kdeme If you would like me to add the accept codes to the `portal_beaconRandomGossip` endpoint, please let me know. We could probably return the same `PutContentResult` type for this endpoint as well.
> 
> The beacon version for put content should actually do the random gossip for now. As it should use the default gossip mechanism for that subnetwork. So this is wrongly implemented still. And the random method should be removed after.

I see, in that case I'll leave it as is and in another PR I'll remove the `portal_beaconRandomGossip` endpoint and update the `portal_beaconPutContent` endpoint so that it uses the randomGossip and returns the acceptMetadata.

portal/network/wire/portal_protocol.nim
bhartnett: I intentionally implemented it this way for the following reasons:
- A different type is used internally vs in the RPC layer to provide some separation so that JSON-RPC types are not used internally.
- The naming of the peerCount field isn't very clear and I like having the successCount field internally which is more descriptive.
- The AcceptMetadata type doesn't contain the peerCount so I would need to return another type from the `offerBatchGetMetadata` function, either a tuple or the PutContentResult object which doesn't look like a good solution. I wouldn't want to put the JSON-RPC `PutContentResult` type inside portal protocol and using a tuple is less clean in my opinion.

My preference would be to leave it as it is but I'm open to changing it if you still prefer an alternative implementation.

kdeme: Yes, I'm a bit divided on this. Originally I also preferred to have the JSON-RPC layer types separated from the p2p layer types.

Now since that the JSON conversion no longer occurs implicitly but we have to add explicitly `useDefaultSerializationIn JrpcConv`, I believe it is less of an issue, and perhaps preferred.

Of course, if the types are not exactly mappable, it makes little sense. 

I do believe that the json-rpc specific code should not live in the p2p code, and because of that never like our current implementation of the RFC trace call, but that is an issue outside the scope of this PR.

bhartnett:
@kdeme Do you think you might still want this changed merged so that the accept metadata can be used to improve the history bridge?

kdeme:
> @kdeme Do you think you might still want this changed merged so that the accept metadata can be used to improve the history bridge?

Yeah, sure :+1:

PR #3346, opened 2025-05-29
Move PooledTransaction from nim-eth to execution client

jangko:


advaita-saha (approved):

execution_chain/core/eip4844.nim
```diff
@@ -14,6 +14,7 @@ import
   results,
   stint,
   ./eip7691,
+  ./pooled_txs,
```
arnetheduck, line 17: what's the dividing line between keeping things in nim-eth and putting in eth1?

jangko: If it's not being used anywhere else but by nimbus-eth1

PR #3351, opened 2025-05-30
Portal bridge: Use startEra and endEra parameters in history bridge backfill audit

bhartnett:
In the Portal history bridge backfill mode we use the `startEra` and `endEra` parameters when `audit=false` but when `audit` is set to true then these parameters are ignored. I think it would be more useful to bound the range of randomly selected block numbers to be within the era parameters. The defaults are still the same so that if no `startEra` and `endEra` parameters are supplied then we should still sample from the same block range as before.

### DIFF
--- a/portal/bridge/history/portal_history_bridge.nim
+++ b/portal/bridge/history/portal_history_bridge.nim
@@ -357,7 +357,6 @@ proc runBackfillLoop(
 proc runBackfillLoopAuditMode(
     bridge: PortalHistoryBridge, era1Dir: string, startEra: uint64, endEra: uint64
 ) {.async: (raises: [CancelledError]).} =
-
   let
     rng = newRng()
     db = Era1DB.new(era1Dir, "mainnet", loadAccumulator())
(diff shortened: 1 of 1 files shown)


portal/bridge/history/portal_history_bridge.nim
```diff
@@ -355,17 +355,22 @@ proc runBackfillLoop(
         continue
 
 proc runBackfillLoopAuditMode(
-    bridge: PortalHistoryBridge, era1Dir: string
+    bridge: PortalHistoryBridge, era1Dir: string, startEra: uint64, endEra: uint64
 ) {.async: (raises: [CancelledError]).} =
+  const eraSize = 8192
```
kdeme: You can use `EPOCH_SIZE` instead

bhartnett:
> Originally the idea was for backfill and audit mode to be separate modes. As in, you can run backfill for any range that seems to be missing on the network, but audit is just generally done over the whole range.
> 
> I can see however how this is not represented well in the current cli help documentation, so we can go with this version,

Yeah that makes sense. Well with this change you can still run the backfill over the whole range as before as long as you don't supply the era start and end parameters. 

If the user does supply the start and end parameters I believe the expectation would be that the audit should be done over that range (unless documented otherwise).

PR #3352, opened 2025-05-30
Portal client: Create metric to track successful and failed offer validations for all sub-networks

bhartnett:


portal/network/state/state_network.nim
```diff
@@ -26,11 +26,6 @@ export results, state_content, hashes
 logScope:
   topics = "portal_state"
 
-declareCounter state_network_offers_success,
```
kdeme, line 29: I had not spotted before that these were prefixed with `state`. Yeah, they probably should not be prefixed as such, we already have the protocol_id label that can be used to filter on and create separate sections on the grafana dashboard.

portal/network/wire/portal_protocol.nim
```diff
@@ -126,6 +126,11 @@ declareHistogram portal_offer_log_distance,
   labels = ["protocol_id"],
   buckets = distanceBuckets
 
+declarePublicCounter portal_subnetwork_offer_validation_successful,
+  "Portal sub-network offers successfully validated", labels = ["protocol_id"]
+declarePublicCounter portal_subnetwork_offer_validation_failed,
```
kdeme: I think we can leave the `subnetwork` part? The label already implies this.

portal/network/state/state_network.nim
bhartnett: Yeah that's right. I originally did it this way because this metric was incremented/used in the state network, not in portal protocol but it turns out that we can use a public metric `declarePublicCounter` with a protocolId and avoid having to create multiple metrics.

portal/network/wire/portal_protocol.nim
bhartnett: Sure, I'll update.

PR #3376, opened 2025-06-10
EIP-7934 : 10MB limit to block size

advaita-saha:


jangko:
The same validation if applied to txpool can be tricky.

execution_chain/core/tx_pool.nim
```diff
@@ -159,9 +159,13 @@ proc assembleBlock*(
     blobsBundle = BlobsBundle(
       wrapperVersion: getWrapperVersion(com, blk.header.timestamp)
     )
+    currentRlpSize = rlp.getEncodedLength(blk.header) + rlp.getEncodedLength(blk.withdrawals)
```
jangko: https://github.com/status-im/nimbus-eth1/blob/6ce4568e628ea33b9e3d55e035e5e68df7417a39/execution_chain/core/tx_pool.nim#L178

withdrawals are fork dependant.

execution_chain/core/tx_pool.nim
```diff
@@ -159,9 +159,13 @@ proc assembleBlock*(
...
 
   for item in pst.packedTxs:
     let tx = item.pooledTx
+    if currentRlpSize > MAX_RLP_BLOCK_SIZE:
+      break
+    currentRlpSize = currentRlpSize + rlp.getEncodedLength(tx.tx)
```
jangko, line 172: Calculating the rlp length like this might works to some extent. But considering how rlp algorithm works, there are and there will be edge cases where this calculation will produce invalid result. To produce a reliable result for all cases, we need a specially crafted `getEncodedLength` or we make a compromise to never exceeds the dangerous limit.

And that dangerous limit is `MAX_RLP_BLOCK_SIZE - 5`. The `5` represent the maximum number of bytes might be produced by the rlp for the length prefix which is 3 bytes for 10 MiB, and additional 2 bytes to compensate for missing rlp list marker when we skip the body. (This is my rough calculation, I may missed something), but you got the point.

And also a break from that loop must be executed _before_ `currentRlpSize` exceeds `MAX_RLP_BLOCK_SIZE - X`.

execution_chain/core/tx_pool.nim
```diff
@@ -159,9 +159,13 @@ proc assembleBlock*(
     blobsBundle = BlobsBundle(
       wrapperVersion: getWrapperVersion(com, blk.header.timestamp)
     )
+    currentRlpSize = rlp.getEncodedLength(blk.header) + rlp.getEncodedLength(blk.withdrawals)
 
   for item in pst.packedTxs:
     let tx = item.pooledTx
+    if currentRlpSize > MAX_RLP_BLOCK_SIZE - 10:
```
advaita-saha:
```suggestion
    if currentRlpSize > MAX_RLP_BLOCK_SIZE - 7:
```

PR #3387, opened 2025-06-12
Portal: default disable the Portal state network

kdeme:
Portal state network will no longer receive support from EF for the foreseeable future. 

It is currently in a development state that is not sufficiently finished to be of any use other than testing the ongoing R&D of itself. That means it is of no use for an end user, thus disabling this by default.


bhartnett (approved):

bhartnett:
We might need to update our website docs as well to indicate that state network must be enabled to run the examples such as eth_call etc.

PR #3399, opened 2025-06-13
add eip-7883 changes

advaita-saha:
https://github.com/ethereum/EIPs/commit/5cdd75157d78522bde8dc58977717c14cd7f2789#diff-deb61419789ec2f1fd580a01375e9c1ab84989138683faf6ab2306283219bcaeL32-R32

### DIFF
--- a/execution_chain/evm/precompiles.nim
+++ b/execution_chain/evm/precompiles.nim
@@ -205,7 +205,7 @@ func modExpFee(c: Computation,
     result = result * result
 
   func mulComplexityEIP7883(maxLen: UInt256): UInt256 =
-    result = 16'u256
+    result = 16.u256
     if maxLen > 32.u256:
       result = maxLen + 7
       result = result div 8
--- a/execution_chain/evm/precompiles.nim
+++ b/execution_chain/evm/precompiles.nim
@@ -205,11 +205,11 @@ func modExpFee(c: Computation,
     result = result * result
 
   func mulComplexityEIP7883(maxLen: UInt256): UInt256 =
-    # gas = ceil(x div 8) ^ 2
-    result = maxLen + 7
-    result = result div 8
-    result = result * result
+    result = 16'u256
     if maxLen > 32.u256:
+      result = maxLen + 7
+      result = result div 8
+      result = result * result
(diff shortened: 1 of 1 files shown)


execution_chain/evm/precompiles.nim
```diff
@@ -205,11 +205,11 @@ func modExpFee(c: Computation,
     result = result * result
 
   func mulComplexityEIP7883(maxLen: UInt256): UInt256 =
-    # gas = ceil(x div 8) ^ 2
-    result = maxLen + 7
-    result = result div 8
-    result = result * result
+    result = 16'u256
```
jangko: looks like `16'u256` is throwing something. Probably need to use `16.u256`.

advaita-saha: yes already did that in the interop branch, didn't get the time to do it here as well

advaita-saha:
```suggestion
    result = 16.u256
```

execution_chain/evm/precompiles.nim
```diff
@@ -205,11 +205,11 @@ func modExpFee(c: Computation,
...
   func mulComplexityEIP7883(maxLen: UInt256): UInt256 =
-    # gas = ceil(x div 8) ^ 2
-    result = maxLen + 7
-    result = result div 8
-    result = result * result
+    result = 16.u256
     if maxLen > 32.u256:
+      result = maxLen + 7
+      result = result div 8
+      result = result * result
```
tersec, line 212: Is there a reason for this `result = result ...` accumulator style? If it were managing to be truly in-place with these large operands, that might be interesting, but it's using formally full calculation then assignment style.

jangko: iirc stint doesn't have `*=` operator.

tersec: Sure, but then why not use a more usual expression style rather than accumulate through `result`? It's not really saving any CPU or memory that I can tell.

jangko: no particular reason, feel free to rewrite it using expression style.

PR #3405, opened 2025-06-16
Fix missing return value in addEthHandlerCapability

jangko:


advaita-saha:
How was it working without returning it ?

jangko:
It works because it starts it's own async loop. The return value needed to terminated the async loop gracefully. But at the top there is also a timer to forcefully terminate the process in case of various reasons involving numerous async loop in nimbus EL.

PR #3406, opened 2025-06-16
Rlpx handshake using the latest version of a protocol capabilities

### LINKED ISSUE #3396: Nimbus-eth1 fails to connect to bootstrap nodes due to encryption/decryption errors
Nimbus-eth1 fails to connect to most or all discovery bootstrap nodes at startup. The logs report encryption/decryption handshake failures during the discv5 handshake process, resulting in no or very few peers being discovered.

- Fails to connect to bootstrap nodes
- Logs contain repeated failed to decrypt handshake packet messages
- Peer discovery stalls or is significantly delayed

jangko:
Turn out the rlpx handshake doesn't care about capabilities version or assume there is only one version for each capabilities.
This PR fix that. And now can connect to all bootnodes from fusaka devnet-1.
The algorithm will chose the latest version of the same capability.

fix #3396

### DIFF
--- a/execution_chain/networking/rlpx.nim
+++ b/execution_chain/networking/rlpx.nim
@@ -1002,6 +1002,10 @@ proc selectCapsByLatestVersion(peer: Peer): seq[ProtocolInfo] =
   # Avoid using multiple capability handshake when connecting to a peer.
   # Use only the latest capability version. e.g. choose eth/69 over eth/68.
   # But other capabilities with different name is okay. e.g. snap/1
+
+  # From the spec:
+  # https://github.com/ethereum/devp2p/blob/bc76b9809a30e6dc5c8dcda996273f0f9bcf7108/rlpx.md#message-id-based-multiplexing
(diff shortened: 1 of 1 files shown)


execution_chain/networking/rlpx.nim
```diff
@@ -998,10 +998,29 @@ proc removePeer(network: EthereumNode, peer: Peer) =
           if observer.protocols.len == 0 or peer.supports(observer.protocols):
             observer.onPeerDisconnected(peer)
 
+proc selectCapsByLatestVersion(peer: Peer): seq[ProtocolInfo] =
+  # Avoid using multiple capability handshake when connecting to a peer.
+  # Use only the latest capability version. e.g. choose eth/69 over eth/68.
+  # But other capabilities with different name is okay. e.g. snap/1
```
tersec, line 1004: Presumably future-proofing here? I'm not sure anything except `eth` is supported in a useful way. But sure, it's good for this not to break with other names.

PR #3414, opened 2025-06-18
portal: pass list item instead of list

chirag-parmar:


portal/tools/eth_data_exporter/parser.nim
```diff
@@ -130,7 +130,7 @@ proc fromJson*(n: JsonNode, name: string, x: var seq[Authorization]) =
   let node = n[name]
   for v in node:
     x = newSeqOfCap[Authorization](node.len)
-    x.add parseAuth(node)
+    x.add parseAuth(v)
```
bhartnett, line 133: This code still doesn't look right. Why does the seq get recreated on each iteration of the loop? 
`x = newSeqOfCap[Authorization](node.len)`

I didn't write this code so perhaps @kdeme has a better idea?

kdeme: I don't know either, I did not write it. It got copied over from premix before that got removed

bhartnett (approved):

Issue #3403, opened 2025-06-13
engineAPI timeout issue from `lighthouse`


advaita-saha:
```bash
Jun 13 19:50:55.225 ERROR Error during execution engine upcheck         error: HttpClient(url: http://snooper-engine:8561/, kind: timeout, detail: operation timed out)
Jun 13 19:50:57.639 WARN  Execution engine call failed                  error: HttpClient(url: http://snooper-engine:8561/, kind: timeout, detail: operation timed out)
Jun 13 19:50:57.645 ERROR Error fetching block for peer                 block_root: 0x6f218219aeee372d6ea12811070a7a90f34ba29738836b59ea2761762b7471b4, error: BlockStreamerError(BlocksByRangeFailure(EngineError(Api { error: HttpClient(url: http://snooper-engine:8561/, kind: timeout, detail: operation timed out) })))
Jun 13 19:50:58.641 ERROR Error during execution engine upcheck         error: HttpClient(url: http://snooper-engine:8561/, kind: timeout, detail: operation timed out)
Jun 13 19:50:59.387 WARN  Execution engine call failed                  error: HttpClient(url: http://snooper-engine:8561/, kind: timeout, detail: operation timed out)
Jun 13 19:51:00.388 WARN  Execution engine call failed                  error: HttpClient(url: http://snooper-engine:8561/, kind: timeout, detail: operation timed out)
Jun 13 19:51:00.388 ERROR Error during execution engine upcheck         error: HttpClient(url: http://snooper-engine:8561/, kind: timeout, detail: operation timed out)
Jun 13 19:51:00.939 INFO  New block received                            slot: 1455, root: 0xca92e3e094e288c434f9a163587a11d28e04868d19b21e230e1482859d1114e6
Jun 13 19:51:01.389 ERROR Error during execution engine upcheck         error: HttpClient(url: http://snooper-engine:8561/, kind: timeout, detail: operation timed out)
Jun 13 19:51:01.940 WARN  Execution engine call failed                  error: HttpClient(url: http://snooper-engine:8561/, kind: timeout, detail: operation timed out), service: "fetch_engine_blobs", block_root: "0xca92e3e094e288c434f9a163587a11d28e04868d19b21e230e1482859d1114e6"
[...]

PR #3417, opened 2025-06-20
Increase rpc http server chunk size to 1 MBytes

jangko:
This will reduce engine api traffic. But don't know if this can help with the overall slowness mentioned in #3403

ping @advaita-saha


tersec:
Given that this is only one buffer in a set of buffers, would one be even able to tell with this PR alone whether it necessarily is enough?

advaita-saha:
Makes the wait a little less, but the full problem is not solved

PR #3427, opened 2025-06-25
Portal: Implementation of new history network

kdeme:
WIP, but starting to implement https://ethresear.ch/t/using-portal-network-protocol-for-history-expiry/22658

TODO (not necessarily all in this PR):
- [x] get receipts/body API
- [x] json-rpc API on top
- [ ] Header Oracle or similar, allow for using (currently not existing) get header json RPC API from EL or just passing headers along when natively integrated
- [x] add network tests
- [x] probably some test vectors generation
- [ ] adapt EL integration + bridging in data from EL
- [x] ~~Might require ContentDb change due to same content id for BlockBody and Receipts. Unless specs change~~ Specs changed so that `ContentDb` change is no longer required. See https://github.com/status-im/nimbus-eth1/pull/3427/commits/2b5cebcee778d7343b71c8479281f96938d44d1d


kdeme:
Specs PR now: https://github.com/ethereum/portal-network-specs/pull/407

PR #3432, opened 2025-06-28
Transaction hashes broadcast handler: use inner tx hash + PooledTransaction size

jangko:
Chasing state root bug, but uncover another bug in the transaction hashes broadcast handler.

### DIFF
--- a/execution_chain/core/executor/process_block.nim
+++ b/execution_chain/core/executor/process_block.nim
@@ -184,6 +184,7 @@ proc procBlkPreamble(
 proc procBlkEpilogue(
     vmState: BaseVMState,
     blk: Block,
+    prevStateRoot: Hash32,
     skipValidation: bool,
     skipReceipts: bool,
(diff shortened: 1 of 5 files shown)


execution_chain/core/executor/process_block.nim
```diff
@@ -286,13 +288,16 @@ proc processBlock*(
   # cases - since each block is bounded in the amount of memory needed, we can
   # run collection once per block instead.
   deferGc:
+    # When there is state root mismatch, we want to show where it arrived from
+    # using actual value stored in DB/txFrame beside showing parent.stateRoot.
+    let prevStateRoot = vmState.ledger.getStateRoot()
```
arnetheduck, line 293: this causes a state root computation which slows down `import` by several times

jangko: ah okay, will revert that part.

PR #3441, opened 2025-07-02
Transform FC module internals into DAG

jangko:
No more `BranchRef`, only `BlockRef`.
Simpler algorithm in many places.
Pruning leaves become more uniform among branches.

### DIFF
--- a/execution_chain/core/chain/forked_chain.nim
+++ b/execution_chain/core/chain/forked_chain.nim
@@ -119,9 +119,14 @@ func findFinalizedPos(
     if fin.number > head.number:
       return err("Invalid finalizedHash: block is newer than head block")
 
-    loopIt(head):
-      if it == fin:
-        return ok(fin)
+    # There is no point traversing the DAG if there is only one branch.
+    # Just return the node.
+    if c.heads.len > 1:
+      loopIt(head):
+        if it == fin:
+          return ok(fin)
+    else:
+      return ok(fin)
 
   err("Invalid finalizedHash: block not in argument head ancestor lineage")
 
@@ -892,15 +897,14 @@ func payloadBodyV1FromBaseTo*(c: ForkedChainRef,
                               last: BlockNumber,
                               list: var seq[Opt[ExecutionPayloadBodyV1]]) =
   var
-    blocks = newSeqOfCap[BlockRef](c.hashToBlock.len)
+    blocks = newSeqOfCap[BlockRef](last-c.base.number+1)
 
   loopIt(c.latest):
-    blocks.add(it)
+    if it.number <= last:
+      blocks.add(it)
 
   for i in countdown(blocks.len-1, 0):
     let y = blocks[i]
-    if y.number > last:
-      return
     list.add Opt.some(toPayloadBody(y.blk))
 
 func equalOrAncestorOf*(c: ForkedChainRef, blockHash: Hash32, headHash: Hash32): bool =


execution_chain/core/chain/forked_chain.nim
```diff
@@ -349,33 +232,66 @@ proc updateFinalized(c: ForkedChainRef, finalized: BlockPos) =
...
-        return true
-      branch = branch.parent
-
-  let finalizedNumber = finalized.number
-  var i = 0
-  while i < c.branches.len:
-    let branch = c.branches[i]
+  func reachable(head, fin: BlockRef): bool =
+    loopIt(head):
+      if it.colored:
```
arnetheduck, line 237: this algorithm can be simplified based on the observation that block numbers are monotonically increasing - basically, when we've identified the finalized block, we walk the parent chain and compare block numbers - if the block number of the examined block is _lower_ than that of the finalized block but is not the finalized block, we know that the branch is stale because the block where it meets the chain must be older than finalized.


```nim
proc isRelevantBranch(head, finalized): bool =
  var blk = head
  while blk != finalized:
    if blk.number < finalized.number:
      return false
    blk = blk.parent
  true
```

arnetheduck: notably, we only need to run isRelevantBranch for branches that are not the canonical head right now

jangko: just tested, it does not work for branches with head lower than the new finalized.

e.g. old fin = A3. stale chain B5-B8.  chain A continue from A4 to A10. new fin = A9.
The simplified algorithm basically not aware of B.

arnetheduck: In your example, the starting point is the list of heads before fin was updated to A9, ie `[A10, B8]` .. then you do:
```nim
var toprune: seq[blockref]
for head in fc.heads:
  if head == canonical:
   continue
  if not isReachable(head, newfin):
    toprune.add(head)
```

b will hit the `if blk.number < fin.number` condition on first iteration and will be removed

jangko: Oops, sorry. I mean not about the head. But how to stop the pruning? When we move backward from head of a stale branch, there is no marker of a junction/forking point.

arnetheduck: you first mark the canonical pruned blocks as finalized - in eth2, we do this by setting parent to nil for every block between old fin and new fin (ie the finalized block history is not kept in memory) - then, when you keep iterating over the parent you will eventually find a blockref with parent == nil and that's your (old) junction

arnetheduck: so basically, first set the range [oldfin, newfin] as finalized and canonical by setting the parent to nil .. then run the above loop to find branches that are no longer viable - then prune those branches, but this time also remove the chain data (headers etc)

jangko: int eth1 we still need parent to calculate base movement. So there is no significant difference with color algorithm.

arnetheduck: well, the base movement computation happens first - you establish the new base, then you clear the parents then you prune the branches - the difference is that you don't need a color field and you don't need to "uncolor" things after you're done - that's the simplification.

arnetheduck: btw, clearing the parent has the added benefit that the GC can release the structure earlier as well

jangko: not all finalized blocks are written to database,

arnetheduck: aah, right - hm - I guess one could use a `finalized` bool to avoid the "uncoloring" in that case - while still a simplification, maybe that's not as attractive.

PR #3448, opened 2025-07-04
Stateless: Make collection of witness keys in ledger configurable at runtime

bhartnett:
This PR enables the collection of witness keys in the ledger at runtime via the hidden statelessProviderEnabled CLI flag. These witness keys will be used to build the execution witnesses which will be stored in the database for each block when statelessProviderEnabled is true.

### DIFF
--- a/execution_chain/common/common.nim
+++ b/execution_chain/common/common.nim
@@ -90,6 +90,11 @@ type
     taskpool*: Taskpool
       ## Shared task pool for offloading computation to other threads
 
+    statelessProviderEnabled*: bool
+      ## Enable the stateless provider. This turns on the features required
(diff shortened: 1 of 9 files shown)


jangko (approved):

execution_chain/common/common.nim
```diff
@@ -90,6 +90,11 @@ type
     taskpool*: Taskpool
       ## Shared task pool for offloading computation to other threads
 
+    statelessProviderEnabled*: bool
+      ## Enable the stateless provider. This turns on the features required
+      ## by stateless clients such as generation and stored of block witnesses
```
bhartnett:
```suggestion
      ## by stateless clients such as generation and storage of block witnesses
```

PR #3455, opened 2025-07-07
Schedule `updateBase` to asynchronous worker.

### LINKED ISSUE #3449: Simplify `updateBase`
`updateBase` currently receives base update jobs that may involve hundreds of blocks - in order to act cooperatively, it splits up the job into chunks of 4 blocks: https://github.com/status-im/nimbus-eth1/blob/65af60139b69dee4b87eef12e079cebea02fb88a/execution_chain/core/chain/forked_chain.nim#L315

Instead of chunking inside `updateBase`, we can perform the chunking outside, thus making `updateBase` itself synchronous and easier to reason about. Further, we can perform the base update lazily. It would work like this:

* A single block or an FCU is processed by the queue worker
* finalization is run eagerly, ie the stale head cleanup runs removing any histories that are not relevant
* `updateBase` is _scheduled_ but not run
* queue worker yields
* `updateBase` runs with a maximum step of 4 (achieved by changing the default [persist batch size](https://github.com/status-im/nimbus-eth1/blob/65af60139b69dee4b87eef12e079cebea02fb88a/execution_chain/config.nim#L345) to 4.
* queue worker yields
* repeat

In the above flow, if finalized moves by 100 blocks, it will take 25 iterations of the loop to update the base to its new value interleaving the work with block processing - as long as the persist batch size is > 1, base will catch up (we should probably clamp the value to at least 2) - alternatively, `updateBase` can be run in a loop by the worker until it's done - either strategy is fine though there's some value in interleaving blocks and base updates.

jangko:
`updateBase` become synchronous and the scheduler will interleave `updateBase` with `importBlock` and `forkChoice`.

The scheduler will move the base at fixed size `PersistBatchSize`.

fix #3449

### DIFF
--- a/execution_chain/core/chain/forked_chain.nim
+++ b/execution_chain/core/chain/forked_chain.nim
@@ -335,32 +335,38 @@ proc processUpdateBase(c: ForkedChainRef) {.async: (raises: [CancelledError]).} 
     let base = c.baseQueue.popFirst()
     c.persistedCount += c.updateBase(base)
 
+  const
+    minLogInterval = 5
+  
   if c.baseQueue.len == 0:
-    # Log only if more than one block persisted
-    # This is to avoid log spamming, during normal operation
-    # of the client following the chain
-    # When multiple blocks are persisted together, it's mainly
-    # during `beacon sync` or `nrpc sync`
-    if c.persistedCount > 1:
-      notice "Finalized blocks persisted",
-        nBlocks = c.persistedCount,
-        base = c.base.number,
-        baseHash = c.base.hash.short,
-        pendingFCU = c.pendingFCU.short,
-        resolvedFin= c.latestFinalizedBlockNumber
-    else:
-      debug "Finalized blocks persisted",
-        nBlocks = c.persistedCount,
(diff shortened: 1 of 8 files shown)


execution_chain/core/chain/forked_chain.nim
```diff
@@ -164,7 +165,7 @@ func calculateNewBase(
   # with large enough step to accomodate for bulk
   # state root verification/bulk persist.
   let distance = target - c.base.number
-  if distance < c.persistBatchSize:
+  if distance < c.persistBatchQueue:
```
arnetheduck: we have to perform state root verification at every updateBase effectively (because we have to invalidate the mpt), so there's no point making this larger than `PersistBatchSize` - ie it's probably better to retain `--persist-batch-size` as the option and remove `PersistBatchQueue`

jangko: https://github.com/status-im/nimbus-eth1/pull/3455/commits/39ef28a3179a8bf2edf5b1a5b2132a2ddc542edf

PR #3457, opened 2025-07-07
proxy: txs, rxs and logs

chirag-parmar:
This PR adds support for block based RPC endpoints by adding the following:

- [x] helper functions to get and verify receipts `rpc/receipts.nim`
- [x] adds transaction, receipts and logs based JSON RPC method handlers ([check this](https://github.com/status-im/nimbus-eth1/issues/3101))
- [x] adds test cases that do a basic sanity check for the API methods implemented

https://github.com/status-im/nimbus-eth1/issues/3101

### DIFF
--- a/execution_chain/tracer.nim
+++ b/execution_chain/tracer.nim
@@ -60,8 +60,7 @@ proc init(
   T.init(com, header.stateRoot)
 
 # -------------------
-
-proc `%`(x: addresses.Address|Bytes32|Bytes256|Hash32): JsonNode =
+proc `%`(x: addresses.Address | Bytes32 | Bytes256 | Hash32): JsonNode =
   result = %toHex(x)
 
 proc toJson(receipt: Receipt): JsonNode =


bhartnett:
It would be good to see some tests added to this PR.

nimbus_verified_proxy/rpc/rpc_eth_api.nim
```diff
@@ -196,13 +198,54 @@ proc installEthApiHandlers*(vp: VerifiedRpcProxy) =
...
+
+  vp.proxy.rpc("eth_getLogs") do(filterOptions: FilterOptions) -> seq[LogObject]:
+    (await vp.getLogs(filterOptions)).valueOr:
+      raise newException(ValueError, error)
+
   # TODO:
   # Following methods are forwarded directly to the web3 provider and therefore
   # are not validated in any way.
   vp.proxy.registerProxyMethod("net_version")
-  vp.proxy.registerProxyMethod("eth_call")
```
bhartnett: @chirag-parmar This removal will enable our implementation of `eth_call` as I mentioned in the meeting.

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,133 @@
...
+  web3/[eth_api_types, eth_api],
+  ../../execution_chain/beacon/web3_eth_conv,
+  ../types,
+  ./blocks
+
+template toLog(lg: LogObject): Log =
+  Log(address: lg.address, topics: lg.topics, data: lg.data)
+
+func toLogs(logs: openArray[LogObject]): seq[Log] =
+  result = newSeqOfCap[Log](logs.len)
```
bhartnett: Usage of `result` should be avoided: https://status-im.github.io/nim-style-guide/language.result.html

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,133 @@
...
+  for lg in logs:
+    result.add(toLog(lg))
+
+func toReceipt(rec: ReceiptObject): Receipt =
+  let isHash = if rec.status.isSome: false else: true
```
bhartnett:
```suggestion
  let isHash = not rec.status.isSome()
```

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,133 @@
...
+
+  var status = false
+  if rec.status.isSome:
+    if rec.status.get() == 1.Quantity:
+      status = true
```
bhartnett: The above 4 lines can be replaced with this:
```suggestion
  let status = rec.status.isSome() and rec.status.get() == 1.Quantity:
```

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,133 @@
...
+        return err(e.msg)
+
+  if rxs.isSome():
+    if orderedTrieRoot(toReceipts(rxs.get())) != header.receiptsRoot:
+      return
+        err("downloaded receipts do not evaluate to the receipts root of the block")
+  else:
+    return err("error downloading the receipts")
+
+  return ok(rxs.get())
```
bhartnett: There appears to be a fair bit of duplicated code in the above two `getReceipts` functions. You could probably put the common code into a shared function.

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,133 @@
...
+          rx = rxs[distinctBase(txIdx)]
+
+        # only add verified logs
+        for rxLog in rx.logs:
+          if rxLog.address == lg.address and rxLog.data == lg.data and
+              rxLog.topics == lg.topics:
+            res.add(lg)
+
+  if res.len == 0:
+    return err("no logs could be verified")
```
bhartnett: Should this actually be an error? What if there are no logs returned for the given `eth_getLogs` query? Normally an empty list would be a valid response.

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,133 @@
...
+    isHash: isHash,
+    status: status,
+    cumulativeGasUsed: rec.cumulativeGasUsed.GasInt,
+    logs: toLogs(rec.logs),
+    logsBloom: rec.logsBloom,
+    receiptType: rec.`type`.get(0.Web3Quantity).ReceiptType,
+  )
+
+func toReceipts(recs: openArray[ReceiptObject]): seq[Receipt] =
+  for r in recs:
```
tersec:
```nim
  recs.mapIt(it.toReceipt)
```

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,133 @@
...
+proc getLogs*(
+    vp: VerifiedRpcProxy, filterOptions: FilterOptions
+): Future[Result[seq[LogObject], string]] {.async.} =
+  let logObjs =
+    try:
+      await vp.rpcClient.eth_getLogs(filterOptions)
+    except CatchableError as e:
+      return err(e.msg)
+
+  var res = newSeq[LogObject]()
```
tersec: This is probably more clearly written as:
```nim
  var res: seq[LogObject]
```

If the expected required capacity is known ahead of time, this can be improved, but just `newSeq()` without some length or capacity usually obfuscates more than provides utility.

nimbus_verified_proxy/rpc/receipts.nim
tersec: In particular, `mapIt` already [implements this optimization](https://github.com/nim-lang/Nim/blob/911a65198475e53fe6f874564223855174a8df8f/lib/pure/collections/sequtils.nim#L1080-L1092).

So can replace the entire thing with:
```nim
func toLogs(logs: openArray[LogObject]): seq[Log] =
  logs.mapIt(it.toLog)
```
also no explicit `return` needed.

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,133 @@
+# nimbus_verified_proxy
+# Copyright (c) 2025 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+import
+  stint,
```
tersec: How does it use `stint`?

nimbus_verified_proxy/rpc/receipts.nim
chirag-parmar: This would just define the type right? it wouldn't initialise the memory would it? and I was under the impression that we use `newSeqOfCap` when we know the capacity ahead of time and `newSeq` when we don't know the capacity ahead of time. Do sequences not require any kind of initialization?

nimbus_verified_proxy/rpc/receipts.nim
chirag-parmar: valid point. added a if clause to skip verification altogether

nimbus_verified_proxy/rpc/receipts.nim
chirag-parmar: noted. made necessary changes

nimbus_verified_proxy/rpc/receipts.nim
tersec: https://github.com/status-im/nimbus-eth1/pull/3457/commits/4c0bda545a3428e6bbfeb156db651c694933f6be

nimbus_verified_proxy/rpc/receipts.nim
tersec: https://github.com/status-im/nimbus-eth1/pull/3457/commits/4c0bda545a3428e6bbfeb156db651c694933f6be

nimbus_verified_proxy/rpc/receipts.nim
tersec: https://nim-lang.org/docs/manual.html#statements-and-expressions-var-statement
> Variables are always initialized with a default value if there is no initializing expression. The default value depends on the type and is always a zero in binary.

Sequences in particular are documented to be initialized to `@[]`.

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,134 @@
...
+  Log(address: lg.address, topics: lg.topics, data: lg.data)
+
+func toLogs(logs: openArray[LogObject]): seq[Log] =
+  logs.mapIt(it.toLog)
+
+func toReceipt(rec: ReceiptObject): Receipt =
+  let isHash = not rec.status.isSome()
+
+  let status =
+    if rec.status.isSome() and rec.status.get() == 1.Quantity: true else: false
```
bhartnett: You can simplify this line to:
```
let status = rec.status.isSome() and rec.status.get() == 1.Quantity
```

nimbus_verified_proxy/rpc/receipts.nim
bhartnett: Was this actually resolved? I can't see the code updated here.

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,134 @@
...
+    if lg.blockHash.isSome() and lg.transactionIndex.isSome() and lg.logIndex.isSome():
+      # exploit sequentiality of logs 
+      if prevBlockHash != lg.blockHash.get():
+        # TODO: a cache will solve downloading the same block receipts for multiple logs
+        rxs = (await vp.getReceipts(lg.blockHash.get())).valueOr:
+          return err(error)
+        prevBlockHash = lg.blockHash.get()
+
+      let
+        txIdx = int(distinctBase(lg.transactionIndex.get()))
```
tersec: `int(distinctBase(Quantity(...)))` is an `int(uint64(...))` which can trigger a `Defect` if out of range.

nimbus_verified_proxy/rpc/receipts.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3457/files#diff-3718119d0e8eaaef0ef91f7daacc999d50e3587bed8e3b9a9b063082c683a53bR97

chirag-parmar: Yes, just not on that line

nimbus_verified_proxy/rpc/receipts.nim
chirag-parmar: Will make the change

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,134 @@
...
+        logIdx =
+          int(distinctBase(lg.logIndex.get())) -
+          int(distinctBase(rxs[txIdx].logs[0].logIndex.get()))
+        rxLog = rxs[txIdx].logs[logIdx]
+
+      if rxLog.address == lg.address and rxLog.data == lg.data and
+          rxLog.topics == lg.topics and
+          match(toLog(lg), filterOptions.address, filterOptions.topics):
+        res.add(lg)
```
bhartnett: If my understanding is correct, it looks like if some but not all of the logs fail verification then the failed ones will be missing and the there is no indication that some of the validations failed.

If so, I think it might be better to just return an err message on the first log to fail verification so that we don't silently fail and exclude some of the data.

chirag-parmar: You are right. But the RPC provider itself could omit a few logs and the client would never know. If omission cannot be verified then the `getLogs` method is inherently considered to be "incomplete". So I took the liberty to not fail the request and utilize the already used bandwidth to return whatever is verified to be true.

chirag-parmar: I am not sure how strict the verified proxy should be in regards to the above. It would definitely make sense if we  decentralise verified proxy to query multiple RPC providers and compare answers between two providers (or if EIP7745 gets implemented in all clients and gets SFI'd in a fork). Until then, the trust assumption of the RPC provider doesn't change if we verify strictly or not. 

But if it feels iffy or not right to not return an error I can change that. LMK

bhartnett: In my opinion we should just make it as strict as possible, as much as the existing types allow. If we know some of the data is invalid we should return an err as soon as possible. As a user I'd want know about it so I could change to a different RPC provider in that case.

chirag-parmar: Okay I'll go ahead and change that

nimbus_verified_proxy/rpc/receipts.nim
chirag-parmar: Correct me if I am wrong, but a bounds checks against `uint32.high` should solve this right?

chirag-parmar: nvm solved it

tersec: https://github.com/status-im/nimbus-eth1/pull/3457/commits/3686fef8f56d231ca0b8022abf58745974265c12

nimbus_verified_proxy/tests/test_api_backend.nim
```diff
@@ -24,6 +24,10 @@ type
     proofs: Table[ProofQuery, ProofResponse]
     accessLists: Table[AccessListQuery, AccessListResult]
     codes: Table[CodeQuery, seq[byte]]
+    blockReceipts: Table[Hash32, seq[ReceiptObject]]
```
tersec, line 27: In theory, the point of defining these `hash` functions is so that the `Table` automatically uses them, per https://nim-lang.org/docs/tables.html#basic-usage-hashing
> What is happening here is that the types used for table keys require to have a `hash()` proc which will convert them to a [Hash](https://nim-lang.org/docs/hashes.html#Hash) value, and the compiler is listing all the hash functions it knows. Additionally there has to be a `==` operator that provides the same semantics as its corresponding `hash` proc.
>
> After you add `hash` and `==` for your custom type everything will work. Currently, however, hash for objects is not defined, whereas `system.==` for objects does exist and performs a "deep" comparison (every field is compared) which is usually what you want. So in the following example implementing only `hash` suffices

i.e. define `hash()` for whatever types you want (`==` will be created automatically typically) and then all four `Table`s here can use whatever actual values one wants to store.

chirag-parmar: You are right, that is the entire reason I implemented the `hash` function but forgot to define the key type for the table. However, the first three tables do actually use `Hash32` from the queries. I'll change the key type for the `logs` table.

tersec: https://github.com/status-im/nimbus-eth1/pull/3457/commits/0b479af9b88d4fbaea591593990c5d7ad7768605

nimbus_verified_proxy/tests/test_api_backend.nim
```diff
@@ -33,6 +37,10 @@ func init*(T: type TestApiState, chainId: UInt256): T =
     proofs: initTable[ProofQuery, ProofResponse](),
     accessLists: initTable[AccessListQuery, AccessListResult](),
     codes: initTable[CodeQuery, seq[byte]](),
+    blockReceipts: initTable[Hash32, seq[ReceiptObject]](),
```
tersec: None of these `initTable`s are necessary/useful

nimbus_verified_proxy/tests/test_api_backend.nim
```diff
@@ -68,12 +80,65 @@ template loadCode*(
...
+        hash(0)
+    addrHash = hash(x.address)
+    topicsHash = hash(x.topics)
+    blockHashHash =
+      if x.blockHash.isSome():
+        hash(x.blockHash.get)
+      else:
+        hash(0)
+
+  return (fromHash xor toHash xor addrHash xor topicsHash xor blockHashHash)
```
tersec: Don't need explicit `return`

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,125 @@
...
+  let logObjs =
+    try:
+      await vp.rpcClient.eth_getLogs(filterOptions)
+    except CatchableError as e:
+      return err(e.msg)
+
+  # store block hashes contains the logs so that we can batch receipt requests
+  var
+    prevBlockHash: Hash32
+    logIndexDiff: int
```
tersec:
```
nimbus-eth1/nimbus_verified_proxy/rpc/receipts.nim(99, 5) Hint: 'logIndexDiff' is declared but not used [XDeclaredButNotUsed]
```

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,125 @@
...
+    prevTxIdx: int
```
tersec:
```
nimbus-eth1/nimbus_verified_proxy/rpc/receipts.nim(100, 5) Hint: 'prevTxIdx' is declared but not used [XDeclaredButNotUsed]
```

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,123 @@
+# nimbus_verified_proxy
+# Copyright (c) 2025 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+import
```
tersec, line 10: should put `{.push gcsafe, raises: [].}` at top, before imports

chirag-parmar: done

nimbus_verified_proxy/tests/test_api_backend.nim
chirag-parmar: https://github.com/nim-lang/Nim/issues/25087

chirag-parmar:
@bhartnett and @tersec any other changes/reviews?

nimbus_verified_proxy/rpc/receipts.nim
```diff
@@ -0,0 +1,125 @@
+# nimbus_verified_proxy
+# Copyright (c) 2025 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push gcsafe, raises: [].}
```
bhartnett, line 8: This should be added to the top of the other files as well. Or you could do a separate PR adding this to all applicable verified proxy files.

chirag-parmar: will make a seperate PR

chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3543

PR #3458, opened 2025-07-08
Stateless:  Witness generation, clearing of witness keys and db persist and get helpers

bhartnett:
Changes in this PR:
- Implemented function to build/generate witnesses using the witness keys in the ledger.
- Refactored/renamed stateless files into better structure.
- Created Witness type which doesn't store the full bytecode or headers, only the hashes which will be used to store the witness in the database without data duplication. The reasoning here is that code is already stored by codeHash and headers are already stored separately.
- Added support to clear the witness keys when calling persist on the ledger.
- Added functions to persist and get witnesses from the database.
- Added more unit tests.


execution_chain/stateless/witness_generation.nim
```diff
@@ -0,0 +1,62 @@
...
+  eth/common,
+  ../db/ledger,
+  ./witness_types
+
+export
+  common,
+  ledger,
+  witness_types
+
+{.push raises: [].}
```
kdeme: We normally put this already before the imports

kdeme (approved):

execution_chain/stateless/witness_generation.nim
bhartnett: I see, does that change the behavior of the `{.push raises: [].}` statement?

PR #3461, opened 2025-07-09
proxy: log filters

chirag-parmar:
- [x] Adds functionality for `eth_newFilter`, `eth_uninstallFilter`, `eth_getFilterLogs` and `eth_getFilterChanges`
- [x] Adds tests for the above methods

Meta Issue: https://github.com/status-im/nimbus-eth1/issues/3101

### DIFF
--- a/nimbus_verified_proxy/rpc/receipts.nim
+++ b/nimbus_verified_proxy/rpc/receipts.nim
@@ -122,7 +122,7 @@ proc verifyLogs*(
       if prevBlockHash != lg.blockHash.get():
         # TODO: a cache will solve downloading the same block receipts for multiple logs
         rxs = (await vp.getReceipts(lg.blockHash.get())).valueOr:
-          return err("Couldn't get block receipt to verify logs")
+          return err(error)
         prevBlockHash = lg.blockHash.get()
       let
         txIdx = distinctBase(lg.transactionIndex.get())


nimbus_verified_proxy/rpc/rpc_eth_api.nim
```diff
@@ -274,7 +276,71 @@ proc installEthApiHandlers*(vp: VerifiedRpcProxy) =
...
       raise newException(ValueError, error)
 
-  # TODO:
+  vp.proxy.rpc("eth_newFilter") do(filterOptions: FilterOptions) -> string:
+    var id: array[8, byte] # 64bits
+
+    if randomBytes(id) != len(id):
+      raise newException(ValueError, "Couldn't assign a identifier for the filter")
+
+    let strId = toHex(id)
```
bhartnett: What happens if the generated id is already used? Sure this is very unlikely to happen but it wouldn't hurt to check for this and generate a second id if there is a clash perhaps.

nimbus_verified_proxy/types.nim
```diff
@@ -74,6 +79,7 @@ type
 
     # TODO: when the list grows big add a config object instead
     # config parameters
+    filterStore*: Table[string, FilterStoreItem]
```
bhartnett, line 84: It appears that this table can grow without limit. I didn't see any upper bound set on the number of filters that can be created. I would suggest setting a max for this. If you are at the max then creating new filters should fail.

nimbus_verified_proxy/rpc/rpc_eth_api.nim
chirag-parmar: good point. Will implement this

nimbus_verified_proxy/types.nim
chirag-parmar: good catch. will change this

PR #3462, opened 2025-07-09
Improves logging: short topics and remove duplicate log for same event

jangko:


execution_chain/networking/discoveryv4.nim
```diff
@@ -28,7 +28,7 @@ export
   enode
 
 logScope:
-  topics = "eth p2p discovery"
+  topics = "p2p-discovery"
```
arnetheduck: afair chronicles treats spaces as separator to create multiple topics, so you can filter for `p2p` separately from `discovery`

arnetheduck:
while we're here: would be nice if the short log for eth1 hashes was the same as eth2, ie the first N chars of the hash and not half/half of beginning/end (makes it easier to find blocks by hash online - they often have autocomplete)

execution_chain/networking/discoveryv4.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/3462/commits/f7070dffdadb00651a2bbadfff5cee2fb9692a69

PR #3463, opened 2025-07-09
proxy: mock transport

chirag-parmar:

Adds a mock transport layer to nimbus proxy for adding unit tests. This PR contains

- eth api backend interface (added similar to the async evm backend)
- `rpc_api_backend` - backend implementation that uses the existing eth api interface from nim-json-rpc
- `test_api_backend` - backend implementation that mocks eth api calls for testing
- `test_blocks` - a sample test that showcases use of test_api_backend
- other supporting changes


nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -101,12 +102,16 @@ proc run*(
...
 
-    # instantiate evm
+    # get network id
     networkId = chainIdToNetworkId(chainId).valueOr:
       raise newException(ValueError, error)
 
+  # set async evm backend
   verifiedProxy.evm = AsyncEvm.init(verifiedProxy.toAsyncEvmStateBackend(), networkId)
 
+  # set rpc backend
```
bhartnett: In my opinion these short comments aren't really necessary because they don't convey any additional information that is not already clear from the code.

nimbus_verified_proxy/rpc_api_backend.nim
```diff
@@ -0,0 +1,47 @@
...
+
+proc initNetworkApiBackend*(vp: VerifiedRpcProxy): EthApiBackend =
+  let
+    ethChainIdProc = proc(): Future[UInt256] {.async.} =
+      await vp.proxy.getClient.eth_chainId()
+
+    getBlockByHashProc = proc(
+        blkHash: Hash32, fullTransactions: bool
+    ): Future[BlockObject] {.async.} =
+      await vp.proxy.getClient.eth_getBlockByHash(blkHash, fullTransactions)
```
bhartnett: You can avoid the unneeded await here in all these functions by returning a raw future and using: {.async: (raises: [], raw: true).}

bhartnett (approved):

nimbus_verified_proxy/tests/test_api_backend.nim
```diff
@@ -0,0 +1,150 @@
...
+    fullBlocks: initTable[Hash32, BlockObject](),
+    blocks: initTable[Hash32, BlockObject](),
+    nums: initTable[Quantity, Hash32](),
+    tags: initTable[string, Hash32](),
+    proofs: initTable[ProofQuery, ProofResponse](),
+    accessLists: initTable[AccessListQuery, AccessListResult](),
+    codes: initTable[CodeQuery, seq[byte]](),
+  )
+
+proc clear*(t: var TestApiState) =
```
tersec: Probably can be `func`

PR #3467, opened 2025-07-13
Update Dockerfile to supported OS

yorickdowne:
My builds are failing because `buster` upstream repos are now defunct. Change this to `bullseye`, the oldest supported LTS.

While at it, change the base images from Ubuntu 20.04/21.04 to Ubuntu 22.04, the oldest supported LTS.

Docker is used to build release images, as well: That is why this needs to use an older LTS, so glibc in the release binaries is at an older version.

### DIFF
--- a/docker/dist/base_image/Dockerfile.amd64
+++ b/docker/dist/base_image/Dockerfile.amd64
@@ -1,5 +1,5 @@
 # Nimbus
-# Copyright (c) 2018-2023 Status Research & Development GmbH
+# Copyright (c) 2018-2025 Status Research & Development GmbH
 # Licensed under either of
 #  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or
 #    http://www.apache.org/licenses/LICENSE-2.0)
@@ -16,7 +16,7 @@
 # it once, upload it to Docker Hub and make sure it's being pulled regularly so
 # it's not deleted after 6 months of inactivity.
 
-FROM ubuntu:20.04
+FROM ubuntu:22.04
 
 SHELL ["/bin/bash", "-c"]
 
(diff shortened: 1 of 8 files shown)


docker/dist/base_image/Dockerfile.amd64
```diff
@@ -16,7 +16,7 @@
 # it once, upload it to Docker Hub and make sure it's being pulled regularly so
 # it's not deleted after 6 months of inactivity.
 
-FROM ubuntu:20.04
+FROM ubuntu:24.04
```
tersec: In general, we aim to keep compatibility of the resulting binaries with glibc versions back to the oldest supported Ubuntu LTS, i.e. currently 22.04. As far as I understand, by building these binaries within Ubuntu 24.04 base images, they pick up that glibc dependency, which means they won't run on older distributions.

This applies to the other `FROM ubuntu:24.04` lines as well -- they should be `FROM ubuntu:22.04`.

yorickdowne: I thought about that and went for 24.04 because these are Dockerfiles - you are saying this is then used to create published binaries as well? In that case yes, should all be 22.04. Let me make that change.

yorickdowne:
Looking at your build pipeline some more: You then use a specific hash of the base build in your Dockerfile. `FROM statusteam/nimbus-eth1:dist_base_eth1_20220326074403@sha256:5c1d2e982d404c08f6bba1f0cc1a678bffae3f6cee80a679edaa0b5e985dc05d
`
Am I thinking right that in order to fully make the change to a new OS requires two PRs? One to update the `base_image`, this one; then another to use the new base image?

kdeme (approved):
Thanks, need the debian change for nightly Portal build.

PR #3481, opened 2025-07-17
Remove state network code from portal code base.

bhartnett:


kdeme:
rebased + removed bridge dir from tool dirs

PR #3482, opened 2025-07-17
Speed up Nimbus shutdown when RPC/HTTP servers are enabled

bhartnett:
I found that Nimbus will hang for a while on shutdown when the RPC server/s are enabled. 

This change speeds up the shutdown process by not bothering to wait for the RPC/HTTP servers to fully stop in the Nimbus `stop` function. I believe this should be safe because the impact of 'killing' these services is low since by the time this stop function is called the database is already closed. There may be more of these services that can be stopped without waiting but I'm not sure which of these if any would be good candidates.

### DIFF
--- a/execution_chain/nimbus_desc.nim
+++ b/execution_chain/nimbus_desc.nim
@@ -61,9 +61,9 @@ proc stop*(nimbus: NimbusNode, conf: NimbusConf) {.async, gcsafe.} =
   trace "Graceful shutdown"
   var waitedFutures: seq[Future[void]]
   if nimbus.httpServer.isNil.not:
-    waitedFutures.add nimbus.httpServer.stop()
+    discard nimbus.httpServer.stop()
   if nimbus.engineApiServer.isNil.not:
-    waitedFutures.add nimbus.engineApiServer.stop()
+    discard nimbus.engineApiServer.stop()
   if conf.maxPeers > 0:
     waitedFutures.add nimbus.networkLoop.cancelAndWait()
   if nimbus.peerManager.isNil.not:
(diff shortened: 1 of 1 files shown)


execution_chain/nimbus_desc.nim
```diff
@@ -61,17 +61,17 @@ proc stop*(nimbus: NimbusNode, conf: NimbusConf) {.async, gcsafe.} =
...
     waitedFutures.add nimbus.networkLoop.cancelAndWait()
   if nimbus.peerManager.isNil.not:
     waitedFutures.add nimbus.peerManager.stop()
   if nimbus.beaconSyncRef.isNil.not:
     waitedFutures.add nimbus.beaconSyncRef.stop()
   if nimbus.metricsServer.isNil.not:
-    waitedFutures.add nimbus.metricsServer.stop()
+    discard nimbus.metricsServer.stop()
   if nimbus.wire.isNil.not:
     waitedFutures.add nimbus.wire.stop()
```
bhartnett, line 76: @jangko Do you think there any more of these services can be stopped without waiting as is done here for the RPC/HTTP services?

jangko: We should investigate why rpc/http server stop is slow. But for now perhaps add a comment/TODO explain why we are not waiting for it.

bhartnett: Actually, I did more testing on the current master branch and I can no longer reproduce the issue. The shutdown speed appears to be ok now even without this change. I noticed the slow shutdown yesterday but perhaps it was related to my local environment, not sure. Closing this PR for now until I can reproduce it again.

bhartnett: I've managed to reproduce this again. It only occurs when there is a port clash locally when starting Nimbus. For example, I have been running a Reth node, and then if I start Nimbus using the same default tcp-port of 30303, then Nimbus doesn't throw any error but then on shutdown it will hang (gets stuck) and doesn't exit.

PR #3486, opened 2025-07-18
Stateless: Use correct pre-state when building witness during import

bhartnett:
When stateless provider is enabled create a new txframe and store the parent txFrame to be used when building the witness.

### DIFF
--- a/execution_chain/core/chain/persist_blocks.nim
+++ b/execution_chain/core/chain/persist_blocks.nim
@@ -45,10 +45,8 @@ type
   Persister* = object
     com: CommonRef
     flags: PersistBlockFlags
-
     vmState: BaseVMState
     stats*: PersistStats
-
     parent: Header
 
   PersistStats* = tuple[blocks: int, txs: int, gas: GasInt]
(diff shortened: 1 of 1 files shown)


execution_chain/core/chain/persist_blocks.nim
```diff
@@ -174,24 +198,6 @@ proc persistBlock*(p: var Persister, blk: Block): Result[void, string] =
...
 
-  if vmState.com.statelessProviderEnabled:
-    let
-      witnessKeys = vmState.ledger.getWitnessKeys()
-      blockHash = header.computeBlockHash()
-      # Get the pre-state from before executing the block of transactions
-      parentTxFrame = CoreDbTxRef(
-        aTx: vmState.ledger.txFrame.aTx.parent,
-        kTx: vmState.ledger.txFrame.kTx.parent
-      )
```
bhartnett, line 185: Using this tx frame as the parent doesn't work correctly when the `chunk-size` is greater than 1 because before this change the parent would be pointing to the last tx frame which was only created/updated once every `chunk-size`  blocks.

PR #3487, opened 2025-07-18
Stateless: Correctly order keys when building witness

bhartnett:
This fixes an issue where the keys in the execution witness were generated in the incorrect order. The witness building is now restructured so that for each address, all of its slot keys are added before moving on to the next address.

### DIFF
--- a/execution_chain/stateless/witness_generation.nim
+++ b/execution_chain/stateless/witness_generation.nim
@@ -30,19 +30,7 @@ proc build*(
     addedCodeHashes = initHashSet[Hash32]()
 
   for key, codeTouched in witnessKeys:
-    let (_, maybeSlot) = key
-    if maybeSlot.isSome():
-      let slot = maybeSlot.get()
-      witness.addKey(slot.toBytesBE())
-
-      let proofs = ledger.getStorageProof(key.address, @[slot])
-      doAssert(proofs.len() == 1)
-      for trieNode in proofs[0]:
-        let nodeHash = keccak256(trieNode)
-        if nodeHash notin addedStateHashes:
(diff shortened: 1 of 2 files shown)


bhartnett:
> Nice that nimbus-eth1 is looking to match up with debug_executionWitness

Yes, that is the plan and actually this part is already implemented. We have the `debug_executionWitness` and `debug_executionWitnessByBlockHash` RPC endpoints implemented and matching the Reth implementation.

Something to point out about Reth's implementation is that for Reth the ordering of the state and keys is non-deterministic because each call to `debug_executionWitness` will produce a different ordering of the data. But I think this is fine because validity of the witnesses is not dependent on the ordering (apart from - for the keys the storage slots must come after each account address).

In order to get access to these execution witnesses using Nimbus EL:
1. Download the era1 files.
2. Run the import command build and store witnesses for each block: `build/nimbus_execution_client import --era1-dir=<era1_dir>  --stateless-provider=true`
3. After running the import up to your desired block, then start the EL with the debug RPC enabled: `build/nimbus_execution_client --rpc --rpc-api=debug`

kevaundray:
Can you say why era1 files are needed? Ie is it not possible to launch up a nimbus-eth1 node on hoodi/mainnet and have debug_executionWitness working for the last 64 blocks?

bhartnett:
> Can you say why era1 files are needed? Ie is it not possible to launch up a nimbus-eth1 node on hoodi/mainnet and have debug_executionWitness working for the last 64 blocks?

The storage of the block witnesses is not yet implemented when syncing via the CL. It's something that I'm planning to implement shortly. For now the only way is using the nimbus import command with the era1 files.

PR #3497, opened 2025-07-22
Adjust for authList v field change to yParity name

kdeme:
Draft as depends on https://github.com/status-im/nim-eth/pull/806

Note that this was giving issues when exporting blocks via EL JSON RPC (other clients) to convert to rlp and store as test vectors (Portal use case). As the Transactions would be invalid as the yParity (v) value would always be 0.
It would however only occur occasionally for some blocks since last hardfork as it applies only to EIP-7702 transactions.

### DIFF
--- a/tools/t8n/testdata/33/txs.json
+++ b/tools/t8n/testdata/33/txs.json
@@ -16,7 +16,7 @@
         "chainId": "0x1",
         "address": "0x000000000000000000000000000000000000aaaa",
         "nonce": "0x1",
-        "v": "0x1",
+        "yParity": "0x1",
         "r": "0xf7e3e597fc097e71ed6c26b14b25e5395bc8510d58b9136af439e12715f2d721",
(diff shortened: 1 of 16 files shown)


tools/t8n/testdata/33/txs.json
```diff
@@ -16,15 +16,15 @@
         "chainId": "0x1",
         "address": "0x000000000000000000000000000000000000aaaa",
         "nonce": "0x1",
-        "v": "0x1",
+        "yParity": "0x1",
```
kdeme, line 19: @jangko How are these files generated? Are they part of generation via https://github.com/ethereum/execution-spec-tests/ ? Or is it something manually added?

jangko: usually copied from geth or manually added

jangko:
requires https://github.com/status-im/nimbus-eth2/pull/7316

PR #3502, opened 2025-07-23
Stateless: Populate chain of headers in execution witness

bhartnett:
In order to prove that the BLOCKHASH opcode execution results are correct (for pre pectra when the block hashes are not fetched from the system contract) we need to collect the chain of block headers from the parent header back to the earliest fetched block number for each block executed.

See here: https://docs.rs/alloy-rpc-types-debug/latest/alloy_rpc_types_debug/struct.ExecutionWitness.html#structfield.headers

In order to implement this I've added an in memory block number to block hash cache in the vmState which is used to find the earliest block number fetched by the BLOCKHASH op code in the EVM for each block which also caches the fetched block hashes for performance reasons.

### DIFF
--- a/execution_chain/db/ledger.nim
+++ b/execution_chain/db/ledger.nim
@@ -719,7 +719,7 @@ template clearWitnessKeys*(ac: LedgerRef) =
 proc getBlockHash*(ac: LedgerRef, blockNumber: BlockNumber): Hash32 =
   ac.blockHashes.get(blockNumber).valueOr:
     let blockHash = ac.txFrame.getBlockHash(blockNumber).valueOr:
-      return default(Hash32)
+      default(Hash32)
 
     ac.blockHashes.put(blockNumber, blockHash)
     blockHash
--- a/execution_chain/stateless/witness_generation.nim
+++ b/execution_chain/stateless/witness_generation.nim
@@ -81,7 +81,7 @@ proc build*(
     parent: Header,
     header: Header): T =
 
-  if parent.stateRoot != default(Hash32):
+  if parent.number > 0:
     doAssert preStateLedger.getStateRoot() == parent.stateRoot
 
   var witness = Witness.build(ledger.getWitnessKeys(), preStateLedger)


execution_chain/evm/types.nim
```diff
@@ -55,6 +55,7 @@ type
     blobGasUsed*      : uint64
     allLogs*          : seq[Log] # EIP-6110
     gasRefunded*      : int64    # Global gasRefunded counter
+    blockHashesCache* : Table[BlockNumber, Hash32]
```
bhartnett: I'm going to change this to a LRUCache so we don't need to clear the cache when the stateless provider is disabled.

bhartnett: I will also move this cache and the related functions into the LedgerRef. Will make testing this logic in isolation easier. Also, most of the caching is already done in the ledger.

jangko:
TBH switching back and forth from regular EL client logic to stateless logic is a bit confusing for me.
If the stateless client feature is still a lot to be added, why not we refactor the EVM to use generics?
We use generics partially in the EVM to choose between tracer on/off codepath. We also use generics in the p2p code and beacon sync while keeping the performance balanced.
If we do the same for stateless codepath, there is no worry for stateless features to interference with regular EL logic performance.
Any stateless features can be added more easily. 
I know the resulted EVM binary size will be doubled, but the development process will be simplified while at the same time both compile time and runtime requirements can be achieved.

bhartnett:
> TBH switching back and forth from regular EL client logic to stateless logic is a bit confusing for me. If the stateless client feature are still a lot to be added, why not we refactor the EVM to use generics? We use generics partially in the EVM to choose between tracer on/off codepath. We also use generics in the p2p code and beacon sync while keeping the performance balanced. If we do the same for stateless codepath, there is no worry for stateless features to interference with regular EL logic performance. Any stateless features can be added more easily. I know the resulted EVM binary size will be doubled, but the development process will be simplified while at the same time both compile time and runtime requirements can be achieved.

The changes I've been working on are not specifically the stateless client (that may end up being a separate binary), but rather I've been working on the stateless provider which generates the block witnesses which would be used by the stateless client/s. So the parts I've been working on are just an additional feature added to the existing EL logic, not alternative logic and nothing in the regular EL is switched off when the stateless provider is enabled. This is similar to the Reth EL which has a runtime CLI parameter which configures enabling the stateless provider at runtime, nothing too complex in my opinion (especially compared to the database layers of the code base). Compile time conditionals don't work in this case because I need/want it to be configurable at runtime.

Actually there isn't any different code paths required for the EVM so I'm not sure why we might want to refactor it to use generics. Perhaps you are referring to the ledger changes? 

The stateless provider is disabled by default so any impact on performance shouldn't be a concern. When the flag is disabled the EL performance should be unchanged.

execution_chain/evm/state.nim
```diff
@@ -214,10 +214,7 @@ func baseFeePerGas*(vmState: BaseVMState): UInt256 =
 
 method getAncestorHash*(
     vmState: BaseVMState, blockNumber: BlockNumber): Hash32 {.gcsafe, base.} =
-  let db = vmState.ledger.txFrame
-  let blockHash = db.getBlockHash(blockNumber).valueOr:
-    return default(Hash32)
-  blockHash
+  vmState.ledger.getBlockHash(blockNumber)
```
bhartnett, line 217: This change should in theory speed up the EVM because the block hashes are now cached in memory. Previously the block hash lookup would always go to the database which has a few layers of logic and then rlp decodes the bytes from the database, eventually doing this just to get a hash:

```
proc getHash(
    db: CoreDbTxRef;
    key: DbKey;
      ): Result[Hash32, string] =
  const info = "getHash()"
  let data = db.get(key.toOpenArray).valueOr:
    if error.error != KvtNotFound:
      warn info, key, error=($$error)
    return err($$error)

  wrapRlpException info:
    return ok(rlp.decode(data, Hash32))
```

jangko:
>  Compile time conditionals don't work in this case because I need/want it to be configurable at runtime.

The `EVM` and it's `BaseEVMState` is re created every time a new block is executed. Before execute a block, the EVM `version` is chosen. Chose the a regular EVM or EVM with additional features. That's how runtime configuration can work.

See this for example how compile time and runtime configuration enable tracer in EVM.
https://github.com/status-im/nimbus-eth1/blob/d63cb74fe51e16fca4675cf4ac6b723a365de6e7/execution_chain/evm/interpreter/op_dispatcher.nim#L119-L133

execution_chain/evm/state.nim
jangko: Right, this is acceptable

bhartnett:
> > Compile time conditionals don't work in this case because I need/want it to be configurable at runtime.
> 
> The `EVM` and it's `BaseEVMState` is re created every time a new block is executed. Before execute a block, the EVM `version` is chosen. Chose the a regular EVM or EVM with additional features. That's how runtime configuration can work.
> 
> See this for example how compile time and runtime configuration enable tracer in EVM.
> 
> https://github.com/status-im/nimbus-eth1/blob/d63cb74fe51e16fca4675cf4ac6b723a365de6e7/execution_chain/evm/interpreter/op_dispatcher.nim#L119-L133

Yes I noticed the `selectVM` macro which converts the runtime `tracingEnabled` parameter into a static compile time parameter. 

Interesting idea but if there are concerns about having this code in the ledger, then I'd rather try and include these additional features in a new custom tracer rather than trying to create another generic option in the EVM code paths. 

Having said that, doing this refactor idea should be a separate PR. The witness keys are already merged and this additional block hashes cache improves the performance of the EVM for the blockhash opcode so it shouldn't be a separate feature anyway.

I also did some basic performance testing using the block import and didn't noticed any slow down when running the import over the first few million blocks.

jangko:
> then I'd rather try and include these additional features in a new custom tracer rather than trying to create another generic option in the EVM code paths.

Use the tracer is a good idea, especially when it related collecting data from the EVM.

PR #3510, opened 2025-07-25
reduce sync error threshold

advaita-saha:
As per testing in devnet, the ideal failing peers seems to be something in between 16-20


advaita-saha:
Issue solved with further peering improvements

PR #3512, opened 2025-07-26
Support BLOCKHASH opcode in Async EVM

bhartnett:
With this change the Async EVM now supports executing calls on contracts which use the BLOCKHASH opcode. Previously this was not supported. 

This PR also updates the Nimbus Verified Proxy as follows:
-  Enables this feature by implementing the `GetBlockHashProc` defined on the `AsyncEvmStateBackend`. The function calls the header store to get the requested blockhash.
- The default header store size is increased to 256 in order to support the BLOCKHASH opcode which can fetch the last 256 block hashes.
- The eth_call endpoint is now enabled. Before this change, eth_call requests would be proxied to the downstream RPC provider without any validation of the data.

### DIFF
--- a/nimbus_verified_proxy/header_store.nim
+++ b/nimbus_verified_proxy/header_store.nim
@@ -48,7 +48,7 @@ func convLCHeader*(lcHeader: ForkedLightClientHeader): Result[Header, string] =
         parentBeaconBlockRoot = Opt.none(Hash32)
 
     when lcDataFork >= LightClientDataFork.Electra:
-      # INFO: there is no visibility of the execution requests hash in light client header 
+      # INFO: there is no visibility of the execution requests hash in light client header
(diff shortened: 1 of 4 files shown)


execution_chain/evm/async_evm.nim
```diff
@@ -14,6 +14,7 @@ import
   chronicles,
   stint,
   results,
+  minilru,
```
chirag-parmar, line 17: Do we need `minilru`?

chirag-parmar (approved):

execution_chain/evm/async_evm.nim
bhartnett: Yes, it's used when iterating over the block hashes cache. See `getBlockHashesCache` function. Didn't add it as an export in the ledger because I didn't want to force that import in all locations that don't use the block cache.

PR #3513, opened 2025-07-27
proxy: block tests

chirag-parmar:
- [x] tests for importing blocks from different forks
- [x] tests for walking blocks
- [x] tests for other methods depending on importing blocks

https://github.com/status-im/nimbus-eth1/issues/3101

### DIFF
--- a/execution_chain/core/executor/process_block.nim
+++ b/execution_chain/core/executor/process_block.nim
@@ -14,7 +14,6 @@ import
   ../../common/common,
   ../../constants,
   ../../utils/utils,
-  ../../constants,
   ../../db/ledger,
   ../../transaction,
   ../../evm/state,


nimbus_verified_proxy/header_store.nim
```diff
@@ -92,6 +92,18 @@ func new*(T: type HeaderStore, max: int): T =
     earliestHash: Opt.none(Hash32),
   )
 
+func clear*(self: HeaderStore) =
+  for key in self.headers.keys():
+    self.headers.del(key)
+
+  for key in self.hashes.keys():
+    self.hashes.del(key)
```
bhartnett: It's faster to just create a new empty object then clearing all keys.

nimbus_verified_proxy/rpc/blocks.nim
```diff
@@ -31,6 +31,14 @@ proc resolveBlockTag*(
       let hLatest = vp.headerStore.latest.valueOr:
         return err("Couldn't get the latest block number from header store")
       ok(hLatest.number)
+    of "finalized":
+      let hFinalized = vp.headerStore.finalized.valueOr:
+        return err("Couldn't get the latest block number from header store")
+      ok(hFinalized.number)
+    of "earliest":
```
bhartnett, line 38: I just noticed that earliest here is just the earliest block in the header store but a user may expect to get the genesis block for this scenario. We should be able to return the genesis block data which is assumed/trusted already.

See the definition of earliest here: https://ethereum.org/en/developers/docs/apis/json-rpc/#default-block

chirag-parmar: the definition of earliest is debatable. https://ethereum.github.io/execution-apis/api-documentation/ here (under `eth_getBlockByNumber`) `earliest` is defined as the earliest block "available" in the **client** being queried. For the proxy this is, technically, the first "finalized" block the header store has seen (I use finalized only because it makes things easy, but in actuality it should be the first block in the queue). For an archive node it is genesis. 

moreover, querying the genesis block over RPC doesn't make a lot of sense. There is more trust involved and it is easier to download it from the repository

bhartnett: Sure, no problem. That sounds reasonable to me.

bhartnett (approved):

nimbus_verified_proxy/header_store.nim
chirag-parmar: makes sense. I'll change it

PR #3528, opened 2025-07-31
Add CI script for docker base image builder

### LINKED ISSUE #3474: Next step in updating Docker images
Now that the base image changes are in, these images need to be created. Could you kick off the relevant CI to update https://hub.docker.com/r/statusteam/nimbus-eth1/tags ?

I'll then create a PR to use the new base images in `docker/dist`

I'm also proposing to remove the base image for ARM and `Dockerfile.arm` - these don't appear to exist, and instead there's a base image for `arm64`, which will continue to exist.

jangko:
fix #3474


jangko:
Not yet, this base images will be rarely build. It will be used by Nightly build CI.
The final product is build by Nightly build CI then can be used by hive etc.

jangko:
> the nomenclature used by EF for docker images is nimbus-eth1:<branch-name>-<commit-hash> & nimbus-eth1:<branch-name>

Right, that is image name for building nimbus-eth1. On the other hand this is base images contains no nimbus-eth1 per se. But the derivatives might be.

PR #3540, opened 2025-08-05
Fix memory usage regression caused by stateroot mismatch fix

bhartnett:
Fixes the memory usage issue introduced by the previous state root mismatch fix here: https://github.com/status-im/nimbus-eth1/pull/3527

Instead of copying the updated values from a parent txFrame into the current txFrame in order to prevent losing the updates (when a snapshot is taken), here we just update the shared db LRU caches directly from the snapshot tables which should have the state updates from previous checkpoints which are copied over when the snapshot is moved.

### DIFF
--- a/execution_chain/db/aristo/aristo_tx_frame.nim
+++ b/execution_chain/db/aristo/aristo_tx_frame.nim
@@ -37,13 +37,6 @@ proc buildSnapshot(txFrame: AristoTxRef, minLevel: int) =
         # `frame` has a snapshot only in the first iteration of the for loop
         txFrame.snapshot = move(frame.snapshot)
 
-        # Copy cached values that are not present in the newer txFrame.
-        # These are needed to update the main caches in the AristoDbRef instance.
-        for k, v in frame.accLeaves:
-          discard txFrame.accLeaves.hasKeyOrPut(k, v)
-        for k, v in frame.stoLeaves:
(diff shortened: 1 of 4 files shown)


execution_chain/db/aristo/aristo_tx_frame.nim
```diff
@@ -217,7 +210,22 @@ with --debug-eager-state-root."""
   #      really run after things have been written (to maintain sync betweeen
   #      in-memory and on-disk state)
 
-  # Copy back updated payloads
+  # Copy back updated payloads into the shared database LRU caches.
+
+  # Copy cached values from the snapshot
```
arnetheduck, line 215: hm, I wonder if it's better to just remove them from `accLeaves` - the problem is that when `acc` is larger than `accLeaves`, all that happens is that the LRU evicts things over and over and at the end contains data that isn't really "last used" but "last in acc iteration order".

to find out, one would have to run a longer benchmark on this code and/or look at what typical acc vs accLeaves sizes are - in the meantime, I think it's good to merge this fix given it's correct for the repro test case and makes the client usable again.

bhartnett: Ok sure, I'll look into that and if it turns out to be faster to just remove the cached values will make another PR.

PR #3541, opened 2025-08-05
Docker base image rework

jangko:
Use up to date tooling e.g. new MacOS X SDK and llvm mingw to replace mex.
Rocksdb prebuilt binary also shipped with base images for their respective OS/CPU.
Base images only contains needed tools, make it slimmer. Build artifact also removed.

Downstream images become for simpler without have to build rocksdb. Build time become much more reasonable.
Except for linux arm64 which is still need tweaking.

Rocksdb binary are cached in CI, make it faster to deploy new base images.


docker/dist/Dockerfile.arm64
```diff
@@ -9,7 +9,7 @@
 # according to those terms.
 
 # The build is reproducible only if this base image stays the same.
-FROM statusim/nimbus-eth1:dist_base_eth1_v2_arm64@sha256:2f3d5dc56108428aca1257f65ce00dd4b5d8212c3927979edf41ffd161f13e63
+FROM statusim/nimbus-eth1:dist_base_eth1_v2_arm64
```
arnetheduck, line 12: an hash prevents someone from updating the tag and thus creating a supply chain attack, ie it's safer to keep the hash here.

jangko: will restore that after successful nightly build

jangko:
> are they automatically updated when bumping nim-rocksdb? ie how does one ensure consistency between versions?

invalidate the cache when there is a new version of nim-rocksdb:

https://github.com/status-im/nimbus-eth1/blob/0a3cf9dd25465af99be0a1beff62072b6f70d422/.github/workflows/build_base_image.yml#L35-L49

and second layer guard:

https://github.com/status-im/nim-rocksdb/blob/a95240da0f674b014d2b732539f12bed566f2446/scripts/build_static_deps.sh#L34-L49

PR #3543, opened 2025-08-06
add header pragma

chirag-parmar:


nimbus_verified_proxy/tests/test_setup.nim
```diff
@@ -6,6 +6,7 @@
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
 {.used.}
+{.push gcsafe, raises: [].}
```
bhartnett: FYI, it's not as important to add these to the tests in my opinion but of course doesn't hurt to add anyway.

Looks like the tests are failing to compile in the CI.

chirag-parmar: yes, a few methods throw errors. It is good because now we have exact specifications

nimbus_verified_proxy/header_store.nim
```diff
@@ -5,7 +5,7 @@
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-{.push raises: [].}
+{.push gcsafe, raises: [].}
```
arnetheduck: See also the linter here: https://github.com/status-im/nimbus-eth2/pull/7353/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR264

The linked PR updates it to accept gcsafe - notably, _after_ raises though. It doesn't greatly matter which order they are in but it's easier to grep when they're in the same order everywhere.

bhartnett: That order is also the same as the style guide: https://status-im.github.io/nim-style-guide/language.proctypes.html?highlight=gcsafe#callbacks-closures-and-forward-declarations-languageproctypes

So @chirag-parmar I would suggest updating to match the order of the style guide as suggest by @arnetheduck so that searching is easier.

chirag-parmar: okay noted.

PR #3553, opened 2025-08-08
Stateless: Speed up witness building by caching nodes when fetching proofs

bhartnett:
Building the block witnesses relies on fetching proofs from the database. 

I found that the slowest sections of code when building proofs were the call to `toNode` to get the `NodeRef` and the rlp encoding but these calls would be repeated multiple times for the same nodes when fetching multiple slots from the same account. Catching the result of these expensive calls results in at least a 2x speed up when fetching multiple slot proofs at once in a batch.

This PR also disables the state root checks in the witness building for an additional speed up. 

I'm planning to use the nodes cache when fetching multiple account proofs in a batch in a future PR when I implement a multiproof function. This should further improve the performance of building witnesses.

### DIFF
--- a/execution_chain/db/aristo/aristo_desc.nim
+++ b/execution_chain/db/aristo/aristo_desc.nim
@@ -61,17 +61,17 @@ type
     ## `sTab[]` tables must correspond to a hash entry held on the `kMap[]`
     ## tables. So a corresponding zero value or missing entry produces an
     ## inconsistent state that must be resolved.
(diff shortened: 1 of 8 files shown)


arnetheduck:
> when fetching multiple slot proofs

if you're fetching slot proofs, wouldn't it make more sense to cache the HashKey rather than the RLP enconding? caching `seq[byte]` means expensive allocations and hashing for every cache entry.

PR #3560, opened 2025-08-14
Replace eth_test with EEST test vectors

jangko:


arnetheduck:
image

unless master is rolled back, this will stay in history :/

tersec:
All EEST test vectors are frozen, or just these?

PR #3563, opened 2025-08-15
devnet EEST tests in local

### LINKED ISSUE #3523: Add support for EEST test vectors for both local test and CI test
The test vectors from [ethereum/tests](https://github.com/ethereum/tests) is not update anymore. From the release page they suggested to us to use [EEST](https://github.com/ethereum/execution-spec-tests) test vectors.

Or we can still use `ethereum/tests` by downloading its released zip instead of cloning submodule which is very long time.

Another benefit from EEST test vectors they also contain engine api test vector in the form of standardized json fixtures files.

So in addition to `test_generalstate_json` and `test_blockchain_json`, we can add test_`engine_api_json`

advaita-saha:
Fixes #3523 
Built on top of #3564 

Adds the following commands for running full EEST releases
- `make -j12 eest_engine_test` : Runs engineAPI tests for Mainnet ( static, stable develop ) & current devnet release of EEST fixtures
- `make -j12 eest_blockchain_test` : Runs block import tests for Mainnet ( static, stable develop ) & current devnet release of EEST fixtures
- `make -j12 eest_full_test` : Runs all of the above tests 

### TODO
- [x] More verbosity in testcase for debugging
- [x] Add blockchain tests
- [x] Add devnet and any new EEST release support
- [x] Make the testing infra extensible to new type of future EEST vectors
- [x] fix failing tests
- [x] ctrl-c should break all tests ( move away from cmd execution )
- [x] Add all eest types to CI cache
- [x] Add `getAppFilename` and remove `debugEcho`


tests/eest/eest_blockchain.nim
```diff
@@ -0,0 +1,102 @@
...
+    (waitFor env.runTest(unit.unit)).isOkOr:
+      debugEcho "TestName: ", unit.name, "RunTest error: ", error
+      testPass = false
+    env.close()
+
+  return testPass
+
+when isMainModule:
+  if paramCount() == 0:
+    debugEcho "Usage: eest_engine vector.json"
```
jangko: iirc nim has something like `appName()` to get the binary name to replace `eest_engine`.

tersec: https://nim-lang.org/docs/os.html#getAppFilename

tersec: But I also wonder why using `debugEcho` here -- it's a trick to avoid side effects, but here either the `debugEcho`s are in module scope, so side effects are allowed, or they're in a `proc`.

advaita-saha: Well this was just moved over from the #3564 
I didn't re-work this part as of yet, sure can shift to `getAppFilename`

advaita-saha: https://github.com/status-im/nimbus-eth1/pull/3563/commits/bf144365235a1ba139d209a64eca25e2179bcc4f

PR #3567, opened 2025-08-19
Deprecate '--custom-network' and merge it's functionality to '--network' 

### LINKED ISSUE #3565: `--network` should allow specifying a folder directly
https://github.com/status-im/nimbus-eth1/blob/166d9dc744f56ac4c8c5bced5762dee0201b2951/nrpc/config.nim#L76

same as nimbus-eth2, `--network=path/to/config` should pick up the network config from the given folder, in addition to the built-in names (`mainnet` etc)

jangko:
nimbus_execution_client:
We cannot remove `--custom-network` yet, there are still two users depends on this: hive tests, and kurtosis interop test. We can remove/disable `--custom-network` after both of hive tests and kurtosis interop test accept the PR to fix them. Now, the `--custom-network` only deprecated and not printed on CLI anymore. `--custom-network` functionality is added to `--network`.

nrpc:
`--custom-network` is removed and merged into `--network`.

fix #3565

### DIFF
--- a/execution_chain/common/chain_config.nim
+++ b/execution_chain/common/chain_config.nim
@@ -664,10 +664,16 @@ func genesisBlockForNetwork*(id: NetworkId): Genesis
   else:
     Genesis()
 
-func networkParams*(id: NetworkId): NetworkParams
-    {.gcsafe, raises: [ValueError, RlpError].} =
-  result.genesis = genesisBlockForNetwork(id)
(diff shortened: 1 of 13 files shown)


jangko:
@barnabasbusa are devnets ELs using the same launcher from https://github.com/ethpandaops/ethereum-package?

Issue #3569, opened 2025-08-19
RLP hash writer causing era1 import error


chirag-parmar:
### Description

Using `wrapEncoding` for encoding leaves in Aristo DB throws the below error for a era1 import test - `./env.sh ./build/nimbus_execution_client import  --era1-dir:tests/replay --max-blocks:1 --network=mainnet --log-level=TRACE --data-dir=./data-test/`

#### Error
```
< INF 2025-08-12 14:17:34.485+05:30 Writing genesis to DB                      blockHash=0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3 stateRoot=0xd7f8974fb5ac78d9ac099b9ad5018bedc2ce0a72dad1827a1709da30580f0544 difficulty=17179869184 gasLimit=5000 timestamp=0 nonce=0x0000000000000042
< DBG 2025-08-12 14:17:34.752+05:30 Core DB persisted                          kvtDur=14us584ns mptDur=27ms875us97ns endDur=235ms958us422ns stateRoot=none()
< DBG 2025-08-12 14:17:34.753+05:30 Reverting to base                          err="no finalized block hash and number"
< INF 2025-08-12 14:17:34.753+05:30 Database initialized                       base="(0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3, 0)" finalized="(0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3, 0)" head="(0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3, 0)"
< NTC 2025-08-12 14:17:34.753+05:30 Importing era1 archive                     start=1 dataDir=./data-test/ era1Dir=tests/replay
< WRN 2025-08-12 14:17:34.754+05:30 getScore()                                 topics="core_db" data=cac9850400000000808080 error="Unsigned integer expected, but the source RLP is a list"
< FAT 2025-08-12 14:17:34.754+05:30 Could not persist block                    blockNumber=1 error="persistHeader: cannot get score"
```

### Steps to reproduce

1. `git switch hash-writer-aristo`
2. `make DEBUG=1 -j12 nimbus_execution_client`
[...]

PR #3572, opened 2025-08-20
Use hash writer for aristo computeKey

### LINKED ISSUE #3164: Use optimised rlp hash writer in `aristo_compute`

chirag-parmar:
closed duplicate: https://github.com/status-im/nimbus-eth1/pull/3226
fixes: https://github.com/status-im/nimbus-eth1/issues/3164

Uses the rlp hash writer for Aristo DB's `computeKey` implementation

### DIFF
--- a/execution_chain/db/aristo/aristo_compute.nim
+++ b/execution_chain/db/aristo/aristo_compute.nim
@@ -14,7 +14,7 @@ import
   std/strformat,
   chronicles,
   eth/rlp,
-  eth/common,
+  eth/common/[accounts_rlp, base_rlp, hashes_rlp],
   results,
   "."/[aristo_desc, aristo_get, aristo_layers],
   ./aristo_desc/desc_backend
--- a/execution_chain/db/aristo/aristo_compute.nim
+++ b/execution_chain/db/aristo/aristo_compute.nim
@@ -117,13 +117,13 @@ func clear(self: var AristoTrieWriter) =
   self.twoPassWriter.clear()
   self.hashWriter.clear()
 
-template appendLeaf(w: var RlpWriter, pfx: NibblesBuf, leafData: auto) =
+template appendLeaf(w: var RlpWriter, pfx: NibblesBuf, leafData: Account | UInt256) =
   w.startList(2)
   w.append(pfx.toHexPrefix(isLeaf = true).data())
   w.wrapEncoding(1)
   w.append(leafData)
 
-template encodeLeaf(w: var AristoTrieWriter, pfx: NibblesBuf, leafData: auto): HashKey =
+template encodeLeaf(w: var AristoTrieWriter, pfx: NibblesBuf, leafData: Account | UInt256): HashKey =
   w.clear()
   w.tracker.appendLeaf(pfx, leafData)
 
--- a/execution_chain/db/aristo/aristo_compute.nim
+++ b/execution_chain/db/aristo/aristo_compute.nim
@@ -113,13 +113,13 @@ func clear(self: var AristoTrieWriter) =
   self.twoPassWriter.clear()
   self.hashWriter.clear()
 
-template appendLeaf(w: var RlpWriter, pfx: NibblesBuf, leafData: untyped) =
+template appendLeaf(w: var RlpWriter, pfx: NibblesBuf, leafData: auto) =
   w.startList(2)
   w.append(pfx.toHexPrefix(isLeaf = true).data())
   w.wrapEncoding(1)
   w.append(leafData)
 
-template encodeLeaf(w: var AristoTrieWriter, pfx: NibblesBuf, leafData: untyped): HashKey =
+template encodeLeaf(w: var AristoTrieWriter, pfx: NibblesBuf, leafData: auto): HashKey =
   w.clear()
   w.tracker.appendLeaf(pfx, leafData)
 
@@ -160,7 +160,7 @@ template appendExt(w: var RlpWriter, pfx: NibblesBuf, branchKey: HashKey) =
   w.append(pfx.toHexPrefix(isLeaf = false).data())
   w.append(branchKey)
 
-template encodeExt(w: var AristoTrieWriter, pfx: NibblesBuf, branchKey: untyped): HashKey =
+func encodeExt(w: var AristoTrieWriter, pfx: NibblesBuf, branchKey: HashKey): HashKey =
   w.clear()
   w.tracker.appendExt(pfx, branchKey)
 
@@ -211,9 +211,9 @@ proc computeKeyImpl(
   # empty state
 
   # Top-most level of all the verticies this hash computation depends on
-  var level = level
-
-  var writer = AristoTrieWriter.init()
+  var
+    level = level
+    writer = AristoTrieWriter.init()
 
   let key =
     case vtx.vType
@@ -241,17 +241,17 @@ proc computeKeyImpl(
           else:
             VOID_HASH_KEY
 
-      writer.encodeLeaf(vtx.pfx):
+      writer.encodeLeaf(vtx.pfx,
         Account(
           nonce: vtx.account.nonce,
           balance: vtx.account.balance,
           storageRoot: skey.to(Hash32),
           codeHash: vtx.account.codeHash,
         )
+      )
     of StoLeaf:
       let vtx = StoLeafRef(vtx)
-      writer.encodeLeaf(vtx.pfx):
-        vtx.stoData
+      writer.encodeLeaf(vtx.pfx, vtx.stoData)
     of Branches:
       # For branches, we need to load the vertices before recursing into them
       # to exploit their on-disk order
@@ -320,9 +320,8 @@ proc computeKeyImpl(
 
       if vtx.vType == ExtBranch:
         let vtx = ExtBranchRef(vtx)
-        writer.encodeExt(vtx.pfx):
-          var bwriter = AristoTrieWriter.init()
-          bwriter.writeBranch(vtx)
+        var bwriter = AristoTrieWriter.init()
+        writer.encodeExt(vtx.pfx, bwriter.writeBranch(vtx))
       else:
         writer.writeBranch(vtx)
 
--- a/execution_chain/db/aristo/aristo_compute.nim
+++ b/execution_chain/db/aristo/aristo_compute.nim
@@ -19,7 +19,7 @@ import
   "."/[aristo_desc, aristo_get, aristo_layers],
   ./aristo_desc/desc_backend
 
-type 
+type
   WriteBatch = tuple[writer: PutHdlRef, count: int, depth: int, prefix: uint64]
 
   AristoTrieWriter = object
@@ -118,9 +118,6 @@ template appendLeaf(w: var RlpWriter, pfx: NibblesBuf, leafData: untyped) =
   w.append(pfx.toHexPrefix(isLeaf = true).data())
   w.wrapEncoding(1)
   w.append(leafData)
-
-  # magical line 
-  discard rlp.encode(leafData)
 
 template encodeLeaf(w: var AristoTrieWriter, pfx: NibblesBuf, leafData: untyped): HashKey =
   w.clear()
--- a/execution_chain/db/aristo/aristo_compute.nim
+++ b/execution_chain/db/aristo/aristo_compute.nim
@@ -130,20 +130,20 @@ func encodeLeafAccount(pfx: NibblesBuf, acc: Account): HashKey =
   let balanceLen = rlpUInt256EncodedLen(acc.balance)
   const sRootLen = 33  # Hash32 always: 0xa0 prefix + 32 bytes
   const cHashLen = 33
-  let accContent = nonceLen + balanceLen + sRootLen + cHashLen
-  let accListLen = rlpListEncodedLen(accContent)
+  let accContentLen = nonceLen + balanceLen + sRootLen + cHashLen
+  let accTotalLen = rlpListEncodedLen(accContentLen)
 
   # wrapped encoding (rlp encoding a rlp encoded item is now ))))
-  let wrapLen = rlpBlobEncodedLen(accListLen)  # blob-wrap the account list
-  let outerContent = hexPfxBlobLen + wrapLen
-  let outerListLen = rlpListEncodedLen(outerContent)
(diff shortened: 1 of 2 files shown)


execution_chain/db/aristo/aristo_compute.nim
```diff
@@ -89,24 +96,81 @@ proc putKeyAtLevel(
...
+template appendLeaf(w: var RlpWriter, pfx: NibblesBuf, leafData: untyped) =
   w.startList(2)
   w.append(pfx.toHexPrefix(isLeaf = true).data())
+  w.wrapEncoding(1)
   w.append(leafData)
-  w.finish().digestTo(HashKey)
 
-template encodeBranch(w: var RlpWriter, vtx: VertexRef, subKeyForN: untyped): HashKey =
+  # magical line 
+  discard rlp.encode(leafData)
```
chirag-parmar: the problematic line

execution_chain/db/aristo/aristo_compute.nim
```diff
@@ -146,37 +210,36 @@ proc computeKeyImpl(
...
-                      keyvtxl[1],
-                      skipLayers = skipLayers,
-                    )
-              level = max(level, sl)
-              skey
-            else:
-              VOID_HASH_KEY
-
-        rlp.encode Account(
+        Account(
```
bhartnett, line 296: Looks like you removed the rlp encoding here.

execution_chain/db/aristo/aristo_compute.nim
```diff
@@ -185,8 +248,7 @@ proc computeKeyImpl(
     of StoLeaf:
       let vtx = StoLeafRef(vtx)
       writer.encodeLeaf(vtx.pfx):
-        # TODO avoid memory allocation when encoding storage data
-        rlp.encode(vtx.stoData)
+        vtx.stoData
```
bhartnett, line 251: And here the rlp encoding is removed.

execution_chain/db/aristo/aristo_compute.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/issues/3569

chirag-parmar: the error that pops up when I remove the line (note: the error pops up in `core_apps.nim` not `aristo_compute.nim`

```
< INF 2025-08-12 14:17:34.485+05:30 Writing genesis to DB                      blockHash=0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3 stateRoot=0xd7f8974fb5ac78d9ac099b9ad5018bedc2ce0a72dad1827a1709da30580f0544 difficulty=17179869184 gasLimit=5000 timestamp=0 nonce=0x0000000000000042
< DBG 2025-08-12 14:17:34.752+05:30 Core DB persisted                          kvtDur=14us584ns mptDur=27ms875us97ns endDur=235ms958us422ns stateRoot=none()
< DBG 2025-08-12 14:17:34.753+05:30 Reverting to base                          err="no finalized block hash and number"
< INF 2025-08-12 14:17:34.753+05:30 Database initialized                       base="(0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3, 0)" finalized="(0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3, 0)" head="(0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3, 0)"
< NTC 2025-08-12 14:17:34.753+05:30 Importing era1 archive                     start=1 dataDir=./data-test/ era1Dir=tests/replay
< WRN 2025-08-12 14:17:34.754+05:30 getScore()                                 topics="core_db" data=cac9850400000000808080 error="Unsigned integer expected, but the source RLP is a list"
< FAT 2025-08-12 14:17:34.754+05:30 Could not persist block                    blockNumber=1 error="persistHeader: cannot get score"
```

execution_chain/db/aristo/aristo_compute.nim
arnetheduck: @bhartnett since you have a good handle on this code, do you have a proposed fix?

bhartnett: @chirag-parmar mentioned that the removal of the `rlp.encode` here is intentional due to how the new rlp encoder works so no I don't have a proposed fixed but I'll take another look and see if I can find a solution.

execution_chain/db/aristo/aristo_compute.nim
```diff
@@ -89,24 +96,78 @@ proc putKeyAtLevel(
 
   ok()
 
-template encodeLeaf(w: var RlpWriter, pfx: NibblesBuf, leafData: untyped): HashKey =
+func init(T: type AristoTrieWriter): T =
+  result.tracker = DynamicRlpLengthTracker()
+  result.tracker.initLengthTracker()
+  result.twoPassWriter = initTwoPassWriter(result.tracker)
+  result.hashWriter = initHashWriter(result.tracker)
```
bhartnett: @chirag-parmar Usage of result return should be avoided, see here: https://status-im.github.io/nim-style-guide/language.result.html

execution_chain/db/aristo/aristo_compute.nim
bhartnett: After merging the latest master branch into this branch the above problem is now fixed. I'm not sure exactly which change in master fixed the problem but it appears to be resolved. The tests pass and the block import is working. @chirag-parmar Perhaps you could test out the changes in this branch and confirm?

I've also put in a few minor improvements not related to the issue FYI but they didn't contribute to the fix.

execution_chain/db/aristo/aristo_compute.nim
bhartnett: > @bhartnett since you have a good handle on this code, do you have a proposed fix?

@arnetheduck This one is fixed. See comment here: https://github.com/status-im/nimbus-eth1/pull/3572#discussion_r2332059707

execution_chain/db/aristo/aristo_compute.nim
chirag-parmar: I'll change this

execution_chain/db/aristo/aristo_compute.nim
chirag-parmar: > @chirag-parmar Perhaps you could test out the changes in this branch and confirm?

It works. Thank you

execution_chain/db/aristo/aristo_compute.nim
```diff
@@ -13,12 +13,19 @@
 import
   std/strformat,
   chronicles,
-  eth/common/[accounts_rlp, base_rlp, hashes_rlp],
+  eth/rlp,
+  eth/common,
```
arnetheduck: why do we need all of `common` here?

execution_chain/db/aristo/aristo_compute.nim
```diff
@@ -89,24 +96,82 @@ proc putKeyAtLevel(
...
+
+func clear(self: var AristoTrieWriter) =
+  self.tracker.clear()
+  self.twoPassWriter.clear()
+  self.hashWriter.clear()
+
+template appendLeaf(w: var RlpWriter, pfx: NibblesBuf, leafData: auto) =
   w.startList(2)
   w.append(pfx.toHexPrefix(isLeaf = true).data())
+  w.wrapEncoding(1)
```
arnetheduck: doesn't this call to `wrapEnconding` depend on `leafData` being of a particular type? if yes, why use `auto`?

execution_chain/db/aristo/aristo_compute.nim
```diff
@@ -89,24 +96,82 @@ proc putKeyAtLevel(
...
   w.append(leafData)
-  w.finish().digestTo(HashKey)
 
-template encodeBranch(w: var RlpWriter, vtx: VertexRef, subKeyForN: untyped): HashKey =
+template encodeLeaf(w: var AristoTrieWriter, pfx: NibblesBuf, leafData: auto): HashKey =
```
arnetheduck: multiple evaluations of `leafData` - `auto` or `untyped` doesn't really matter in this context, they behave the same

execution_chain/db/aristo/aristo_compute.nim
bhartnett: I'm not sure about `wrapEncoding` but the reason I changed this line was that we are passing a type account or slot and don't need to pass in a block of statements. Perhaps it would be better to make it even more strict and use `Account | UInt256` as the type.

execution_chain/db/aristo/aristo_compute.nim
bhartnett: Yes this change wasn't related to reducing the evaluations. I just wanted to remove the untyped parameter.

execution_chain/db/aristo/aristo_compute.nim
```diff
@@ -154,7 +154,7 @@ template appendExt(w: var RlpWriter, pfx: NibblesBuf, branchKey: HashKey) =
   w.append(pfx.toHexPrefix(isLeaf = false).data())
   w.append(branchKey)
 
-template encodeExt(w: var AristoTrieWriter, pfx: NibblesBuf, branchKey: untyped): HashKey =
+func encodeExt(w: var AristoTrieWriter, pfx: NibblesBuf, branchKey: HashKey): HashKey =
```
bhartnett: This change is the one which reduces the evaluations of branchKey from 2 per call to 1.

execution_chain/db/aristo/aristo_compute.nim
bhartnett: @chirag-parmar

chirag-parmar: I think I left it in there from the experimentation I was doing to solve the error. I'll remove it

execution_chain/db/aristo/aristo_compute.nim
arnetheduck: the same thing applies to `subKeyForN` and `leafData` - they need to either be stored in a variable or encodeBranch/encodeLeaf need to be made `func`

bhartnett: Good point. Updated as suggested.

execution_chain/db/aristo/aristo_compute.nim
arnetheduck: ok - it feels odd that it should be needed cc @chirag-parmar - ie `append` should be able to figure this out ideally - but it seems to work, so ..

execution_chain/db/aristo/aristo_compute.nim
```diff
@@ -89,24 +96,86 @@ proc putKeyAtLevel(
...
-  for (n {.inject.}, subvid {.inject.}) in vtx.allPairs():
-    w.append(subKeyForN)
+  for key in hashKeys:
+    w.append(key)
   w.append EmptyBlob
-  w.finish().digestTo(HashKey)
 
-template encodeExt(w: var RlpWriter, pfx: NibblesBuf, branchKey: HashKey): HashKey =
+template encodeBranch(w: var AristoTrieWriter, rvtx: VertexRef, subKeyForN: untyped): HashKey =
+  var hashKeys: array[16, HashKey]
```
arnetheduck:
```suggestion
  var hashKeys {.noinit.}: array[16, HashKey]
```

arnetheduck:
so .. the final thing to do with this branch is to run a benchmark: using an unhashed snapshot like https://eth1-db.nimbus.team/mainnet-20250831-nohash.tar.gz, run the node pre/post to measure how long it takes to create the full hash tree (it's printed when starting the node) - afair, 25% of cpu time was spent allocating rlp encodings, so we should see a good boost here.

execution_chain/db/aristo/aristo_compute.nim
chirag-parmar: Apologies for the late reply. the call to `wrapEncoding` doesn't depend on the `leafData` being of a particular type. Only leaf nodes have nested serializations i.e. Serializing an object/list/tuple that already has a serialized rlp value in it.

chirag-parmar: > append should be able to figure this out ideally

I didn't understand what it must be able to figure out.

arnetheduck: I think what I would have expected maybe is a call like `appendEncoded(leafData)` which does the extra round of encoding of the passed-in item - `wrapEncoding` looks like an internal low-level operation, specially with that number being passed in, because presumably it must be followed up with actual data - that gives the caller an unnecessary opportunity to use it the wrong way

chirag-parmar:
> Unfortunately with these changes the stateroot computation appears to be slower than the master branch. 

I'm not sure why we are measuring the compute time. The purpose of the hash writer was to save on memory and not time. It still does improve on time a little as seen [here](https://github.com/status-im/nim-eth/pull/765#issuecomment-2591766025) but only for large data objects like entire blocks. 

> @chirag-parmar Perhaps you could do a benchmark as well to confirm these results?

I'll try and run the benchmarks and measure the memory usage too. But I wouldn't expect results too far off from @bhartnett 's benchmarks.

arnetheduck:
> hash writer was to save on memory

not really - it's both - memory allocations _take_ time and the idea is to avoid creating work for the garbage collector, which ultimately saves both time and memory, when done right. 25% of _time_ is spent allocating and collecting, and by creating the hashes directly we should be able to avoid that work entirely.

execution_chain/db/aristo/aristo_compute.nim
chirag-parmar: makes sense actually. https://github.com/status-im/nim-eth/issues/813

bhartnett:
This alternative implementation https://github.com/status-im/nimbus-eth1/pull/4115 which uses the RlpArrayBufWriter to pre-allocated the rlp output buffer on the stack performs as well as this hash writer implementation even after these recent improvements, but the arraybuf version has a simpler and more re-usable implementation. 

I suggest that we can close this PR @chirag-parmar

PR #3577, opened 2025-08-21
fix crash in invalid fc state

### LINKED ISSUE #3576: Unable to load chain, and crash in fusaka-devnet-3
```bash
INF 2025-08-21 11:29:46.678+00:00 Database initialized                       base="(0x87eb9f7501e9af70e8a0f242aceb73338d1298d6eee00fc31e483f293372418a, 185547)" finalized="(0xb00f8cc1f32c824201b7dc749b1c2177ec0619ecf8b454463c95ca8143ba7c06, 185539)" head="(0x87eb9f7501e9af70e8a0f242aceb73338d1298d6eee00fc31e483f293372418a, 185547)"
/root/nimbus-eth1/vendor/nim-json-rpc/json_rpc/client.nim(865) main
/root/nimbus-eth1/vendor/nim-json-rpc/json_rpc/client.nim(853) NimMain
/root/nimbus-eth1/execution_chain/nimbus_execution_client.nim(274) _ZN23nimbus_execution_client3runE3refIN11nimbus_desc26NimbusNodecolonObjectType_EEN6config10NimbusConfE
/root/nimbus-eth1/execution_chain/nimbus_execution_client.nim(49) _ZN23nimbus_execution_client13basicServicesE3refIN11nimbus_desc26NimbusNodecolonObjectType_EEN6config10NimbusConfE3refIN6common25CommonRefcolonObjectType_EE
/root/nimbus-eth1/execution_chain/core/chain/forked_chain/chain_serialize.nim(245) _ZN15chain_serialize11deserializeE3refIN10chain_desc30ForkedChainRefcolonObjectType_EE
/root/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/chcks.nim(28) raiseIndexError2
/root/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/fatal.nim(53) _ZN6system8sysFatalE8typeDescIN10exceptions11IndexDefectEE6string
/root/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/excpt.nim(349) _ZN6system18rawWriteStackTraceE3varI3seqIN6system15StackTraceEntryEEE
/root/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/stacktraces.nim(62) _ZN11stacktraces30auxWriteStackTraceWithOverrideE3varI3seqIN6system15StackTraceEntryEEE
Error: unhandled exception: index 121 not in 0 .. 91 [IndexDefect]
```

This is a weird case and should not occur if everything is going correctly. While restarts in nimbus-eth1, some nodes went into this
The index defect occurs in 
https://github.com/status-im/nimbus-eth1/blob/ecf67fe964108a945777458a752c96336845b6a1/execution_chain/core/chain/forked_chain/chain_serialize.nim#L245

The array is initialised in 
https://github.com/status-im/nimbus-eth1/blob/ecf67fe964108a945777458a752c96336845b6a1/execution_chain/core/chain/forked_chain/chain_serialize.nim#L232

So this occurrence is weird, but might have happened during moving the db up one layer
One way is to check for `state.numBlocks < state.latest` and do a `fc.reset(prevBase)`

advaita-saha:
Fixes #3576

### DIFF
--- a/execution_chain/core/chain/forked_chain/chain_serialize.nim
+++ b/execution_chain/core/chain/forked_chain/chain_serialize.nim
@@ -236,7 +236,8 @@ proc deserialize*(fc: ForkedChainRef): Result[void, string] =
      state.base > state.numBlocks: 
     fc.reset(prevBase)
     return err("Invalid state: latest block is greater than number of blocks")
-
+  
+  # Sanity Checks for all the heads in FC state
   for head in state.heads:
     if head > state.numBlocks:
       fc.reset(prevBase)
--- a/execution_chain/core/chain/forked_chain/chain_serialize.nim
+++ b/execution_chain/core/chain/forked_chain/chain_serialize.nim
@@ -231,9 +231,16 @@ proc deserialize*(fc: ForkedChainRef): Result[void, string] =
   let prevBase = fc.base
   var blocks = newSeq[BlockRef](state.numBlocks)
 
-  if state.latest > state.numBlocks:
+  # Sanity Checks for the FC state
+  if state.latest > state.numBlocks or
+     state.base > state.numBlocks: 
     fc.reset(prevBase)
     return err("Invalid state: latest block is greater than number of blocks")
+
+  for head in state.heads:
+    if head > state.numBlocks:
+      fc.reset(prevBase)
+      return err("Invalid state: heads greater than number of blocks")
 
   try:
     for i in 0..<state.numBlocks:
--- a/execution_chain/core/chain/forked_chain/chain_serialize.nim
+++ b/execution_chain/core/chain/forked_chain/chain_serialize.nim
@@ -231,6 +231,10 @@ proc deserialize*(fc: ForkedChainRef): Result[void, string] =
   let prevBase = fc.base
   var blocks = newSeq[BlockRef](state.numBlocks)
 
+  if state.latest > state.numBlocks:
+    fc.reset(prevBase)
+    return err("Invalid state: latest block is greater than number of blocks")
+
   try:
     for i in 0..<state.numBlocks:
       let


execution_chain/core/chain/forked_chain/chain_serialize.nim
```diff
@@ -231,6 +231,10 @@ proc deserialize*(fc: ForkedChainRef): Result[void, string] =
   let prevBase = fc.base
   var blocks = newSeq[BlockRef](state.numBlocks)
 
+  if state.latest > state.numBlocks:
+    fc.reset(prevBase)
+    return err("Invalid state: latest block is greater than number of blocks")
```
arnetheduck, line 238: is the root cause understood here? why would this happen?

advaita-saha: Not really identified, still trying to pin point the cause
This can only happen IMO sometimes if the `nimbus-eth1` process is force killed without `-SIGINT`.

Another factor maybe moving the db to the top level ( compliant with eth2 ) in the devnet didn't occur correctly due to some issue.

Also this didn't occur on all nodes

jangko: I think `state.base` and `state.heads` should also checked against `state.numBlocks`.

jangko (approved):

execution_chain/core/chain/forked_chain/chain_serialize.nim
arnetheduck: > Another factor maybe moving the db to the top level ( compliant with eth2 ) in the devnet didn't occur correctly due to some issue.

this sounds unreasonable: the location of the database has no influence on the _contents_ of the database - this error is about an inconsistent database state caused by .. something. that something, we have to investigate because it should not happen - in the meantime, we should log a big ugly TODO warning here so that it's not lost

PR #3578, opened 2025-08-22
Add test covering HashKey rlp encoding when data contains embedded trie node

bhartnett:
For some reason when decoding some trie nodes when verifying proofs and fetching one of the HashKey values from a branch node, the value was encoded as a list when it should be a blob. 

This appears to fix this issue. Perhaps `appendRawBytes` in the default rlp writer is broken.

Edit:
The above 'fix' was not actually correct (see comment [here](https://github.com/status-im/nimbus-eth1/pull/3578#issuecomment-3222691713)) and so I've reverted that change and I've also added a test case which covers this code path. The test fails when the rlp encoding of the HashKey type is incorrectly modified for the embedded trie node scenario.


execution_chain/db/aristo/aristo_desc/desc_identifiers.nim
```diff
@@ -218,7 +218,7 @@ func hash*(a: HashKey): Hash =
 
 func append*(w: var RlpWriter; key: HashKey) =
   if 1 < key.len and key.len < 32:
-    w.appendRawBytes key.data
```
bhartnett: @chirag-parmar Do you know of any reason why this might be encoding values as a list rather than a blob?

bhartnett: This is the code that decodes the trie nodes and expects a blob to be present in each of the branch node indexes, not a list: https://github.com/status-im/nimbus-eth1/blob/d584afb9925d059cde163800d538d48fd168cb2f/execution_chain/db/aristo/aristo_proof.nim#L243

arnetheduck:
One of the bigger concerns here is that the tests are green even _after_ this change - it means that this core piece of logic is untested and, most likely, both of which are red flags - thus, the first thing to do would be to find/implement some solid test vectors and link to a _spec_ to document how it should be encoded and go from there.

execution_chain/db/aristo/aristo_desc/desc_identifiers.nim
chirag-parmar: This is weird.

All I can say is that `HashKey` is being encoded as a list because the overloading resolution algorithm at the time of compilation uses the generic method defined here https://github.com/status-im/nim-eth/blob/master/eth/rlp/writer.nim#L202 for serializing it. At the time of compilation, for some unknown reason, it doesn't `mixin` the `append` method defined in `desc_identifiers.nim`. 

 I have been facing a similar issue https://github.com/status-im/nimbus-eth1/pull/3572. Even here, the score gets serialized as a list (because it is an object: UInt256) UNLESS I include that one magical line of code.
 
 So I think the change you made has nothing to do with the incorrect serialization. This is just my educated guess (even `writeBlob` calls `appendRawBytes` internally)

bhartnett: > This is weird.
> 
> All I can say is that `HashKey` is being encoded as a list because the overloading resolution algorithm at the time of compilation uses the generic method defined here https://github.com/status-im/nim-eth/blob/master/eth/rlp/writer.nim#L202 for serializing it. At the time of compilation, for some unknown reason, it doesn't `mixin` the `append` method defined in `desc_identifiers.nim`.
> 
> I have been facing a similar issue #3572. Even here, the score gets serialized as a list (because it is an object: UInt256) UNLESS I include that one magical line of code.
> 
> So I think the change you made has nothing to do with the incorrect serialization. This is just my educated guess (even `writeBlob` calls `appendRawBytes` internally)

I see, well thanks for the info. I'll keep looking into it as it would be good to know definitively the root cause.

bhartnett:
> One of the bigger concerns here is that the tests are green even _after_ this change - it means that this core piece of logic is untested and, most likely, unused, both of which are red flags - thus, the first thing to do would be to find/implement some solid test vectors and link to a _spec_ to document how it should be encoded and go from there.

The rlp serialization of the HashKey type impacts the stateRoot and storageRoot calculations and therefore this is tested when syncing or running the block import with stateRoot checks enabled. But yes the unit tests did not cover this specific scenario where the HashKey type contains an embedded trie node which only occurs under certain conditions. I've just added a test case which covers this and verified that it breaks when the HaskKey serialization is modified as was previously done in this PR.

PR #3580, opened 2025-08-22
Add command line option for marking a manual sync target finalised

mjfh:
why
  If it is known to be finalised, it can be set. The implications are
  that with the given sync target no CL is needed to keep the memory footprint
  small.

### DIFF
--- a/execution_chain/config.nim
+++ b/execution_chain/config.nim
@@ -507,19 +507,20 @@ type
         name: "jwt-secret" .}: Option[InputFile]
 
       beaconSyncTarget* {.
-        separator: "\pBEACON SYNC OPTIONS:"
+        hidden
         desc: "Manually set the initial sync target specified by its 32 byte" &
               " block hash (e.g. as found on etherscan.io) represented by a" &
               " hex string"
-        name: "beacon-sync-target" .}: Option[string]
+        name: "debug-beacon-sync-target" .}: Option[string]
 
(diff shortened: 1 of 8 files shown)


execution_chain/config.nim
```diff
@@ -508,12 +508,19 @@ type
...
+              " hex string"
         name: "beacon-sync-target" .}: Option[string]
 
+      beaconSyncTargetIsFinal* {.
+        defaultValue: false
+        desc: "If the sync taget is finalised (e.g. as stated on" &
+              " etherscan.io) this can be set here. For a non-finalised" &
+              " manual sync target it is advisable to run this EL against a" &
+              " CL which will result in a smaller memory footprint"
+        name: "beacon-sync-target-is-final".}: bool
```
arnetheduck: both of these options should be `hidden` and start with `debug-` - it's not something we want to keep in a public release

mjfh: This was modelled like in `reth` where it is public. But I can set it hidden of course if you want so?

arnetheduck: yeah, let's keep them hidden for now .. in the future, we should probably add a way to create "debug" options instead of "hidden" options, to document that these may be removed and / or changed without notice.

In general, the option is not something uses should "need", when they follow the default ux

mjfh: ok, makes sense

PR #3582, opened 2025-08-25
fix: correct fork transition logic in lastFork function

MozirDmitriy:
Fix the logical impossibility in lastFork function where the condition
fork > currentFork was used in a countdown loop, making it impossible
to find previous forks. Changed to fork < currentFork to correctly
identify the last hard fork before the current one.


advaita-saha:
This calculates the ending fork ( i.e last )
Not the previous fork

advaita-saha:
The misleading comment has been corrected in the PR https://github.com/status-im/nimbus-eth1/pull/3563

PR #3585, opened 2025-08-26
fix: queueUpdateBase batching to iterate from base to prevQueuedBase

GarmashAlex:
- Iterate from new base target back toward prevQueuedBase for correct step scheduling
- Pre-allocate steps capacity based on (base.number - prevQueuedBase.number)
- Keeps batch size semantics via PersistBatchSize; no behavior changes elsewhere


jangko:
close in favor of #3741

PR #3589, opened 2025-08-26
proxy: state tests

chirag-parmar:
- [ ] tests for getting accounts and storages
- [ ] tests for evm based methods


nimbus_verified_proxy/tests/data/download_proofs.sh
```diff
@@ -0,0 +1,132 @@
+curl --request POST \
+  --url https://eth.blockrazor.xyz \
+  --header 'Content-Type: application/json' \
+  --data '{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "eth_getBlockByNumber",
+  "params": ["latest", true]
+}' | jq '.result' > proof_block.json
```
tersec: This download the `proof_block.json` and other parts of this script download `storage_proof.json` as well as `code.json`.

It does look like these are small and useful enough to keep in the GitHub repository itself, but this `download_proofs.sh` script maybe doesn't belong in the `tests/data/` directory? One possibility is https://github.com/status-im/nimbus-eth1/tree/master/scripts

nimbus_verified_proxy/tests/data/download_proofs.sh
```diff
@@ -0,0 +1,132 @@
...
+  --url https://eth.blockrazor.xyz \
+  --header 'Content-Type: application/json' \
+  --data '{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "eth_getCode",
+  "params": ["0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2", "latest"]
+}' | jq '.result' > code.json
+
+echo "Account Balance:\n"
```
tersec: This prints a literal `\n` for me (Debian Linux, default `bash`):
```
$ echo "Account Balance:\n"
Account Balance:\n
```

https://www.shellcheck.net/wiki/SC2028 suggests to replace all these `echo`s which have embedded escape sequences to `printf`s, which are specified to expand them:
```
$ printf "Account Balance:\n"
Account Balance:
$
```

nimbus_verified_proxy/tests/test_utils.nim
```diff
@@ -42,6 +42,27 @@ proc getLogsFromJson*(
   let logs = JrpcConv.decode(logBytes.get, seq[LogObject])
   return logs
 
+proc getProofFromJson*(
+    filepath: string
+): ProofResponse {.raises: [SerializationError].} =
+  var proofBytes = readAllBytes(filepath)
```
tersec: Why are `proofBytes`, `filebytes` (in both `getAccessListFromJson` and `getCodeFromJson` `var`? It doesn't appear to be modified. They can be `let`s instead.

nimbus_verified_proxy/tests/test_state.nim
```diff
@@ -0,0 +1,86 @@
...
+  ../header_store,
+  ./test_utils,
+  ./test_api_backend
+
+suite "test state verification":
+  let
+    ts = TestApiState.init(1.u256)
+    vp = startTestSetup(ts, 1, 1, 8897)
+
+  test "test EVM based methods":
```
tersec: EVM-based

nimbus_verified_proxy/tests/data/download_proofs.sh
tersec: It looks like `nimbus_verified_proxy/tests/data` has a couple of other shell scripts already, and I'd kind of say the same thing about them, it's reasonable/useful to show how the test files were obtained/derived, but also, they aren't themselves test data.

nimbus_verified_proxy/tests/data/download_proofs.sh
```diff
@@ -0,0 +1,132 @@
+curl --request POST \
```
tersec: It's convenient for shell scripts to have the shbang line, and also for things like these where it's generating data the correctness and completeness of which is important, using `set -Eeuo pipefail` will try to make explicit many of the failure modes bash/zsh-like shell scripts can have, rather than silently failing to run some of their commands.

nimbus_verified_proxy/tests/test_utils.nim
```diff
@@ -42,6 +42,27 @@ proc getLogsFromJson*(
...
+  let proof = JrpcConv.decode(proofBytes.get, ProofResponse)
+  return proof
```
tersec: Both of the last two lines can be replaced by the shorter, simpler:
```nim
  JrpcConv.decode(proofBytes.get, ProofResponse)
```
i.e. implicit return; the `proof` variable really does nothing here.

A similar transformation can apply to the others. Just
```nim
  JrpcConv.decode(filebytes.get, AccessListResult)
```
instead of
```nim
  let accessList = JrpcConv.decode(filebytes.get, AccessListResult)
  return accessList
```
and similarly for `getCodeFromJson`:
```nim
  JrpcConv.decode(filebytes.get, string).hexToSeqByte()
```

nimbus_verified_proxy/tests/data/download_proofs.sh
chirag-parmar: moved them to `nimbus_verified_proxy/tests/scripts`. They would be irrelevant outside the testing environment of verified proxy so best to keep it local rather than grow a global scripts folder.

execution_chain/version.nim
```diff
@@ -1,30 +1,30 @@
...
-
-  NimbusMinor* = 1
-  ## is the minor number of Nimbus' version.
-
-  NimbusPatch* = 0
-  ## is the patch number of Nimbus' version.
-
-  NimbusVersion* = $NimbusMajor & "." & $NimbusMinor & "." & $NimbusPatch
-  ## is the version of Nimbus as a string.
+# Copyright (c) 2025 Status Research & Development GmbH
```
tersec: What's going on with these `execution_chain/version.nim` changes?

chirag-parmar: not really sure. Working on resolving it atm. My guess is that it is some whitespace issue

chirag-parmar: The file had the Windows CR(carriage return) on every line. Removing those caused the diff

nimbus_verified_proxy/tests/test_state.nim
```diff
@@ -0,0 +1,86 @@
...
+
+    # load proof also loads the block
+    ts.loadProof(address, @[slot], blk, proof)
+    # upload the same proof to resolve for accounts
+    ts.loadProof(address, @[], blk, proof)
+    ts.loadCode(address, blk, contractCode)
+    # this is for optimistic state fetch
+    ts.loadAccessList(tx, blk, accessList)
+
+    discard vp.headerStore.add(convHeader(blk), blk.hash)
```
tersec: It would be better to check that these two `discard`s are both `isOk` rather than `discard` the results.

chirag-parmar: the test would fail to execute if the `add` function fails. But I get what you mean, will add the checks

tersec: Sure, but it's also to test the return value of `add` and `finalizeStatus`

tersec (approved):

nimbus_verified_proxy/tests/test_state.nim
chirag-parmar: oh good catch. I found an inconsistency in the definition of `add` methods. It is supposed to return `Result[void, string]` but it is still `Result[bool, string]`. Opening another PR for this

chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3614

PR #3593, opened 2025-08-27
Discovery: Filter out non 'eth' nodes and regularly update 'eth' field.

### LINKED ISSUE #3557: Nimbus EL spam Lodestar with failed connection attempts
Copied from discord:

Lodestar is noticing that during peer discovery with discv5 enabled, Nimbus EL nodes spam them with failed connection attempts. I'm not sure if there's a better way but if nothing else, checking for the https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.4/specs/phase0/p2p-interface.md#eth2-field ENR field and filtering out those peers would avoid this.

I've looked into Geth's source code and can't find any obvious logic around avoiding this, so presumably there is a better way
It would be better to use allowlists/allowconditions for this kind of thing, rather than excluding specific fingerprints (eth2)
If nothing else, I'm fairly sure Geth isn't checking for that ENR field

jangko:
This should fix #3557 

The async worker will regularly update `localNode` enr with `("eth", forkId)` every 15 secs if there is changes in the `forkId`.
It will make other easier for other node to peer with nimbus-eth1.

And the returned node by discovery V5 also get filtered by "eth" field and compatible `forkId`.
This will filter out e.g. `eth2` nodes,  and nodes with different chain in advance before actually using rlpx to connect them.

### DIFF
--- a/execution_chain/common/hardforks.nim
+++ b/execution_chain/common/hardforks.nim
@@ -343,6 +343,7 @@ type
     byBlock: seq[uint64]
     byTime: seq[uint64]
     genesisCRC: uint32
+    cache: seq[ForkID]
 
 func newID*(calc: ForkIdCalculator, head, time: uint64): ForkID =
   var hash = calc.genesisCRC
@@ -361,6 +362,24 @@ func newID*(calc: ForkIdCalculator, head, time: uint64): ForkID =
     return (hash, fork)
 
   (hash, 0'u64)
+
+func compatible*(calc: var ForkIdCalculator, forkId: ForkID): bool =
+  if calc.cache.len == 0:
+    calc.cache = newSeqOfCap[ForkID](calc.byBlock.len + calc.byTime.len)
+    var hash = calc.genesisCRC
+    for fork in calc.byBlock:
+      hash = crc32(hash, fork.toBytesBE)
+      calc.cache.add( (hash, fork) )
+
+    for fork in calc.byTime:
+      hash = crc32(hash, fork.toBytesBE)
+      calc.cache.add( (hash, fork) )
+
+  for id in calc.cache:
+    if id == forkId:
+      return true
+
+  false
 
 func initForkIdCalculator*(map: ForkTransitionTable,
                            genesisCRC: uint32,
--- a/execution_chain/networking/eth1_discovery.nim
+++ b/execution_chain/networking/eth1_discovery.nim
@@ -15,6 +15,7 @@ import
   chronicles,
   results,
   metrics,
+  eth/common/base_rlp,
   ./discoveryv5,
   ./discoveryv4,
   ./eth1_enr
@@ -37,9 +38,14 @@ type
   AddressV4 = discoveryv4.Address
   AddressV5 = discoveryv5.Address
 
+  UpdaterHook* = object
+    forkId*: proc(): ForkID {.noSideEffect, raises: [].}
+    compatibleForkId*: proc(id: ForkID): bool {.noSideEffect, raises: [].}
+
   Eth1Discovery* = ref object
     discv4: DiscV4
     discv5: DiscV5
+    hook: UpdaterHook
 
 #------------------------------------------------------------------------------
 # Private functions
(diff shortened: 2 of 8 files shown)


execution_chain/networking/eth1_discovery.nim
```diff
@@ -37,9 +38,14 @@ type
...
   AddressV5 = discoveryv5.Address
 
+  UpdaterHook* = object
+    forkId*: proc(): ForkID {.noSideEffect, raises: [].}
+    compatibleForkId*: proc(id: ForkID): bool {.noSideEffect, raises: [].}
+
   Eth1Discovery* = ref object
     discv4: DiscV4
     discv5: DiscV5
+    hook: UpdaterHook
```
kdeme: I would rename this to indicate better that it provides the fork id context for filtering. Or at least add something forkId or ENR related to the name.

Also, because within `eth1_discovery` it is not even really used for the ENR update. 

And `updateForkID` does not access it (`ForkID` is passed as parameter), so the hook is mostly used in ENR update functionality in the `peer_pool` code.

kdeme (commented):
looks mostly good to me, small comment.

I did wonder, but did not look this up in the code, if there is no existing fork id update code where this ENR update can be easily pushed from.

execution_chain/networking/eth1_discovery.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/3593/commits/d08db06b4975e8baf7e26bea084c1ab97be473da

jangko:
> I did wonder, but did not look this up in the code, if there is no existing fork id update code where this ENR update can be easily pushed from.

No easy place to update the ENR. Other places such as Engine API newPayload will create problem with test code. Put it in p2p code is less intrusive.

execution_chain/common/hardforks.nim
```diff
@@ -362,6 +363,24 @@ func newID*(calc: ForkIdCalculator, head, time: uint64): ForkID =
 
   (hash, 0'u64)
 
+func compatible*(calc: var ForkIdCalculator, forkId: ForkID): bool =
+  if calc.cache.len == 0:
+    calc.cache = newSeqOfCap[ForkID](calc.byBlock.len + calc.byTime.len)
+    var hash = calc.genesisCRC
```
tersec: `hash` has special meaning within Nim, in theory shouldn't cause an issue here, but might be clearer to choose a variable name that's not literally `hash`.

tersec (approved):

tersec:
Kurtosis failures look unrelated, but it'd still be nicer if they ran;
```
Err:32 https://apt.fury.io/kurtosis-tech  InRelease
  Could not connect to apt.fury.io:443 (13.248.213.92), connection timed out Could not connect to apt.fury.io:443 (15.197.149.68), connection timed out Could not connect to apt.fury.io:443 (76.223.57.73), connection timed out Could not connect to apt.fury.io:443 (3.33.241.96), connection timed out
Fetched 8332 kB in 39s (216 kB/s)
Reading package lists...
Building dependency tree...
Reading state information...
W: Failed to fetch https://apt.fury.io/kurtosis-tech/InRelease  Could not connect to apt.fury.io:443 (13.248.213.92), connection timed out Could not connect to apt.fury.io:443 (15.197.149.68), connection timed out Could not connect to apt.fury.io:443 (76.223.57.73), connection timed out Could not connect to apt.fury.io:443 (3.33.241.96), connection timed out
8 packages can be upgraded. Run 'apt list --upgradable' to see them.
W: Some index files failed to download. They have been ignored, or old ones used instead.

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package kurtosis-cli
```

execution_chain/common/hardforks.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/3593/commits/5b550a9a9e5acff56451c4f93f967f8c5cae7700

PR #3595, opened 2025-08-27
bump nimbus-eth2 and remove old fulu field in Portal beacon test

kdeme:
Those old beacon tests only test with altair test data and the digest fields are arbitrary.


arnetheduck:
also need to bump all other vendor deps to (at least) the same version as eth2

PR #3596, opened 2025-08-27
Update data directory defaults to include namespace, network

arnetheduck:
Align with eth2 to enable the safe use of a shared data directory between eth1/2 and other nimbus, when using defaults.

This also updates the execution client to ensure filesystem permissions on the data directory, similar to the beacon node.

### DIFF
--- a/execution_chain/config.nim
+++ b/execution_chain/config.nim
@@ -885,20 +885,11 @@ func dbOptions*(conf: NimbusConf, noKeyCache = false): DbOptions =
 proc makeConfig*(cmdLine = commandLineParams()): NimbusConf
     {.raises: [CatchableError].} =
   ## Note: this function is not gc-safe
-
-  # The try/catch clause can go away when `load()` is clean
-  try:
-    {.push warning[ProveInit]: off.}
-    result = NimbusConf.load(
-      cmdLine,
-      version = NimbusBuild,
-      copyrightBanner = NimbusHeader
-    )
-    {.pop.}
-  except CatchableError as e:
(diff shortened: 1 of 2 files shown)


nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
```diff
@@ -38,31 +25,37 @@ type
...
   # Storage
-  dataDir* {.
-    desc: "The directory where nimbus_verified_proxy will store all data",
-    defaultValue: defaultVerifiedProxyDataDir(),
-    defaultValueDesc: $defaultDataVerifiedProxyDirDesc,
-    abbr: "d",
-    name: "data-dir"
-  .}: OutDir
+  dataDirFlag* {.
+    desc: "The directory where nimbus will store all blockchain data"
```
kdeme, line 50: Is it intentionally called general nimbus here over nimbus_verified_proxy?

kdeme (approved):

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
arnetheduck: oh, uh ... of course! I did not copy-paste this from somewhere else at all :)

that said, I think it's fine for the purpose.. 

that's funny about `{..}` btw, that you can omit the `,` but nph will put them back - if one project uses nph and the other doesn't, it's harder to compare configs .. https://github.com/status-im/nim-confutils/issues/114 would be nice here

PR #3606, opened 2025-08-29
txFrame HashKeys (kMap) memory optimization 

bhartnett:
This PR adds a flag to `txFrameBegin` which enables moving the kMap from the parent to the new frame. In the ForkedChain we always move the kMap unless the `block number = block expected to be persisted + 1` or the `persistBatchSize < 2`.

This should have the effect of keeping the kMap entries only for the txFrames of the blocks which will be persisted. These blocks benefit the most from having the kMap entries available for the state root computation.

### DIFF
--- a/tests/test_aristo/test_tx_frame.nim
+++ b/tests/test_aristo/test_tx_frame.nim
@@ -108,3 +108,25 @@ suite "Aristo TxFrame":
       check:
         tx.fetchAccountRecord(acc1[0]).isOk()
         tx.fetchAccountRecord(acc2[0]).isErr() # Doesn't exist in tx2
+
+  test "Frames using moveParentHashKeys parameter":
+    let
+      tx0 = db.txFrameBegin(db.baseTxFrame())
+      tx1 = db.txFrameBegin(tx0)
+
+    check:
+      tx0.mergeAccountRecord(acc1[0], acc1[1]).isOk()
+      tx1.mergeAccountRecord(acc2[0], acc2[1]).isOk()
+      tx0.fetchStateRoot() != tx1.fetchStateRoot()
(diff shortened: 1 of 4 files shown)


tests/test_aristo/test_tx_frame.nim
```diff
@@ -108,3 +108,25 @@ suite "Aristo TxFrame":
...
+    check:
+      tx0.mergeAccountRecord(acc1[0], acc1[1]).isOk()
+      tx1.mergeAccountRecord(acc2[0], acc2[1]).isOk()
+      tx0.fetchStateRoot() != tx1.fetchStateRoot()
+      tx0.kMap.len() == 0
+      tx1.kMap.len() == 1
+
+    let tx2 = db.txFrameBegin(tx1, moveParentHashKeys = true)
+    check:
+      tx1.kMap.len() == 0
```
arnetheduck, line 128: a few more things to test: 

* that after moving, we still compute the same state roots on both moved-from and moved-to frame
* that after computing state root in moved-from block, the moved-to block, when persisted, uses the right keys (ie the ones from updates in its own frame and not the ones that were (re)computed as part of computing the moved-from state root - basically, that merging keys _again_ is correct
* that we (still) can persist moved-from frames and load them later on - they will have a "gappy" key table which is fine, ie the gaps will be filled in lazily

bhartnett:
> interesting, this came out a lot smaller a change than I expected ;)

Yeah, turned out to be a simple change.

tests/test_aristo/test_tx_frame.nim
bhartnett: Tests updated as suggested.

PR #3607, opened 2025-08-29
proxy: better error handling

chirag-parmar:
### Description:

This PR rectifies all the improper error handling in verified proxy. For this, it could be seen as a refactoring PR. It additionally also brings in new error definitions that will prove to be useful in the future.

### Notes:
1. Uses `Result` everywhere which is translated to exceptions only at the topmost entrypoint of the proxy. This makes sure the proxy can be imported as a library without the need to handle exceptions with `try-except` blocks.
2. Uses differentiated errors which will come in handy later for scoring endpoints.

### DIFF
--- a/nimbus_verified_proxy/nimbus_verified_proxy.nim
+++ b/nimbus_verified_proxy/nimbus_verified_proxy.nim
@@ -40,8 +40,9 @@ type ProxyError = object of CatchableError
 proc verifyChainId(
     engine: RpcVerificationEngine
 ): Future[void] {.async: (raises: [CancelledError]).} =
+  const ZERO_U256 = 0.u256 # pushed construction of zero to compile time
   let providerId = (await engine.backend.eth_chainId()).valueOr:
-    0.u256
+    ZERO_U256
 
   # This is a chain/network mismatch error between the Nimbus verified proxy and
   # the application using it. Fail fast to avoid misusage. The user must fix


kdeme (commented):
This PR really changes the current usage of `Result` into the direction of exceptions to deal with error handling in NVP.

While it does look logically correct (I didn't check everything), and exception handling has been improved a lot with usages of raises and the addition of async raise, we have been mostly going in the other direction in our projects. That is, towards the usage of `Result`.

For reasoning behind this see also https://status-im.github.io/nim-style-guide/errors.html

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -31,14 +31,17 @@ import
...
     engine: RpcVerificationEngine
-): Future[void] {.async: (raises: []).} =
-  let providerId =
-    try:
-      await engine.backend.eth_chainId()
-    except CatchableError:
-      0.u256
+): Future[void] {.async: (raises: [CancelledError]).} =
+  let providerId = (await engine.backend.eth_chainId()).valueOr:
+    0.u256
```
tersec: It'd be a bit more efficient to construct this at compiletime, either via `static(0.u256)` or if that doesn't work, use a `const` (which definitely will).

chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3607/commits/f4381a8d8d7daf3a1400b0083117cd3fd868c00e

PR #3608, opened 2025-08-29
Beacon sync maint update

mjfh:


execution_chain/config.nim
```diff
@@ -493,6 +493,13 @@ type
         defaultValueDesc: "\"jwt.hex\" in the data directory (see --data-dir)"
         name: "jwt-secret" .}: Option[InputFile]
 
+      beaconSyncInitPeersMin* {.
```
arnetheduck, line 496: why would you want to do this?

mjfh: Apart from hard coding for testing, for a given `base`, the only way to control both, the length of the syncer sprint and the min number of active peers is using this option together with the sync `target` option. Without this min-peer option, the `target` would typically fetch the full header list already before multi-peer mode is reached.

Using the CL to set the target, the multi-peer mode can be emulated by holding back (i.e. late start.) But here the length of the syncer sprint for given `base` is out of control for testing,

arnetheduck: > the full header list

wait, what? we necessarily must be re-examining the list of available peers on a regular basis and rebalance the pending requests regardless of how many of them were connected at the start - fixing this early on is not acceptable.

Imagine that we get connected to a single slow peer, and then to a second, fast peer - in this case, we should be requesting a fresh batch of headers from the second peer as soon as it connects and not let the first peer monopolise the full header download.

Basically, the requests for headers must continually be in a round-robin with some maximum number of headers per peer per request - the number of headers should come from an estimate of how fast the peer is starting with some fairly low number and increasing if the peer responds quickly, up to some fixed limit, similar to the basic TCP window size algorithm.

The need for this option sounds like an implementation gap that should be addressed by improving the algorithm rather than adding a workaround option.

PR #3609, opened 2025-08-29
NimbusVersion in one place for all version related code

jangko:
.nimble file, nimbus-eth1 binaries, docker built release files, etc now get the 'NimbusVersion' string from one place: 'version.nim'.

This PR also fix nightly build failure due to recent changes that import nim vm unfriendly stuff into version.nim

### DIFF
--- a/execution_chain/version_info.nim
+++ b/execution_chain/version_info.nim
@@ -1,0 +1,29 @@
+# Copyright (c) 2018-2025 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
+{.push raises: [].}
+
+import
+  std/[strformat],
+  stew/byteutils, ./compile_info, beacon_chain/buildinfo,
+  ./version
+
+export
+  version
+
+const
+  NimbusName* = "nimbus-eth1"
+  ## project name string
+
+  GitRevisionBytes* = hexToByteArray[4](GitRevision)
+
+  FullVersionStr* = "v" & NimbusVersion & "-" & GitRevision
+
+  ClientId* = &"{NimbusName}/{FullVersionStr}/{hostOS}-{hostCPU}/Nim-{NimVersion}"
+
+  ShortClientId* = NimbusName & "/" & FullVersionStr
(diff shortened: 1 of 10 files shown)


nimbus.nimble
```diff
@@ -7,8 +7,11 @@
 
 mode = ScriptMode.Verbose
 
+# Get the version from one location
+import ./execution_chain/version as ver
+
 packageName   = "nimbus"
-version       = "0.1.0"
+version       = NimbusVersion
```
arnetheduck: importing versions is not going to work with newer nimble versions, ie there's a new declarative parser that does not support this

execution_chain/version_info.nim
```diff
@@ -0,0 +1,29 @@
...
+  ./version
+
+export
+  version
+
+const
+  NimbusName* = "nimbus-eth1"
+  ## project name string
+
+  GitRevisionBytes* = hexToByteArray[4](GitRevision)
```
arnetheduck, line 25: this this the same number of bytes that we use in eth2? the two should be aligned

jangko: I don't know about nimbus-eth2 requirement. But this `GitRevisionBytes` is used in one of engine_api method, and the spec requires it must be 4 bytes.

jangko: https://github.com/ethereum/execution-apis/blob/main/src/engine/identification.md#clientversionv1

arnetheduck: ok - that link to the spec should be a code comment then, so someone doesn't change it by  accident

jangko: https://github.com/status-im/nimbus-eth1/pull/3609/commits/23ba32c7724c01b6334f5deb66438563faf8ddd6

nimbus.nimble
jangko: https://github.com/status-im/nimbus-eth1/pull/3609/commits/8326207621ee7db1bd393d35a92bcb3252b2a632

execution_chain/version_info.nim
```diff
@@ -0,0 +1,31 @@
...
+
+export
+  version
+
+const
+  NimbusName* = "nimbus-eth1"
+  ## project name string
+
+  # Please keep it 4 bytes long, used in `engine_ClientVersionV1`
+  # https://github.com/ethereum/execution-apis/blob/main/src/engine/identification.md#clientversionv1
```
tersec: Better to link to a specific commit, because this not a stable URL: https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/identification.md#clientversionv1

jangko: https://github.com/status-im/nimbus-eth1/pull/3609/commits/205e9639c6e31331afc444adb726a6e790eb08ae

PR #3611, opened 2025-09-01
Add support for TOML config file

### LINKED ISSUE #3591: TOML config file support
Similar to eth2, eth1 should learn to load configuration [from toml](https://github.com/status-im/nimbus-eth2/blob/92defe0fe6c67255cef3687514c1ba80b197f3b3/beacon_chain/nimbus_binary_common.nim#L220), preferably reusing makeBannerAndConfig.

jangko:
fix #3591


execution_chain/config.nim
```diff
@@ -885,18 +889,78 @@ func dbOptions*(conf: NimbusConf, noKeyCache = false): DbOptions =
     rdbPrintStats = conf.rdbPrintStats,
   )
 
+#-------------------------------------------------------------------
+# TOML serializer overloads of SecondarySources
+#-------------------------------------------------------------------
+
+proc readValue*(r: var TomlReader, val: var OutDir)
```
arnetheduck: these must be imported from confutils or they will conflict

jangko: https://github.com/status-im/nimbus-eth1/pull/3611/commits/e05f38bfdcf10324b30439f2e7ff50bf689604ca

execution_chain/config.nim
```diff
@@ -885,18 +890,70 @@ func dbOptions*(conf: NimbusConf, noKeyCache = false): DbOptions =
...
+#-------------------------------------------------------------------
+# TOML serializer overloads of SecondarySources
+#-------------------------------------------------------------------
+
+proc readValue*(r: var TomlReader, val: var NetworkParams)
+       {.gcsafe, raises: [IOError, SerializationError].} =
+  # Not actually parse it, only to silence compiler
+  discard r.parseAsString()
+
+proc readValue*(r: var TomlReader, val: var Port)
```
arnetheduck: also these ..

execution_chain/config.nim
```diff
@@ -885,18 +890,70 @@ func dbOptions*(conf: NimbusConf, noKeyCache = false): DbOptions =
...
+       {.gcsafe, raises: [IOError, SerializationError].} =
+  val = r.parseInt(int64).Port
+
+proc readValue*(r: var TomlReader, val: var IpAddress)
+       {.gcsafe, raises: [IOError, SerializationError].} =
+  try: val = parseIpAddress(r.parseAsString())
+  except ValueError as exc:
+    raise newException(SerializationError, exc.msg)
+
+proc readValue*(r: var TomlReader, val: var enr.Record)
```
arnetheduck: ...and these would ideally move to nim-eth because eth2 needs the same

execution_chain/config.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/3611/commits/fa1162ffa25d2a7f5bb7191b62ae369591bfddca

execution_chain/config.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/3611/commits/fa1162ffa25d2a7f5bb7191b62ae369591bfddca

tests/test_configuration.nim
```diff
@@ -9,14 +9,23 @@
...
   ../execution_chain/config,
   ../execution_chain/common/[chain_config, context, manager],
   ./test_helpers
 
+func `==`*(a, b: OutDir): bool =
+  a.string == b.string
+
+func `==`*(a, b: NatConfig): bool =
+  let size = min(sizeof(a), sizeof(b))
+  cmpMem(a.unsafeAddr, b.unsafeAddr, size) == 0
```
arnetheduck: this is broadly not safe, since there may be padding in the object (now or in the future) that leaves a few bytes undefined - specially in case objects that have more complex alignment rules

jangko: https://github.com/status-im/nimbus-eth1/pull/3611/commits/1da4f71431ddc966c0b1f91666227506414a3bae

jangko: https://github.com/status-im/nimbus-eth1/pull/3611/commits/1da4f71431ddc966c0b1f91666227506414a3bae

arnetheduck:
regarding NetworkParams, instead of assigning them to the config object, an option is to remove them entirely from `NimbusConf` and instead have a function that returns them as a fresh instance - this is probably the better long-term option since then, there is no special "update-the-config-object-after-parsing" to be done - this will also make it easier to compose config objects.

PR #3613, opened 2025-09-01
proxy: fee-based methods

chirag-parmar:
- [x] implement functionality for `eth_gasPrice`, `eth_maxPriorityFeePerGas` and `eth_blobBaseFee`
- [x] write tests to test the implemented functionality

### DIFF
--- a/execution_chain/utils/utils.nim
+++ b/execution_chain/utils/utils.nim
@@ -78,7 +78,7 @@ func sumHash*(hashes: varargs[Hash32]): Hash32 =
   ctx.finish result.data
   ctx.clear()
 
-proc sumHash*(body: BlockBody): Hash32 {.gcsafe, raises: []} =
+proc sumHash*(body: BlockBody): Hash32 {.gcsafe, raises: [].} =
   let txRoot = calcTxRoot(body.transactions)
   let ommersHash = keccak256(rlp.encode(body.uncles))
   let wdRoot = if body.withdrawals.isSome:
--- a/nimbus_verified_proxy/rpc/fees.nim
+++ b/nimbus_verified_proxy/rpc/fees.nim
@@ -18,14 +18,16 @@ import
   ./blocks,
   ./transactions
 
+func avgWoOverflow(a: uint64, b: uint64): uint64 =
+  return a div 2 + b div 2 + ((a mod 2 + b mod 2) div 2)
+
 func median(prices: var openArray[GasInt]): GasInt =
   if prices.len > 0:
     sort(prices)
     let middle = prices.len div 2
     if prices.len mod 2 == 0:
       # prevent overflow
-      let addedAmt = prices[middle].uint64 + prices[middle - 1].uint64
-      return (addedAmt div 2).GasInt
+      return avgWoOverflow(prices[middle], prices[middle - 1]).GasInt
     else:
       return prices[middle]
 
--- a/nimbus_verified_proxy/rpc/fees.nim
+++ b/nimbus_verified_proxy/rpc/fees.nim
@@ -16,7 +16,7 @@ import
   ../../execution_chain/beacon/web3_eth_conv,
   ../types,
   ./blocks,
-  ./transactions.nim
+  ./transactions
 
 func median(prices: var openArray[GasInt]): GasInt =
   if prices.len > 0:
--- a/nimbus_verified_proxy/rpc/fees.nim
+++ b/nimbus_verified_proxy/rpc/fees.nim
@@ -12,7 +12,7 @@ import
   chronos,
   eth/common/eth_types_rlp,
   web3/[eth_api_types, eth_api],
-  std/[sequtils, algorithm],
+  std/algorithm,
   ../../execution_chain/beacon/web3_eth_conv,
   ../types,
   ./blocks,
--- a/nimbus_verified_proxy/rpc/fees.nim
+++ b/nimbus_verified_proxy/rpc/fees.nim
@@ -47,18 +47,19 @@ proc suggestGasPrice*(vp: VerifiedRpcProxy): Future[Result[GasInt, string]] {.as
 
   ok(max(minGasPrice, median(prices)))
 
-proc suggestMaxPriorityGasPrice*(vp: VerifiedRpcProxy): Future[Result[GasInt, string]] {.async.} =
+proc suggestMaxPriorityGasPrice*(
+    vp: VerifiedRpcProxy
+): Future[Result[GasInt, string]] {.async.} =
   let
     blk = (await vp.getBlock(blockId("latest"), true)).valueOr:
       return err(error)
     txs = blk.transactions.toTransactions().valueOr:
       return err(error)
 
-  var prices  = newSeqOfCap[GasInt](64)
+  var prices = newSeqOfCap[GasInt](64)
 
   for tx in txs:
     if tx.maxPriorityFeePerGas > GasInt(0):
       prices.add(tx.maxPriorityFeePerGas)
 
   ok(median(prices))
-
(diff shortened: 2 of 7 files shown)


execution_chain/utils/utils.nim
```diff
@@ -78,7 +78,7 @@ func sumHash*(hashes: varargs[Hash32]): Hash32 =
   ctx.finish result.data
   ctx.clear()
 
-proc sumHash*(body: BlockBody): Hash32 {.gcsafe, raises: []} =
+proc sumHash*(body: BlockBody): Hash32 {.gcsafe, raises: [].} =
```
tersec, line 81: Why does this have to specify `raises: []` again?

```nim
{.push raises: [].}
```
is already at the beginning of the module.

execution_chain/utils/utils.nim
```diff
@@ -78,7 +78,7 @@ func sumHash*(hashes: varargs[Hash32]): Hash32 =
...
   let txRoot = calcTxRoot(body.transactions)
   let ommersHash = keccak256(rlp.encode(body.uncles))
   let wdRoot = if body.withdrawals.isSome:
```
tersec, line 84: I know this isn't what's being changed by this PR, mainly, but if cleaning up the pragma syntax, then I'd also note that
```nim
  let wdRoot = if body.withdrawals.isSome:
    calcWithdrawalsRoot(body.withdrawals.get)
  else: EMPTY_ROOT_HASH
```
is fine, sort of, but it does the `isSome`/`get` pattern, which is only runtime-safe, while:
```nim
  let wdRoot = body.withdrawals.valueOr:
    EMPTY_ROOT_HASH
```
is shorter, and avoids the explicit `.get`. Of course, `.valueOr` is implemented in a broadly similar way, but it can be checked for bugs in a centralized way.

The other
```nim
proc sumHash*(header: Header): Hash32 =
  let wdRoot = if header.withdrawalsRoot.isSome:
                 header.withdrawalsRoot.get
               else: EMPTY_ROOT_HASH
  sumHash(header.txRoot, header.ommersHash, wdRoot)
```
has the same pattern and could be improved in the same way.

nimbus_verified_proxy/rpc/rpc_eth_api.nim
```diff
@@ -356,6 +359,41 @@ proc installEthApiHandlers*(vp: VerifiedRpcProxy) =
...
+
+    if header.blobGasUsed.isNone():
+      raise newException(ValueError, "blobGasUsed missing from latest header")
+    if header.excessBlobGas.isNone():
+      raise newException(ValueError, "excessBlobGas missing from latest header")
+    let blobBaseFee =
+      getBlobBaseFee(header.excessBlobGas.get, com, com.toEVMFork(header)) *
+      header.blobGasUsed.get.u256
+    if blobBaseFee > high(uint64).u256:
+      raise newException(ValueError, "blobBaseFee is bigger than uint64.max")
```
tersec, line 382: In Ethereum terms, it doesn't fit in a `Quantity`, which is an official Ethereum spec type. `uint64` is, while it's how `Quantity` is defined, incidental here.

execution_chain/utils/utils.nim
chirag-parmar: it would be better if we make this change in a seperate PR, no? it would keep this organized

nimbus_verified_proxy/rpc/rpc_eth_api.nim
chirag-parmar: why does it not? the base fee can take value upto 1ETH per gas as 64bit unsigned integer

chirag-parmar: https://github.com/ethereum/go-ethereum/pull/29140/files#diff-9886da3412b43831145f62cec6e895eb3613a175b945e5b026543b7463454603R369

chirag-parmar: ~I'll change it to `UInt256`~. 

Well `nim-web3` defines it as a `Quantity` looks like the changes has to be done in `nim-web3` https://github.com/status-im/nim-web3/blob/master/web3/eth_api.nim#L34

nimbus_verified_proxy/rpc/fees.nim
```diff
@@ -0,0 +1,65 @@
...
+  ../types,
+  ./blocks,
+  ./transactions
+
+func median(prices: var openArray[GasInt]): GasInt =
+  if prices.len > 0:
+    sort(prices)
+    let middle = prices.len div 2
+    if prices.len mod 2 == 0:
+      # prevent overflow
```
tersec, line 29: How does this prevent overflow? If's the straightforward `(x+y)/2` method, not the variation along the lines of `min(x,y) + (max(x,y)-min(x,y)) div 2` (probably there are ways to write this to exploit not requiring `min`/`max`).

execution_chain/utils/utils.nim
tersec: Sure, yeah.

nimbus_verified_proxy/rpc/rpc_eth_api.nim
tersec: It's likely the web3 API defines it that way.

nimbus_verified_proxy/rpc/fees.nim
chirag-parmar: good catch, changed it

nimbus_verified_proxy/rpc/rpc_eth_api.nim
tersec: > why does it not? the base fee can take value upto 1ETH per gas as 64bit unsigned integer

My concern is the error message, not the logic, i.e. the user-facing verbiage. `uint64` is a programming artifact specific to Nimbus. Here, the idea of a 64-bit `Quantity` is something Ethereum defines and is something which is explainable to an Ethereum user.

If, it turns out, that the web3 API seems not to define the actual return value on this, that's fine. But it's also worth checking, yeah, what other ELs do in their actual `eth_blobBaseFee` implementations. The Geth link is to an internal function, not what they're willing to return to web3 API users.

chirag-parmar: https://github.com/status-im/nim-web3/issues/222

chirag-parmar: > My concern is the error message, not the logic, i.e. the user-facing verbiage. uint64 is a programming artifact specific to Nimbus. Here, the idea of a 64-bit Quantity is something Ethereum defines and is something which is explainable to an Ethereum user

Ah ok gotcha.

> If, it turns out, that the web3 API seems not to define the actual return value on this, that's fine.

the official spec defines the return value as only a hex string

I'll have a look at other ELs and change the error message

execution_chain/utils/utils.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/issues/3618

execution_chain/utils/utils.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/issues/3618

nimbus_verified_proxy/rpc/rpc_eth_api.nim
chirag-parmar: both geth and reth return `UInt256` serialized into a hex string. We oughta do the same thing. Opened an issue https://github.com/status-im/nim-web3/issues/222. I think it would be better if we merge this PR first and then I move on to that issue. Resolving that issue might entail more work outside the scope of verified proxy and would block this PR. @tersec what do you think?

tersec: Agree, yeah.

PR #3619, opened 2025-09-03
Beacon sync remove some test options

mjfh:


execution_chain/sync/beacon.nim
```diff
@@ -25,6 +25,10 @@ export
 logScope:
   topics = "beacon sync"
 
+var beaconSyncConfigHook = BeaconSyncConfigHook(nil)
```
arnetheduck, line 28: globals make the code not gc-safe which in turn makes the code unusable when multithreading - this needs another solution

PR #3632, opened 2025-09-05
CI: extract tag

chirag-parmar:
The release script for github actions had a `Extract branch name` step that expected branch names in the format `refs/heads/<branch_name>` but when used on tags the prefix is `refs/tags/<tag_name>`. This causes an invalid directory error.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,7 +52,9 @@ jobs:
 
       - name: Extract branch name
         id: extract_branch
-        run: echo "branch_name=${GITHUB_REF#refs/heads/}" >> $GITHUB_OUTPUT
+        run: | 
+          ref_name=$(echo "$GITHUB_REF" | sed -e 's#^refs/heads/##' -e 's#^refs/tags/##')
+          echo "branch_name=$ref_name" >> $GITHUB_OUTPUT
 
       - name: Build project
         id: make_dist
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,9 +52,7 @@ jobs:
 
       - name: Extract branch name
         id: extract_branch
-        run: | 
-          ref_name=$(echo "$GITHUB_REF" | sed -e 's#^refs/heads/##' -e 's#^refs/tags/##')
-          echo "branch_name=$ref_name" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


jangko:
Lgtm. But I wonder should we also check the tag to match the binary version. Because it looks weird if the release tarball have different version compared to released binary. Probably we to automate the release procedure too because it involves 3 step. 1 increment binary version. 2 nimble package version. 3 git tag version. And that is error prone

chirag-parmar:
> Probably we need to automate the release procedure too because it involves 3 step. 1 increment binary version. 2 nimble package version. 3 git tag version. And that is error prone if done manually.

Agreed.

Also something that is bothering me is that we are rebuilding for release despite normal build workflows running. Is there a reasons for this? can we not just upload artefacts from the respective build jobs and the release workflow just downloads and packages them?

tersec:
> Also something that is bothering me is that we are rebuilding for release despite normal build workflows running. Is there a reason for this? can we not just upload artefacts from the respective build jobs and the release workflow just downloads and packages them?

One thing this enables is potentially faster releases, because it just doesn't have to do as much as the normal build workflows. It also avoids questions of, is the cache/artefacts available, et cetera, or defining lifetimes of them which for 99% of builds won't be relevant but for specific tagged builds will be.

It's possible but Nimbus is supposed to have reproducible builds, so it should be an optimization, not a correctness point, and an optimization that creates potentially fraught failure modes due to asynchrony with other build jobs, or theoretically arbitrary delays (longer than artefacts last? who knows) after a commit is done but before it's tagged, at which it's possible there won't be any artefacts to grab.

> 1 increment binary version. 2 nimble package version. 3 git tag version. And that is error prone if done manually.

Well, git tags are an external artifact which isn't even immutable; one can, in technical terms in git, social conventions aside, freely add/remove git tags. So I'm not sure if it's possible to inextricably intertwine them with (1)/(2).

For (1)/(2), it should be possible to derive them from a single source of truth, yes.

advaita-saha:
This PR looks good to me
Specially considering 
> whatever it is, it should be exactly the same as nimbus-eth2 (regardless if it takes adjustments to eth2 or eth1 or both)

Maybe we can automate this process, but for now I think we should go with this
And later maybe next time update the release process for both eth1 and eth2, so that there is no difference in between them

For now let's get this merged and tested

PR #3634, opened 2025-09-05
proxy: release CI

chirag-parmar:


.github/workflows/nimbus_verified_proxy.yml
```diff
@@ -88,6 +90,25 @@ jobs:
...
+          if-no-files-found: warn
+ 
+      - name: Upload release checksum artifacts
+        if: success() && startsWith(github.ref, 'refs/tags/')
+        uses: actions/upload-artifact@v4
+        with:
+          name: proxy-${{ matrix.os }}-${{ matrix.cpu }}-checksum
+          path: |
+            build/nimbus_verified_proxy.sha512sum
+          if-no-files-found: warn
```
advaita-saha: this means the release will misbehave when the current verified proxy CI hasn't finished it's build?
i.e release might use stale artifacts

.github/workflows/release.yml
```diff
@@ -133,9 +133,22 @@ jobs:
           cat linux-arm64-checksum/* >> release_notes.md
           echo '# Windows AMD64' >> release_notes.md
           cat windows-amd64-checksum/* >> release_notes.md
-          echo '# macos ARM64' >> release_notes.md
+          echo '# MacOS ARM64' >> release_notes.md
```
tersec: Official capitalization is `macOS`

.github/workflows/release.yml
```diff
@@ -133,9 +133,22 @@ jobs:
...
+          echo '## Verified Proxy SHA512 checksums' >> release_notes.md
+          echo '' >> release_notes.md
+          echo '```' >> release_notes.md
+          echo '# Linux AMD64' >> release_notes.md
+          cat proxy-linux-amd64-checksum/* >> release_notes.md
+          echo '# Linux ARM64' >> release_notes.md
+          cat proxy-linux-arm64-checksum/* >> release_notes.md
+          echo '# Windows AMD64' >> release_notes.md
+          cat proxy-windows-amd64-checksum/* >> release_notes.md
+          echo '# MacOS ARM64' >> release_notes.md
```
tersec: also `macOS`.

.github/workflows/nimbus_verified_proxy.yml
chirag-parmar: Ah right, I just realised github workflows are not pipelined. Should I go ahead and pipeline the workflows or should I just rebuild the verified proxy for the release job(rebuilding seems wasteful)?

kdeme: > Ah right, I just realised github workflows are not pipelined. Should I go ahead and pipeline the workflows or should I just rebuild the verified proxy for the release job(rebuilding seems wasteful)?

Whatever gets selected here, please also think ahead in the sense that nimbus_portal_client will also need to be added to the release binaries eventually

PR #3642, opened 2025-09-08
Beacon sync trace replay test tools

mjfh:


tools/syncer/syncer_test_client_replay.nim
```diff
@@ -0,0 +1,147 @@
...
+    fileName: string    # capture file name
+    noStopQuit: bool    # capture modifier argument
+    fakeImport: bool    # ditto
+    syncTicker: bool    # ..
+
+let
+  cmdName = getAppFilename().extractFilename()
+
+# ------------------------------------------------------------------------------
+# Private helpers, command line parsing tools
```
arnetheduck, line 51: we already have a command line parsing tool in the form of confutils - why not use that here? We want to avoid these custom ways of doing things as much as possible as they increase maintenance overhead over time.

tools/syncer/helpers/nimbus_el_wrapper.nim
```diff
@@ -0,0 +1,37 @@
...
+#    http://www.apache.org/licenses/LICENSE-2.0)
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT) or
+#    http://opensource.org/licenses/MIT)
+# at your option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+## Wrapper to expose `run()` from `nimbus_execution_client.nim` without
+## marking is exportable.
+
+include # (!)
```
arnetheduck: instead of this hack, the regular nimbus runner should be refactored in such a way that it doesn't need to be included - ie running nimbus as a library is a reasonable feature that is also useful for the common nimbus executable.

mjfh: would be enough to make `run()` public

arnetheduck: well, that is the minimum viable approach indeed, though it's worth considering whether that is a reasonable approach from a maintenance perspective.

tools/syncer/replay/replay_reader/reader_gunzip.nim
```diff
@@ -0,0 +1,352 @@
...
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or
+#    http://www.apache.org/licenses/LICENSE-2.0)
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT) or
+#    http://opensource.org/licenses/MIT)
+# at your option. This file may not be copied, modified, or distributed except
+# according to those terms.
+
+## Incremental unzip based on `Stream` input (derived from
+## `test/replay/unzip.nim`.)
```
arnetheduck: instead of adding yet another copy-pasted implementation of gunzip, why not turn this into a library proper? or, much better from a maintenance perspective, _reuse_ our existing compression infrastructure such as `snappy` which is used in both era files and other places.

tools/syncer/replay/replay_reader/reader_reclog.nim
```diff
@@ -0,0 +1,352 @@
...
+# -----------
+
+func toStrSeq(n: int; w: TraceSyncActvFailed): seq[string] =
+  var res = newSeqOfCap[string](15)
+  res.addX("=ActvFailed", n, w)
+  res.add "base=" & w.baseNum.bnStr
+  res.add "latest=" & w.latestNum.bnStr
+  res
+
+func toStrSeq(n: int; w: TraceSyncActivated): seq[string] =
```
arnetheduck: why do we need a new format for this file? why not reuse our existing infrastructure, for example using json_serialization which is interoperable with other tooling and requires much less maintenance?

mjfh: Would be the right time to try that.

tools/syncer/replay/replay_reader/reader_gunzip.nim
mjfh: The original gzip wrapper was buggy. But this might have improved tough. I'd rather have a look at snappy which should be able to compress records while writing.

arnetheduck: well, each such wrapper adds maintenance overhead, _specially_ if it's buggy, because when making unrelated changes, the tooling code must still be maintained by others not familiar with it - in tools that are not used on a regular basis but are rather auxiliary, it's important to focus on reusing existing infrastructure as much as possible since the cost of maintaining this code otherwise becomes significant. 

If it turns out in the end that gzip is the right format to use here (because it can be motivated by for example some existing tooling using gzip in particular and not something else), the right approach is to _fix the existing gzip reader_ and extract it to a reusable  and isolated component - not introduce another copy that potentially is also buggy but in different ways.

This means reusing existing file formats, avoid using custom compression, logging or command line parsing, reusing standards like era and/or json as much as is feasible, and not adding large swathes of copy-pasted and/or otherwise hacky code that someone else down the line will have to understand, in order to keep compatible.

mjfh: OK, I will move the gunzip code to the `nim-zlib` pkg. I see now that the main difference to the `nim-zlib` provided gzip/gunzip tools is that they only work on a single block of data. The version I need (and use) inflates incrementally for an input of `Stream` type.

PR #3649, opened 2025-09-08
Fix invalid result error access + move empty txs check

kdeme:
Check for empty txs in getBlockTransactionData and getExecutionPayloadBodyV1.

First step as more locations potentially require this on usage of also getBlockTransactions and getBlockTransactionHashes.

Can also add additionally check for rootHash calc.

### DIFF
--- a/execution_chain/db/core_db/core_apps.nim
+++ b/execution_chain/db/core_db/core_apps.nim
@@ -402,6 +402,11 @@ proc getTransactions*(
     var res: seq[Transaction]
     for encodedTx in db.getBlockTransactionData(txRoot):
       res.add(rlp.decode(encodedTx, Transaction))
+
+    # Txs not there in db - Happens during era1/era import, when we don't store txs and receipts
+    if (res.len == 0 and txRoot != zeroHash32):
+      return err("No transactions found in db for txRoot " & $txRoot)
+
     return ok(move(res))
 
 proc getBlockBody*(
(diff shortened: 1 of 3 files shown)


execution_chain/db/core_db/core_apps.nim
```diff
@@ -402,6 +402,11 @@ proc getTransactions*(
     var res: seq[Transaction]
     for encodedTx in db.getBlockTransactionData(txRoot):
       res.add(rlp.decode(encodedTx, Transaction))
+
+    # Txs not there in db - Happens during era1/era import, when we don't store txs and receipts
```
arnetheduck, line 406: I guess the idea here is that they can be fetched from the era file at some point.

execution_chain/db/payload_body_db.nim
```diff
@@ -48,6 +48,10 @@ proc getExecutionPayloadBodyV1*(
   for encodedTx in db.getBlockTransactionData(header.txRoot):
     body.transactions.add TypedTransaction(encodedTx)
 
+  # Txs not there in db - Happens during era1/era import, when we don't store txs and receipts
```
arnetheduck, line 51: it will also happen when we prune the db on an ongoing basis and leave only the headers

execution_chain/db/core_db/core_apps.nim
kdeme: Yes, my original suggestion on pruning was rather to have the pruning block number stored somewhere and go fetch data below that number immediately from era or Portal. Instead of trying on the db first.

Perhaps we can still go with that approach.

execution_chain/db/payload_body_db.nim
kdeme: yes, see also https://github.com/status-im/nimbus-eth1/pull/3649#discussion_r2348684264

PR #3657, opened 2025-09-10
--bootstrap-nodes and --static-peers now accept multi formats

### LINKED ISSUE #3644: Empty bootstrap nodes on startup
```
DBG 2025-09-08 13:55:23.178+02:00 Added bootstrap node                       topics="eth p2p discv5" uri=enr:
```

jangko:
Both of them now accept enr, enode, and path to file contains enr/enode.
The file format can be simple list or EF bootnodes yaml subset.
Also solve empty bootstrap nodes on startup issue 3644.
Each enodes will go to discovery v4 and enrs go to discovery v5.

fix #3644


execution_chain/networking/bootnodes.nim
```diff
@@ -1,47 +1,166 @@
...
+  for line in splitLines(text):
+    if line.startsWith("- enr:") or
+       line.startsWith("- enode:"):
+      result[i] = line.substr(2)
+      inc i
+
+    if line.startsWith("enr:") or
+       line.startsWith("enode:"):
+      result[i] = line
+      inc i
```
kdeme: From what I've seen in eth-clients repos, ENR and Enode strings are always in separate files.

With that in mind, why bring the files together here to then later split them again? Makes more sense to me to keep the parsing separate. This would also make removal of Enodes easier if/when they get deprecated.

execution_chain/networking/bootnodes.nim
```diff
@@ -1,47 +1,166 @@
...
+{.push raises: [].}
+
+import
+  std/[strutils, os],
+  stew/[io2, base64],
+  eth/p2p/discoveryv5/enr,
+  ./discoveryv4/enode
+
+type
+  Bootnodes* = object
```
kdeme: We seem to be calling this often BootstrapNodes and Bootnodes. 

Considering the eth-clients files are called `bootstrap_nodes.yaml` and our config options also bootstrap, perhaps name it everywhere bootstrap nodes.

execution_chain/config.nim
```diff
@@ -221,44 +221,26 @@ type
 
     bootstrapNodes {.
       separator: "\pNETWORKING OPTIONS:"
-      desc: "Specifies one or more bootstrap nodes(as enode URL) to use when connecting to the network"
+      desc: "Specifies one or more bootstrap nodes to use when connecting to the network"
+      longDesc: "Accepts the following formats or combination\n" &
+        "- enr\n" &
+        "- enode\n" &
+        "- Path to file contains enr or enode(plain list or yaml format)"
```
kdeme: Supporting both ENR and Enode here does sound like the best way to support this with decent UI.

But it might be a good idea to remain consistent with CL and portal node options where the file remains a separate flag.

execution_chain/networking/bootnodes.nim
kdeme: I guess part of the reason for this is the current bootstrapNodes flag which allows for any of the files.

tests/networking/test_enode.nim
```diff
@@ -94,13 +94,21 @@ suite "ENode":
...
-    func runBNTest(bns: openArray[string]): bool =
-      for z in bns:
-        let res = ENode.fromString(z)
-        if res.isErr: return false
-      true
+    var boot: Bootnodes
+    check getBootnodes("mainnet", boot) == Result[void, string].ok()
+    check getBootnodes("holesky", boot) == Result[void, string].ok()
+    check getBootnodes("sepolia", boot) == Result[void, string].ok()
+    check getBootnodes("hoodi", boot) == Result[void, string].ok()
```
kdeme: Can just do `isOk()` instead of this more verbose version?

kdeme (commented):
I general I agree with the approach taken here to use the existing ENRs for discv5 instead of the hack conversion of Enodes.
And for the bootstrap flag it sounds like a good plan to support both.

Not fully sure on the file flag version. I believe we should remain consistent with CL/Portal or vice versa. 

With this change, can https://github.com/status-im/nimbus-eth1/blob/master/execution_chain/networking/eth1_enr.nim#L21-L43 be removed by the way?

execution_chain/networking/bootnodes.nim
jangko: fixed

execution_chain/config.nim
jangko: fixed: keep both `--bootstrap-file` and `--static-peers-file`

tests/networking/test_enode.nim
jangko: fixed

execution_chain/networking/bootnodes.nim
jangko: fixed

PR #3664, opened 2025-09-11
shift commithash calc from eth2 to native eth1

advaita-saha:


execution_chain/version_info.nim
```diff
@@ -9,14 +9,62 @@
...
 import
-  std/[strformat], stew/byteutils, ./compile_info, beacon_chain/buildinfo, ./version
+  std/[os, strutils, strformat],
+  stew/byteutils,
+  ./compile_info,
+  ./version
 
 export version
 
+# TODO: Unify and use the buildinfo from nimbus-eth2
```
arnetheduck, line 19: why not just do this at once, instead of introducing this intermediate PR? it's not hard to do and avoids the double work of first doing this hack and then doing the right thing.

advaita-saha: fusaka-devnet was not able to deploy new eth1 branches, and all of our eth1 nodes were broken

Issue #3659, opened 2025-09-10
Crash during hoodi sync


advaita-saha:
```bash
DBG 2025-09-10 09:35:51.253+00:00 Validating block                           topics="forked chain" blkHash=0x12c9b6224a90798acedb1c713af479e39c6921f4555532321f01d21b53acff21 blk="(parentHash: 0xc93b0d92b5c0330c1d8da92fa44152eb7b0fe5e57e44ee57417d11e5caf54649, coinbase: 0x6cf99dff5c5bd8e865f663fbfdf7f66981ef2a7f, stateRoot: 0xbf6e45a8301759207b05fe74a7085491a7222ef56f32c0d7a74006f734970509, transactionsRoot: 0x4379ed7af005cbfea7a87121fbf129b704aaa03bcb33bdfc837853ad6950a1fa, receiptsRoot: 0x3e91fad9947aeefc06c7c8d460f666793640ba0fd37e7d98c4f8d779bc7a71be, number: 1182670, gasLimit: 59941351, gasUsed: 9444255, nonce: 0x0000000000000000, baseFeePerGas: ok(1029492987), withdrawalsRoot: ok(0xfed4ec4276d3aea3a7c7b73f5fe99fefc4adae2b913f896cf0222bbe3ef2229a), blobGasUsed: ok(524288), excessBlobGas: ok(0), parentBeaconBlockRoot: ok(0x70f89b5df554932d533f7808c49b135003e1c21e016136a07e009358f90e4eee), requestsHash: ok(0xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855))" parentTxFrame=130599130292848 txFrame=130598958304272
DBG 2025-09-10 09:35:52.199+00:00 Discovery receive error                    topics="p2p" discv4="disc: msg missing hash" discv5="Invalid protocol id"
DBG 2025-09-10 09:35:52.200+00:00 Received JSON-RPC request                  topics="JSONRPC-HTTP-SERVER" address=127.0.0.1:34816 len=343
DBG 2025-09-10 09:35:52.200+00:00 Processing JSON-RPC request                topics="json-rpc-router" id=1284 name=engine_forkchoiceUpdatedV3
[...]

PR #3666, opened 2025-09-11
Fix db.txId == 0  AssertionDefect when using snapshots and persist with stateroot check

bhartnett:
This fix is related to this issue: https://github.com/status-im/nimbus-eth1/issues/3659

The problem is that the assertions in the Aristo mpt and kvt init_common modules incorrectly assume that no nested db sessions can be created but when calling persist and supplying a stateroot, nested sessions do get created under certain conditions when using snapshots.

I opted to remove the verifyIxId related code because it's not used anywhere, the assertions are wrong and it's unlikely to provide any benefit for debugging purposes.

### DIFF
--- a/execution_chain/db/aristo/aristo_init/memory_db.nim
+++ b/execution_chain/db/aristo/aristo_init/memory_db.nim
@@ -57,13 +57,13 @@ when extraTraceMessages:
 # Private helpers
 # ------------------------------------------------------------------------------
 
-func newSession(db: MemBackendRef): MemPutHdlRef =
-  new result
+template newSession(db: MemBackendRef): MemPutHdlRef =
+  MemPutHdlRef()
 
-func getSession(hdl: PutHdlRef; db: MemBackendRef): MemPutHdlRef =
+template getSession(hdl: PutHdlRef; db: MemBackendRef): MemPutHdlRef =
   hdl.MemPutHdlRef
 
-func endSession(hdl: PutHdlRef; db: MemBackendRef): MemPutHdlRef =
+template endSession(hdl: PutHdlRef; db: MemBackendRef): MemPutHdlRef =
   hdl.MemPutHdlRef
 
 # ------------------------------------------------------------------------------


tests/test_aristo/test_tx_frame.nim
```diff
@@ -108,3 +108,31 @@ suite "Aristo TxFrame":
...
         tx.fetchAccountRecord(acc2[0]).isErr() # Doesn't exist in tx2
+
+  # This test case reproduces a bug which triggers an:
+  # `db.txId == 0`  [AssertionDefect]
+  # This occurs when the txId inside the database is (incorrectly) expected to always
+  # be equal to zero when starting a batch.
+  # See the related issue here: https://github.com/status-im/nimbus-eth1/issues/3659
+  # and the related file here: https://github.com/status-im/nimbus-eth1/blob/master/execution_chain/db/aristo/aristo_init/init_common.nim
+  # When passing in a stateroot to the persist call it is possible
+  # for a nested batch to be created withing the call to compute the stateroot
```
arnetheduck, line 119: this nested call is probably a bug to address .. the state root check should be moved into forkedchain or somewhere outside of the writing now that it's possible to do so, so that it happens _before_ opening a writing session - nested sessions are tricky semantically.

bhartnett: Sure, will do. Did you have thoughts on whether this cleanup is still a good idea? I think we should still remove this code because it is very possible/likely there will be other scenarios where concurrent sessions/batches might occur. 

For example every time we try to compute a state root even outside of the call to persist, a session might get created and we also concurrently sync and persist data via the header_chain_cache (HeaderChainRef) which writes directly to the kvt part of the db.

The thing is the db.txId value is shared between all these sessions that could in theory happen concurrently (which can produce a similar result as a nested session) but this id will get reset back to zero after finishing a session which can interfere and cause a crash in another session.

arnetheduck: > whether this cleanup is still a good idea

it's probably not a good idea in its present form - ie nested transactions are not really well-supported / deliberately accounted for -  with tests and well-defined semantics of what to do when the inner session succeeds but the "outer" fails, for example when running out of disk space - the idea behind frames is to ensure that we never leave the database in an inconsistent state.


> which writes directly to the kvt part of the db.

I think this is a design flaw in the header cache, which is a result of legacy workarounds rather than being a deliberate strategy that works well within the context. 

It was meant to be _entirely_ separate, with its own transaction lifetimes independent of the rest - instead, it looks like it got added as a hack on top of the current hacks, if it depends on the same txframe session.

At the core lies the fundamental flaw that we use txframe to achieve "fake" transactionality without really doing it correctly - nested sessions are an example of that - the header cache and the state db don't have a deliberate atomicity assumption in the higher abstraction levels of the code - they look kind of independent - but then they get roped into the atomicity at the lower levels because there's no clear separation of concerns in the API.

Basically, what all this means is that no part of the code should create long-lived sessions, that is using the core txframe, and no part of the code should create reentrant sessions (what the state root check incorrectly does).

The longer-term way to correct this is to further split the kvt so that its transaction boundaries cleanly match those of the higher-level code. The header cache, since it manages its own consistency, should not be sharing any resources with the rest of the db in such a world. It should, in all relevant aspects, be treated as a separate db.

The header cache could also be rewritten to write to the "correct" destination tables directly instead of being a separate db - but that's a different question entirely, ie it would require that the header sync was written with an entirely different approach (in that world, header sync would be "healing" gaps in the regular db instead of the current approach where it keeps its own caches - but in that world, it would have to participate correctly in the forkedchain fork handling as well).

Issue #3661, opened 2025-09-10
commit hash not updating on version_info


advaita-saha:
The unification PR https://github.com/status-im/nimbus-eth1/pull/3596
Started using the `GitRevision` from `nimbus-eth2`

How the `GitRevision` [code](https://github.com/status-im/nimbus-eth2/blob/b0bc9cda6ba07a69bcfb81bc84a0cf93925abea7/beacon_chain/buildinfo.nim#L36-L63) is structured we would be fetching the git commit hash from `nimbus-eth2`

This results in `nimbus-eth1` having a commit hash of the vendored `nimbus-eth2`
Which by extension is passed in every blocks we build and p2p layer, and also used by testing infra ( which now can't distinguish between different eth1 builds )

## Block built by us
<img width="1689" height="740" alt="Image" src="https://github.com/user-attachments/assets/741c3fdb-d258-4462-a34b-d5c58d0ec082" />

## Nimbus-eth2 vendored by us

<img width="720" height="450" alt="Image" src="https://github.com/user-attachments/assets/50a843ee-648d-4fff-b0d6-675ab9ebca95" />

kdeme:
This got resolved in https://github.com/status-im/nimbus-eth1/pull/3670

PR #3670, opened 2025-09-13
bump `nimbus-eth2` and fix commithash + copyright banner

advaita-saha:
Full fix for #3661


kdeme (approved):

advaita-saha:
@kdeme Can you look at the portal tests ?

tersec:
```
/home/runner/work/nimbus-eth1/nimbus-eth1/portal/tests/beacon_network_tests/test_beacon_content.nim(190, 1) template/generic instantiation of `suite` from here
/home/runner/work/nimbus-eth1/nimbus-eth1/portal/tests/beacon_network_tests/test_beacon_content.nim(197, 5) Error: the field 'bellatrix' is not accessible.
```
is due to:
- https://github.com/status-im/nimbus-eth2/pull/7478

Because
https://github.com/status-im/nimbus-eth1/blob/1ed687158d27462284a82a69dfa731f9dff9750b/portal/tests/beacon_network_tests/test_beacon_content.nim#L190-L201 won't work anymore.

That change is intentional, and while it's of course technically possible to revert it, maintaining the design it's based on into new forks doesn't make sense. I already haven't for Gloas.

https://github.com/status-im/nimbus-eth1/blob/1ed687158d27462284a82a69dfa731f9dff9750b/portal/tests/beacon_network_tests/test_beacon_content.nim#L30-L45 shows another approach, which remains viable, and if the exact contents of `forkDigests` is really that unimportant, the configuration information provided to `ForkDigests.init` can be done on the basis of convenience as well, e.g., something like `ForkDigests.init(getMetadataForNetwork("mainnet"), default(Hash32))` should work fine.

portal/tests/beacon_network_tests/test_beacon_content.nim
```diff
@@ -191,14 +191,8 @@ suite "Beacon Content Keys and Values":
...
   # fine for the different forks.
-  const forkDigests = ForkDigests(
-    phase0: ForkDigest([0'u8, 0, 0, 1]),
-    altair: ForkDigest([0'u8, 0, 0, 2]),
-    bellatrix: ForkDigest([0'u8, 0, 0, 3]),
-    capella: ForkDigest([0'u8, 0, 0, 4]),
-    deneb: ForkDigest([0'u8, 0, 0, 5]),
-    electra: ForkDigest([0'u8, 0, 0, 6]), # TODO fulu: ForkDigest([0'u8, 0, 0, 7]),
-  )
+  let forkDigests =
```
tersec, line 194: Would be slightly surprised if this can't also be `const`, even.

tersec: At least `nimbus-eth2` definitely builds these presets in at compile-time.

advaita-saha: Actually tried that but fails with 
```bash
Building: build/all_portal_tests
/home/advaita/development/nim/nimbus-eth1/portal/tests/beacon_network_tests/test_beacon_content.nim(190, 1) template/generic instantiation of `suite` from here
/home/advaita/development/nim/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/memory.nim(10, 5) Error: cannot 'importc' variable at compile time; c_memcpy
make: *** [Makefile:294: all_portal_tests] Error 1
```

tersec: Oh, ok, one of the C libraries. `c_memcpy` can/should probably be `when nimvm`-checked to avoid depending on in runtime, but for this kind of test code it's not worth waiting for and isn't that important.

advaita-saha: Also this is in tests, doesn't matter much

tersec: Can't do anything with `hash_tree_root()` or `eth2digest()` in `const`/`static` currently. Definitely out of scope of this PR to address.

kdeme (approved):

PR #3678, opened 2025-09-15
Provide main program API so it can be extended (e.g. for debugging.)

mjfh:
This allows to import the main module `nimbus_execution_client` as   fully functional client and update/modify command line features as  well as handler features.

### DIFF
--- a/.github/workflows/nightly_build.yml
+++ b/.github/workflows/nightly_build.yml
@@ -29,7 +29,7 @@ jobs:
       run:
         shell: bash
 
-    name: '${{ matrix.os }}-${{ matrix.cpu }}'
+    name: 'build-${{ matrix.os }}-${{ matrix.cpu }}'
     runs-on: ubuntu-latest
 
     steps:
@@ -71,6 +71,31 @@ jobs:
           name: ${{ matrix.os }}-${{ matrix.cpu }}-checksum
           path: ./dist/${{ steps.make_dist.outputs.archive_dir }}/build/nimbus_execution_client.sha512sum
           retention-days: 2
+
+  run-binary:
+    needs: [build, matrix_config]
+
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJSON(needs.matrix_config.outputs.matrix) }}
+
+    defaults:
+      run:
+        shell: bash
+
+    name: 'run-binary-${{ matrix.os }}-${{ matrix.cpu }}'
+    runs-on: ${{ matrix.builder }}
+
+    steps:
+      - name: Download artefacts
+        uses: actions/download-artifact@v5
+        with:
+          name: ${{ matrix.os }}-${{ matrix.cpu }}-archive
+
+      - name: Run binary
+        run: |
+          tar -xvzf nimbus-eth1-${{ matrix.os }}-${{ matrix.cpu }}-nightly-latest.tar.gz --strip-components=1
+          build/nimbus_execution_client --help
 
   nightly-build-release:
     name: Nightly build release
--- a/Makefile
+++ b/Makefile
@@ -124,7 +124,9 @@ VERIF_PROXY_OUT_PATH ?= build/libverifproxy/
 	dist \
 	eest \
 	t8n \
-	t8n_test
+	t8n_test \
+	evmstate \
+	evmstate_test
 
 ifeq ($(NIM_PARAMS),)
 # "variables.mk" was not included, so we update the submodules.
--- a/docker/dist/Dockerfile.linux-amd64
+++ b/docker/dist/Dockerfile.linux-amd64
@@ -9,7 +9,7 @@
 # according to those terms.
 
 # The build is reproducible only if this base image stays the same.
-FROM statusim/nimbus-eth1:dist_base_eth1_v2_linux_amd64@sha256:4eb9d6800889ea4b8422aecd82c0d6af45d6d2318a0be3777e4643d812078ed1
+FROM statusim/nimbus-eth1:dist_base_eth1_v2_linux_amd64@sha256:efa42e9735d99e78807ebcceb85b123142e0f4971b087ce66e4e89534b8fb424
 
 SHELL ["/bin/bash", "-c"]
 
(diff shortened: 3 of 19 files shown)


execution_chain/nimbus_execution_client.nim
```diff
@@ -212,7 +212,12 @@ proc preventLoadingDataDirForTheWrongNetwork(db: CoreDbRef; conf: NimbusConf) =
       expected=calculatedId
     quit(QuitFailure)
 
-proc run(nimbus: NimbusNode, conf: NimbusConf) =
+# ------------
+
+proc runExeClient*(nimbus: NimbusNode, conf: NimbusConf) =
```
arnetheduck: missing `{.push raises: [].}` - most likely, there are exceptions coming out here as well that need to be annotated

execution_chain/nimbus_execution_client.nim
```diff
@@ -288,7 +336,8 @@ proc run(nimbus: NimbusNode, conf: NimbusConf) =
   of NimbusCmd.`import`:
     importBlocks(conf, com)
   of NimbusCmd.`import-rlp`:
-    waitFor importRlpBlocks(conf, com)
+    QuitFailure.onCancelledException("Import of RLP blocks cancelled"):
```
arnetheduck, line 339: for a `CancelledError` to be raised here, `runClient` would have to be an async function that something cancels - since this is not the case, `CancelledException` can never be raised here and the code ends up being misleading

arnetheduck: see also https://status-im.github.io/nim-chronos/concepts.html#cancellation

mjfh: Hmm, the function `importRlpBlocks()` has annotation 
    `{.async: (raises: [CancelledError]).}`
 unlike `importBlocks()` which is neither `async` nor raises it any catchable errors.

arnetheduck: it has the annotation because the compiler cannot tell from which context the function is being called, whether it's actually possible that it's cancelled. `CancelledError` happens when you call `cancel` on a future - if you did:

```nim
let fut = importRlpBlocks(conf, com)
fut.cancel()
waitFor(fut)
```

this would _maybe_ raise a CancelledError depending on the what happened inside of `importRlpBlocks` before the cancellation happened.

Since nothing calls cancel on the future returned by `importRlpBlocks`, a `CancellationError` cannot be raised here - just like `if key in mytable: let value = mytable[key]` causes a raises effect without actually ever raising.

The correct way to handle cases like this is to a) understand if an exception can actually be raised, and then b) ignore it with an assert, ie `try: waitFor ... except CancelledError: raiseAssert "Nothing cancels the future"`

execution_chain/nimbus_execution_client.nim
```diff
@@ -227,28 +279,24 @@ proc run(nimbus: NimbusNode, conf: NimbusConf) =
       fatal "Cannot load Kzg trusted setup from file", msg=res.error
       quit(QuitFailure)
 
-  let coreDB =
-    # Resolve statically for database type
-    AristoDbRocks.newCoreDbRef(
+  # The constructor `newCoreDbRef()` calls `addExitProc()` which in turn
+  # accesses a global variable holding a call back function. This function
+  # `addExitProc()` is synchronised against an internal tread lock and is
+  # considered safe, here.
```
arnetheduck, line 285: this usage of gcsafe is in fact unsafe - gcsafe is about allocating memory from different threads and because `addExitProc` performs an allocation, it is not safe to ignore.

in particular, gcsafe is _not_ the same as "thread safe" - it's about allocating memory from different heaps and while we use the gc, heaps are thread-local.

execution_chain/nimbus_execution_client.nim
```diff
@@ -309,16 +358,24 @@ proc run(nimbus: NimbusNode, conf: NimbusConf) =
...
-  # Processing command line arguments
-  let conf = makeConfig()
-
-  # Set up logging before everything else
-  setupLogging(conf.logLevel, conf.logStdout, none(OutFile))
+  # The function `setupLogging()` calls `setTopicState()` which in turn
+  # accesses a global `Table` variable. The latter function is synchronised
+  # against an internal lock via a `guard` annotation and is considered
+  # thread safe, here.
+  {.gcsafe.}:
```
arnetheduck, line 376: same as above - a lock is not sufficient for `gcsafe`

arnetheduck: https://github.com/status-im/nim-chronicles/pull/187 - this is the proper fix for the gcsafety issue, ie ensuring that allocations happen only from one thread and not from multiple. 

alternatively, the function should simply propagate `gcsafe` upwards.

execution_chain/nimbus_execution_client.nim
mjfh: Ok, what to do?

execution_chain/nimbus_execution_client.nim
mjfh: ditto

execution_chain/nimbus_execution_client.nim
arnetheduck: fix the code or don't mark the function `gcsafe` - ie someone calling this function should *know* that it's not gcsafe and take appropriate action (for example making sure it's called only from one thread. Casting away the gcsafety marker means lying to the consumer of the function which must never be done.

execution_chain/nimbus_execution_client.nim
arnetheduck: ditto: I just linked an example of how you can *make* code gcsafe (by not using garbage-collected types or ensuring only safe operations that don't copy/allocate/refcount are used)

execution_chain/nimbus_execution_client.nim
mjfh: Which code should I fix? The problem is buried in the Aristo code with a global hook (I did not write this code.) Actually, this hook should relocated to some local descriptor. But this might be more than a simple change.

arnetheduck: well, the easiest thing in this case is probably to expose the stat printing as a public function and call it just before the application terminates, from the main module - as the TODO notes, it was a hack at the time and part of making the code callable as a library is cleaning up the relevant todos with correct code

mjfh: I will have a look at that

PR #3681, opened 2025-09-15
Clear parent when pruning branches

### LINKED ISSUE #3674: Clear parent when pruning branches
https://github.com/status-im/nimbus-eth1/blob/58c297fb7bd92b099151f12409e76aebdd05372e/execution_chain/core/chain/forked_chain.nim#L266 - here, blocks are removed from the cache but still hold a reference to the parent - this reference can keep blocks in memory longer than they need to be there - every time a block is removed, it should see its parent cleared (taking care during iteration)

jangko:
fix #3674

### DIFF
--- a/execution_chain/core/chain/forked_chain/chain_branch.nim
+++ b/execution_chain/core/chain/forked_chain/chain_branch.nim
@@ -34,17 +34,29 @@ template number*(b: BlockRef): BlockNumber =
   b.blk.header.number
 
 func `==`*(a, b: BlockRef): bool =
-  a.hash == b.hash
+  if a.isNil.not and b.isNil.not:
+    a.hash == b.hash
+  else:
+    false
 
 template isOk*(b: BlockRef): bool =
   b.isNil.not
 
-template loopIt*(init: BlockRef, body: untyped) =
+template loopItImpl(condition: untyped, init: BlockRef, body: untyped) =
   block:
     var it{.inject.} = init
-    while it.isOk:
+    while it.condition:
       body
       it = it.parent
+
+template loopItAndClearParentImpl(condition: untyped, init: BlockRef, body: untyped) =
+  block:
+    var it{.inject.} = init
+    while it.condition:
+      body
+      let tmp = it
+      it = it.parent
+      tmp.parent = nil
 
 template stateRoot*(b: BlockRef): Hash32 =
   b.blk.header.stateRoot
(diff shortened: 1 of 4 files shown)


execution_chain/core/chain/forked_chain/chain_branch.nim
```diff
@@ -56,15 +68,20 @@ const
 template finalize*(b: BlockRef) =
   b.index = DAG_NODE_FINALIZED
 
-template notFinalized*(b: BlockRef) =
+template unFinalize*(b: BlockRef) =
```
arnetheduck: this naming is off .. once something is finalized, it can never change state - in other words, everything starts out as being not finalized and the only direction in which the state change can go is from "not finalized" to "finalized" - the actions that "make" something become finalized are either that the CL or the user says so.


_in theory_ we can remember in the database if something ever "became" finalized but I don't know that it's worth the bother due to the recursive nature of finality (ie final block finalizes everything before it).

arnetheduck: in other words, the only operations we should have in the api is basically `init` that creates a blockref and `finalize` which marks the block (and its ancestors) final.

execution_chain/core/chain/forked_chain/chain_branch.nim
```diff
@@ -56,15 +68,20 @@ const
...
   b.index = DAG_NODE_CLEAR
 
-template finalized*(b: BlockRef): bool =
-  b.index == DAG_NODE_FINALIZED
+template notFinalized*(b: BlockRef): bool =
+  b.index != DAG_NODE_FINALIZED
+
+template loopIt*(init: BlockRef, body: untyped) =
```
arnetheduck: can't this be an iterator? the api is somewhat unusual and therefore a bit harder to read

execution_chain/core/chain/forked_chain/chain_branch.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/3681/commits/f235d1b4d5c49285841562be49f0ef64a04245c1

execution_chain/core/chain/forked_chain/chain_branch.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/3681/commits/f235d1b4d5c49285841562be49f0ef64a04245c1

execution_chain/core/chain/forked_chain/chain_branch.nim
```diff
@@ -34,37 +34,49 @@ template number*(b: BlockRef): BlockNumber =
...
 
 template finalize*(b: BlockRef) =
   b.index = DAG_NODE_FINALIZED
 
-template notFinalized*(b: BlockRef) =
-  b.index = DAG_NODE_CLEAR
+template notFinalized*(b: BlockRef): bool =
+  b.index != DAG_NODE_FINALIZED
+
+iterator loopIt*(init: BlockRef): BlockRef =
```
arnetheduck: not sure if it's for this PR, maybe call this `ancestors`?

execution_chain/core/chain/forked_chain/chain_branch.nim
```diff
@@ -34,37 +34,49 @@ template number*(b: BlockRef): BlockNumber =
...
-    var it{.inject.} = init
-    while it.isOk:
-      body
-      it = it.parent
+template loopItImpl(condition: untyped, init: BlockRef) =
+  var it = init
+  while it.condition:
+    yield it
+    it = it.parent
+
```
arnetheduck:
```suggestion
    let next = it.parent
    yield it
    it = next
```

this allows the parent to change in the loop so it can be used for parent-clearing

arnetheduck (approved):

execution_chain/core/chain/forked_chain/chain_branch.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/3681/commits/904638f728eecdde128f62585eaa0f4e4061ae98

execution_chain/core/chain/forked_chain/chain_branch.nim
jangko: https://github.com/status-im/nimbus-eth1/pull/3681/commits/904638f728eecdde128f62585eaa0f4e4061ae98

arnetheduck:
when deserializing, where is the `index` updated to its 0/1 value?

jangko:
will add

PR #3683, opened 2025-09-16
Stateless: Implement stateless execution of blocks

bhartnett:
This adds a new `statelessProcessBlock` function which executes a block in memory by using the execution witness to populate an in memory database. The aristo `putSubtrie` function is used load the subtrie of trie nodes into the database before executing the block. The witness is validated for correctness before starting the execution. After executing the block the stateroot is compared to the stateroot in the header of the current block.

In order to verify the correctness of the generated execution witnesses, stateless execution can be enabled at runtime by setting the `--stateless-witness-validation=true` flag (note that this flag depends on the `--stateless-provider=true` being set as well).


tests/eest/test_stateless_execution.nim
```diff
@@ -0,0 +1,33 @@
...
+  std/os,
+  unittest2,
+  ./eest_helpers,
+  ./eest_blockchain
+
+const
+  baseFolder = "tests/fixtures"
+  eestType = "blockchain_tests"
+  eestReleases = [
+    "eest_stable"
```
advaita-saha: We are not using stable anymore
For mainnet => `eest_develop`
For devnet => `eest_devnet`

bhartnett: Thanks, will update.

PR #3691, opened 2025-09-18
EVM: add DIV by zero test case

jangko:
Also make both DIV and SDIV become consistent. Because top == rhs in both cases, so we do not need to set top to zero anymore when rhs is zero.


jangko:
Not a bug in test suite or EVM, only some inconsistencies lead to confusion.
https://discord.com/channels/613988663034118151/748198686911234170/1417906126430933104

PR #3695, opened 2025-09-19
Fix crash due to missing checkpoint on shutdown during sync

bhartnett:
I would sometimes get this error when stopping Nimbus with Ctrl-C while syncing:
```
../csu/libc-start.c(392) 
../sysdeps/nptl/libc_start_call_main.h(58) 
/home/user/development/status-im/nimbus-eth1/vendor/nim-json-rpc/json_rpc/client.nim(875) main
/home/user/development/status-im/nimbus-eth1/vendor/nim-json-rpc/json_rpc/client.nim(863) NimMain
/home/user/development/status-im/nimbus-eth1/execution_chain/nimbus_execution_client.nim(332) _ZN23nimbus_execution_client12runExeClientE3refIN11nimbus_desc26NimbusNodecolonObjectType_EEN6config10NimbusConfE
/home/user/development/status-im/nimbus-eth1/execution_chain/db/core_db/base.nim(102) _ZN4base7persistE3refIN9base_desc25CoreDbRefcolonObjectType_EE3refIN9base_desc27CoreDbTxRefcolonObjectType_EE
/home/user/development/status-im/nimbus-eth1/execution_chain/db/aristo/aristo_tx_frame.nim(119) _ZN15aristo_tx_frame7persistE3refIN11aristo_desc27AristoDbRefcolonObjectType_EE3refIN12desc_backend25PutHdlRefcolonObjectType_EE3refIN11aristo_desc27AristoTxRefcolonObjectType_EE
/home/user/development/status-im/nimbus-eth1/execution_chain/db/aristo/aristo_tx_frame.nim(127) _ZN15aristo_tx_frame7persistE3refIN11aristo_desc27AristoDbRefcolonObjectType_EE3refIN12desc_backend25PutHdlRefcolonObjectType_EE3refIN11aristo_desc27AristoTxRefcolonObjectType_EE
/home/user/development/status-im/nimbus-eth1/vendor/nim-results/results.nim(979) _ZN7results6expectE3varI6ResultIN9hardforks12BlobScheduleE4voidEE6string
/home/user/development/status-im/nimbus-eth1/vendor/nim-results/results.nim(438) _ZN7results17raiseResultDefectE6string
/home/user/development/status-im/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/excpt.nim(349) _ZN6system18rawWriteStackTraceE3varI3seqIN6system15StackTraceEntryEEE
/home/user/development/status-im/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/stacktraces.nim(62) _ZN11stacktraces30auxWriteStackTraceWithOverrideE3varI3seqIN6system15StackTraceEntryEEE
Error: unhandled exception: `checkpoint` before persisting frame [ResultDefect]
```

This fixes the problem by making sure there is a checkpoint completed before persist.

### DIFF
--- a/execution_chain/nimbus_execution_client.nim
+++ b/execution_chain/nimbus_execution_client.nim
@@ -329,6 +329,7 @@ proc runExeClient*(nimbus: NimbusNode, conf: NimbusConf) {.gcsafe.} =
         txFrame = fc.baseTxFrame
       fc.serialize(txFrame).isOkOr:
         error "FC.serialize error: ", msg=error
+      txFrame.checkpoint(fc.base.blk.header.number)
       com.db.persist(txFrame)
     com.db.finish()
 
--- a/execution_chain/nimbus_execution_client.nim
+++ b/execution_chain/nimbus_execution_client.nim
@@ -327,10 +327,12 @@ proc runExeClient*(nimbus: NimbusNode, conf: NimbusConf) {.gcsafe.} =
       let
         fc = nimbus.fc
         txFrame = fc.baseTxFrame
+
       fc.serialize(txFrame).isOkOr:
         error "FC.serialize error: ", msg=error
-      txFrame.checkpoint(fc.base.blk.header.number)
+      txFrame.checkpoint(fc.base.blk.header.number, skipSnapshot = true)
       com.db.persist(txFrame)
+
     com.db.finish()
 
   case conf.cmd


execution_chain/nimbus_execution_client.nim
```diff
@@ -327,9 +327,12 @@ proc runExeClient*(nimbus: NimbusNode, conf: NimbusConf) {.gcsafe.} =
       let
         fc = nimbus.fc
         txFrame = fc.baseTxFrame
+
       fc.serialize(txFrame).isOkOr:
         error "FC.serialize error: ", msg=error
+      txFrame.checkpoint(fc.base.blk.header.number, skipSnapshot = true)
```
jangko, line 333: Actually I have no clue why this is needed. And this is a sign of bad design decision. A random developer/maintainer will have difficult time debugging this kind of bug in the future. Can we at least have some safety net to help the next people to figure out why it have to be like this?

bhartnett: Not sure I understand what you mean. Are you suggesting I add a comment about why the checkpoint is required here? Or something else?

bhartnett: A checkpoint is required on every txFrame before persisting, see here: https://github.com/status-im/nimbus-eth1/blob/c859f168f8c18b52b86772d257c84e80d0d70c1e/execution_chain/db/aristo/aristo_tx_frame.nim#L127

The checkpoint sets the block number in the txFrame which is eventually stored in the database.

bhartnett: After the call to persist the txFrame.blockNumber is reset: https://github.com/status-im/nimbus-eth1/blob/c859f168f8c18b52b86772d257c84e80d0d70c1e/execution_chain/db/aristo/aristo_tx_frame.nim#L239

So when trying to persist the the txFrame after serializing the forked chain, the crash is triggered.

jangko: > A checkpoint is required on every txFrame before persisting.

If that is the case, instead of only relying on runtime crash, why not force user to e.g. pass block number when calling `persist` and put `checkpoint` call into `persist`? And we can avoid this kind of bug during compile time.

bhartnett: > If that is the case, instead of only relying on runtime crash, why not force user to e.g. pass block number when calling `persist` and put `checkpoint` call into `persist`? And we can avoid this kind of bug during compile time.

Sure, that sounds like a reasonable idea but I think it is beyond the scope of this PR. Changing the persist API would impact a few other locations and really I was just hoping to fix the crash with this change. Mainly wanted to get your input on whether the checkpoint of the block number in the forked chain base is correct.

jangko (approved):

Issue #3699, opened 2025-09-19
EL getting into tx nonce mismatch


advaita-saha:
```bash
INF 2025-09-19 09:18:56.164+00:00 Blocks import error (skip remaining)       topics="beacon sync" n=0 iv=#1173217..#1173256 nBlocks=40 nthBn=#1173217 nthHash=ed24f5c06580 base=#1173216 head=#1173216 blkFailCount=1 error="(excp: ENoException, name: , msg: Error processing tx with index 0:invalid tx: account nonce mismatch. txNonce=1051, accNonce=1049, elapsed: 2ms419us618ns)"
```

While this is difficult is reproduce, this has popped up twice, once in devnet fleet, and once in nimbus hoodi fleet.
Since unable to reproduce this, best would be to work with a db copy from where it is last seen ( i.e the fleet ). cc @jakubgs 

The p2p layer indeed lack block validation, but this issue I am sure is not occurring due to misbehaving peers. As this is occurring across all peers once it has happened. And also happening across restarts.
Which means we are stuck with a corrupted database

Need to figure out a way to recover from this and also prevent the same.

bhartnett:
> I don't remember seeing this before - could it be the recent txframe snapshot optimization?

Perhaps it could be related but I am still yet to confirm the cause. It appears that the state in the database is corrupted in some way which is very odd.

Basically the account which submitted the failing transaction with address `0x03d284fdf9a80a71c98107b59c35a0e493d69fae` has incorrect leaf data but the state root for that block at 1173216 is correct with value: `0xcdd2ea1fac83fa44bba01da8c341afeab6ca416ad295dea043cb1b202038d115`

I fetched the trie node data for that account at that block using `eth_getProof` using the local database and then ran `verifyProof`. The verification failed and after debugging it turns out that only the first two trie nodes are correct, the third trie node fails to match the next hash.

bhartnett:
I suspect the problem is related to this section of code in aristo compute key: https://github.com/status-im/nimbus-eth1/blob/70b88c1ccc007416f151efe8155627a17cb4e301/execution_chain/db/aristo/aristo_compute.nim#L78

Looks like this code can write `VertexRef`s to the database as a part of updating the `HashKey`s but these vertexes get fetched using `getKeyRc` which can return the vertex from the in memory cache layers. Now if the vertex was updated to a bad value in the in memory cache later such as when processing a bad block then in theory a bad vertex could get written to the database when doing a state root check. 

[...]

PR #3705, opened 2025-09-23
Prevent database corruption during stateroot computation

bhartnett:
This fix is related to this issue: https://github.com/status-im/nimbus-eth1/issues/3699

I have found that when computing the stateroot it is possible for in memory vertexes (that have a level lower than the baseTxFrame) to be written to the database and this is very likely to be the cause of the database corruption. When computing the stateroot we should only write directly to the database if the vertex was just fetched from the database. 

I confirmed that in the forked chain tests this scenario was occurring (but without the data corruption) when processing a single linear chain of blocks. The problem is basically that the snapshot data can contain vertexes older than base after the base is updated. It seems likely that when dealing with multiple branches/forks, this scenario could lead to a database corruption. 

With this change, when computing the stateroot we only allow writing vertexes to the database that were fetched directly from the database.

### DIFF
--- a/execution_chain/core/chain/forked_chain.nim
+++ b/execution_chain/core/chain/forked_chain.nim
@@ -493,8 +493,6 @@ proc validateBlock(c: ForkedChainRef,
 
   c.writeBaggage(blk, blkHash, txFrame, receipts)
 
-
-
   let newBlock = c.appendBlock(parent, blk, blkHash, txFrame, move(receipts))
 
   for i, tx in blk.transactions:
--- a/execution_chain/core/chain/forked_chain.nim
+++ b/execution_chain/core/chain/forked_chain.nim
@@ -482,13 +482,18 @@ proc validateBlock(c: ForkedChainRef,
     parentTxFrame=cast[uint](parentFrame),
     txFrame=cast[uint](txFrame)
 
+  # Update the snapshot before processing the block so that any vertexes in snapshots
+  # from lower levels than the baseTxFrame are removed from the snapshot before running
+  # the stateroot computation.
+  c.updateSnapshot(blk, txFrame)
+
   var receipts = c.processBlock(parent, txFrame, blk, blkHash, finalized).valueOr:
     txFrame.dispose()
     return err(error)
 
   c.writeBaggage(blk, blkHash, txFrame, receipts)
 
-  c.updateSnapshot(blk, txFrame)
+
 
   let newBlock = c.appendBlock(parent, blk, blkHash, txFrame, move(receipts))
 
--- a/execution_chain/core/chain/forked_chain.nim
+++ b/execution_chain/core/chain/forked_chain.nim
@@ -482,16 +482,12 @@ proc validateBlock(c: ForkedChainRef,
     parentTxFrame=cast[uint](parentFrame),
     txFrame=cast[uint](txFrame)
 
-  # Update the snapshot before processing the block so that any vertexes in snapshots
-  # from lower levels than the baseTxFrame are removed from the snapshot before running
-  # the stateroot computation.
-  c.updateSnapshot(blk, txFrame)
-
   var receipts = c.processBlock(parent, txFrame, blk, blkHash, finalized).valueOr:
     txFrame.dispose()
     return err(error)
 
   c.writeBaggage(blk, blkHash, txFrame, receipts)
+  c.updateSnapshot(blk, txFrame)
 
   let newBlock = c.appendBlock(parent, blk, blkHash, txFrame, move(receipts))
 
--- a/execution_chain/core/chain/forked_chain.nim
+++ b/execution_chain/core/chain/forked_chain.nim
@@ -482,12 +482,16 @@ proc validateBlock(c: ForkedChainRef,
     parentTxFrame=cast[uint](parentFrame),
     txFrame=cast[uint](txFrame)
 
+  # Update the snapshot before processing the block so that any vertexes in snapshots
+  # from lower levels than the baseTxFrame are removed from the snapshot before running
+  # the stateroot computation.
+  c.updateSnapshot(blk, txFrame)
+
   var receipts = c.processBlock(parent, txFrame, blk, blkHash, finalized).valueOr:
     txFrame.dispose()
     return err(error)
 
   c.writeBaggage(blk, blkHash, txFrame, receipts)
-  c.updateSnapshot(blk, txFrame)
 
   let newBlock = c.appendBlock(parent, blk, blkHash, txFrame, move(receipts))
 
--- a/execution_chain/core/chain/forked_chain/chain_serialize.nim
+++ b/execution_chain/core/chain/forked_chain/chain_serialize.nim
@@ -126,16 +126,12 @@ proc replayBlock(fc: ForkedChainRef;
     parentFrame = parent.txFrame
     txFrame = parentFrame.txFrameBegin
 
-  # Update the snapshot before processing the block so that any vertexes in snapshots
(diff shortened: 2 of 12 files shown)


execution_chain/core/chain/forked_chain/chain_serialize.nim
```diff
@@ -126,12 +126,16 @@ proc replayBlock(fc: ForkedChainRef;
     parentFrame = parent.txFrame
     txFrame = parentFrame.txFrameBegin
 
+  # Update the snapshot before processing the block so that any vertexes in snapshots
+  # from lower levels than the baseTxFrame are removed from the snapshot before running
```
arnetheduck: this comment looks somewhat concerning - there should be _no_ verticies from below baseTxFrame in memory at any point - when we persist, we move everything below baseTxFrame to the on-disk database and empty the baseTxFrame, breaking its "link" with txframe below. In fact, once we've performed persist, any "levels" below baseTxFrame are no longer connected and should not be reachable so how come they're still being added to?

bhartnett: The data in each snapshot contains a level (integer). Every time we create a snapshot on a txFrame the snapshot is moved from the older frame but the level from which the vertex originated is kept in the snapshot so that even after we persist and clean up all the lower level txFrames (as you've mentioned) the vertex data from the lower level can remain in the snapshot which is part of a newer (not yet persisted) txFrame.

bhartnett: Note that in this scenario (when importing blocks via forked chain sync) the txFrame that get persisted (during updateBase) lags behind the txFrame at the head which holds the snapshot. In other words we are not persisting the txFrame which is holding the snapshot but rather an older txFrame with a lower level.

arnetheduck: Ah, I think I see the problem - this used to be cleared out by persist but I guess the way snapshots are being built now influences it.

in that case, shouldn't this be fixable by comparing the level of the snapshot with the basetxframe level and discarding on access if it's lower? that would remove the need to call updateSnapshot (which feels more like a workaround - the level stuff should be transparent to the user of the api, it's an internal optimization mechanism)

arnetheduck: or maybe not even discarding, but returning "dblevel" in that scenario .. ie if a snapshot was written to the database and there's no "newer" information in higher levels, it seems to me it should still be valid? what am I missing?

bhartnett: Well the main point of this change is that we shouldn't be writing vertexes to the db during the stateroot computation, that are older/at a lower level then the database level. My reasoning for this is that it could cause the db corruption if the vertex being written happens to be older then what is already in the database. If we never deal with multiple branches/forks then this is probably safe because the full history of writes would be in the snapshot, but in practice when dealing with multiple branches and reorgs we have a dag of txFrames and multiple distinct snapshots. If I return `dbLevel` then the older vertex would get written to the db as was done before this change so no point in doing that.

tests/test_aristo/test_tx_frame.nim
```diff
@@ -227,3 +227,50 @@ suite "Aristo TxFrame":
...
+
+    # Verify that getting the state root of the level 3 txFrame does not impact
+    # the persisted state in the database.
+    let stateRootBefore = tx3.fetchStateRoot().get()
+    try:
+      discard tx4.fetchStateRoot().get()
+    except AssertionDefect:
+      # Should throw assertion defect here because tx4 is at lower level than
+      # the baseTxFrame level.
+      discard
```
arnetheduck:
```suggestion
    expect(Defect):
      discard tx4.fetchStateRoot()
```

execution_chain/core/chain/forked_chain.nim
```diff
@@ -482,13 +482,18 @@ proc validateBlock(c: ForkedChainRef,
...
+  c.updateSnapshot(blk, txFrame)
+
   var receipts = c.processBlock(parent, txFrame, blk, blkHash, finalized).valueOr:
     txFrame.dispose()
     return err(error)
 
   c.writeBaggage(blk, blkHash, txFrame, receipts)
 
-  c.updateSnapshot(blk, txFrame)
+
```
arnetheduck:
```suggestion
```

tests/test_aristo/test_tx_frame.nim
```diff
@@ -227,3 +227,50 @@ suite "Aristo TxFrame":
...
+    # level 2
+    let tx4 = db.txFrameBegin(tx1)
+    for i in 300..<400:
+      let acc = makeAccount(i.uint64)
+      check tx4.mergeAccountRecord(acc[0], acc[1]).isOk()
+    tx4.checkpoint(2, skipSnapshot = false)
+
+    block:
+      let batch = db.putBegFn().expect("working batch")
+      db.persist(batch, tx3) # after this the baseTxFrame is at level 4
```
arnetheduck: 4 or 3?

bhartnett: Typo, should be 3. Will update.

execution_chain/core/chain/forked_chain.nim
```diff
@@ -482,14 +482,17 @@ proc validateBlock(c: ForkedChainRef,
     parentTxFrame=cast[uint](parentFrame),
     txFrame=cast[uint](txFrame)
 
+  # Update the snapshot before processing the block so that any vertexes in snapshots
+  # from lower levels than the baseTxFrame are removed from the snapshot before running
+  # the stateroot computation.
+  c.updateSnapshot(blk, txFrame)
```
bhartnett: @jangko Will moving this updateSnapshot call up above processBlock cause any issues that you are aware of? It seems ok in my testing.

execution_chain/core/chain/forked_chain/chain_serialize.nim
bhartnett: > in that case, shouldn't this be fixable by comparing the level of the snapshot with the basetxframe level and discarding on access if it's lower? that would remove the need to call updateSnapshot (which feels more like a workaround - the level stuff should be transparent to the user of the api, it's an internal optimization mechanism)

@arnetheduck Yes I like this idea. I've updated the the aristo layers to filter out any vertexes and hash keys that have a level lower than the base level. Would be good to get another review from you after these latest changes.

execution_chain/db/aristo/aristo_desc.nim
```diff
@@ -218,6 +218,8 @@ iterator rstack*(tx: AristoTxRef, stopAtSnapshot = false): AristoTxRef =
 
     if stopAtSnapshot and tx.snapshot.level.isSome():
       break
+    if stopAtBaseLevel and tx.level <= tx.db.baseTxFrame().level:
```
arnetheduck: I can't think of any valid sequence of txframe interation that would lead to this scenario - as such, this would probably be better off as an `assert` - the question though is whether this is a good place for it - in particular, the `get` operation runs very frequently so I'm not convinced we want to burden it the performance implications we'd see here - instead, we could focus on catching this in `persist`  which has to examine all entries anyway.

bhartnett: Good point, I agree that it shouldn't occur under normal conditions but it probably could occur when dealing with multiple heads/forks if base is updated/advanced forward on one chain but then the txFrames on another chain would become invalid so if they get used before getting cleaned up then it might happen. 

If this did happen it would be a bug in our code and not something that we need to support. I agree that it should be an assert so that we can catch these types of issues though and considering that these code paths are hotspots, I'll go with my previous change.

PR #3709, opened 2025-09-24
proxy: refactor

chirag-parmar:
Verified proxy consists fo two components the verifying algorithms and the proxy. This PR logically splits the verification engine from the proxy. The proxy is further broken down into seperate client and server implementations that connect to the verification engine as backend and frontend resp.


nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
```diff
@@ -57,17 +57,46 @@ type VerifiedProxyConf* = object
...
-  # In order to support the BLOCKHASH opcode for eth_call we need at least
-  # MAX_PREV_HEADER_DEPTH headers in the header cache.
-  cacheLen* {.
+  accountCacheLen* {.
     hidden,
-    desc: "Length of the header cache maintained in memory",
+    desc: "Length of the accounts cache maintained in memory",
+    defaultValue: 128,
+    defaultValueDesc: "128",
+    name: "dbg=acc-cache-len"
```
kdeme:
```suggestion
    name: "dbg-acc-cache-len"
```

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
```diff
@@ -57,17 +57,46 @@ type VerifiedProxyConf* = object
...
+  .}: int
+
+  codeCacheLen* {.
+    hidden,
+    desc: "Length of the code cache maintained in memory",
+    defaultValue: 64,
+    defaultValueDesc: "64",
+    name: "dbg-code-cache-len"
```
kdeme: I checked and  `nimbus_beacon_node`, `nimbus_execution_client`, `nimbus_portal_client`, consensus light client all use `debug` for this and not `dbg`. So lets keep those consistent.

nimbus_verified_proxy/engine/accounts.nim
```diff
@@ -125,18 +125,18 @@ proc getAccount*(
...
 
   return account
 
 proc getCode*(
-    lcProxy: VerifiedRpcProxy,
+    engine: RpcVerificationEngine,
     address: Address,
     blockNumber: base.BlockNumber,
     stateRoot: Root,
 ): Future[Result[seq[byte], string]] {.async.} =
```
kdeme: missing raises

kdeme: And on many other locations

nimbus_verified_proxy/engine/receipts.nim
```diff
@@ -43,11 +43,11 @@ func toReceipts(recs: openArray[ReceiptObject]): seq[Receipt] =
   recs.mapIt(it.toReceipt)
 
 proc getReceipts(
-    vp: VerifiedRpcProxy, header: Header, blockTag: types.BlockTag
+    engine: RpcVerificationEngine, header: Header, blockTag: types.BlockTag
 ): Future[Result[seq[ReceiptObject], string]] {.async.} =
```
kdeme: Missing raises for example here also.

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
chirag-parmar: good catch. thanks

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
chirag-parmar: makes sense. I just wanted to make them shorter. But can go back for consistency

nimbus_verified_proxy/engine/accounts.nim
chirag-parmar: these functions don't raise any exception. I have wrapped everything in a try catch block. The raises (empty) pragma is activated for the entire file (not visible in these changes)

nimbus_verified_proxy/engine/receipts.nim
chirag-parmar: same as above

nimbus_verified_proxy/engine/accounts.nim
kdeme: I did see the try/catch blocks, and assumed that because of these the procs won't actually raise something. Nevertheless, somebody could change/add something in the future that does raise.

And the `{.push raises: [], gcsafe.}`  that you apply  at top is not the same as async raises. Those you will need to apply here individually to make sure no exception raises from async procs.

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
tersec: `--debug-foo` has a specific semantics, per https://nimbus.guide/options.html#available-options which states:
> Any `debug`-prefixed flags are considered ephemeral and subject to removal without notice.

Which isn't technically stated to apply to `--dbg-foo`.

nimbus_verified_proxy/engine.nim
tersec, line 1: How does this file relate to
https://github.com/status-im/nimbus-eth1/blob/775421c4862fe95de9933fcdd0c457c6dbbdf07f/nimbus_verified_proxy/engine/engine.nim#L1-L31
?

It seems almost identical, except for the `import` formatting and that this version appears to construct a `let engine = RpcVerificationEngine` to return but never actually returns it, which causes Nim to by default return `default(T)`, or `default(RpcVerificationEngine)`.

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
```diff
@@ -57,17 +57,46 @@ type VerifiedProxyConf* = object
...
 
-  # In-Memory Cache Size
-  # In order to support the BLOCKHASH opcode for eth_call we need at least
-  # MAX_PREV_HEADER_DEPTH headers in the header cache.
-  cacheLen* {.
+  accountCacheLen* {.
     hidden,
-    desc: "Length of the header cache maintained in memory",
+    desc: "Length of the accounts cache maintained in memory",
+    defaultValue: 128,
```
tersec, line 63: With all these `defaultValue`/`defaultValueDesc` pairs, two points:

1. when the option is `hidden`, it's fine to just leave out the `defaultValueDesc`, or even the `desc` fields. They'll never be visible anyway; that's what `hidden` means, that they don't show up in `--help`. If you find it useful as a sort of internal code documentation, that's fine too, but that's all they're really doing.

2. but more generally, if or when one does want to have these `defaultValue: foo` and `defaultValueDesc: $foo` pairs, where it's the same `foo` in each case, then something like https://github.com/status-im/nimbus-eth2/blob/27ef2f10ab8c05ea3794b57fab5d33bcec712048/beacon_chain/conf.nim#L47-L56 where
```nim
const
  # TODO: How should we select between IPv4 and IPv6
  # Maybe there should be a config option for this.
  defaultAdminListenAddress* = (static parseIpAddress("127.0.0.1"))
  defaultSigningNodeRequestTimeout* = 60
  defaultBeaconNode* = "http://127.0.0.1:" & $defaultEth2RestPort
  defaultBeaconNodeUri* = parseUri(defaultBeaconNode)
  defaultGasLimit* = 45_000_000
  defaultAdminListenAddressDesc* = $defaultAdminListenAddress
  defaultBeaconNodeDesc = $defaultBeaconNode
```
can then be (re)used symbolically in both `defaultValue`/`defaultValueDesc` pairs so they can never get out of sync with each other.

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -39,6 +39,23 @@ proc cleanup*(ctx: ptr Context) =
   dealloc(ctx.configJson)
   freeShared(ctx)
 
+proc verifyChaindId(engine: RpcVerificationEngine): Future[void] {.async.} =
+  let localId = engine.chainId
+
+  let providerId =
+    try:
+      await engine.backend.eth_chainId()
+    except CatchableError as e:
```
tersec:
```
nimbus-eth1/nimbus_verified_proxy/nimbus_verified_proxy.nim(48, 30) Hint: 'e' is declared but not used [XDeclaredButNotUsed]
```
Can just omit the `as e`, e.g.
```
    except CatchableError:
```
Separately, there are a lot of `CatchableError` references throughout, but that can be addressed via a separate comment. As far as this specific one goes, it's just the `XDeclaredButNotUsed`.

nimbus_verified_proxy/rpc_api_backend.nim
```diff
@@ -8,71 +8,105 @@
...
+  case backend.kind:
+  of Http:
+    backend.httpClient
+  of WebSocket:
+    backend.wsClient
+
+proc init*(T: type JsonRpcBackend, url: Web3Url): T =
+  var backend: JsonRpcBackend
+
+  if url.kind == HttpUrl:
```
tersec: Can directly use a `let` here by using an `if` expression rather than `if` statement and therefore never notionally have a point in the Nim code where the `backend` exists but isn't initialized. But, even better, because
https://github.com/status-im/nimbus-eth1/blob/775421c4862fe95de9933fcdd0c457c6dbbdf07f/nimbus_verified_proxy/nimbus_verified_proxy_conf.nim#L19-L22
is an enum, can exploit [the exhaustiveness of case statements/expressions](https://nim-lang.org/docs/manual.html#statements-and-expressions-case-statement), and get compile-time checking that all cases are handled, too:
```nim
proc init*(T: type JsonRpcBackend, url: Web3Url): T =
  var backend =
    case url.kind
    of HttpUrl:
      JsonRpcBackend(
        kind: Http,
        httpClient: newRpcHttpClient(),
        url: url.web3Url
    of WsUrl:
      JsonRpcBackend(
        kind: WebSocket,
        wsClient: newRpcWebSocketClient(),
        url: url.web3Url
```

nimbus_verified_proxy/engine/engine.nim
tersec, line 1: How does this file relate to
https://github.com/status-im/nimbus-eth1/blob/775421c4862fe95de9933fcdd0c457c6dbbdf07f/nimbus_verified_proxy/engine/engine.nim#L1-L31
?

It seems almost identical, except for the `import` formatting and that this version appears to construct a `let engine = RpcVerificationEngine` to return but never actually returns it, which causes Nim to by default return `default(T)`, or `default(RpcVerificationEngine)`.

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
```diff
@@ -57,17 +57,46 @@ type VerifiedProxyConf* = object
...
+    defaultValueDesc: "128",
```
tersec: With all these `defaultValue`/`defaultValueDesc` pairs, two points:

1. when the option is `hidden`, it's fine to just leave out the `defaultValueDesc`, or even the `desc` fields. They'll never be visible anyway; that's what `hidden` means, that they don't show up in `--help`. If you find it useful as a sort of internal code documentation, that's fine too, but that's all they're really doing.

2. but more generally, if or when one does want to have these `defaultValue: foo` and `defaultValueDesc: $foo` pairs, where it's the same `foo` in each case, then something like https://github.com/status-im/nimbus-eth2/blob/27ef2f10ab8c05ea3794b57fab5d33bcec712048/beacon_chain/conf.nim#L47-L56 where
```nim
const
  # TODO: How should we select between IPv4 and IPv6
  # Maybe there should be a config option for this.
  defaultAdminListenAddress* = (static parseIpAddress("127.0.0.1"))
  defaultSigningNodeRequestTimeout* = 60
  defaultBeaconNode* = "http://127.0.0.1:" & $defaultEth2RestPort
  defaultBeaconNodeUri* = parseUri(defaultBeaconNode)
  defaultGasLimit* = 45_000_000
  defaultAdminListenAddressDesc* = $defaultAdminListenAddress
  defaultBeaconNodeDesc = $defaultBeaconNode
```
can then be (re)used symbolically in both `defaultValue`/`defaultValueDesc` pairs so they can never get out of sync with each other.

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
chirag-parmar: oh okay. I don't think any of the debug options I marked are ephemeral. I certainly wouldn't want them to be removed without notice. But they aren't also options exposed to the user. I think i'll just remove the prefix and mark them hidden.

nimbus_verified_proxy/engine/accounts.nim
chirag-parmar: good point. I'll change them

nimbus_verified_proxy/nimbus_verified_proxy.nim
chirag-parmar: good catch

nimbus_verified_proxy/engine/engine.nim
chirag-parmar: relate to what?

it does return engine(EDIT: as per our private discussion), no? or is this some special case scenario where I have to explicitly mention `return`?

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
tersec: It's not that they're intrinsically ephemeral or would be removed without (your) notice, it's basically a question of are they indefinitely end-user supportable. If they're not something we want to commit to, indefinitely, they should have `--debug-foo`.

Anything without `--debug-foo`, with some exceptions (we'll remove Holesky soon for example, but that's external support disappearing in a coordinated way across the ecosystem) might becomes a permanent compatibility surface. In some cases that's exactly what one wants -- that's how people use software. In some cases, not as much.

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -39,6 +39,23 @@ proc cleanup*(ctx: ptr Context) =
   dealloc(ctx.configJson)
   freeShared(ctx)
 
+proc verifyChaindId(engine: RpcVerificationEngine): Future[void] {.async.} =
+  let localId = engine.chainId
```
tersec: Copying a `UInt256` is somewhat involved. One approach we can use is:
```nim
  template localId: untyped = engine.chainId`
```
to get a local variable-like ergonomics but without the inefficiency.

nimbus_verified_proxy/engine/engine.nim
tersec: Ah, this was mistaken/confused, yeah.

nimbus_verified_proxy/json_rpc_frontend.nim
```diff
@@ -0,0 +1,195 @@
...
+    parsedUrl = parseUri(url.web3Url)
+    hostname = if parsedUrl.hostname == "": "127.0.0.1" else: parsedUrl.hostname
+    port =
+      if parsedUrl.port == "":
+        8545
+      else:
+        parseInt(parsedUrl.port)
+    listenAddress = initTAddress(hostname, port)
+
+  if url.kind == HttpUrl:
```
tersec: `HttpUrl`/`WsUrl` are `enum` values, so it's nicer to use `case` rather than `if` to exploit Nim's exhaustiveness compile-time exhaustiveness checking, and separately to use an expression to avoid the semi-initialized `server` at the top which might accidentally get used as the code gets maintained e.g., remove `  var server: JsonRpcServer` and replace this section with
```nim
  case url.kind
  of HttpUrl:
    JsonRpcServer(
      kind: Http, httpServer: newRpcHttpServer([listenAddress], RpcRouter.init(), auth)
    )
  of WsUrl:
    var server =
      JsonRpcServer(kind: WebSocket, wsServer: newRpcWebSocketServer(listenAddress))

    server.wsServer.router = RpcRouter.init()
    server
```
which constructs the server and returns it in one step, minimizing the accessible intermediate-in-construction stages.

nimbus_verified_proxy/json_rpc_frontend.nim
```diff
@@ -0,0 +1,195 @@
...
+    server =
+      JsonRpcServer(kind: WebSocket, wsServer: newRpcWebSocketServer(listenAddress))
+
+    server.wsServer.router = RpcRouter.init()
+
+  server
+
+proc start*(server: JsonRpcServer): Result[void, string] =
+  try:
+    if server.kind == Http:
```
tersec: `Http` and `WebSocket` are, similarly, `enum` values, so `case` is preferred for its exhaustiveness checking to `if`:
```nim
    case server.kind
    of Http:      server.httpServer.start()
    of WebSocket: server.wsServer.start()
```
This way, if the possible kinds of servers ever change (either increase or decrease, doesn't matter), then Nim will catch it automatically rather than at runtime with some parts of the code mysteriously not triggering because they don't match some new enum value.

nimbus_verified_proxy/json_rpc_frontend.nim
```diff
@@ -0,0 +1,195 @@
...
+      server.httpServer.start()
+    elif server.kind == WebSocket:
+      server.wsServer.start()
+  except CatchableError as e:
+    return err(e.msg)
+
+  ok()
+
+proc getServer(server: JsonRpcServer): RpcServer =
```
tersec: This can be `func`

nimbus_verified_proxy/engine/blocks.nim
```diff
@@ -83,20 +83,20 @@ proc walkBlocks(
...
-      if vp.headerStore.contains(nextHash):
-        vp.headerStore.get(nextHash).get()
+      if engine.headerStore.contains(nextHash):
+        engine.headerStore.get(nextHash).get()
       else:
         let blk =
           try:
-            await vp.rpcClient.eth_getBlockByHash(nextHash, false)
+            await engine.backend.eth_getBlockByHash(nextHash, false)
           except CatchableError as e:
```
tersec, line 100: I'm going to semi-arbitrarily use this `CatchableError` to ask, is there really no more specific set of errors to catch, across this PR/verified proxy codebase? Ideally combined with async exception specifications we would be able to have a compile-time verification of all the possible exceptions that could be thrown.

nimbus_verified_proxy/engine/utils.nim
```diff
@@ -0,0 +1,25 @@
+# nimbus_verified_proxy
+# Copyright (c) 2025 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [], gcsafe.}
+
+import stint, chronos, ./types, chronicles
```
tersec: This doesn't appear to log anything? Chronicles will get reported as used because of `logScope`, but it doesn't appear to be doing anything else.

Similarly, this module doesn't do anything asynchronous or use a `Future` or an `await`. Not sure why it imports `chronos` either.

nimbus_verified_proxy/engine/utils.nim
```diff
@@ -0,0 +1,25 @@
...
+  if chainId == 1.u256:
+    ok(1.u256)
+  elif chainId == 11155111.u256:
+    ok(11155111.u256)
+  elif chainId == 17000.u256:
+    ok(17000.u256)
+  elif chainId == 560048.u256:
+    ok(560048.u256)
+  else:
+    return err("Unknown chainId")
```
tersec: Don't need `return` here, since it's already set up as an `if` expression rather than an `if` statement.

nimbus_verified_proxy/engine/utils.nim
tersec: Or for that matter `./types`? It's basically a `results` + `stint` dependency right now. If it needs more those can be added as needed.

nimbus_verified_proxy/engine/blocks.nim
arnetheduck: nim-json-rpc raises CatchableError throughtout, it has not yet been updated to `async: (raises..)` afair

nimbus_verified_proxy/json_rpc_frontend.nim
chirag-parmar: good point

nimbus_verified_proxy/engine/blocks.nim
chirag-parmar: I didn't have this specific knowledge but it explains the problem. I mostly relied on the compiler, which always complained `unlisted exception: CatchableError`, so.... I added the try catch block (@tersec if you remember the C codegen error with futures and BlockTag a few months back, that's when I tidied up all these things)

tersec: Yes, incidentally, that's been addressed and backported to `version-2-2`, so Nim v2.2.6 will fix it:
- https://github.com/nim-lang/Nim/pull/25160

tersec: > nim-json-rpc raises CatchableError throughtout, it has not yet been updated to `async: (raises..)` afair

In any case, will resolve this then, fixing `nim-json-rpc` or waiting for said fix is beyond the scope of this PR; the answer is basically, no, there's not, so these `CatchableError`s are the best one can do for the moment.

nimbus_verified_proxy/engine/utils.nim
chirag-parmar: good catch. Initially, there were more procedures/functions within this file that used those imports. Will remove them

nimbus_verified_proxy/nimbus_verified_proxy.nim
arnetheduck: that said, almost every time you do catch CatchableError in async code, you must also propagate CancelledError, ie catch and reraise it: https://www.youtube.com/watch?v=1bBVQaAD_jI

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
chirag-parmar: noted

nimbus_verified_proxy/nimbus_verified_proxy.nim
chirag-parmar: This code was just copied over from another file. I rectified the issue though. Adding a template would also increase the compilation time better to just use `engine.chainId` directly.

nimbus_verified_proxy/nimbus_verified_proxy.nim
chirag-parmar: Did you intend to attach some other link to this comment? or is there a metaphorical meaning to the video?

kdeme: :oncoming_police_car:

nimbus_verified_proxy/nimbus_verified_proxy.nim
tersec: Sure, either works.

nimbus_verified_proxy/engine/accounts.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3709/commits/606c8db86935950a420da304e808693c974cb152

nimbus_verified_proxy/nimbus_verified_proxy.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3709/commits/606c8db86935950a420da304e808693c974cb152 @arnetheduck returning cancelled error now. changed it wherever `CatchableError` was caught but no other error was propagated.

PR #3712, opened 2025-09-24
Port eth2 binary `main` style to eth1

arnetheduck:
Making the two projects more similar makes it easier to maintain a cohesive user experience over time:

* remove metrics logging - this is not viable for the amount of metrics that we have (no other project has this either)
* remove `EthContext` - looks like a refactoring leftover
* reuse more `nimbus_binary_common` utilities instead of homegrown versions
* bump eth2

In particular, these changes are sufficient to support #3646 .


arnetheduck:
Draft until https://github.com/status-im/nimbus-eth2/pull/7526 has been merged

execution_chain/config.nim
```diff
@@ -320,7 +294,7 @@ type
       separator: "\pPERFORMANCE OPTIONS",
       defaultValue: 0,
       desc: "Number of worker threads (\"0\" = use as many threads as there are CPU cores available)"
-      name: "num-threads" .}: uint
+      name: "num-threads" .}: int
```
tersec, line 297: Why allow negative numbers? If anything, eth2 should disallow them

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -305,9 +305,18 @@ proc run*(
...
 
-when isMainModule:
-  {.pop.}
-  var config =
-    makeBannerAndConfig("Nimbus verified proxy " & fullVersionStr, VerifiedProxyConf)
-  {.push raises: [].}
+# noinline to keep it in stack traces
+proc main() {.noinline, raises: [CatchableError].} =
+  const
+    banner = "Nimbus verified proxy " & fullVersionStr
```
chirag-parmar: Can we import `execution_chain/version_info` and use `FullVersionStr`? proxy will remain in sync with execution_chain that way

tersec:
> Draft until [status-im/nimbus-eth2#7526](https://github.com/status-im/nimbus-eth2/pull/7526) has been merged

Merged.

nimbus_verified_proxy/nimbus_verified_proxy.nim
kdeme: Also `nimbus_portal_client` will require this eventually (fairly soon). Should be sufficient to import just `FullVersionStr` from execution client version file?

execution_chain/networking/netkeys.nim
```diff
@@ -0,0 +1,52 @@
...
+import std/[strutils, os], stew/[io2, byteutils], results, eth/common/keys
+
+proc containsOnlyHexDigits(hex: string): bool =
+  const HexDigitsX = HexDigits + {'x'}
+  for c in hex:
+    if c notin HexDigitsX:
+      return false
+  true
+
+proc getNetKeys*(rng: var HmacDrbgContext, netKey: string): Result[KeyPair, string] =
```
kdeme, line 21: portal client has code with a similar goal but obviously differently implemented :sweat_smile: : https://github.com/status-im/nimbus-eth1/blob/master/portal/common/common_utils.nim#L81

Probably something we want to also make common in another iteration (and potentially with rework of nimbus-eth2 keystore to also use that...)

nimbus_verified_proxy/nimbus_verified_proxy.nim
chirag-parmar: bette done that way because the releases will be clubbed together. UPDATE: And it makes more sense to maintain one version file rather than three

PR #3714, opened 2025-09-25
Fix sync performance regression

bhartnett:
This reverts the recent txFrame HashKeys (kMap) memory optimization change. 

It turns out that this change slowed down the sync speed because when syncing we check the state root for most blocks (when the finalized flag is not set) not just the block to be persisted.

When running the 'Replay mainnet era, multiple FCU' forked chain test I found that disabling the memory optimization feature provided a large speed up.

Also manually tested the sync on hoodi before and after and the sync improved a bit. Before was getting around 80 blocks imported per log message (approx every 5 seconds) and after got around 120 blocks per log message. The improvement likely depends on the workload.


arnetheduck:
Hold on, but the two things are orthogonal: creating snapshots and checking state root. We  can skip the state root check and still perform this optimization.

arnetheduck:
What I mean is that I don't understand why the memory optimization would have this impact - I think it's worth investigating this more before reverting the change - ie running both versions through vtune and see what's different, for example.

When we're not verifying the state root for every block, like we do when syncing (ie we verify every `persistBatchSize`), there should be no keys to move so the memory optimization should effectively be a no-op. The mem-opt should actually only come into practical play when we're synced past finality, which is also the time when we might encounter forks in general, which is where it's extra important to conserve memory.

bhartnett:
@arnetheduck It turns out that the slow down in the stateroot computation is caused by this: https://github.com/status-im/nimbus-eth1/blob/742e2298bf5bd2753a868832b92e8605d1e04b68/execution_chain/db/aristo/aristo_layers.nim#L193

When copying the content of a txFrame into a snapshot here the hashkey is set to VOID_HASH_KEY when the kMap of the frame doesn't contain the rvid. Moving the kMap causes this to occur in the parent (moved from) frame. This has the effect of forcing the key to be recalculated in the stateroot computation which slows the process down significantly. 

We should be able to fix this issue by modifying this copyFrom function in some way so that these VOID_HASH_KEY values are over written with the values that are present in the moved to frame.

PR #3717, opened 2025-09-26
Improve performance of stateroot computation when using kMap memory optimization

bhartnett:
It turns out that the slow down in the stateroot computation is caused by this: https://github.com/status-im/nimbus-eth1/blob/742e2298bf5bd2753a868832b92e8605d1e04b68/execution_chain/db/aristo/aristo_layers.nim#L193

When copying the content of a txFrame into a snapshot here the hashkey is set to VOID_HASH_KEY when the kMap of the frame doesn't contain the rvid. Moving the kMap as a memory optimisation (see [here](https://github.com/status-im/nimbus-eth1/pull/3667)) causes this to occur in the parent (moved from) frame. This has the effect of forcing the key to be recalculated in the stateroot computation which slows the process down significantly. 

The fix is to copy the moved kMap values into the snapshot vtx records.

### DIFF
--- a/tests/test_aristo/test_tx_frame.nim
+++ b/tests/test_aristo/test_tx_frame.nim
@@ -138,6 +138,55 @@ suite "Aristo TxFrame":
       tx3.kMap.len() == 0
       tx2.fetchStateRoot() == tx3.fetchStateRoot()
 
+  test "Frames using moveParentHashKeys parameter with snapshots":
+    let
+      tx0 = db.txFrameBegin(db.baseTxFrame())
+      tx1 = db.txFrameBegin(tx0)
+
+    check:
+      tx0.mergeAccountRecord(acc1[0], acc1[1]).isOk()
+      tx1.mergeAccountRecord(acc2[0], acc2[1]).isOk()
+      tx0.fetchStateRoot() != tx1.fetchStateRoot()
+      tx0.kMap.len() == 0
+      tx1.kMap.len() == 1
+    tx1.checkpoint(1, skipSnapshot = false)
+
+    let tx2 = db.txFrameBegin(tx1, moveParentHashKeys = true)
+    tx2.checkpoint(2, skipSnapshot = false)
+
+    block:
+      var hashKey: HashKey
+      for v in tx2.kMap.values():
+        if v.isValid():
+          hashKey = v
+      check hashKey != default(HashKey)
+      var snapshotContainsHashKey = false
+      for k, v in tx2.snapshot.vtx:
+        if v[1] == hashKey:
+          snapshotContainsHashKey = true
+
+      check:
+        snapshotContainsHashKey
+        tx1.fetchStateRoot() == tx2.fetchStateRoot()
+
+    let tx3 = db.txFrameBegin(tx2, moveParentHashKeys = false)
+    tx3.checkpoint(3, skipSnapshot = false)
+
+    block:
+      var hashKey: HashKey
+      for v in tx2.kMap.values():
+        if v.isValid():
+          hashKey = v
+      check hashKey != default(HashKey)
+      var snapshotContainsHashKey = false
+      for k, v in tx3.snapshot.vtx:
+        if v[1] == hashKey:
+          snapshotContainsHashKey = true
+
+      check:
+        snapshotContainsHashKey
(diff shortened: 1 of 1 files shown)


execution_chain/db/aristo/aristo_layers.nim
```diff
@@ -192,6 +192,10 @@ proc copyFrom*(snapshot: var Snapshot, tx: AristoTxRef) =
     do:
       snapshot.vtx[rvid] = (vtx, VOID_HASH_KEY, tx.level)
 
+  for rvid, key in tx.kMap:
```
arnetheduck, line 195: this looks like it has the unfortunate effect of sometimes repeating the assignment (for the values in `sTab`) .. not sure if there's a way to avoid that, that isn't costly (an extra sTab lookup is costly for example)

arnetheduck (approved):

execution_chain/db/aristo/aristo_layers.nim
arnetheduck: aka can we do this "somewhere else" in the pipeline and avoid some of the double work?

kdeme:
I have been testing this branch briefly on mainnet sync:

**before**
```
INF 2025-09-26 08:04:14.839+00:00 Imported blocks                            topics="beacon sync" nImported=20 nUnproc=20069 nStagedQ=0 base=#23413793 head=#23413923 target=#23433992 targetHash=002fa72be241 nSyncPeers=1
INF 2025-09-26 08:05:24.216+00:00 Imported blocks                            topics="beacon sync" nImported=20 nUnproc=20049 nStagedQ=0 base=#23413813 head=#23413943 target=#23433992 targetHash=002fa72be241 nSyncPeers=1
INF 2025-09-26 08:06:53.790+00:00 Imported blocks                            topics="beacon sync" nImported=24 nUnproc=20025 nStagedQ=0 base=#23413837 head=#23413967 target=#23433992 targetHash=002fa72be241 nSyncPeers=1
INF 2025-09-26 08:08:14.611+00:00 Imported blocks                            topics="beacon sync" nImported=20 nUnproc=20005 nStagedQ=0 base=#23413857 head=#23413987 target=#23433992 targetHash=002fa72be241 nSyncPeers=1
INF 2025-09-26 08:09:44.753+00:00 Imported blocks                            topics="beacon sync" nImported=25 nUnproc=19980 nStagedQ=0 base=#23413881 head=#23414012 target=#23433992 targetHash=002fa72be241 nSyncPeers=2
```

Reaching between 0.25 and 0.30 bps (based on log timestamps)

**With this PR**

```
INF 2025-09-26 08:45:53.850+00:00 Imported blocks                            topics="beacon sync" nImported=40 nUnproc=31653 nStagedQ=2 base=#23414193 head=#23414323 target=#23445976 targetHash=7255e7ef6230 nSyncPeers=1
INF 2025-09-26 08:46:14.274+00:00 Imported blocks                            topics="beacon sync" nImported=41 nUnproc=31612 nStagedQ=0 base=#23414233 head=#23414364 target=#23445976 targetHash=7255e7ef6230 nSyncPeers=1
INF 2025-09-26 08:46:51.574+00:00 Imported blocks                            topics="beacon sync" nImported=61 nUnproc=31551 nStagedQ=0 base=#23414293 head=#23414425 target=#23445976 targetHash=7255e7ef6230 nSyncPeers=1
INF 2025-09-26 08:47:25.762+00:00 Imported blocks                            topics="beacon sync" nImported=49 nUnproc=31502 nStagedQ=0 base=#23414345 head=#23414474 target=#23445976 targetHash=7255e7ef6230 nSyncPeers=1
INF 2025-09-26 08:48:01.886+00:00 Imported blocks                            topics="beacon sync" nImported=55 nUnproc=31447 nStagedQ=0 base=#23414397 head=#23414529 target=#23445976 targetHash=7255e7ef6230 nSyncPeers=1

```

Reaching between 1.4 and 2 bps on a short test.

execution_chain/db/aristo/aristo_layers.nim
bhartnett: Hmm, yeah that's true but I can't think of a better place to put this logic. Maybe it's ok here because this only runs when creating a snapshot and generally we are only dealing with two frames when building the snapshot (the parent and the current).

PR #3721, opened 2025-09-26
Speed up sync speed by skipping state root checks

bhartnett:
If I'm not mistaken we should be able to skip the stateroot checks when importing blocks via the syncer because the header chain should have already been verified and therefore the blocks are known to be finalized.

This sets the finalized flag to true when calling `importBlock` on the `ForkedChain` module. This flag basically just disables the stateroot check. Note that the stateroot is still always checked before persisting to the database every `persistBatchSize` blocks.

### DIFF
--- a/execution_chain/sync/beacon/worker/blocks/blocks_import.nim
+++ b/execution_chain/sync/beacon/worker/blocks/blocks_import.nim
@@ -37,8 +37,13 @@ proc importBlock*(
       B=ctx.chain.baseNumber.bnStr, L=ctx.chain.latestNumber.bnStr
   else:
     try:
-      (await ctx.chain.queueImportBlock(blk, finalized = true)).isOkOr:
-        return err((ENoException,"",error,Moment.now()-start))
+      # At this point the header chain has already been verifed and so we know
+      # the block is finalized as long as the block number is less than or equal
+      # to the latest finalized block. Setting the finalized flag to true here
+      # has the effect of skipping the stateroot check for performance reasons.
+      let isFinalized = blk.header.number <= ctx.chain.latestFinalizedBlockNumber
+      (await ctx.chain.queueImportBlock(blk, isFinalized)).isOkOr:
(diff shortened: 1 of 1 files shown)


execution_chain/sync/beacon/worker/blocks/blocks_import.nim
```diff
@@ -37,7 +37,7 @@ proc importBlock*(
       B=ctx.chain.baseNumber.bnStr, L=ctx.chain.latestNumber.bnStr
   else:
     try:
-      (await ctx.chain.queueImportBlock blk).isOkOr:
+      (await ctx.chain.queueImportBlock(blk, finalized = true)).isOkOr:
```
advaita-saha: Would be nice to have a comment here why we are doing this

arnetheduck:
Interesting, I thought it was already doing this but it must have been lost in translation .. anyway, the block is finalized only if its number is `< latestFinalizedBlockNumber` - merely coming from the syncer is not enough.

execution_chain/sync/beacon/worker/blocks/blocks_import.nim
bhartnett: Sure, will update.

advaita-saha: https://github.com/status-im/nimbus-eth1/pull/3721/commits/cfa8e97fad48dca50932ac30bd8e57e60a86fa54

mjfh:
In the current implementation, the syncer does not know whether the block is finalised or not. It will opportunistically try to import everything. Setting the import flag `true` is  not correct as the syncer does not know.

mjfh (commented):
The `finalised` flag is meaningless for the syncer.

All blocks passed on form the syncer to the `FC` module ate tainted because the bodies are fetched from untrusted sources. They must be sanitised/checked by the `FC` module (which is in a better position to do that than the syncer.)

PR #3730, opened 2025-09-29
Remove call to dispose when block validation fails

bhartnett:
The call to dispose is causing a segfault during block import because it appears that the same block may get imported/processed multiple times if the validation fails the first time. On the second validation there is a crash when creating the snapshot because the txFrame has already been disposed.

### DIFF
--- a/execution_chain/core/chain/forked_chain.nim
+++ b/execution_chain/core/chain/forked_chain.nim
@@ -486,9 +486,7 @@ proc validateBlock(c: ForkedChainRef,
   # the stateroot computation.
   c.updateSnapshot(blk, txFrame)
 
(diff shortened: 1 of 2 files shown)


execution_chain/core/chain/forked_chain.nim
```diff
@@ -486,9 +486,7 @@ proc validateBlock(c: ForkedChainRef,
   # the stateroot computation.
   c.updateSnapshot(blk, txFrame)
 
-  var receipts = c.processBlock(parent, txFrame, blk, blkHash, finalized).valueOr:
-    txFrame.dispose()
```
bhartnett, line 490: @jangko What was the purpose of having the dispose call here? Do you think it would be ok to remove it?

bhartnett:
It turns out that this is not the ideal fix. I have an alternative PR coming so closing this.

PR #3736, opened 2025-10-02
Fix segfault when database directory is locked / in use

bhartnett:


execution_chain/nimbus_execution_client.nim
```diff
@@ -341,12 +341,12 @@ proc main*(config = makeConfig(), nimbus = NimbusNode(nil)) {.noinline.} =
     except CancelledError:
       raiseAssert "Never cancelled"
   defer:
-    waitFor metricsServer.stopMetricsServer()
+    if metricsServer.isSome():
```
arnetheduck, line 344: doesn't make sense: https://github.com/status-im/nimbus-eth2/blob/stable/beacon_chain/nimbus_binary_common.nim#L378

bhartnett: I see. Well this did prevent/fix the seg fault in my testing locally for some reason.

arnetheduck: might need another round of digging then - probably just changed one UB for another

bhartnett: After reverting my change the seg fault is triggered when running two Nimbus ELs on the same data dir. This is the stack trace:
```
Traceback (most recent call last, using override)
/home/user/development/status-im/nimbus-eth1/execution_chain/nimbus_execution_client.nim(371) nimbus_execution_client
/home/user/development/status-im/nimbus-eth1/execution_chain/nimbus_execution_client.nim(352) _ZN23nimbus_execution_client4mainEN6config10NimbusConfE3refIN11nimbus_desc26NimbusNodecolonObjectType_EE
/home/user/development/status-im/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/excpt.nim(438) nimLeaveFinally
/home/user/development/status-im/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/excpt.nim(428) _ZN6system20reportUnhandledErrorE3refIN6system9ExceptionEE
/home/user/development/status-im/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/excpt.nim(380) _ZN6system23reportUnhandledErrorAuxE3refIN6system9ExceptionEE
/home/user/development/status-im/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/excpt.nim(652) signalHandler
/home/user/development/status-im/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/excpt.nim(334) _ZN6system18rawWriteStackTraceE3varI6stringE
/home/user/development/status-im/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system/stacktraces.nim(59) _ZN11stacktraces30auxWriteStackTraceWithOverrideE3varI6stringE
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
Segmentation fault (core dumped)
```

Turns out that it fails at the end of the defer block in the nimLeaveFinally nim function:
```
  defer:
    # if metricsServer.isSome():
    let fut = metricsServer.stopMetricsServer()

    waitFor fut
```

Might be a nim bug perhaps?

arnetheduck: sounds like it, yes - check 2.2.6, else this needs investigation and reporting

bhartnett: I just tested using the recent nim 2.2.6 nimbus-build-system update and confirmed that the problem is fixed when using that version of nim. I guess the problem was a nim bug after all. I can see 2 SIGSEGV related fixes in the release notes: https://nim-lang.org/blog/2025/10/31/nim-226.html

arnetheduck: yeah, unfortunately it looks like there's a new bug in 2.2.6, that needs investigation: https://github.com/status-im/nimbus-eth1/actions/runs/19056332522/job/54427222420?pr=3809

tersec: Yes, already looking at this one.

mjfh: I had this for my internal PR documentation with the suggestion to resolve the `defer` and move the destructor to the end of the `main()` function:

      When an uncaught exception (e.g. from an `assert`) is thrown,
      moving the `defer` to the end it  makes the difference. With an
      explicitly stated destructor at the end, it will  not be run in case
      of an exception.
    
      Unfortunately, the metrics destructor will in many cases crash with a
      SIGFAULT after an exception was thrown. It will provoke a secondary
      function stack trace for the SEGFAULT which will overwrite any
      previous stack trace for the primary exception.

We discussed this at the team meeting three weeks ago.

bhartnett: The stack trace was only lost/replaced because of the nim bug which caused a SIGFAULT when running the defer/finally block. Since the nim bug is now fixed then this lost stack trace should no longer be an issue and so I don't see any reason to restructure the code.

PR #3737, opened 2025-10-02
Remove snapshots cache from ForkedChain

bhartnett:


arnetheduck:
> Switching branches 

It's not so much about switching branches, but rather creating a new branch from a recent block, _most often_ `head - 1` - ie the vast majority of branches happen because the "current head" gets orphaned or that the proposer is a block late and tries to build on the previous head instead.

PR #3741, opened 2025-10-03
Fix OverflowDefect crash in block validation

bhartnett:
This fixes an underflow defect crash which was triggered when `c.baseDistance + c.persistBatchSize` is less than `latestFinalizedBlockNumber`. 

I also found that the `queueUpdateBase` function had the const `PersistBatchSize` value hard coded so updates were still applied every 4 blocks even when setting a custom `persistBatchSize`.

Resolves https://github.com/status-im/nimbus-eth1/issues/3507

### DIFF
--- a/execution_chain/core/block_import.nim
+++ b/execution_chain/core/block_import.nim
@@ -83,7 +83,7 @@ proc importRlpBlocks*(importFile: string,
 proc importRlpBlocks*(conf: NimbusConf, com: CommonRef): Future[void] {.async: (raises: [CancelledError]).} =
   # Both baseDistance and persistBatchSize are 0,
   # we want changes persisted immediately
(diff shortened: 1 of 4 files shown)


bhartnett:
I was curious to see how the persist batch size parameter setting can effect the import performance and so ran a few tests which all imported the first 50000 blocks on hoodi. Here are the results:

**Test 1**
Batch size - 4 (default)
Run time - 41 mins

**Test 2**
Batch size - 100
Run time - 41 mins

**Test 3**
Batch size - 1000
Run time - Losing peers so cancelled the test

**Test 4**
Batch size - 20
Run time - 33 mins

It appears that increasing the batch size too much has a negative effect on performance due to the loss of peers and issues with networking but a small increase in the batch size may actually improve performance (depending on the size of the blocks). The performance improvement when using a batch size of 20 is likely due to the less frequent database persist and stateroot computations.

PR #3747, opened 2025-10-06
Use mcl to boost alt_bn128 precompiles performance

### LINKED ISSUE #2713: Add `mcl` (and possibly `constantine`) backends for alt_bn128 precompiles
One of the largest bottlenecks during syncing post [eip-1108](https://eips.ethereum.org/EIPS/eip-1108) are the bncurve precompiles - nim-bncurve can no longer keep up.

As an alternative, we could switch to `mcl` which is a classic battletested option that is [fast](https://hackmd.io/@gnark/eccbench).

Worth exploring is also [constantine](https://github.com/mratsim/constantine), though this should be added only once we have an mcl backend, so that we can run comparative analysis.

`nim-bncurve` should be kept as a pure-nim fallback option.

### LINKED ISSUE #3378: Perf: EcAdd precompile extremely slow
## Description

https://github.com/NethermindEth/gas-benchmarks reveal nimbus `EcAdd` precompiles perform really badly. Below screenshot is a comparison against other clients. 

Image

Dashboard Link: https://grafana.observability.ethpandaops.io/d/feo4ronhsqv40d/opcodes-benchmarking?orgId=1&from=now-6h&to=now&timezone=browser&var-posgreSQL=benuragv7iuwwb&var-ClientName=$__all&var-TestTitle=EcAdd%20with%20%280,%200%29&var-TestTitle=EcAdd%20with%20%281,%202%29&var-TestTitle=EcAdd%20with%2032-byte%20coordinates&var-TestTitle=EcAdd12CACHABLE&var-TestTitle=EcAdd32ByteCoordinatesCACHABLE&var-TestTitle=EcAddMarius1&refresh=auto

jangko:
Linux amd64, with and without avx512 instruction:
|Operation    |      Nim-bn254 |    mcl-x64    | mcl-avx512    | nim->mcl-x64(%) | nim->mcl-avx512(%) |
|:------------|---------------:|--------------:|--------------:|----------------:|-------------------:|
|G1 + G1      |    511.494.241 |     4.714.393 |     4.834.672 | faster 99,08           | faster 99,05 |
|G1 * FR	    |    718.740.303 |    70.999.718 |    71.017.095 | faster 90,12           | faster 90,12 |
|G1 G2 pairing|	25.122.616.596 | 5.310.392.374 | 4.661.134.042 | faster 78,86           | faster 81,45 |

No data available for arm64 yet.

TODO:
- [x] nightly/release build for arm64
- [x] docker build for arm64

fix #3378
fix #2713

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,9 +73,14 @@ jobs:
         run: |
           gcc --version
           #export ZERO_AR_DATE=1 # avoid timestamps in binaries
-          DEFAULT_MAKE_FLAGS="-j${ncpu} ROCKSDB_CI_CACHE=RocksBinCache"
+          if [[ '${{ matrix.cpu }}' == 'arm64' ]]; then
+            # mcl on arm64 needs clang compiler
+            NIMFLAGS="--cc:clang"
+          fi
+          DEFAULT_MAKE_FLAGS="-j${ncpu} ROCKSDB_CI_CACHE=RocksBinCache NIMFLAGS=${NIMFLAGS}"
           make ${DEFAULT_MAKE_FLAGS} all test_import build_fuzzers check_revision
           # clear some space
           find . -type d -name "nimcache" -exec rm -rf {} +
           # "-static" option will not work for osx unless static system libraries are provided
           make ${DEFAULT_MAKE_FLAGS} test t8n_test eest_full_test
+
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,11 +73,7 @@ jobs:
         run: |
           gcc --version
           #export ZERO_AR_DATE=1 # avoid timestamps in binaries
-          #if [[ '${{ matrix.cpu }}' == 'arm64' ]]; then
-          #  # mcl on arm64 needs clang compiler
-          #  NIMFLAGS="--cc:clang"
-          #fi
-          DEFAULT_MAKE_FLAGS="-j${ncpu} ROCKSDB_CI_CACHE=RocksBinCache NIMFLAGS=${NIMFLAGS}"
+          DEFAULT_MAKE_FLAGS="-j${ncpu} ROCKSDB_CI_CACHE=RocksBinCache"
           make ${DEFAULT_MAKE_FLAGS} all test_import build_fuzzers check_revision
           # clear some space
           find . -type d -name "nimcache" -exec rm -rf {} +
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -73,9 +73,14 @@ jobs:
         run: |
           gcc --version
           #export ZERO_AR_DATE=1 # avoid timestamps in binaries
-          DEFAULT_MAKE_FLAGS="-j${ncpu} ROCKSDB_CI_CACHE=RocksBinCache"
+          if [[ '${{ matrix.cpu }}' == 'arm64' ]]; then
+            # mcl on arm64 needs clang compiler
```
tersec: Is this true for arm64 Linux as well?

jangko: Yes, mcl have no arm64's assembly module, only provide clang .ll intermediate representation. And no C fallback too.

tersec: https://llvm.org/docs/DeveloperPolicy.html#ir-backwards-compatibility notes
> The textual format is not backwards compatible. We don’t change it too often, but there are no specific promises.

So this would introduce some potentially specific compatibility issues (not supporting both slightly older/newer clang versions for example, or preventing updates to newer clang versions because mcl isn't ready).

https://llvm.org/docs/OpaquePointers.html#version-support describes one example. There's some deprecation period, but especially if wanting to support both oldest-available-clang-from-almost-5-years-ago and newer clangs might be challenge. For example, https://packages.ubuntu.com/jammy/clang from Ubuntu 22.04 (EOL 2027) is clang 14:
> However, using opaque pointers in the optimization pipeline is **not** fully supported. This release can be used to make out-of-tree code compatible with opaque pointers, but opaque pointers should **not** be enabled in production.

while https://packages.debian.org/trixie/clang from current Debian stable is clang 19:
> Only opaque pointers are supported. Typed pointers are not supported.

I'm not sure how it'd be possible to support building on both at once.

jangko: in that case, mcl have the generator, and we can regenerate the .ll file if needed

tersec: At any given time, though, wouldn't one or the other versions exist? Then the same issue exists, that those .ll files either work on Ubuntu 22.04 LTS or modern Debian or Ubuntu, but not both. Unless we start simply requiring .ll file regeneration as part of the build process (which also requires clang to take advantage of, of course).

jangko: https://github.com/status-im/nimbus-eth1/actions/runs/18305852975
https://github.com/status-im/nim-mcl/pull/5
https://github.com/status-im/nimbus-eth1/pull/3747/commits/a7551a78752ea3f5e066feab92b6d4449061d48c

Shipping .o files for linux arm64 seems reduce the complexity

tersec: Ok, it's a least-bad approach. I guess it re-enables gcc-based building too?

jangko: Yes

PR #3750, opened 2025-10-07
Implement dynamic adjustment of ForkedChain persistBatchSize

bhartnett:


execution_chain/core/chain/forked_chain.nim
```diff
@@ -352,6 +353,31 @@ with --debug-eager-state-root."""
...
+      targetTime = 1 # seconds
+      batchSizeLowerBound = 5
+      batchSizeUpperBound = 995
+
+    let
+      finishTime = EthTime.now()
+      runTime = finishTime - startTime
+
+
+    if runTime < targetTime and c.persistBatchSize <= batchSizeUpperBound:
```
jangko: `EthTime.now()` resolution is one second. `runTime < targetTime(1 second)` will not works. And `persistBatchsize` will never increased, right?

bhartnett: It does work because if the run time is less than one second then runTime = 0. In other words it appears to truncate or round down the value since the precision only supports seconds. 

I've been testing it and verified that both increasing the size and decreasing the batch size does indeed work.

jangko: Maybe add some comment to avoid confusion?

bhartnett: Sure, will do

bhartnett: Actually, I might update this to measure the time in milliseconds instead of seconds.

bhartnett: I've updated the code to use milliseconds precision instead.

PR #3751, opened 2025-10-07
don't trust values from p2p layers

### LINKED ISSUE #3677: Sanitize remote values before using
https://github.com/status-im/nimbus-eth1/blob/58c297fb7bd92b099151f12409e76aebdd05372e/execution_chain/sync/wire_protocol/handler.nim#L190 - if a remote peer sends a large number, this will over-allocate and crash the application

advaita-saha:
Fixes #3677

### DIFF
--- a/execution_chain/sync/wire_protocol/handler.nim
+++ b/execution_chain/sync/wire_protocol/handler.nim
@@ -187,7 +187,7 @@ proc getBlockHeaders*(ctx: EthWireRef,
     chain = ctx.chain
 
   var
-    list = newSeqOfCap[Header](req.maxResults)
+    list = newSeqOfCap[Header](min(req.maxResults, MAX_HEADERS_SERVE))
     header = chain.blockHeader(req.startBlock).valueOr:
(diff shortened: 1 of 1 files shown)


execution_chain/sync/wire_protocol/handler.nim
```diff
@@ -187,7 +187,7 @@ proc getBlockHeaders*(ctx: EthWireRef,
     chain = ctx.chain
 
   var
-    list = newSeqOfCap[Header](req.maxResults)
+    list = newSeqOfCap[Header](min(req.maxResults, MAX_HEADERS_SERVE))
```
bhartnett, line 190: I wonder if there are more locations similar to this that need to be handled correctly as well?

advaita-saha: Yeah I did look for them at a glace to look for other places
Couldn't find anyplace else with these type of problem ( might be another type )

I am going through the p2p layer, will fix if I find any more

PR #3760, opened 2025-10-10
Implement snapshots cache in aristo database

bhartnett:
This PR adds a snapshots cache/queue to the aristo database instance. 

The main purpose of this change is to support limiting the number of snapshots that can be created in total and also to support clearing out old data from snapshots just after persisting. 

When we persist a txFrame to the database the base level increases but any snapshots that were not persisted may still have vertexes with a level lower than the base level. This stale data should be removed to further save memory and also prevent issues caused by reading stale data.

If we decided to go ahead with this idea then ideally it should be merged just after this PR: https://github.com/status-im/nimbus-eth1/pull/3737

### DIFF
--- a/execution_chain/core/chain/forked_chain.nim
+++ b/execution_chain/core/chain/forked_chain.nim
@@ -506,19 +506,18 @@ proc validateBlock(c: ForkedChainRef,
     parentTxFrame=cast[uint](parentFrame),
     txFrame=cast[uint](txFrame)
 
-  # Checkpoint creates a snapshot of ancestor changes in txFrame - it is an
-  # expensive operation, specially when creating a new branch (ie when blk
-  # is being applied to a block that is currently not a head).
-  # Create the snapshot before processing the block so that any vertexes in snapshots
-  # from lower levels than the baseTxFrame are removed from the snapshot before running
-  # the stateroot computation.
-  parentFrame.checkpoint(parent.blk.header.number, skipSnapshot = false)
+
 
   var receipts = c.processBlock(parent, txFrame, blk, blkHash, finalized).valueOr:
     txFrame.dispose()
     return err(error)
 
   c.writeBaggage(blk, blkHash, txFrame, receipts)
+
+  # Checkpoint creates a snapshot of ancestor changes in txFrame - it is an
+  # expensive operation, specially when creating a new branch (ie when blk
+  # is being applied to a block that is currently not a head).
+  parentFrame.checkpoint(blk.header.number, skipSnapshot = false)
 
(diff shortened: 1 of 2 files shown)


execution_chain/db/aristo/aristo_init/init_common.nim
```diff
@@ -86,6 +86,7 @@ proc initInstance*(db: AristoDbRef): Result[void, AristoError] =
   db.txRef = AristoTxRef(db: db, vTop: vTop, snapshot: Snapshot(level: Opt.some(0)))
   db.accLeaves = LruCache[Hash32, AccLeafRef].init(ACC_LRU_SIZE)
   db.stoLeaves = LruCache[Hash32, StoLeafRef].init(ACC_LRU_SIZE)
+  db.maxSnapshots = 10
```
bhartnett, line 89: This can be made configurable if/when needed.

execution_chain/core/chain/forked_chain/chain_serialize.nim
```diff
@@ -143,6 +135,11 @@ proc replayBlock(fc: ForkedChainRef;
 
   fc.writeBaggage(blk.blk, blk.hash, txFrame, receipts)
 
+  # Checkpoint creates a snapshot of ancestor changes in txFrame - it is an
+  # expensive operation, specially when creating a new branch (ie when blk
+  # is being applied to a block that is currently not a head).
+  parentFrame.checkpoint(blk.header.number, skipSnapshot = false)
```
arnetheduck, line 141: `txFrame.checkpoint`.. no?

execution_chain/core/chain/forked_chain.nim
```diff
@@ -506,20 +506,19 @@ proc validateBlock(c: ForkedChainRef,
...
   var receipts = c.processBlock(parent, txFrame, blk, blkHash, finalized).valueOr:
     txFrame.dispose()
     return err(error)
 
   c.writeBaggage(blk, blkHash, txFrame, receipts)
 
+  # Checkpoint creates a snapshot of ancestor changes in txFrame - it is an
+  # expensive operation, specially when creating a new branch (ie when blk
+  # is being applied to a block that is currently not a head).
+  parentFrame.checkpoint(blk.header.number, skipSnapshot = false)
```
arnetheduck, line 520: ..and here..

execution_chain/core/chain/forked_chain/chain_serialize.nim
bhartnett: Yes should be `txFrame.checkpoint`. Will fix in my next PR.

PR #3781, opened 2025-10-21
Add blob schedule fallback test case and fix fallback logic

### LINKED ISSUE #3764: bpo fallback option if parameters don't change
https://github.com/eth-clients/hoodi/pull/23
https://github.com/eth-clients/sepolia/pull/119
https://github.com/eth-clients/holesky/pull/136

Osaka fork does not change the amount of blobs, so its been removed from the blob schedule. The idea is that the bpo forks should be different from the regular forks. We should account for this logic.

## Action
Write tests for this, and check for bugs in the fallback code path

jangko:
fix #3764

The fallback logic only applied to chain config parsed from json file. Built in blob schedule still need explicit value.

### DIFF
--- a/execution_chain/common/chain_config.nim
+++ b/execution_chain/common/chain_config.nim
@@ -395,22 +395,52 @@ proc validateChainConfig(conf: ChainConfig): bool =
     if cur.time.isSome:
       lastTimeBasedFork = cur
 
+func numBPOForks(): int {.compileTime.} =
+  for x in Prague..HardFork.high:
+    if toLowerAscii($x).startsWith("bpo"):
+      inc result
+
+func getBPOForks(N: static[int]): array[N, HardFork] {.compileTime.} =
+  var i = 0
+  for x in Prague..HardFork.high:
+    if toLowerAscii($x).startsWith("bpo"):
+      result[i] = x
+      inc i
+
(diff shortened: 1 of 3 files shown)


advaita-saha:
> Built in blob schedule still need explicit value

Ahh nice catch

execution_chain/common/chain_config.nim
```diff
@@ -395,22 +395,52 @@ proc validateChainConfig(conf: ChainConfig): bool =
...
+      if conf.blobSchedule[fork].isNone:
+        conf.blobSchedule[fork] = conf.blobSchedule[prevFork]
+      if conf.blobSchedule[fork].value.baseFeeUpdateFraction == 0:
+        # Set fallback to Cancun's baseFeeUpdateFraction and prevent division by zero
+        warn "baseFeeUpdateFraction not set, fallback to Cancun's", fork=fork
+        conf.blobSchedule[fork].value.baseFeeUpdateFraction = 3_338_477'u64
+      prevFork = fork
+
+  setBlobScheduleWithFallback(RegularForks)
+  setBlobScheduleWithFallback(BPOForks)
```
advaita-saha, line 443: I mean good to have, but shouldn't technically need to have blob parameter fallback for a bpo fork
Which is a fork specifically designed to change only blob parameters lol

PR #3782, opened 2025-10-21
proxy: rest based light client

chirag-parmar:
meta: #3101 

Description:

This PR ports the existing light client within nimbus-eth2 for use in the verified proxy with light client updates downloaded purely via the REST API. This is beneficial for using the proxy in a browser. In doing so, the PR modifies the behaviour of the light client slightly as noted below.

Notes:
1. **Syncing**: The original light client starts with a bootstrap and syncs blocks from bootstrap to head. It then follows the chain optimistically. [here](https://github.com/status-im/nimbus-eth2/blob/stable/beacon_chain/nimbus_light_client.nim#L230) and [here](https://github.com/status-im/nimbus-eth2/blob/stable/beacon_chain/nimbus_light_client.nim#L279). The verified proxy version of the light client uses the bootstrap update and subsequent sync committee updates only for establishing the (current and next) sync committee. Once established, however, it starts following the chain directly from the current head. From the perspective of trust, optimistically following the chain from the current head is the same as syncing the chain and then following it optimistically (since the validation is only dependent on the current sync committee which doesn't change within the current sync committee period) [here](https://github.com/status-im/nimbus-eth2/blob/stable/beacon_chain/spec/light_client_sync.nim#L76)

2. **Peers and Endpoints**: The original light client uses multiple peers concurrently to download a particular update. The verified proxy version uses the same code but abstracts out rotating/randomizing peers (in the case of NVP it is API endpoints) to the backend implementation through the use of `reqId`s. The modified light client uses unique request ids for all concurrent requests which can then be used to rotate/randomize and score API endpoints. The parallelization parameter for the light client is hardcoded to 2.

3. **Downloading Updates**: The original light client used a mix of light client updates and p2p gossip to maintain the canonical chain. The verified proxy version purely uses light client updates to maintain the canonical chain. Hence, it runs the manager loop on a per slot frequency rather than the per sync committee period frequency of the original client(not fully true, it is quite arbitrary but can be considered the average case scenario)

Other Changes:

- [x] remove deprecated warnings
- [x] remove extra file (c_frontend)
- [x] fix nits (`FullVersionStr`)

### DIFF
--- a/nimbus_verified_proxy/lc/lc_manager.nim
+++ b/nimbus_verified_proxy/lc/lc_manager.nim
@@ -307,8 +307,7 @@ proc loop(self: LightClientManager) {.async: (raises: [CancelledError]).} =
   # try atleast twice
   let
     NUM_RETRIES = 2
-    RETRY_TIMEOUT =
-      chronos.seconds(int64(self.timeParams.SLOT_DURATION) div (NUM_RETRIES + 1))
+    RETRY_TIMEOUT = self.timeParams.SLOT_DURATION div (NUM_RETRIES + 1)
 
   while true:
     let
@@ -384,7 +383,7 @@ proc loop(self: LightClientManager) {.async: (raises: [CancelledError]).} =
         continue
 
     # check for updates every slot
-    await sleepAsync(chronos.seconds(int64(self.timeParams.SLOT_DURATION)))
+    await sleepAsync(self.timeParams.SLOT_DURATION)
 
 proc start*(self: var LightClientManager) =
   ## Start light client manager's loop.


nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
```diff
@@ -195,34 +147,24 @@ proc parseCmdArg*(T: type Web3Url, p: string): T {.raises: [ValueError].} =
...
-    agentString: pc.agentString,
-    discv5Enabled: pc.discv5Enabled,
-    directPeers: pc.directPeers,
-    trustedBlockRoot: pc.trustedBlockRoot,
-    web3Urls: @[EngineApiUrlConfigValue(url: pc.backendUrl.web3Url)],
-    jwtSecret: none(InputFile),
-    stopAtEpoch: 0,
-  )
+proc completeCmdArg*(T: type UrlList, val: string): seq[string] =
+  return @[]
```
tersec: Don't need `return` here

nimbus_verified_proxy/lc/lc_manager.nim
```diff
@@ -0,0 +1,393 @@
...
+    finalityUpdateVerifier: finalityUpdateVerifier,
+    optimisticUpdateVerifier: optimisticUpdateVerifier,
+    isLightClientStoreInitialized: isLightClientStoreInitialized,
+    isNextSyncCommitteeKnown: isNextSyncCommitteeKnown,
+    getFinalizedSlot: getFinalizedSlot,
+    getOptimisticSlot: getOptimisticSlot,
+    getBeaconTime: getBeaconTime,
+  )
+
+# https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.3/specs/altair/light-client/p2p-interface.md#getlightclientbootstrap
```
tersec: If these URLs had been just copy/pasted from somewhere else, sure, at least within this PR this (and the other v1.6.0 alpha.3 references, and v1.4.0-beta.5, and etc) seem to arise from nowhere existing?

If they're being added newly, then better to ensure they're consistent with the current v1.6.0-beta.1.

nimbus_verified_proxy/lc/lc_manager.nim
```diff
@@ -0,0 +1,393 @@
...
+    self.updateVerifier
+  elif E.V is ForkedLightClientFinalityUpdate:
+    self.finalityUpdateVerifier
+  elif E.V is ForkedLightClientOptimisticUpdate:
+    self.optimisticUpdateVerifier
+  else:
+    static:
+      doAssert false
+
+iterator values(v: auto): auto =
```
tersec, line 177: As long as one is careful not to export this -- `values` with an `auto` type will otherwise cause potential issues with things importing this module.

nimbus_verified_proxy/lc/lc_manager.nim
```diff
@@ -0,0 +1,393 @@
...
+      optimisitcEpoch = (optimisticSlot mod SLOTS_PER_EPOCH)
+
+    # Obtain bootstrap data once a trusted block root is supplied
+    if not self.isLightClientStoreInitialized():
+      let trustedBlockRoot = self.getTrustedBlockRoot()
+
+      # reattempt bootstrap download in 2 seconds
+      if trustedBlockRoot.isNone:
+        debug "TrustedBlockRoot unavaialble re-attempting bootstrap download"
+        await sleepAsync(chronos.seconds(2))
```
tersec: Should these different 2 second sleeps be consistent with each other, or are the both incidentally 2 seconds? If they should be consistent with each other, then better to extract it as a constant and use that constant in both, so that if one is changed, the other is automatically changed.

nimbus_verified_proxy/lc/lc_manager.nim
```diff
@@ -0,0 +1,393 @@
...
+        UpdatesByRange, (startPeriod: finalizedPeriod, count: uint64(count))
+      )
+
+    # check and download optimistic update
+    if optimisticSlot < currentSlot:
+      debug "Downloading light client optimistic updates", slot = currentSlot
+      let didProgress = await self.query(OptimisticUpdate)
+      if not didProgress:
+        # retry in 2 seconds
+        await sleepAsync(chronos.seconds(2))
```
tersec: Should these 2 second retry timers be intrinsically consistent?

nimbus_verified_proxy/lc/lc_manager.nim
```diff
@@ -0,0 +1,393 @@
...
+    downloadOptimistic = true
+    downloadFinality = false
+    didOptimisticProgress = false
+    didFinalityProgress = false
+
+  while true:
+    let
+      wallTime = self.getBeaconTime()
+      currentSlot = wallTime.slotOrZero()
+      currentEpoch = (currentSlot mod SLOTS_PER_EPOCH)
```
tersec: The CL has a `epoch` helper for this.

nimbus_verified_proxy/lc/lc_manager.nim
```diff
@@ -0,0 +1,393 @@
...
+      currentPeriod = currentSlot.sync_committee_period
+      finalizedSlot = self.getFinalizedSlot()
+      finalizedPeriod = finalizedSlot.sync_committee_period
+      finalizedEpoch = (finalizedSlot mod SLOTS_PER_EPOCH)
+      optimisticSlot = self.getOptimisticSlot()
+      optimisticPeriod = optimisticSlot.sync_committee_period
+      optimisitcEpoch = (optimisticSlot mod SLOTS_PER_EPOCH)
```
tersec: `optimisitcEpoch` has a typo, but more importantly, should be able to use `optimisticSlot.epoch`

nimbus_verified_proxy/lc/lc_manager.nim
```diff
@@ -0,0 +1,393 @@
...
+
+  while true:
+    let
+      wallTime = self.getBeaconTime()
+      currentSlot = wallTime.slotOrZero()
+      currentEpoch = (currentSlot mod SLOTS_PER_EPOCH)
+      currentPeriod = currentSlot.sync_committee_period
+      finalizedSlot = self.getFinalizedSlot()
+      finalizedPeriod = finalizedSlot.sync_committee_period
+      finalizedEpoch = (finalizedSlot mod SLOTS_PER_EPOCH)
```
tersec: `finalizedSlot.epoch`

nimbus_verified_proxy/lc/lc_manager.nim
```diff
@@ -0,0 +1,393 @@
...
+      optimisticSlot = self.getOptimisticSlot()
+      optimisticPeriod = optimisticSlot.sync_committee_period
+      optimisitcEpoch = (optimisticSlot mod SLOTS_PER_EPOCH)
+
+    # Obtain bootstrap data once a trusted block root is supplied
+    if not self.isLightClientStoreInitialized():
+      let trustedBlockRoot = self.getTrustedBlockRoot()
+
+      # reattempt bootstrap download in 2 seconds
```
tersec: As a general point, how does this scale with different slot times? Should these 2 second timers be the same with 5, 6, or 12 second slots or should they scale?

nimbus_verified_proxy/json_lc_backend.nim
```diff
@@ -0,0 +1,138 @@
...
+  client.peers.setLen(0)
+  client.urls.setLen(0)
+
+proc getEthLCBackend*(client: LCRestClient): EthLCBackend =
+  let
+    getLCBootstrapProc = proc(
+        reqId: uint64, blockRoot: Eth2Digest
+    ): Future[NetRes[ForkedLightClientBootstrap]] {.async: (raises: [CancelledError]).} =
+      let
+        peer = client.peers[reqId mod uint64(client.peers.len)]
```
tersec: The CL has a helper, `lenu64`, maybe available here

nimbus_verified_proxy/json_lc_backend.nim
```diff
@@ -0,0 +1,138 @@
...
+          except CatchableError as e:
+            raise newException(CancelledError, e.msg)
+
+      ok(res)
+
+    getLCUpdatesProc = proc(
+        reqId: uint64, startPeriod: SyncCommitteePeriod, count: uint64
+    ): Future[LightClientUpdatesByRangeResponse] {.async: (raises: [CancelledError]).} =
+      let
+        peer = client.peers[reqId mod uint64(client.peers.len)]
```
tersec: if `lenu64` is available, this is exactly what it's for

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -222,89 +209,16 @@ proc run*(
...
-  # launch a async routine
-  asyncSpawn runOnSecondLoop()
-
   # run an infinite loop and wait for a stop signal
   while true:
     poll()
     if ctx != nil and ctx.stop:
       # Cleanup
-      waitFor network.stop()
+      waitFor lcRestClient.closeAll()
```
tersec: If there are going to be multiple `waitFor`s in this function, is it possible to make the function itself async, use `await`, and have only one `waitFor` on the outside?

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
```diff
@@ -12,7 +12,8 @@ import
   json_rpc/rpcproxy, # must be early (compilation annoyance)
   json_serialization/std/net,
   beacon_chain/conf_light_client,
-  beacon_chain/nimbus_binary_common
+  beacon_chain/nimbus_binary_common,
+  std/strutils
```
tersec: Usually put stdlib imports first

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
```diff
@@ -114,70 +124,12 @@ type VerifiedProxyConf* = object
...
-  discv5Enabled* {.desc: "Enable Discovery v5", defaultValue: true, name: "discv5".}:
-    bool
-
-  directPeers* {.
-    desc: "The list of priviledged, secure and known peers to connect and maintain the connection to, this requires a not random netkey-file. In the complete multiaddress format like: /ip4/<address>/tcp/<port>/p2p/<peerId-public-key>. Peering agreements are established out of band and must be reciprocal."
-    name: "direct-peer" .}: seq[string]
+  # (Untrusted) web3 provider
+  # No default - Needs to be provided by the user
+  lcEndpoints* {.
+    desc: "command seperated URLs of the light client data provider",
```
tersec: separated

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
```diff
@@ -195,34 +147,24 @@ proc parseCmdArg*(T: type Web3Url, p: string): T {.raises: [ValueError].} =
...
+proc parseCmdArg*(T: type UrlList, p: string): T {.raises: [ValueError].} =
+  let urls = p.split(',')
+
+  for u in urls:
+    let
+      parsed = parseUri(u)
+      normalizedScheme = parsed.scheme.toLowerAscii()
+
+    if not (normalizedScheme == "http" or normalizedScheme == "https"):
+      raise newException(ValueError, "Light Client Endpoint should be a http(s) url")
```
tersec: URL is an acronym and capitalized as such

nimbus_verified_proxy/json_lc_backend.nim
```diff
@@ -0,0 +1,138 @@
...
+const
+  MaxMessageBodyBytes* = 128 * 1024 * 1024 # 128 MB (JSON encoded)
+  BASE_URL = "/eth/v1/beacon/light_client"
+
+type
+  LCRestPeer = ref object
+    score: int
+    restClient: RestClientRef
+
+  LCRestClient* = ref object
```
kdeme: Perhaps a bit of a nit but I find this naming scheme a bit confusing.

`LCRestClient` contains peers which are `LCRestPeer`s which are actually `RestClient`s.

I would probably have called it `LCRestClientManager`/`Pool`/`Provider`.

Doesn't matter that much I realize as in the end it is the `EthLCBackend` that gets used over the code base.

nimbus_verified_proxy/lc_backend.nim
```diff
@@ -0,0 +1,138 @@
...
+    ): Future[NetRes[ForkedLightClientBootstrap]] {.async: (raises: [CancelledError]).} =
+      let
+        peer = client.peers[reqId mod uint64(client.peers.len)]
+        res =
+          try:
+            await peer.restClient.getLightClientBootstrap(
+              blockRoot, client.cfg, client.forkDigests
+            )
+          except CatchableError as e:
+            raise newException(CancelledError, e.msg)
```
kdeme: What's the deal exactly with these `CancelledError`s being raised on `CatchableError` (I only noticed now there are already cases of this being done in NVP code base)? 

Is this intentional?  If so, how is this handled? seems very risky to use in such way.

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
kdeme: I think we should try to give this a more user-friendly name. I don't think end users will know what a light client data provider is. And then abbreviating to LC makes it even worse?

Maybe `beacon-node-api-endpoint` (a bit long) or just `beacon-node-api`? This is probably better known as the REST api is also officially named as such: https://ethereum.github.io/beacon-APIs/ .

We probably do want to clarify in the description that the endpoint needs to have support for serving LC data.

nimbus_verified_proxy/json_lc_backend.nim
```diff
@@ -0,0 +1,138 @@
...
+              client.cfg, client.forkDigests
+            )
+          except CatchableError as e:
+            raise newException(CancelledError, e.msg)
+
+      ok(res)
+
+    updateScoreProc = proc(reqId: uint64, value: int) =
+      let peer = client.peers[reqId mod uint64(client.peers.len)]
+      peer.score += value
```
etan-status: These functions assume that the peers list doesn't get modified in-between requests. Otherwise, the score is added to an incorrect peer, and also earlier peers would get selected more frequently by the modulo.

nimbus_verified_proxy/lc_backend.nim
```diff
@@ -0,0 +1,138 @@
...
+          try:
+            await peer.restClient.getLightClientFinalityUpdate(
+              client.cfg, client.forkDigests
+            )
+          except CatchableError as e:
+            raise newException(CancelledError, e.msg)
+
+      ok(res)
+
+    getLCOptimisticProc = proc(
```
etan-status, line 127: Code duplication could be reduced with a template that handles peer selection and error handling. Given it's only 4 requests and new forks don't really add more of these, also okay to keep it as is.

nimbus_verified_proxy/lc_backend.nim
```diff
@@ -0,0 +1,138 @@
...
+    .} =
+      let
+        peer = client.peers[reqId mod uint64(client.peers.len)]
+        res =
+          try:
+            await peer.restClient.getLightClientFinalityUpdate(
+              client.cfg, client.forkDigests
+            )
+          except CatchableError as e:
+            raise newException(CancelledError, e.msg)
```
etan-status: CancelledError has a very specific meaning in Chronos, not sure if this is appropriate to use here. Should this be ResponseError instead?

nimbus_verified_proxy/lc/lc.nim
```diff
@@ -0,0 +1,158 @@
...
+    if lightClient.onFinalizedHeader != nil:
+      lightClient.onFinalizedHeader(lightClient, lightClient.getFinalizedHeader)
+
+  proc onOptimisticHeader() =
+    if lightClient.onOptimisticHeader != nil:
+      lightClient.onOptimisticHeader(lightClient, lightClient.getOptimisticHeader)
+
+  # initialize without dumping 
+  lightClient.processor = LightClientProcessor.new(
+    false, ".", ".", cfg, genesis_validators_root, finalizationMode, lightClient.store,
```
etan-status: These would be more readable by first assigning `false` and `"."` to named constants. 

```
const
  dumpEnabled = false
  dumpDirInvalid = "."
  dumpDirIncoming = "."

...
dumpEnabled, dumpDirInvalid, dumpDirIncoming,
cfg, genesis_validators_root, ...
```

nimbus_verified_proxy/lc/lc.nim
```diff
@@ -0,0 +1,158 @@
...
+func getOptimisticHeader*(lightClient: LightClient): ForkedLightClientHeader =
+  withForkyStore(lightClient.store[]):
+    when lcDataFork > LightClientDataFork.None:
+      var header = ForkedLightClientHeader(kind: lcDataFork)
+      header.forky(lcDataFork) = forkyStore.optimistic_header
+      header
+    else:
+      default(ForkedLightClientHeader)
+
+proc new*(
```
etan-status, line 50: As this is heavily inspired by `createLightClient` in nimbus-eth2, maybe better to integrate `lc_backend` there, and make it so that `Eth2Node` is no longer required (but instead, any backend(s) can be passed). Avoids code duplication across repos with one copy of them guaranteed to become stale over time

nimbus_verified_proxy/lc_backend.nim
```diff
@@ -0,0 +1,138 @@
+# nimbus_verified_proxy
```
etan-status, line 1: The `json_` name isn't future-proof, as you can also use SSZ for lower bandwidth using the `restAccept` param e.g., for `getLightClientBootstrap`, without changing the interface (the calling code doesn't really care whether it's JSON)

nimbus_verified_proxy/lc_backend.nim
```diff
@@ -0,0 +1,138 @@
...
+  beacon_chain/spec/forks,
+  ./lc/lc_manager,
+  ./nimbus_verified_proxy_conf
+
+logScope:
+  topics = "SSZLCRestClient"
+
+const
+  MaxMessageBodyBytes* = 128 * 1024 * 1024 # 128 MB (JSON encoded)
+  BASE_URL = "/eth/v1/beacon/light_client"
```
etan-status, line 26: https://ethereum.github.io/beacon-APIs/#/Events/eventstream

There's also an event stream that could be used to avoid repeated polling. (not required initially though)

nimbus_verified_proxy/json_lc_backend.nim
```diff
@@ -0,0 +1,138 @@
...
+            await peer.restClient.getLightClientOptimisticUpdate(
+              client.cfg, client.forkDigests
+            )
+          except CatchableError as e:
+            raise newException(CancelledError, e.msg)
+
+      ok(res)
+
+    updateScoreProc = proc(reqId: uint64, value: int) =
+      let peer = client.peers[reqId mod uint64(client.peers.len)]
```
etan-status: These functions assume that the peers list doesn't get modified in-between requests. Otherwise, the score is added to an incorrect peer, and also earlier peers would get selected more frequently by the modulo.

nimbus_verified_proxy/lc_backend.nim
```diff
@@ -0,0 +1,138 @@
...
+              client.cfg, client.forkDigests
+            )
+          except CatchableError as e:
+            raise newException(CancelledError, e.msg)
+
+      ok(res)
+
+    updateScoreProc = proc(reqId: uint64, value: int) =
+      let peer = client.peers[reqId mod uint64(client.peers.len)]
+      peer.score += value
```
etan-status: Does this need saturation to avoid over/underflow? (signed int overflow is UB in C)

nimbus_verified_proxy/lc_backend.nim
```diff
@@ -0,0 +1,138 @@
...
+logScope:
+  topics = "SSZLCRestClient"
+
+const
+  MaxMessageBodyBytes* = 128 * 1024 * 1024 # 128 MB (JSON encoded)
+  BASE_URL = "/eth/v1/beacon/light_client"
+
+type
+  LCRestPeer = ref object
+    score: int
```
etan-status, line 30: The score doesn't seem to do anything?

nimbus_verified_proxy/lc/lc.nim
```diff
@@ -0,0 +1,158 @@
...
+  lightClient.manager = LightClientManager.init(
+    rng, getTrustedBlockRoot, bootstrapVerifier, updateVerifier, finalityVerifier,
+    optimisticVerifier, isLightClientStoreInitialized, isNextSyncCommitteeKnown,
+    getFinalizedSlot, getOptimisticSlot, getBeaconTime,
+  )
+
+  lightClient
+
+proc setBackend*(lightClient: LightClient, backend: EthLCBackend) =
+  lightClient.manager.backend = backend
```
etan-status, line 148: Or, just pass the backend with the init call.

nimbus_verified_proxy/lc/lc_manager.nim
```diff
@@ -0,0 +1,393 @@
...
+  UpdateScoreProc = proc(id: uint64, value: int) {.gcsafe, raises: [].}
+
+  EthLCBackend* = object
+    getLightClientBootstrap*: LightClientBootstrapProc
+    getLightClientUpdatesByRange*: LightClientUpdatesByRangeProc
+    getLightClientFinalityUpdate*: LightClientFinalityUpdateProc
+    getLightClientOptimisticUpdate*: LightClientOptimisticUpdateProc
+    updateScore*: UpdateScoreProc
+
+  LightClientManager* = object
```
etan-status, line 72: Overall, to reduce code duplication, maybe also better to move this inside nimbus-eth2 (as a library), rather than having a fork here that's almost the same (but not exactly the same).

nimbus_verified_proxy/lc/lc_manager.nim
```diff
@@ -0,0 +1,393 @@
...
+    if currentEpoch > finalizedEpoch + 2:
+      debug "Downloading light client finality updates", slot = currentSlot
+      let didProgress = await self.query(FinalityUpdate)
+      if not didProgress:
+        # retry in two seconds
+        await sleepAsync(chronos.seconds(2))
+        continue
+
+    # check for updates every slot
+    await sleepAsync(chronos.seconds(int64(SECONDS_PER_SLOT)))
```
etan-status: SECONDS_PER_SLOT is gone. Use cfg.timeParams.SLOT_DURATION instead.

Or, `beaconClock.fromNow(currentSlot + 1)`, this also accounts for the processing itself taking some time (otherwise this will slowly drift later and later into a slot and eventually skip one).

nimbus_verified_proxy/lc/lc_manager.nim
```diff
@@ -0,0 +1,393 @@
...
+      let didProgress = await self.query(Bootstrap, trustedBlockRoot.get)
+
+      # reattempt bootstrap download in 2 seconds
+      if not didProgress:
+        debug "Re-attempting bootstrap download"
+        await sleepAsync(chronos.seconds(2))
+
+      continue
+
+    # check and download sync committee updates
```
etan-status, line 336: nextLightClientSyncTask no good?

nimbus_verified_proxy/lc/lc_manager.nim
```diff
@@ -0,0 +1,393 @@
...
+    # check and download finality update
+    if currentEpoch > finalizedEpoch + 2:
+      debug "Downloading light client finality updates", slot = currentSlot
+      let didProgress = await self.query(FinalityUpdate)
+      if not didProgress:
+        # retry in two seconds
+        await sleepAsync(chronos.seconds(2))
+        continue
+
+    # check for updates every slot
```
etan-status, line 385: If this leads to rate-limiting, the events endpoint could be useful as an alternative. There's `nextLcSyncTaskDelay` to compute a delay that allows following the sync committee in time while being subscribed to events. That's the strategy I'm using on eth-light.xyz

nimbus_verified_proxy/libverifproxy/verifproxy.nim
```diff
@@ -38,20 +37,14 @@ proc runContext(ctx: ptr Context) {.thread.} =
...
     let rpcAddr = jsonNode["RpcAddress"].getStr()
     let myConfig = VerifiedProxyConf(
-      listenAddress: some(defaultListenAddress),
       eth2Network: some(jsonNode["Eth2Network"].getStr()),
       trustedBlockRoot: Eth2Digest.fromHex(jsonNode["TrustedBlockRoot"].getStr()),
-      backendUrl: parseCmdArg(Web3Url, jsonNode["Web3Url"].getStr()),
-      frontendUrl: parseCmdArg(Web3Url, jsonNode["Web3Url"].getStr()),
+      backendUrl: parseCmdArg(Web3Url, jsonNode["backendUrl"].getStr()),
+      frontendUrl: parseCmdArg(Web3Url, jsonNode["frontendUrl"].getStr()),
+      lcEndpoints: parseCmdArg(UrlList, jsonNode["lcEndpoints"].getStr()),
```
etan-status: Some of these are PascalCase while others are camelCase. 

In nimbus-eth2, we have `--external-beacon-api-url` to specify the "lcEndpoints"

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
etan-status: `--external-beacon-api-url` is the one in nimbus-eth2, maybe use the same one for consistency?

nimbus_verified_proxy/lc/lc_manager.nim
chirag-parmar: yes

nimbus_verified_proxy/lc/lc_manager.nim
chirag-parmar: yes

nimbus_verified_proxy/lc/lc_manager.nim
chirag-parmar: you are right they should scale

nimbus_verified_proxy/nimbus_verified_proxy.nim
chirag-parmar: makes sense

nimbus_verified_proxy/json_lc_backend.nim
chirag-parmar: makes sense. changing the naming

nimbus_verified_proxy/lc/lc_manager.nim
chirag-parmar: It is a great strategy but then it would make rotating between endpoints more complex. Also, if messages are initiated from within the proxy we can reserve the control over bandwidth and privacy.

nimbus_verified_proxy/libverifproxy/verifproxy.nim
chirag-parmar: You are right, the casing is inconsistent. I'll change it.

nimbus_verified_proxy/lc_backend.nim
chirag-parmar: you are right

nimbus_verified_proxy/lc_backend.nim
chirag-parmar: This is intentional. The API used for light client updates (defined here `beacon_chain/spec/eth2_apis/rest_light_client_calls.nim`) doesn't use proper error specifications. Therefore we catch `CatchableError` and propagate a cancelled error in place. Refer this discussion: https://github.com/status-im/nimbus-eth1/pull/3709#discussion_r2380818638

chirag-parmar: same for everywhere else in verified proxy

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
chirag-parmar: using `--external-beacon-api-urls`

nimbus_verified_proxy/lc_backend.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3782#discussion_r2472027034 let's continue the discussion here

nimbus_verified_proxy/lc_backend.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3782#discussion_r2469578566 let's continue this discussion here

nimbus_verified_proxy/lc_backend.nim
etan-status: Just because there is precedence in other modules doesn't make it more correct. `CancelledError` is used in Chronos when you `cancel` a future. It indicates that the entire sync process should get cancelled at the next opportunity. By transforming random errors to a `CancelledError`, you change the meaning from "there was some issue with one particular request" to "please cancel the entire sync process and propagate cancellation".

kdeme: What etan said basically.

But regarding that discussion you link, I think what @arnetheduck meant there is that catching `CatchableError` will make you also catch the `CancelledError`, which would stop the flow of raising a cancellation.  So instead you want to catch all **but** the `CancelledError`.

What is done here now however is transforming all exceptions into a CancelledError.

So I think what you want instead is something like:

```Nim
try:
  <async code>
except CancelledError as e:
  raise e
except CatchableError as e:
  return err(e.msg)
```

chirag-parmar: That makes a lot more sense. Apologies, for misunderstanding the instruction. I'll rectify this.

nimbus_verified_proxy/lc/lc_manager.nim
chirag-parmar: The logic of nextLightClientSyncTask is slightly different than what I want(it will only download a finalized updated once per sync committee). As you said it works great if I am subscribing for optimictic and finalized updates through the events endpoint but I don't wanna use the events endpoint to be able to rotate endpoints and in the future also use onion sites

nimbus_verified_proxy/json_lc_backend.nim
chirag-parmar: Good catch will change this

nimbus_verified_proxy/lc_backend.nim
chirag-parmar: the entire scoring structure needs to be redone in a separate PR. The scores aren't even used here atm. I could have added it to this PR but it would have made it unnecessarily large.

nimbus_verified_proxy/lc_backend.nim
chirag-parmar: true. https://github.com/status-im/nimbus-eth1/pull/3782#discussion_r2476411114

nimbus_verified_proxy/lc/lc.nim
chirag-parmar: Can be done, but this styling shows that the backend is independent of the light client core algo. There is some kind of syntactic intuition to this in my perspective. But I'm open to changing this as well. In some ways it is like passing the `trustedBlockRoot` outside the init call.

nimbus_verified_proxy/lc/lc.nim
chirag-parmar: code duplication is definitely avoided if these changes are made to the nimbus-eth2 repo directly. But I think the way we use the light client in verified proxy differs from the actual light client in nimbus-eth2. For ex, the core sync logic is changed (with optimistic updates and finalized updates downloaded in the same sync loop). 

This matters because, every time there is a change required, updates would have to be pushed to nimbus-eth2 then nimbus-eth1. Also for the most part the verified proxy reuses a lot of light client code avoiding duplication (processor, light_client_calls, etc.). It is only the sync loop and light client initialization that is changed, which isn't a problem IMO.

nimbus_verified_proxy/lc_backend.nim
chirag-parmar: @kdeme https://github.com/status-im/nimbus-eth1/pull/3607

nimbus_verified_proxy/lc/lc_manager.nim
chirag-parmar: let's move the discussion here https://github.com/status-im/nimbus-eth1/pull/3782#discussion_r2480039580

nimbus_verified_proxy/lc/lc.nim
etan-status: Could be achieved with a flag passed during initialization

chirag-parmar: ok makes sense. I'll open a PR on `nimbus-eth2` but let's not delay this PR in favor of that. When that gets merged in I'll refactor the code in `nimbus-eth1` and remove all duplicated logic. What do you think?

etan-status: sounds good! thanks <3

chirag-parmar: https://github.com/status-im/nimbus-eth2/pull/7704

nimbus_verified_proxy/json_lc_backend.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3782/commits/34b20edad15329dcff72f04914940f432782e22a

nimbus_verified_proxy/json_lc_backend.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3782/commits/34b20edad15329dcff72f04914940f432782e22a

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
```diff
@@ -114,70 +124,12 @@ type VerifiedProxyConf* = object
...
-    bool
-
-  directPeers* {.
-    desc: "The list of priviledged, secure and known peers to connect and maintain the connection to, this requires a not random netkey-file. In the complete multiaddress format like: /ip4/<address>/tcp/<port>/p2p/<peerId-public-key>. Peering agreements are established out of band and must be reciprocal."
-    name: "direct-peer" .}: seq[string]
+  # (Untrusted) web3 provider
+  # No default - Needs to be provided by the user
+  beaconApiUrls* {.
+    desc: "command separated URLs of the light client data provider",
+    name: "external-beacon-api-urls"
```
etan-status: or, same flag multiple time, instead of comma-separated?

etan-status: I think that's how in nimbus-eth2 the bootnodes work

etan-status (approved):
I still think lc.nim and lc_manager.nim has too much overlap with the one in eth2, could be done there with flag instead of copy. But the feature is useful enough to have to not block it just for that. OK from my side, but @tersec should also check.

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
chirag-parmar: ok

PR #3788, opened 2025-10-23
Fix problem with manual setting `pendingFCU`

mjfh:


arnetheduck:
> a second register

you can just look up the block by number and take the hash from there - `pendingFCU` is indeed only supposed to be used for _resolving_ a fin block number after which it's no longer useful - the second important point being that the finalized block _number_ only increases  monotonically, so if pendingFCU resolves to an older block, it should ignore that update and warn that it was attempted.

Issue #3794, opened 2025-10-24
Invalid forkid checking for discv5 eligibleNode check


kdeme:
compatibleForkId always fails, also for correct forks:

https://github.com/status-im/nimbus-eth1/blob/master/execution_chain/networking/eth1_discovery.nim#L111

Bug is in `compatibleForkId`.

This causes discv5 nodes to be never used for rlpx connection setup.

(noticed in fact on looking at these metrics added https://github.com/status-im/nimbus-eth1/pull/3792)

PR #3796, opened 2025-10-24
Fix ForkId compatible bug

kdeme:
WIP, initial bugfix is fine, but testing like we should is not there yet, and might reveal more issues.


kdeme:
https://github.com/status-im/nimbus-eth1/issues/3794

Issue #3795, opened 2025-10-24
Missing eip-2124 / EIP-6122 testing for fork combination possibilities


kdeme:
ForkId logic as described here https://eips.ethereum.org/EIPS/eip-2124#stale-software-examples 
is not properly tested and I am not sure that our way of implementation (https://github.com/status-im/nimbus-eth1/blob/master/execution_chain/common/hardforks.nim#L371-L387) covers all that (aside from the existing bug of https://github.com/status-im/nimbus-eth1/issues/3794)

See also test cases here: https://github.com/ethereum/go-ethereum/blob/0413af40f60290cf689b4ecca4e51fef0ec11119/core/forkid/forkid_test.go#L158

or in EIPS
https://eips.ethereum.org/EIPS/eip-2124
https://eips.ethereum.org/EIPS/eip-6122
https://eips.ethereum.org/EIPS/eip-2364

It is also the case that we might need to do some changes to be able to test this in that same way.

kdeme:
Fixed in https://github.com/status-im/nimbus-eth1/pull/3803 and https://github.com/status-im/nimbus-eth1/pull/3796

PR #3803, opened 2025-10-31
Fix forkId compatibility according to EIP-2124 and EIP-6122

kdeme:
fixing https://github.com/status-im/nimbus-eth1/issues/3795


kdeme:
image

Peering from fresh start (genesis) with this change.

This PR is the left run, master is the right run.

PR #3810, opened 2025-11-04
Rework ForkId to get rid of ChainForkId and its conversions

kdeme:
And be more consistent in the naming + use fields as per EIPs

relying on https://github.com/status-im/nim-eth/pull/826


execution_chain/networking/peer_pool.nim
```diff
@@ -162,22 +161,22 @@ proc lookupPeers(p: PeerPoolRef) {.async: (raises: [CancelledError]).} =
     # to be later processed by connection worker
     await p.discovery.lookupRandomNode(p.connQueue)
 
-func updateForkID(p: PeerPoolRef) =
+func updateForkId(p: PeerPoolRef) =
   if p.forkId.isNil:
     return
 
   let forkId = p.forkId()
   if p.lastForkId == forkId:
```
kdeme, line 169: For some reason the `==` here fails with
```
Expression: x.hash == y.hash
  [1] x.hash: Bytes4
  [2] y.hash: Bytes4
```
without the added `==` here: 
https://github.com/status-im/nim-eth/pull/826/commits/21ba8923a822a60c40261c97c15d0b8a234c0a40#diff-fdc22bb681374a71a5a754cf46c5ade4b1c54c35c66f163533e86879bab9228aR190-R193

It shouldn't be needed though, and in fact in testing equals win forkid_test.nim, it does work without the added `==`.

kdeme: Actually, it works without the specific ForkId `==` added, by adding `eth/common/base` import in p2p.nim (or exporting base in peer_pool.nim. So I'll go that route.

Depends on order I guess and hitting p2p.nim first,
basically instantiation here https://github.com/status-im/nimbus-eth1/blob/master/execution_chain/networking/p2p.nim#L116 ?

PR #3846, opened 2025-11-27
Run all EEST engine tests, not just ones in static folder

bhartnett:


scripts/eest_ci_cache.sh
```diff
@@ -67,8 +67,8 @@ download_and_extract() {
   rm -rf "${dest_dir}/blockchain_tests/static"
 
   mkdir -p "${dest_dir}/engine_tests"
-  mv "${dest_dir}/blockchain_tests_engine/static/state_tests/"* "${dest_dir}/engine_tests" 2>/dev/null || true
-  rm -rf "${dest_dir}/blockchain_tests_engine"
```
bhartnett, line 71: This line was deleting all the other tests not in the static folder and effectively reducing the number of tests which are run.

PR #3860, opened 2025-12-04
Reduce down stateless_execution_helpers to compile for baremetal

lewis1revill:


execution_chain/stateless/stateless_execution_helpers.nim
```diff
@@ -68,33 +40,48 @@ func decodeRlp*(T: type Block, rlpBytes: openArray[byte]): Result[T, string] =
...
-proc statelessProcessBlockJsonFiles*(
-    witnessJsonFilePath: string, com: CommonRef, blockJsonFilePath: string
-): Result[void, string] =
-  let
-    witnessJson = ?readFileToStr(witnessJsonFilePath)
-    blkJson = ?readFileToStr(blockJsonFilePath)
-  statelessProcessBlockJson(witnessJson, com, blkJson)
+  statelessProcessBlockRlp(witnessRlpBytes, blkRlpBytes)
+
+proc main() =
```
arnetheduck, line 58: the main function also goes inside the isMainModule block or it will cause an "unused private function" warning on import

PR #3861, opened 2025-12-04
Block import improvements

bhartnett:
This PR improves a few things in the block import:
- Change naming of flags and parameters to use positive naming which makes the code easier to understand.
- The `FullValidation` flag was not being used and the fullValidation debug parameter did nothing previously. The flag is now used to apply the per block validation.


CASABECI (commented):
Successful

PR #3865, opened 2025-12-05
EIP-7928: Implement debug_getBadBlocks RPC endpoint

bhartnett:


execution_chain/rpc/debug.nim
```diff
@@ -107,6 +115,17 @@ proc getBlockAccessList*(
...
+
+  let txFrame = chain.txFrame(blockHash).txFrameBegin()
+  defer:
+    txFrame.dispose()
+
+  let totalDifficulty = txFrame.getScore(blockHash).valueOr:
+    return chain.baseTxFrame().headTotalDifficulty()
+
+  return totalDifficulty
+
```
kdeme, line 128: Note that this can / will have to be removed when/if https://github.com/status-im/nimbus-eth1/pull/3882 gets merged and bal-devnet-0 gets rebased

bhartnett: No problem, I'll handle that when I merge master into my branch, once that PR is merged.

PR #3873, opened 2025-12-11
Move rpc evm call code to support zkEVM bare metal compilation

bhartnett:
This moves the rpc related code in call_evm.nim out into a separate file call_evm_rpc.nim so that the stateless execution function won't end up importing the RPC related code which is using the current time.


bhartnett:
> what's wrong with `now`?

now is deprecated in nim-eth in order to remove the dependency on the std/times code when importing eth/common/times. I just picked a different name to avoid the naming clash.

kdeme:
fyi: My reasoning was based on the amount of usage in the code base of `EthTime` coming from  `now()` vs from other sources.
Felt justified to pull it out of that code. But if you think that it affects too much the UX of `now`, I'm fine just keeping it behind the conditionals too.

PR #3881, opened 2025-12-17
Disable taskpools at compile time when threads are not enabled

bhartnett:
With this we disable the taskpools code in the EL at compile time when `--threads:off` is set. 

We want to be able to compile the stateless execution function on bare metal for the zkEVM integration but the task pools code doesn't compile on bare metal and is not needed in that context either so we should support disabling it.

### DIFF
--- a/execution_chain/evm/async_evm.nim
+++ b/execution_chain/evm/async_evm.nim
@@ -103,7 +103,6 @@ proc init*(
 ): T =
   let com = CommonRef.new(
     DefaultDbMemory.newCoreDbRef(),
-    taskpool = nil,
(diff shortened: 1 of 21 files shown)


execution_chain/common/common.nim
```diff
@@ -27,9 +26,12 @@ export
   evmforks,
   genesis,
   hardforks,
-  taskpools,
   utils
 
+when compileOption("threads"):
```
bhartnett, line 31: @tersec Thoughts on using this switch? Is this the recommended way to check if threads are enabled at compile time?

PR #3882, opened 2025-12-17
Remove totalDifficulty from JSON-RPC API BlockObject

kdeme:
This was removed from the specifications on Sep 4, 2024.

+ remove related functionality.

Cannot remove yet headTotalDifficulty call as this is used in eth68 protocol. Once we remove this protocol (eth69 is latest) that call can be removed also.

Draft as depends on https://github.com/status-im/nim-web3/pull/231

### DIFF
--- a/portal/tools/eth_data_exporter.nim
+++ b/portal/tools/eth_data_exporter.nim
@@ -109,7 +109,7 @@ proc cmdExportEra1(config: ExporterConf) =
       var headerRecords: seq[historical_hashes_accumulator.HeaderRecord]
       for blockNumber in startNumber .. endNumber:
         let
-          (header, body, totalDifficulty) = (
+          (header, body) = (
             waitFor noCancel(client.getBlockByNumber(blockId(blockNumber)))
           ).valueOr:
             error "Failed retrieving block, skip creation of era1 file",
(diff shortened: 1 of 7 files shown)


bhartnett:
> @bhartnett heads-up on having to remove this also in https://github.com/status-im/nimbus-eth1/blob/bal-devnet-0/execution_chain/rpc/debug.nim#L304 when/if this gets merged and you rebase your branch

Yep sure, I'll make sure the merge is done correctly when I merge master into my branch.

portal/tools/eth_data_exporter.nim
```diff
@@ -122,14 +122,16 @@ proc cmdExportEra1(config: ExporterConf) =
               blockNumber, error
             break writeFileBlock
 
+        # Note: It is no longer possible to retrieve the total difficulty via JSON-RPC
```
arnetheduck, line 125: why was it removed? maybe this needs an upstream revert?

kdeme: See reasoning here: https://github.com/ethereum/execution-apis/pull/570

The only applicable use case identified there was Block Explorers.
It does not look like it had any push back, not sure if any other discussion on this took place, I did not follow that repo at that time.


I would like to remove this "generate era1 files via JSON-RPC requests" functionality here now completely. It was not very viable in the first place to generate all era1 files this way because it is very slow via JSON-RPC. I just created that back in the day to verify a few era1 files so that I could compare with the ones generated via geth.

I was keeping it around for now until we would actually add this directly in `nec` (back when I added this, `nec` could not fully sync at all, which is why I went for the JSON-RPC way).
But I can delete it all already now too.

arnetheduck: ok - I see the two things as separate - I think the right thing to do for the API would have been to make it optional post-merge but required for pre-merge blocks - it would probably make sense to make that PR to the api repo - separately, we can get rid of its usage in the block explorer (ie and divert to era files or whatever)

arnetheduck: pursuant to this point, I think we should leave it as an optional and fill it in to the merge block, as far as implementation goes - otherwise, the endpoint is not well-defined for pre-merge blocks which is ugly.

kdeme: Yes, making it optional will work for passing the `execution-apis` tests. Not sure if it would trigger issues on hive. But I will take that route for Nimbus.

> I think the right thing to do for the API would have been to make it optional post-merge but required for pre-merge blocks

While I do agree with this, teams have made the move long time back. Window of making this revert/adjustment probably has passed.

PR #3883, opened 2025-12-18
Adjust for optional totalDifficulty in JSON-RPC BlockObject

kdeme:
Going for optionally `totalDifficulty` instead of removal.

See reasoning in discussion here: https://github.com/status-im/nimbus-eth1/pull/3882#discussion_r2629779319

Depends on https://github.com/status-im/nim-web3/pull/233

### DIFF
--- a/execution_chain/rpc/rpc_utils.nim
+++ b/execution_chain/rpc/rpc_utils.nim
@@ -391,8 +391,5 @@ proc getEthConfigObject*(com: CommonRef,
   return res
 
 proc getTotalDifficulty*(chain: ForkedChainRef, blockHash: Hash32): Opt[UInt256] =
-  let txFrame = chain.txFrame(blockHash).txFrameBegin()
-  defer:
-    txFrame.dispose()
-
-  txFrame.getScore(blockHash)
+  # Note: It's ok to use baseTxFrame for TD as this is for historical blocks
+  chain.baseTxFrame().getScore(blockHash)


execution_chain/rpc/rpc_utils.nim
```diff
@@ -389,3 +389,10 @@ proc getEthConfigObject*(com: CommonRef,
     res.last = Opt.none(ConfigObject)
 
   return res
+
+proc getTotalDifficulty*(chain: ForkedChainRef, blockHash: Hash32): Opt[UInt256] =
+  let txFrame = chain.txFrame(blockHash).txFrameBegin()
+  defer:
+    txFrame.dispose()
+
+  txFrame.getScore(blockHash)
```
kdeme: I want to note that, ideally here `getScore` returns a `None` in case the block referenced is post-merge.

However from reading the code it would seem that is not the case (but it was probably the long term plan).

One could assume that getScore here would return failure if block post merge: https://github.com/status-im/nimbus-eth1/blob/08efe955b20ab0bbeae685414af86cf804e6d1a0/execution_chain/db/core_db/core_apps.nim#L217

But this comment indicates that the score is still stored also for each post merge block: https://github.com/status-im/nimbus-eth1/blob/08efe955b20ab0bbeae685414af86cf804e6d1a0/execution_chain/db/core_db/core_apps.nim#L606-L613

I have not tested this yet but I would assume that it will still find a score.

execution_chain/rpc/rpc_utils.nim
```diff
@@ -389,3 +389,10 @@ proc getEthConfigObject*(com: CommonRef,
     res.last = Opt.none(ConfigObject)
 
   return res
+
+proc getTotalDifficulty*(chain: ForkedChainRef, blockHash: Hash32): Opt[UInt256] =
+  let txFrame = chain.txFrame(blockHash).txFrameBegin()
```
kdeme: Using txFrame here for the fork-aware view, even though this value is not needed for current forks (post merge forks). I assumed it was useful also when using it for uncle blocks (pre-merge). But after looking into the code it is not. I will adjust to the `baseTxFrame` version.

kdeme: @bhartnett I think it is not needed to do the txFrame version as you have added here: https://github.com/status-im/nimbus-eth1/blob/bal-devnet-0/execution_chain/rpc/debug.nim#L119-L122

As this value is not supposed to change anymore. (and ideally does not get set post merge, but ok)

bhartnett: @kdeme Good point. In that case should it just always return zero?

PR #3891, opened 2026-01-05
proxy: enable multiple web3 endpoints

chirag-parmar:
### Description

This PR enables using multiple endpoints for the web3 backend. For every request a random url is picked from a pool of urls.

### DIFF
--- a/nimbus_verified_proxy/json_rpc_backend.nim
+++ b/nimbus_verified_proxy/json_rpc_backend.nim
@@ -74,7 +74,14 @@ proc getClientFromPool(pool: JsonRpcClientPool): JsonRpcClient =
   var randomNum: uint64
   pool.rng[].generate(randomNum)
 
+  # NOTE: we use the mod operator to bring the random number into range
+  # this introduces a bias in the output distribution but is negligible
+  # for this use case. The bias becomes insignificant when score filters
+  # are used to select clients in the future.
   pool.clients[randomNum mod uint64(pool.clients.len)]
+
+template resolveClientFromPool(pool: JsonRpcClientPool): RpcClient =
+  pool.getClientFromPool().resolveClient()
 
 proc getEthApiBackend*(pool: JsonRpcClientPool): EthApiBackend =
   let
@@ -82,7 +89,7 @@ proc getEthApiBackend*(pool: JsonRpcClientPool): EthApiBackend =
         async: (raises: [CancelledError])
     .} =
       try:
-        ok(await pool.getClientFromPool().resolveClient().eth_chainId())
+        ok(await pool.resolveClientFromPool().eth_chainId())
       except CancelledError as e:
         raise e
       except RpcPostError as e:
@@ -100,7 +107,7 @@ proc getEthApiBackend*(pool: JsonRpcClientPool): EthApiBackend =
         blkHash: Hash32, fullTransactions: bool
     ): Future[EngineResult[BlockObject]] {.async: (raises: [CancelledError]).} =
       try:
-        let res = await pool.getClientFromPool().resolveClient().eth_getBlockByHash(
+        let res = await pool.resolveClientFromPool().eth_getBlockByHash(
           blkHash, fullTransactions
         )
         if res.isNil():
@@ -123,7 +130,7 @@ proc getEthApiBackend*(pool: JsonRpcClientPool): EthApiBackend =
         blkNum: BlockTag, fullTransactions: bool
     ): Future[EngineResult[BlockObject]] {.async: (raises: [CancelledError]).} =
       try:
-        let res = await pool.getClientFromPool().resolveClient().eth_getBlockByNumber(
+        let res = await pool.resolveClientFromPool().eth_getBlockByNumber(
           blkNum, fullTransactions
         )
         if res.isNil():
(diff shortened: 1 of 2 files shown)


nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
```diff
@@ -146,6 +146,15 @@ proc parseCmdArg*(T: type Web3Url, p: string): T {.raises: [ValueError].} =
...
     )
 
+proc parseCmdArg*(T: type seq[Web3Url], p: string): T {.raises: [ValueError].} =
+  let urls = p.split(',')
+  var parsedWeb3Urls: seq[Web3Url]
+
+  for u in urls:
+    parsedWeb3Urls.add(parseCmdArg(Web3Url, u))
+
+  parsedWeb3Urls
```
tersec: The entire `var parsedWeb3Urls` / `...` / `parsedWeb3Urls` can be condensed, while being made clearer, to `urls.mapIt(parseCmdArg(Web3Url, it))`.

nimbus_verified_proxy/json_rpc_backend.nim
```diff
@@ -9,52 +9,80 @@
...
+    await client.resolveClient().close()
+
+  pool.clients.setLen(0)
+  pool.urls.setLen(0)
+
+proc getClientFromPool(pool: JsonRpcClientPool): JsonRpcClient =
+  var randomNum: uint64
+  pool.rng[].generate(randomNum)
+
+  pool.clients[randomNum mod uint64(pool.clients.len)]
```
tersec, line 81: This is the usual biased `mod`-based random number generator. It's fine here insofar as the bias is very small, enough to be undetectable, and it's not depending on cryptographic-level uniformity, but it might be worth noting the general reasoning in a comment.

nimbus_verified_proxy/json_rpc_backend.nim
```diff
@@ -72,7 +100,9 @@ proc getEthApiBackend*(client: JsonRpcClient): EthApiBackend =
         blkHash: Hash32, fullTransactions: bool
     ): Future[EngineResult[BlockObject]] {.async: (raises: [CancelledError]).} =
       try:
-        let res = await client.getClient().eth_getBlockByHash(blkHash, fullTransactions)
+        let res = await pool.getClientFromPool().resolveClient().eth_getBlockByHash(
```
tersec: Worth having a helper function `resolveClientFromPool`? There are at least 11 instances of this `pool.getClientFromPool().resolveClient()` pattern.

chirag-parmar: worth it for sure. A template would create a shorthand

nimbus_verified_proxy/json_rpc_backend.nim
chirag-parmar: will do

nimbus_verified_proxy/nimbus_verified_proxy_conf.nim
tersec: https://github.com/status-im/nimbus-eth1/pull/3891/commits/f17af164612436faf7fb3230faa148d9c517f302

nimbus_verified_proxy/json_rpc_backend.nim
tersec: https://github.com/status-im/nimbus-eth1/pull/3891/commits/f17af164612436faf7fb3230faa148d9c517f302

nimbus_verified_proxy/json_rpc_backend.nim
tersec: https://github.com/status-im/nimbus-eth1/pull/3891/commits/f17af164612436faf7fb3230faa148d9c517f302

nimbus_verified_proxy/json_rpc_backend.nim
kdeme: The bias should be fine here indeed, but if you do want to get rid of it (or just for general reference in future), we have a non-bias version here: https://github.com/status-im/nim-eth/blob/master/eth/p2p/discoveryv5/random2.nim#L20

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -132,26 +120,18 @@ proc run(
...
+  let status = jsonRpcServer.start()
   if status.isErr():
     raise newException(ProxyError, status.error.errMsg)
 
   # adding endpoints will also start the backend
   if lcRestClientPool.addEndpoints(config.beaconApiUrls).isErr():
     raise newException(ProxyError, "Couldn't add endpoints for light client queries")
 
-  # verify chain id that the proxy is connected to
-  await engine.verifyChainId()
```
kdeme, line 153: Does it not make sense to keep this quick check in place for the the rpc endpoints?

chirag-parmar: I could add it back in but the check will have to be made for all endpoints. Once we have scoring enabled all endpoints responding with wrong chain data should be filtered automatically. Moreover, it only provides a small user experience upgrade, doesn't minimize trust in any significant way.

chirag-parmar: Is this resolved or should I add it back in?

kdeme: I'll leave it up to you as I understand that scoring will include this and more.

PR #3892, opened 2026-01-05
proxy: download blocks parallely while walking the chain

chirag-parmar:


nimbus_verified_proxy/libverifproxy/setup.nim
```diff
@@ -60,6 +60,7 @@ proc load(T: type VerifiedProxyConf, configJson: string): T {.raises: [ProxyErro
       of "Auto": StdoutLogKind.Auto
       else: StdoutLogKind.None
     maxBlockWalk = jsonNode.getOrDefault("maxBlockWalk").getInt(1000)
+    parallelBlockDownloads = jsonNode.getOrDefault("parallelBlockDownloads").getInt(10)
```
tersec: For this (and all the other `jsonNode.getOrDefault("foo").getInt(...)` lines), because it's tied to an architecture-size specific integer (`int`, which is 32-bit on 32-bit platforms and 64-bit on 64-bit platforms), it creates a tricky-to-discover/test failure mode. Instead of using [`getInt`](https://nim-lang.org/docs/json.html#getInt%2CJsonNode%2Cint), it's more predictable across 32/64-bit platforms to use [`getBiggestInt`](https://nim-lang.org/docs/json.html#getBiggestInt%2CJsonNode%2CBiggestInt), which returns  `BiggestInt`, [currently an `int64`](https://nim-lang.org/docs/ctypes.html#BiggestInt) regardless of whether it's on a 32-bit or 64-bit platform.

It does say that it's
> an alias for the biggest signed integer type the Nim compiler supports. Currently this is `int64`, but it is platform-dependent in general.

But at least it's no more, and currently less, platform-dependent than `getInt`/`int`.

nimbus_verified_proxy/libverifproxy/setup.nim
```diff
@@ -78,6 +79,7 @@ proc load(T: type VerifiedProxyConf, configJson: string): T {.raises: [ProxyErro
     storageCacheLen: storageCacheLen,
     codeCacheLen: codeCacheLen,
     accountCacheLen: accountCacheLen,
+    parallelBlockDownloads: uint64(parallelBlockDownloads),
```
tersec: If `parallelBlockDownloads` is negative, this will silently map due to two's complement into the upper half of the `uint64` range, e.g., `echo -1'i64.uint64` results in `2**64 - 1`, or `high(uint64)`. This is how unsigned integers work in Nim: they don't range-check.

nimbus_verified_proxy/engine/blocks.nim
```diff
@@ -108,34 +106,62 @@ proc walkBlocks(
...
+
+    for i in nextNum - numDownloads + 1 .. nextNum:
+      let header =
+        if engine.headerStore.contains(i):
+          engine.headerStore.get(i).get()
+        else:
+          let
+            blk =
+              ?(
+                await engine.backend.eth_getBlockByNumber(
```
tersec: How is this parallel? It's `await` in a `for` loop.

chirag-parmar: oops, my bad. https://github.com/status-im/nimbus-eth1/pull/3892/commits/2b3447989651c9016820d098dee82aecb3aa405f

tersec: https://github.com/status-im/nimbus-eth1/pull/3892/commits/2b3447989651c9016820d098dee82aecb3aa405f

nimbus_verified_proxy/libverifproxy/setup.nim
chirag-parmar: but the consumer of these integers also uses `int`. Wouldn't it make more sense to be consistent. What is the tricky-to-discover/test failure mode you are talking about? any examples?

EDIT: also why should the bit width of these configuration values be platform independent?

nimbus_verified_proxy/libverifproxy/setup.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3892/commits/978bf02969f0b7b982540194150c0ce9ce8da204

nimbus_verified_proxy/engine/blocks.nim
```diff
@@ -108,34 +106,59 @@ proc walkBlocks(
...
+        downloadedHeaders[blk.hash] = h
+
+    for j in 0 ..< futs.len:
+      let unverifiedHeader =
+        if engine.headerStore.contains(nextHash):
+          engine.headerStore.get(nextHash).get()
+        else:
+          try:
+            downloadedHeaders[nextHash]
+          except KeyError as e:
```
tersec: `e` isn't used here; can just use `except KeyError`.

nimbus_verified_proxy/engine/blocks.nim
```diff
@@ -108,34 +106,59 @@ proc walkBlocks(
...
+    nextHash = sourceHash # sourceHash is already the parent hash
+    nextNum = sourceNum - 1
+    downloadedHeaders: Table[Hash32, Header]
+    futs: seq[Future[EngineResult[BlockObject]]]
+
+  while nextNum > targetNum:
+    futs = @[]
+    downloadedHeaders.clear()
+
+    while nextNum > targetNum and uint64(futs.len) < engine.parallelBlockDownloads:
```
tersec, line 119: This seems like it won't actually download the `targetNum` block itself? (`>` vs `>=`)

nimbus_verified_proxy/engine/blocks.nim
```diff
@@ -108,34 +106,59 @@ proc walkBlocks(
...
+      if not engine.headerStore.contains(nextNum):
+        let tag = BlockTag(kind: bidNumber, number: Quantity(nextNum))
+        futs.add(engine.backend.eth_getBlockByNumber(tag, false))
 
-        trace "getting next block",
-          hash = nextHash,
-          number = blk.number,
-          remaining = distinctBase(blk.number) - targetNum
+      nextNum -= 1
```
tersec, line 124: See https://github.com/status-im/nimbus-eth1/pull/3892/changes#diff-223d8876a66a7bfacccc6d0d0c52b706341981ee91c954e03c5c4d64a268701fR119 for context, somewhat: because it won't try for `targetNum` itself this underflow/wraparound won't hit. But if `targetNum == 0`, and `nextNum` manages to hit 0, then `nextNum -= 1` wraps to `high(uint64)`, and this loop might not exit, because `nextNum > targetNum` will be true again.

nimbus_verified_proxy/libverifproxy/setup.nim
tersec: Yeah, it's good to be consistent and minimize the integer type conversions required (exception: safe conversions which don't change signedness and don't decrease range, so e.g., `uintfoo` to `uint64` or `intfoo` to `int64`). So in general `int` is a type which might be approached cautiously. Sometimes it's basically required, e.g., anything which ends up as an index into an `array` or `seq` directly or indirectly, has to be an `int` anyway to work. But sometimes there's no strong constraint of that sort and one can get this consistency without using platform-dependent types.

> EDIT: also why should the bit width of these configuration values be platform independent?

A couple of main advantages:

1. it means that the end-user UX and functionality is as much the same as possible regardless of whether it's compiled as 32-bit or 64-bit. For the rest of the `nimbus-eth1` repo, practically it will almost always be used having been compiled as 64-bit code, but the NVP has a decent chance of being used in some 32-bit applications due to its target embedded functionality/API; and

2. arguably at least as importantly, it decreases the scenarios wherein someone tests it on a 32-bit platform and some issue proves difficult to reproduce on the 64-bit platforms which run in CI, on the machines on which we're developing all this software, etc. It's easier for minimal differences to exist which can cause either UX or debugging isues.

nimbus_verified_proxy/engine/blocks.nim
chirag-parmar: that's the intended logic. Since we are just verifying if the blockhash is part of the canonical chain the parent hash of one block before would give us the info we need.

tersec: Ok

nimbus_verified_proxy/engine/blocks.nim
tersec: So this won't trigger either

nimbus_verified_proxy/engine/blocks.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3892/commits/7138369aa07342a7ca74f98a961533e9a5b92ca4

nimbus_verified_proxy/libverifproxy/setup.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3892/changes/f1eecb2062b1515711824a6a5749421853810666

Issue #3899, opened 2026-01-07
Unify client starting process including import of blocks


advaita-saha:
Currently we need to import `era1/era` files, or even the `rlp block files` first and then separately start the client.
Pretty much other clients can do this, with a flag where it can consume blocks from files before starting the client.

The ideal flow would be to check if `era/era1/rlp` files are there i.e those flags are passed or not.
If yes then load the blocks from there, and then start the client.
Maybe also add a debug flag to signal if the chain segment that we are loading is finalized or not.  

This whole things creates a simple workflow where users can just pass the flag where the era1/era/rlp files are and then not to worry again about if the import completed, and we need to start the client actually now, it would do that on it's own.

Another added benefit that comes from this feature would be easier testing in hive. Now in hive the chain segment that is imported is not fully finalized, rather a part of the chain segment is finalized and rest of the chain segment is not. This would iron out the issues with nimbus-eth1 and hive therefore, since just after the import we get one fCU call determining the head and finalized. But stopping the client after the import we don't know which is the head the finalized.

Issue #3793, opened 2025-10-24
Update release, kurtosis etc to use `nimbus`


arnetheduck:
[nimbus](#3646) offers a superset of the `nimbus_execution_client` functionality and should be used for release binaries etc..

Where relevant, `nimbus executionClient` (or `nimbus execution_client` - both work, need to pick one) replace the existing `nimbus_execution_client`.

advaita-saha:
Need to change EF/ethpandaops client builder first 
And our docker images, before moving ahead with these changes

advaita-saha:
Also please let's do `nimbus execution_client`, for consistency
And also because camel case in this scenario look really weird ( atleast in my eyes )

arnetheduck:
we recommend it for `trustedNodeSync` - I don't have any particular opinion beyond using the same for both

advaita-saha:
Once #3909 is merged

- [x] Update Dockerfile to build the `nimbus` binary
- [x] Update kurtosis to use the single binary
- [x] Update hive script to use the single binary 
- [x] Maybe consistent versioning of EL and single binary 
- [x] Update `nimbus-eth1` release CI to release the single binary

PR #3909, opened 2026-01-13
address Unify client starting process for rlp

RazorClient:
addresses https://github.com/status-im/nimbus-eth1/issues/3899

- introduces rlp import as optional during client boot up process instead of previous seperate import-rlp which helps in hive testing of different finality points in the chain segment
- introduces the flag  ```--bootstrap-blocks-file:<path>  — RLP-encoded block files to import before startup.```
 - introduces the flag  ```--debug-bootstrap-finalized:<bool> ( default false) — if true, treat those bootstrap treated segments as finalized when importing.```

### DIFF
--- a/scripts/check_revision.sh
+++ b/scripts/check_revision.sh
@@ -42,7 +42,7 @@ if [[ $COPYRIGHT_YEAR_START_GET != *$COPYRIGHT_YEAR_START_EXPECT*  ]]; then
 fi
 
 LAUNCHING_INFO_EXPECT="Launching execution client"
-LAUNCHING_INFO_GET=$($EL --help | head -n 1)
+LAUNCHING_INFO_GET=$($EL import | head -n 1)
 
 if [[ $LAUNCHING_INFO_GET != *$LAUNCHING_INFO_EXPECT*  ]]; then
   echo "\"Launching execution client\" should appear at first line"


RazorClient:
hey cleaned up a bit by removing the stuff in nim_import and also discarding the import arm as executionclientconf how has all the options that import used to have by itself 

will need a bit of input about the finalization part and how to reduce the complexity there @advaita-saha

execution_chain/nimbus_execution_client.nim
```diff
@@ -390,12 +390,37 @@ proc main*(config = makeConfig(), nimbus = NimbusNode(nil)) {.noinline.} =
...
+
+        try:
+          (waitFor importRlpFiles(files, com, config.bootstrapBlocksFinalized)).isOkOr:
+            fatal "Failed importing bootstrap RLP blocks", msg = error, finalized = config.bootstrapBlocksFinalized
+            quit(QuitFailure)
+        except CancelledError:
+          raiseAssert "Bootstrap import future should not be cancelled"
+
+      notice "Pre-start import complete"
+
```
advaita-saha: This needs to be further simplified if we are not dealing with era import, and only care about rlp import

RazorClient: addressed

execution_chain/core/block_import.nim
```diff
@@ -74,31 +76,35 @@ proc importRlpBlocks*(blocksRlp:seq[byte],
...
   ok()
 
-proc importRlpBlocks*(importFile: string,
-                     chain: ForkedChainRef,
-                     finalize: bool): Future[Result[void, string]] {.async: (raises: [CancelledError]).} =
+proc importRlpBlocks*(
+    importFile: string,
+    chain: ForkedChainRef,
+    finalize: bool
+  ): Future[Result[void, string]] {.async: (raises: [CancelledError]).} =
```
advaita-saha: Change not required

execution_chain/core/block_import.nim
```diff
@@ -16,16 +16,18 @@ import
...
-                      chain: ForkedChainRef,
-                      finalize: bool):
-                        Future[Result[void, string]] {.async: (raises: [CancelledError]).} =
+# Only parse the RLP data and feed blocks into the ForkedChainRef.
+# Optionally finalize via fork-choice when requested.
+proc importRlpBlocks*(
+    blocksRlp: seq[byte],
+    chain: ForkedChainRef,
+    finalize: bool
+  ): Future[Result[void, string]] {.async: (raises: [CancelledError]).} =
```
advaita-saha: Change not required

execution_chain/core/block_import.nim
```diff
@@ -74,31 +76,35 @@ proc importRlpBlocks*(blocksRlp:seq[byte],
...
-      warn "Error when importing blocks", msg=error
-      # Finalize the existing chain in case of rlp read error
-      (await chain.forkChoice(chain.latestHash, chain.latestHash)).isOkOr:
-        error "Error when finalizing chain", msg=error
+  for blocksFile in files:
+    (await importRlpBlocks(blocksFile, chain, config.bootstrapBlocksFinalized)).isOkOr:
+      warn "Error when importing blocks", msg = error
+      if config.bootstrapBlocksFinalized:
+        (await chain.forkChoice(chain.latestHash, chain.latestHash)).isOkOr:
+          error "Error when finalizing chain", msg = error
```
advaita-saha: what's the point of creating the `files` variable ?

advaita-saha (changes requested):

execution_chain/core/block_import.nim
RazorClient: linted and fixed

execution_chain/core/block_import.nim
RazorClient: the problem was 
```
importRlpBlocks(blocksFile, chain, config.bootstrapBlocksFinalized)
  [1] blocksFile: InputFile
  [2] chain: ForkedChainRef
  [3] config.bootstrapBlocksFinalized: bool

Expected one of (first mismatch at [position]):
[1] proc importRlpBlocks(blocksRlp: seq[byte]; chain: ForkedChainRef; finalize: bool): InternalRaisesFuture[
    Result[void, string], (CancelledError,)]
[1] proc importRlpBlocks(config: ExecutionClientConf; com: CommonRef): InternalRaisesFuture[
    void, (CancelledError,)]
[1] proc importRlpBlocks(importFile: string; chain: ForkedChainRef; finalize: bool): InternalRaisesFuture[
    Result[void, string], (CancelledError,)]
```
    
now just made it so that 
```
for blocksFile in config.bootstrapBlocksFile:
  let filePath = string(blocksFile)
```

scripts/check_revision.sh
```diff
@@ -42,7 +42,7 @@ if [[ $COPYRIGHT_YEAR_START_GET != *$COPYRIGHT_YEAR_START_EXPECT*  ]]; then
 fi
 
 LAUNCHING_INFO_EXPECT="Launching execution client"
-LAUNCHING_INFO_GET=$($EL import-rlp | head -n 1)
+LAUNCHING_INFO_GET=$($EL --help | head -n 1)
```
advaita-saha:
```suggestion
LAUNCHING_INFO_GET=$($EL import | head -n 1)
```

execution_chain/conf.nim
```diff
@@ -512,6 +511,18 @@ type
               " CL which will result in a smaller memory footprint"
         name: "debug-beacon-sync-target-is-final".}: bool
 
+      bootstrapBlocksFile* {.
+        desc: "Import RLP encoded block files before starting the client"
+        defaultValue: @[]
+        name: "bootstrap-blocks-file" .}: seq[InputFile]
```
kdeme, line 517: It is my understanding that this is a way of importing mostly used for testing? And thus really just a dev option?

I mean, the UX of these files is not really there compared to era (where to find them, command line repeat, upfront validation not existing?, etc). Hence, to not clutter cli UX here, would it not be better to be a debug/hidden flag?

RazorClient: yeah does make sense 
making it hidden like the rockdb params would be the best idea then imo

execution_chain/nimbus_execution_client.nim
```diff
@@ -390,11 +396,6 @@ proc main*(config = makeConfig(), nimbus = NimbusNode(nil)) {.noinline.} =
   case config.cmd
   of NimbusCmd.`import`:
     importBlocks(config, com)
-  of NimbusCmd.`import - rlp`:
```
bhartnett, line 393: Removing this will have broken the hive tests. See here: https://github.com/ethereum/hive/blob/0976a64d8d4e522a4eb5a8c30d0b0b9c8e8fd41d/clients/nimbus-el/nimbus.sh#L91.

So hive needs to be updated as well.

@advaita-saha @RazorClient

RazorClient: The general idea was to transition the hive tests to the single binary's commands as per https://github.com/status-im/nimbus-eth1/issues/3793; however, as pointed out, the test will break 
so question is, revert this or update Hive?

cc: @advaita-saha

advaita-saha: > Removing this will have broken the hive tests. See here: https://github.com/ethereum/hive/blob/0976a64d8d4e522a4eb5a8c30d0b0b9c8e8fd41d/clients/nimbus-el/nimbus.sh#L91.
> 
> 
> 
> So hive needs to be updated as well.
> 
> 
> 
> @advaita-saha @RazorClient 

Aware, hive sync tests are already broken
Will make a PR to harmonise everything along with transition to single binary

bhartnett: FYI, for BALs we use the eels/consume-rlp hive tests which use this rlp import command as well. And these have been working fine up to this point.

advaita-saha: Yes
ethereum/sync tests fails

I made a very bandaid like fix for that

advaita-saha: @bhartnett https://github.com/ethereum/hive/pull/1381

PR #3910, opened 2026-01-13
Bump nim-eth and accompanying changes

kdeme:
Relies on https://github.com/status-im/nimbus-eth2/pull/7850


bhartnett:
> Well, first this needs to go in [status-im/nimbus-eth2#7850](https://github.com/status-im/nimbus-eth2/pull/7850). And I'm not sure if that is sufficiently safe, probably should get a fix for that issue first. But yes, that means nim-eth repo is kind of not bump-able for now.

This is blocking a separate update which I need to get into nim-eth and nimbus-eth1 so perhaps the nim-eth change which was merged should be reverted for now so that other updates can continue? The CI fails without having this PR merged when I use the newest version of nim-eth.

PR #3911, opened 2026-01-14
EIP-7928: Block level access lists

bhartnett:
This PR contains all the changes required to implement EIP-7928 which includes:
- Added BAL tracking into the evm to collect the required data for BAL building.
- BAL pre and post execution validation.
- Engine API changes.
- Added two new RPC endpoints debug_getBlockAccessList and debug_getBadBlocks.

BALs are only enabled at Amsterdam and the tracking is disabled otherwise so this change should have minimal impact pre-Amsterdam.


execution_chain/common/chain_config.nim
```diff
@@ -508,7 +508,7 @@ func defaultBlobSchedule*(): array[Cancun..HardFork.high, Opt[BlobSchedule]] =
     Bpo3  : Opt.none(BlobSchedule),
     Bpo4  : Opt.none(BlobSchedule),
     Bpo5  : Opt.none(BlobSchedule),
-    Amsterdam: Opt.none(BlobSchedule),
+    Amsterdam: Opt.some(BlobSchedule(target: 14'u64, max: 21'u64, baseFeeUpdateFraction: 11_684_671'u64)),
```
etan-status, line 511: Just as a side note, Kurtosis genesis generator by default doesn't configure Bpo1 and Bpo2, so Amsterdam is set to Osaka values. This is not necessarily more / less correct, it just is a difference in defaults.

etan-status: One still has to be careful, that if _no BPO_ is scheduled, that it doesn't randomly assume default BPOs. But if that's wrong, it wasn't a regression in this specific PR as Bpo1/Bpo2 are already assumed to be part of the 'default blob schedule'.

PR #3917, opened 2026-01-16
update Dockerfile to build nimbus single binary

advaita-saha:
### Primary changes to running _**Nimbus Execution Client**_ 

Every flag remains same, which were used earlier. The only thing changes is how the client is started, so an extra argument is passed while running the execution client

-----------

_Changes from_
```bash
./build/nimbus_execution_client --network=<> --data-dir=<> . . . . . 
```
_Changes to_
```bash
./build/nimbus executionClient --network=<> --data-dir=<> . . . . . 
```

Addresses #3793


tersec:
Kurtosis issues?

bhartnett:
Building the `nimbus` target is quite a bit slower than building `nimbus_execution_client`. For the use cases such as hive that just require the execution client I wonder if there is much benefit of using the single binary. The code paths tested in both scenarios should mostly be the same.

Also we should keep the usage of `nimbus_execution_client` at least for development purposes so perhaps it might be worth keeping a dockerfile that builds just the execution client.

PR #3922, opened 2026-01-19
proxy: export backend to C API and unified method call

chirag-parmar:
### Description:
This PR brings in the following changes:
1. `userData` field for all exported functions to allows linking of concurrent requests with their responses.
2. Unified method call `nvp_call` which takes the RPC method name as a parameter along with other RPC parameters
3. Exporting the engine backend via an abstracted out transport function that passes the RPC method name and params as string. This allows defining the transport on the C side.

### DIFF
--- a/nimbus_verified_proxy/libverifproxy/types.nim
+++ b/nimbus_verified_proxy/libverifproxy/types.nim
@@ -110,6 +110,6 @@ func packArg*[T](arg: T): Result[string, string] {.raises: [].} =
 
 proc alloc*(str: string): cstring =
   var ret = cast[cstring](allocShared(str.len + 1))
-  copyMem(addr ret[0], addr str[0], str.len)
+  copyMem(ret, str.cstring, str.len)
   ret[str.len] = '\0'
   return ret


nimbus_verified_proxy/libverifproxy/verifproxy.h
```diff
@@ -93,41 +94,56 @@ void stopVerifProxy(Context *ctx);
...
+/**
+ * call any RPC method
+ *
+ * @param ctx       Context pointer.
+ * @param reqId     Request ID
+ * @param name      Name of the RPC method
+ * @param params    parameters required for the RPC method
+ * @param cb        Callback invoked with a hex block number.
+ */
+void nvp_call(Context *ctx, unsigned int reqId, char* name, char* params, CallBackProc cb);
```
arnetheduck: in general, rather than an integer, the custom in C apis is to add a `void*` or `uintptr_t` "user data" field that gets passed around - see the nwaku ffi.

integers are smaller than pointers and therefore not convenient in C.

chirag-parmar: fair point. More dynamic that way. I'll make the necessary changes

chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3922/commits/5f3ca132c4967fb11cc38c465d46059114dec8f8

arnetheduck: https://github.com/logos-messaging/logos-messaging-nim/blob/master/library/libwaku.h#L34 - waku places it last though - since there's no reason not to, better follow their lead

chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3922/commits/6073e17e6a826df8fe126c41184417fc55bbbf41

nimbus_verified_proxy/libverifproxy/types.nim
```diff
@@ -23,11 +30,86 @@ type
...
+# generalized overloading
+func packArg*[T](arg: T): Result[string, string] {.raises: [].} =
+  try:
+    ok(JrpcConv.encode(arg))
+  except CatchableError as e:
+    err("Parameter coudln't be encoded: " & e.msg)
+
+proc alloc*(str: string): cstring =
+  var ret = cast[cstring](allocShared(str.len + 1))
+  let s = cast[seq[char]](str)
```
tersec: Why do this `cast` if the only thing one is doing with it is to index into each character?  Can [index into `string`s directly](https://nim-lang.org/docs/system.html#%5B%5D.t%2Cstring%2Cint).

nimbus_verified_proxy/libverifproxy/types.nim
```diff
@@ -1,16 +1,23 @@
...
-# Copyright (c) 2025 Status Research & Development GmbH
+# Copyright (c) 2025-2026 Status Research & Development GmbH
 # Licensed and distributed under either of
 #   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-import std/lists, ../engine/types, chronos
+import
+  std/[strutils, lists],
```
tersec: Generally prefer lexicographically ordered.

nimbus_verified_proxy/libverifproxy/types.nim
```diff
@@ -1,16 +1,23 @@
 # nimbus_verified_proxy
-# Copyright (c) 2025 Status Research & Development GmbH
+# Copyright (c) 2025-2026 Status Research & Development GmbH
 # Licensed and distributed under either of
 #   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-import std/lists, ../engine/types, chronos
+import
```
tersec, line 10: Nicer to add `{.push raises: [], gcsafe.}` at the top, before the imports, than have to repeat them everywhere.

chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3922/commits/1c56e2f535458952f23ab127229f4b54331218d2

nimbus_verified_proxy/libverifproxy/types.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3922/commits/1c56e2f535458952f23ab127229f4b54331218d2

nimbus_verified_proxy/libverifproxy/types.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3922/commits/628cc4a6e2d1afef8cd346c23c56f8df2bdc889d

nimbus_verified_proxy/libverifproxy/types.nim
```diff
@@ -23,11 +32,84 @@ type
...
+    err("Parameter of type " & $argType & " coudln't be decoded: " & e.msg)
+
+# generalized overloading
+func packArg*[T](arg: T): Result[string, string] {.raises: [].} =
+  try:
+    ok(JrpcConv.encode(arg))
+  except CatchableError as e:
+    err("Parameter coudln't be encoded: " & e.msg)
+
+proc alloc*(str: string): cstring =
```
tersec, line 111: Can be `func`.

nimbus_verified_proxy/libverifproxy/types.nim
```diff
@@ -23,11 +32,84 @@ type
...
+  var ret = cast[cstring](allocShared(str.len + 1))
+  copyMem(addr ret[0], addr str[0], str.len)
```
tersec:
```nim
proc alloc*(str: string): cstring =
  var ret = cast[cstring](allocShared(str.len + 1))
  copyMem(addr ret[0], addr str[0], str.len)
  ret[str.len] = '\0'
  return ret

discard alloc("")
```
results in
```
nimbus-eth2/v.nim(7) a
nimbus-eth2/v.nim(3) _ZN1a5allocE6string
nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system/chcks.nim(28) raiseIndexError2
nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system/fatal.nim(53) _ZN6system8sysFatalE8typeDescIN10exceptions11IndexDefectEE6string
nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system/excpt.nim(349) _ZN6system18rawWriteStackTraceE3varI3seqIN6system15StackTraceEntryEEE
nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system/stacktraces.nim(62) _ZN11stacktraces30auxWriteStackTraceWithOverrideE3varI3seqIN6system15StackTraceEntryEEE
Error: unhandled exception: index out of bounds, the container is empty [IndexDefect]
```
because `addr str[0]` doesn't work.

For example, `stew/ptrops` [handles this](https://github.com/status-im/nim-stew/blob/b3b3760c3d89c88cd8e291fd64d4e66980f6b3a3/stew/ptrops.nim#L52-L54) with:
```nim
func baseAddr*[T](x: openArray[T]): ptr T =
  # Return the address of the zero:th element of x or `nil` if x is empty
  if x.len == 0: nil else: cast[ptr T](x)
```
But that doesn't solve the problem, because then
```nim
  copyMem(addr ret[0], nil, str.len)
```
becomes UB, because `memcpy()`, becomes UB, e.g., https://en.cppreference.com/w/cpp/string/byte/memcpy notes
> If any of the following conditions is satisfied, the behavior is undefined:
>
> - `dest` or `src` is a [null pointer](https://en.cppreference.com/w/cpp/language/pointer.html#Null_pointers) or [invalid pointer](https://en.cppreference.com/w/cpp/language/pointer.html#Invalid_pointers).

in C++ and, likewise [in C](https://en.cppreference.com/w/c/string/byte/memcpy.html):
> The behavior is undefined if access occurs beyond the end of the dest array. If the objects overlap (which is a violation of the [restrict](https://en.cppreference.com/w/c/language/restrict.html) contract)(since C99), the behavior is undefined. The behavior is undefined if either `dest` or `src` is an invalid or null pointer.

This is true even when copying 0 bytes, e.g., look at any of the commentary on the UB-ness of `memcpy(NULL, NULL, 0)`. It might [become not-undefined at some future point](https://developers.redhat.com/articles/2024/12/11/making-memcpynull-null-0-well-defined), but it's not yet.

nimbus_verified_proxy/libverifproxy/types.nim
chirag-parmar: I think `allocShared` has side effects

nimbus_verified_proxy/libverifproxy/types.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/3922/commits/75284eac11ae7dbbb9763d6509503b73da27bb4d

PR #3923, opened 2026-01-19
Add nimbus_portal_client binary to the release procedure

kdeme:


.github/workflows/release.yml
```diff
@@ -76,18 +76,22 @@ jobs:
...
           NVP_TAR="nimbus_verified_proxy-${OS}-${CPU}-${TAG}-$(git rev-parse --short=8 HEAD).tar.gz"
+          NPC_TAR="nimbus_portal_client-${OS}-${CPU}-${TAG}-$(git rev-parse --short=8 HEAD).tar.gz"
 
           # Pack only the relevant binaries (and their checksums) from build/
           # Add additional files (LICENSE, README, configs) if desired with extra -C args.
-          tar -czf "${NEC_TAR}" --exclude="build/nimbus_verified_proxy*" -C "${ARCHIVE_DIR}" .
-          tar -czf "${NVP_TAR}" --exclude="build/nimbus_execution_client*" -C "${ARCHIVE_DIR}" .
+          tar -czf "${NEC_TAR}" --exclude="build/nimbus_verified_proxy*" --exclude="build/nimbus_portal_client*" -C "${ARCHIVE_DIR}" .
+          tar -czf "${NVP_TAR}" --exclude="build/nimbus_execution_client*" --exclude="build/nimbus_portal_client*" -C "${ARCHIVE_DIR}" .
+          tar -czf "${NPC_TAR}" --exclude="build/nimbus_execution_client*" --exclude="build/nimbus_verified_proxy*" --exclude="build/librocksdb.dll" -C "${ARCHIVE_DIR}" .
```
kdeme, line 85: To avoid future adds perhaps we should work include based here instead (e.g. use --wildcards)

.github/workflows/release.yml
```diff
@@ -117,6 +121,20 @@ jobs:
           path: ./dist/${{ steps.make_dist.outputs.nvp_tar }}.sha512sum
           retention-days: 2
 
+      - name: Upload nimbus portal client tarball artefact
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.os }}-${{ matrix.cpu }}-nimbus-portal-client
```
kdeme, line 127: used full name here compared to other two because why not.

I can alter all to be consistent if wanted, or just call it portal.

docker/dist/README.md.tpl
```diff
@@ -1,8 +1,8 @@
 # Binary Nimbus distribution
 
-This binary distribution of the Nimbus Execution Client & Verified Proxy package is compiled
-in a [reproducible way](https://reproducible-builds.org/) from source files
-hosted at https://github.com/status-im/nimbus-eth1.
+The binary distribution of the Nimbus Execution Client, Nimbus Verified Proxy and
```
kdeme, line 3: Perhaps here should just call it more generally "the binary distributions of Nimbus are compiled in a ...", to avoid having to adjust each time.

kdeme:
Closing this as we are currently not sure yet if we want to support a stand-alone portal client in the long term.

Use cases currently are be:
1. To be able to run multiple stand-alone portal clients with only one EL: This is how our Portal fleet is run currently and is a rather short term use case to be able to have a lot of nodes up and running.
2.  Use in combination with a portal bridge to inject data (how we initially populated the network with data). Eventually this should become purely EL steered.
3. Run in standalone mode to only read from the network (this does not allow you to validate the data). Glados uses this for auditing purposes.
4. Have a standalone mode so that it can be used by other EL clients.

Only point 4 is really long term and a thus a long term reason to support the release of a binary.
First step before this would/could be  to first integrate the portal client in nimbus_execution_client / nimbus unified client. 

The situation does change if we were to launch & support a header network with headers with proofs, as at that moment you can validate the data without an EL.

PR #3928, opened 2026-01-21
Apply bal without executing block for finalized blocks

bhartnett:
As a sync optimization we can skip executing the block when we have a block access list available for a finalized block. In this case we simply apply the BAL and then persist the ledger changes without executing the transactions. 

The function which applies the BAL to the state, loops through each account and writes the last update for each type of change (balance, nonce, code, storage). This should be much faster than executing the full block.

This feature is also used when replaying blocks on startup which should speed up that part which was previously rather slow.


bhartnett:
We should be able to apply the BAL in this way when downloading the state and catching up during snap sync, assuming that we don't need to keep receipts for those blocks. Once we have implemented snap sync we can revisit this.

PR #3953, opened 2026-02-02
pruner

advaita-saha:
- [x] Convert to eth2 similar tracking i.e tail
- [x] organize input flags properly ( make debug where ideal )
- [x] Use the similar eth2 retention time
- [x] Make `import` command to update the tail in db

### DIFF
--- a/ci/hive.groovy
+++ b/ci/hive.groovy
@@ -74,6 +74,11 @@ pipeline {
         """
       }
     }
+    stage('Build hiveproxy') {
+      steps {
+        sh 'docker build -t hive/hiveproxy:latest ${WORKSPACE}/hive/hiveproxy'
+      }
+    }
     stage('Run Hive Tests') {
       parallel {
         stage('sync reth-nimbus') {
--- a/portal/bridge/beacon/portal_beacon_bridge.nim
+++ b/portal/bridge/beacon/portal_beacon_bridge.nim
@@ -1,5 +1,5 @@
 # Nimbus
-# Copyright (c) 2023-2025 Status Research & Development GmbH
+# Copyright (c) 2023-2026 Status Research & Development GmbH
 # Licensed and distributed under either of
 #   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
@@ -248,32 +248,31 @@ proc gossipHistoricalSummaries(
     return err("No historical_summaries found")
 
   let summariesForked = summariesOpt.get()
-  withForkyHistoricalSummariesWithProof(summariesForked):
-    when historicalFork >= HistoricalSummariesFork.Electra:
-      let
-        epoch = forkySummaries.slot.epoch()
-        forkDigest = forkDigestAtEpoch(forkDigests[], epoch, cfg)
-        summariesWithProof = HistoricalSummariesWithProof(
-          epoch: epoch,
-          historical_summaries: forkySummaries.historical_summaries,
-          proof: forkySummaries.proof,
-        )
-        contentKey = encode(historicalSummariesContentKey(epoch.uint64))
-        content = encodeSsz(summariesWithProof, forkDigest)
+  case summariesForked.kind
+  of HistoricalSummariesFork.Electra:
+    let
+      forkySummaries = summariesForked.electraData
+      epoch = forkySummaries.slot.epoch()
+      forkDigest = forkDigestAtEpoch(forkDigests[], epoch, cfg)
+      summariesWithProof = HistoricalSummariesWithProof(
+        epoch: epoch,
+        historical_summaries: forkySummaries.historical_summaries,
+        proof: forkySummaries.proof,
+      )
+      contentKey = encode(historicalSummariesContentKey(epoch.uint64))
+      content = encodeSsz(summariesWithProof, forkDigest)
 
-      try:
-        let peers = await portalRpcClient.portal_beaconRandomGossip(
-          contentKey.asSeq().toHex(), content.toHex()
-        )
-        info "Beacon historical_summaries gossiped", peers, epoch
+    try:
+      let peers = await portalRpcClient.portal_beaconRandomGossip(
+        contentKey.asSeq().toHex(), content.toHex()
+      )
+      info "Beacon historical_summaries gossiped", peers, epoch
 
-        ok()
-      except CatchableError as e:
-        err("JSON-RPC error: " & $e.msg)
-    elif historicalFork >= HistoricalSummariesFork.Capella:
-      err("No historical summaries pre-Electra should be gossiped")
-    else:
-      err("No historical summaries pre-Capella")
+      ok()
+    except CatchableError as e:
+      err("JSON-RPC error: " & $e.msg)
+  of HistoricalSummariesFork.Capella:
+    err("No historical summaries pre-Electra should be gossiped")
 
 proc runBeacon*(config: PortalBridgeConf) {.raises: [CatchableError].} =
   notice "Launching Nimbus Portal beacon chain bridge", cmdParams = commandLineParams()


execution_chain/pruner.nim
```diff
@@ -13,12 +13,16 @@ import
...
   results,
+  eth/common/times,
   ./db/core_db,
   ./common
 
 logScope:
   topic = "pruner"
 
+const
+  RetentionPeriod = 6 * 30 * 24 * 60 * 60'u64 # ~6 months in seconds
```
tersec: Is this supposed to be ~6 months, or specifically align with [`MIN_EPOCHS_FOR_BLOCK_REQUESTS`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/phase0/p2p-interface.md#configuration) (33,024 epochs)? There was discussion of this but I'm not sure if/how it was resolved.

execution_chain/pruner.nim
```diff
@@ -0,0 +1,125 @@
...
+      await sleepAsync(pruner.loopDelay)
+      continue
+
+    notice "Background pruner: starting cycle",
+      fromBlock = begin, toBlock = start, cutoffTimestamp = cutoff.uint64
+
+    var currentBlock = begin
+    var reachedRetentionWindow = false
+    var prevBaseNumber = start
+    var prunedSincePersist = 0'u64
```
tersec: 4 `var`s probably warrants a `var` block.

execution_chain/pruner.nim
```diff
@@ -0,0 +1,125 @@
...
+      start = pruner.com.db.baseTxFrame.getSavedStateBlockNumber()
+      begin = pruner.com.db.baseTxFrame.getHistoryExpired()
+      cutoff = EthTime(EthTime.now().uint64 - RetentionPeriod)
+
+    if begin >= start:
+      await sleepAsync(pruner.loopDelay)
+      continue
+
+    notice "Background pruner: starting cycle",
+      fromBlock = begin, toBlock = start, cutoffTimestamp = cutoff.uint64
```
tersec: It's somewhat more typesafe to use `distinctBase` instead of explicit things such as `uint64`, because it guarantees at callsite that it's a runtime-safe integer conversion.

execution_chain/db/core_db/core_apps.nim
```diff
@@ -678,6 +685,142 @@ proc getCodeByHash*(db: CoreDbTxRef, codeHash: Hash32): Result[seq[byte], string
...
+    blockHash: Hash32;
+      ): Result[void, string] =
+  const info = "deleteBlockBody()"
+  let header = ?db.getBlockHeader(blockHash)
+
+  try:
+    db.deleteTransactions(header.transactionsRoot)
+    db.deleteUncles(header.ommersHash)
+    if header.withdrawalsRoot.isSome:
+      db.deleteWithdrawals(header.withdrawalsRoot.get(EMPTY_ROOT_HASH))
```
tersec: Can use [`isErrOr`](https://github.com/arnetheduck/nim-results/blob/df8113dda4c2d74d460a8fa98252b0b771bf1f27/results.nim#L1233-1265).

execution_chain/db/core_db/core_apps.nim
```diff
@@ -678,6 +685,142 @@ proc getCodeByHash*(db: CoreDbTxRef, codeHash: Hash32): Result[seq[byte], string
...
+    db: CoreDbTxRef;
+    blockNumber: BlockNumber;
+      ): Result[void, string] =
+  const info = "deleteBlockBody()"
+  let header = ?db.getBlockHeader(blockNumber)
+
+  db.deleteTransactions(header.transactionsRoot)
+  db.deleteUncles(header.ommersHash)
+  if header.withdrawalsRoot.isSome:
+    db.deleteWithdrawals(header.withdrawalsRoot.get())
```
tersec: [`isErrOr`](https://github.com/arnetheduck/nim-results/blob/df8113dda4c2d74d460a8fa98252b0b771bf1f27/results.nim#L1233-1265)

execution_chain/db/core_db/core_apps.nim
```diff
@@ -678,6 +685,142 @@ proc getCodeByHash*(db: CoreDbTxRef, codeHash: Hash32): Result[seq[byte], string
...
+    blockNumber: BlockNumber;
+      ): Result[void, string] =
+  const info = "deleteBlockBodyAndReceipts()"
+  let
+    header = ?db.getBlockHeader(blockNumber)
+
+  db.deleteTransactions(header.transactionsRoot)
+  db.deleteUncles(header.ommersHash)
+  if header.withdrawalsRoot.isSome:
+    db.deleteWithdrawals(header.withdrawalsRoot.get())
```
tersec: [`isErrOr`](https://github.com/arnetheduck/nim-results/blob/df8113dda4c2d74d460a8fa98252b0b771bf1f27/results.nim#L1233-1265)

execution_chain/db/storage_types.nim
```diff
@@ -67,6 +68,10 @@ func dataDirIdKey*(): DbKey {.inline.} =
   result.data[0] = byte ord(dataDirId)
   result.dataEndPos = 1
 
+func historyExpiryIdKey*(): DbKey {.inline.} =
```
tersec: Is there a performance advantage to this? see https://status-im.github.io/nim-style-guide/language.inline.html

execution_chain/conf.nim
```diff
@@ -575,6 +581,24 @@ type
         defaultValue: false
         name: "debug-store-slot-hashes".}: bool
 
+    of NimbusCmd.`prune`:
+      purgeLimit* {.
+        desc: "Purge history in database till this block number"
```
tersec: "through" would be more idiomatic (till/until are a bit heavy on the time-metaphor)

execution_chain/nimbus_prune.nim
```diff
@@ -0,0 +1,60 @@
...
+
+proc prune*(config: ExecutionClientConf, com: CommonRef) =
+
+  let
+    start = com.db.baseTxFrame().getSavedStateBlockNumber()
+    begin = com.db.baseTxFrame().getHistoryExpired()
+    batchSize = 150'u64
+
+  notice "Current database at", blockNumber = start
+  notice "Purging all block bodies till", start
```
tersec: But for logging, this is a bit more nuanced, because it should work with both JSON and non-JSON logging. For JSON-logging, it's not going to be necessarily concatenated in the correct order for this to read properly in the logs.

execution_chain/nimbus_prune.nim
```diff
@@ -0,0 +1,60 @@
...
+  not ProcessState.stopIt(notice("Shutting down", reason = it))
+
+proc prune*(config: ExecutionClientConf, com: CommonRef) =
+
+  let
+    start = com.db.baseTxFrame().getSavedStateBlockNumber()
+    begin = com.db.baseTxFrame().getHistoryExpired()
+    batchSize = 150'u64
+
+  notice "Current database at", blockNumber = start
```
tersec: Similar issue with "at"/JSON logging.

execution_chain/nimbus_prune.nim
```diff
@@ -0,0 +1,60 @@
...
+    txFrame = com.db.baseTxFrame().txFrameBegin()
+
+  
+  while running() and currentBlock <= start:
+    txFrame.deleteBlockBodyAndReceipts(currentBlock).isOkOr:
+      warn "Failed", blkNum=currentBlock, error
+    
+    if (currentBlock mod batchSize) == 0:
+      checkpoint()
+      notice "Deletion of blocks persisted", blks=currentBlock
```
tersec: Maybe simplify to "Blocks deleted"? Or is there a useful intermediate state to note where block deletion happens but has not yet been persisted?

execution_chain/db/core_db/core_apps.nim
```diff
@@ -678,6 +685,142 @@ proc getCodeByHash*(db: CoreDbTxRef, codeHash: Hash32): Result[seq[byte], string
...
+      ): BlockNumber =
+  const info = "getHistoryExpired()"
+  let
+    key = historyExpiryIdKey()
+    blkNum = db.getOrEmpty(key.toOpenArray).valueOr:
+      warn info, error=($$error)
+      @[]
+
+  if blkNum.len == 0:
+    return BlockNumber(0)
```
tersec: `uint64.fromBytesLE` is pretty fast; is this special case useful?

On x86 (i.e. also little-endian, but so is arm64/aarch64):
```nim
import stew/endians2
var blkNum: array[8, byte]
echo uint64.fromBytesLE(blkNum)
```
becomes
```c
static N_INLINE(void,
                _ZN6system7copyMemE7pointer7pointer25range09223372036854775807)(
    void *dest_p0, void *source_p1, NI size_p2) {
  nimCopyMem(dest_p0, source_p1, size_p2);
}

/* ... */

static N_INLINE(
    NU64,
    _ZN8endians29fromBytesE8typeDescI6uInt64E9openArrayI5uInt8EN6system10EndiannessE)(
    NU8 *x_p1, NI x_p1Len_0,
    tyEnum_Endianness__TrZv6w9aMrTts1VtK8jHPpQ endian_p2) {
  NU64 result;
  result = (NU64)0;
  {
    if (!!((((NI)8) <= x_p1Len_0)))
      goto LA3_;
    _ZN10assertions16failedAssertImplE6string(
        ((NimStringDesc *)&TM__ELnAjkZEKwpyn84oKQpo5Q_2));
  }
LA3_:;

  if (((NI)0) < 0 || ((NI)0) >= x_p1Len_0) {
    raiseIndexError2(((NI)0), x_p1Len_0 - 1);
  }
  _ZN6system7copyMemE7pointer7pointer25range09223372036854775807(
      ((void *)((&result))), ((void *)((&x_p1[((NI)0)]))), ((NI)8));

  {
    if (!!((endian_p2 == ((tyEnum_Endianness__TrZv6w9aMrTts1VtK8jHPpQ)0))))
      goto LA7_;

    result = _ZN8endians29swapBytesE6uInt64(result);
  }
LA7_:;
  popFrame();
  return result;
}

/* ... */

static N_INLINE(NU64,
                _ZN8endians211fromBytesLEE8typeDescI6uInt64E9openArrayI5uInt8E)(
    NU8 *x_p1, NI x_p1Len_0) {
  NU64 result;
  result = (NU64)0;
  result =
      _ZN8endians29fromBytesE8typeDescI6uInt64E9openArrayI5uInt8EN6system10EndiannessE(
          x_p1, x_p1Len_0, ((tyEnum_Endianness__TrZv6w9aMrTts1VtK8jHPpQ)0));
  popFrame();
  return result;
}

/* ... */

    nimZeroMem((void *)T1_, sizeof(tyArray__nHXaesL0DJZHyVS07ARPRA));
    T2_ = (NU64)0;
    T2_ = _ZN8endians211fromBytesLEE8typeDescI6uInt64E9openArrayI5uInt8E(
        blkNum__a_u3, 8);
    T1_[0] = _ZN7dollars7dollar_E6uInt64(T2_);
    echoBinSafe(T1_, 1);
```

i.e. it has two bounds checks (is the size of the `openArray` correct, i.e. replacing this check for 0 in a sense, and is the endian matching).

Branch prediction should remove the latter completely, but it would probably be better if `stew/endians2` had a `static` version of this to avoid pointless runtime code cost.

tersec: And `nimCopyMem(dest_p0, source_p1, size_p2);` should become a single 8-byte `mov` instruction or ARM equivalent.

execution_chain/nimbus_prune.nim
```diff
@@ -0,0 +1,60 @@
...
+
+proc running(): bool =
+  not ProcessState.stopIt(notice("Shutting down", reason = it))
+
+proc prune*(config: ExecutionClientConf, com: CommonRef) =
+
+  let
+    start = com.db.baseTxFrame().getSavedStateBlockNumber()
+    begin = com.db.baseTxFrame().getHistoryExpired()
+    batchSize = 150'u64
```
tersec: How is this batch size derived?

execution_chain/pruner.nim
```diff
@@ -0,0 +1,125 @@
...
+    batchSize: batchSize,
+    loopDelay: loopDelay,
+  )
+
+proc pruneLoop(pruner: BackgroundPrunerRef) {.async: (raises: [CancelledError]).} =
+  while true:
+    let
+      start = pruner.com.db.baseTxFrame.getSavedStateBlockNumber()
+      begin = pruner.com.db.baseTxFrame.getHistoryExpired()
+      cutoff = EthTime(EthTime.now().uint64 - RetentionPeriod)
```
tersec: This is [defined in `nim-eth`](https://github.com/status-im/nim-eth/blob/2b6cfd8f91174e26d8ad945a15b33d5c997d47f1/eth/common/times.nim#L14-L19) as
```nim
# Getting the current time is not supported when compiling to bare metal
when not defined(`any`) and not defined(standalone):
  from std/times import getTime, toUnix

  proc now*(_: type EthTime): EthTime =
    getTime().toUnix.EthTime
```
regarding which a couple of points:

-  if one is going to just convert this back to a `uint64`, that amounts to doing `getTime().toUnix.EthTime.uint64`, but then the `EthTime` conversion is pointless, and can just do `getTime().toUnix.uint64` directly without `eth/common/times`
- is there an expectation that this pruner should run in an OS-free/standalone/bare-metal environment?

execution_chain/conf.nim
```diff
@@ -523,6 +524,11 @@ type
         defaultValue: false
         name: "debug-bootstrap-finalized" .}: bool
 
+      backgroundPruning* {.
+        desc: "Enable background pruning of expired block bodies and receipts"
+        defaultValue: false
+        name: "background-pruning" .}: bool
```
arnetheduck: this should be a debug option - ie eventually we should have history and archive modes just like eth2

execution_chain/conf.nim
```diff
@@ -575,6 +581,24 @@ type
         defaultValue: false
         name: "debug-store-slot-hashes".}: bool
 
+    of NimbusCmd.`prune`:
```
arnetheduck: not sure this is worth exposing - ie pruning manually is probably not something we care to support .. the problem being of course that if you're running a node, you want it to be pruning continuously as the chain progresses and this cannot be done in a separate process

execution_chain/pruner.nim
```diff
@@ -0,0 +1,125 @@
...
+    com: CommonRef
+    batchSize: uint64
+    loopDelay: chronos.Duration
+    loopFut: Future[void].Raising([CancelledError])
+
+proc init*(
+    T: type BackgroundPrunerRef,
+    com: CommonRef,
+    batchSize = 40'u64,
+    loopDelay = chronos.seconds(3),
```
arnetheduck: not sure why this is pruning every 3s since blocks are only produced every 12 .. but more than that, we likely want to control pruning so it doesn't happen in the hot window when a block is about to arrive (ie just after slot start).

execution_chain/pruner.nim
arnetheduck: indeed, this should probably match eth2 so that the two keep roughly the same horizon

execution_chain/db/core_db/core_apps.nim
```diff
@@ -678,6 +685,142 @@ proc getCodeByHash*(db: CoreDbTxRef, codeHash: Hash32): Result[seq[byte], string
...
+proc setHistoryExpired*(
+    db: CoreDbTxRef;
+    blockNumber: BlockNumber;
+      ) =
+  const info = "setHistoryExpired()"
+  let value = blockNumber.toBytesLE()
+  db.put(historyExpiryIdKey().toOpenArray, value).isOkOr:
+    warn info, blockNumber, error=($$error)
+
+proc getHistoryExpired*(
```
arnetheduck: in eth2 we call this the tail, just like the head is the first block we have, the tail becomes the last block we keep track of .. probably worth taking a look at the terminology of eth2 (and implementation) of eth2 and keep this similar

execution_chain/pruner.nim
```diff
@@ -0,0 +1,125 @@
...
+
+    await sleepAsync(pruner.loopDelay)
+
+
+proc start*(pruner: BackgroundPrunerRef) =
+  pruner.loopFut = pruner.pruneLoop()
+
+proc stop*(pruner: BackgroundPrunerRef) {.async: (raises: []).} =
+  if not pruner.loopFut.isNil:
+    await noCancel pruner.loopFut.cancelAndWait()
```
arnetheduck: no need for `noCancel` with `cancelAndWait` (it already ignores cancellations)

execution_chain/pruner.nim
advaita-saha: Yes that's why the persist control lives with the FC module, on when it decides to persist.
After trying different variation of self persist, leaving the control within FC module was the best option for performance and non-blocking other async process

12s sounds right, during following the chain with a synced node
But sounds slow while syncing ( maybe can do a dynamic one )

execution_chain/conf.nim
advaita-saha: Agreed
But I want something like `ncli_db` on the eth1 side
Which helps us modify the db ( prune can be a part of that ) for debugging and emergency purposes

advaita-saha: Will add a similar interface later 
for now removing the prune command

execution_chain/nimbus_prune.nim
advaita-saha: Removed prune command ref https://github.com/status-im/nimbus-eth1/pull/3953#discussion_r2811274867

execution_chain/conf.nim
advaita-saha: Removed prune command ref https://github.com/status-im/nimbus-eth1/pull/3953#discussion_r2811274867

execution_chain/db/storage_types.nim
advaita-saha: Not sure in this context, but all the others had `{.inline.}`
https://github.com/status-im/nimbus-eth1/blob/b1ac3ae3afe53420087cc2c57be96065b34a6180/execution_chain/db/storage_types.nim#L46-L75

execution_chain/nimbus_prune.nim
advaita-saha: Removed prune command ref https://github.com/status-im/nimbus-eth1/pull/3953#discussion_r2811274867

execution_chain/nimbus_prune.nim
advaita-saha: Removed prune command ref https://github.com/status-im/nimbus-eth1/pull/3953#discussion_r2811274867

execution_chain/nimbus_prune.nim
advaita-saha: In the case of the `prune` command, it was just random
In case of the background pruner, it was adjusted based on how much time it takes to prune and blocks other async processes 

Removed prune command ref https://github.com/status-im/nimbus-eth1/pull/3953#discussion_r2811274867

execution_chain/pruner.nim
advaita-saha: Surely can align with this
Comes to around ~147days i.e ~4.8months

execution_chain/db/core_db/core_apps.nim
advaita-saha: You mean on each `deleteTransactions`, `deleteUncles` . . . . . ?
```nim
db.<deleteFoo>(<header.foo>).isErrOr:
```

These delete functions doesn't return a `Result`, mainly because we don't have a count of txs in the header
https://github.com/status-im/nimbus-eth1/blob/f5cf4e79d4fd5599155a1c80a8528144652bd980/execution_chain/db/core_db/core_apps.nim#L714-L725

execution_chain/pruner.nim
advaita-saha: 1. Sure that makes sense
2. `this pruner`, the node is not expected to run in bare metal, so also not the pruner

execution_chain/db/storage_types.nim
tersec: Yes, I'm not sure it makes sense for them either, but this PR doesn't otherwise affect them.

execution_chain/db/core_db/core_apps.nim
advaita-saha: https://github.com/status-im/nimbus-eth1/pull/3953/commits/103b214ecf83a1ec2bd106ad32634f4a8e7f59b4

execution_chain/db/storage_types.nim
```diff
@@ -67,6 +69,14 @@ func dataDirIdKey*(): DbKey {.inline.} =
   result.data[0] = byte ord(dataDirId)
   result.dataEndPos = 1
 
+func tailIdKey*(): DbKey {.inline.} =
+  result.data[0] = byte ord(tail)
+  result.dataEndPos = 1
+
+func prunerStateKey*(): DbKey {.inline.} =
```
tersec, line 76: It seems like both `tailIdKey` and `prunerStateKey` are simply constants.

advaita-saha: Yes the keys are constant, since we are storing only 1 value for each of those keys
Similar to `dataDirIdKey` and others

i.e one global value in the db

PR #3956, opened 2026-02-02
Use intops-powered Stint and bncurve, introduce intops a new dependency

moigagoo:
Stint and bncurve are migrating to the new [intops](https://github.com/vacp2p/nim-intops/) library:
- https://github.com/status-im/nim-stint/pull/170
- https://github.com/status-im/nim-bncurve/pull/16

The tests pass, the benchmarks are good (some even got better).

However, we need to validate that no projects depending in Stint or bncurve break with this change.

This PR introduces the new versions of Stint and bncurve that use intops as well as adds intops itself as a dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,11 +267,8 @@ latest commits when they are not ready to be used in the superproject.
 Adding the submodule "https://github.com/status-im/foo" to "vendor/foo":
 
 ```bash
-vendor/nimbus-build-system/scripts/add_submodule.sh status-im/foo
-# or
-./env.sh add_submodule status-im/foo
-# want to place it in "vendor/bar" instead?
-./env.sh add_submodule status-im/foo vendor/bar
+git submodule add --force https://github.com/status-im/foo.git "vendor/foo"
+git config -f .gitmodules submodule.vendor.foo.ignore untracked
 ```
 
 Removing the submodule "vendor/bar":


.gitmodules
```diff
@@ -9,10 +9,10 @@
   ignore = dirty
   branch = master
 [submodule "vendor/nim-stint"]
-  path = vendor/nim-stint
-  url = https://github.com/status-im/nim-stint.git
-  ignore = dirty
-  branch = master
+	path = vendor/nim-stint
+	url = https://github.com/moigagoo/nim-stint.git
```
tersec: It would be preferable not to switch until/unless this gets merged into the `status-im` org version of `stint`, not a personal fork.

moigagoo: @tersec yeah, sure, I don't intend to mere it like. I'll set the status to Draft to avoid confusion.

It's like a chicken-and-egg problem: I can't merge the stint or bncurve PRs until I verify they don't harm nimbus performance, but I can't verify them on Nimbus until I merge them.

The plan is to:
1. ✅ Make sure the nimbus tests pass with the updated versions of stint and bncurve.
2. Verify the benchmarks are OK (I'll do that locally since nimbus benchmarks are not run on PRs)
3. Merge stint and bncurve PRs
4. Update this PR and merge it

PR #3988, opened 2026-02-16
introduce multiarch manifest push in release

advaita-saha:


.github/workflows/release.yml
```diff
@@ -139,9 +137,39 @@ jobs:
...
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Create and push multi-arch manifest (versioned)
+        run: |
+          TAG="${{ steps.extract_tag.outputs.tag }}"
+          docker manifest create statusim/nimbus-eth1:${TAG} \
+            --amend statusim/nimbus-eth1:linux-amd64-${TAG} \
+            --amend statusim/nimbus-eth1:linux-arm64-${TAG}
+          docker manifest push statusim/nimbus-eth1:${TAG}
```
tersec: Interesting, in `nimbus-eth2` a couple of days ago I had to either disable `--provenance`  which had been newly added to some GitHub Actions build pipeline setting in the last couple of weeks or use `docker buildx imagetools create`; see
- https://github.com/docker/build-push-action/issues/1459
- https://github.com/docker/buildx/issues/1509

If this works, this is fine, there's nothing intrinsically better about the `buildx imagetools create` approach, but I wonder if this will also stop working soon.

tersec: https://github.com/status-im/nimbus-eth2/commit/fa7a87e83baa2a9a0f9ff52ea8102230c1c0b1ec has the commit itself.

advaita-saha: Wasn't aware of this, thanks for pointing it out
Will cross-check with demo pushes in my repo setup to make sure things are working

advaita-saha: Both works but better to migrate to image tools which is the standard now

tersec: https://github.com/status-im/nimbus-eth1/pull/3988/commits/f3cd040984052e9b31998e501e0559f29fb9e30a

PR #3990, opened 2026-02-17
Parallel state root computation

bhartnett:
The approach taken here to parallelize the state root computation, works by creating a separate task for each subtrie of the root branch node and each of these tasks can run in a separate thread. The main thread then calls sync on each future to collect each of the hash key values for each task before finishing the computation of the root node in the main thread. 

In theory the best case speed up is 16x (if there are at least 16 cores available) as we create up to 16 tasks but in practice it looks like a 2-4x speed up depending on the workload due to the overhead of locking and passing data between threads. If the root node happens to not be a branch node (or extension branch) then only single threaded computation is used. For now, the parallel computation is only used when computing the state root and not when computing storage roots or other hash keys because the overhead of the tasks and multi-threading may not be worth it in those cases.

The locking is only done inside the aristo_compute module so the synchronization only applies to the compute key code and doesn't effect other parts of the code base which are still single threaded. Parallel state root computation is disabled by default for now. It can be enabled later after more testing. The approach here is to only write data to the database from the main thread and the thread pool runs the tasks which compute the hash keys and then copy them into a queue (single producer single consumer thread safe circular buffer, 2 per sub task) which the main thread then reads from and writes the keys back to either the database or an in memory layer. 

A read-write lock is added to each txFrame and the locking is done at this level to reduce contention, as reads can run at the same time as writes to a different txFrame/layer. When passing the data between threads we only need to copy each item once to push the data into the queue, when popping the data we move it to avoid the copy. When writing a hashkey to an in memory layer we only need to copy the hashkey between threads but when writing to the database we serialize the vertex into a vertex buf and copy to the main thread, then write to the database using a new `putVtxBlobFn` which just writes the vertex buf bytes directly to the database. 

I implemented a quick proof of concept thread safe lru cache which added a read write lock to the existing minilru type but the performance was not great and I found that bypassing the lrucache when hiting the rocksdb backend provided much better performance, since rocksdb is already optimised to support concurrent reads. For this reason the `GetVtxFlag.NoPutCache` flag was added to the getKeyBe call to avoid writing to the caches from the child threads when executing in parallel.


execution_chain/db/aristo/aristo_compute.nim
```diff
@@ -76,17 +86,19 @@ proc putKeyAtLevel(
 
   if level >= db.db.baseTxFrame().level:
     let txRef = db.deltaAtLevel(level)
-    txRef.layersPutKey(rvid, vtx, key)
+    withLock(db.lock):
+      txRef.layersPutKey(rvid, vtx, key)
```
arnetheduck: this approach unfortunately won't work: layersPutKey (and similar functions below for loading) might allocate memory from the thread-local pool that later would be referenced by another thread - this will eventually lead to crashes. In general, `ref`, `seq` etc cannot travel across thread boundaries, meaning that all memory must either be pre-allocated, manually managed (ie via createShared) or released upon completion of the task running in a thread.

bhartnett: Yes, I've run into the crashes already. Perhaps the tasks should only write the updates into the database and any other keys that need to be computed with a level higher than base need to be done in the main thread (not in a task).

arnetheduck: It's a tricky problem in the current architecture - to compute the hash, you need to load adjacent nodes - that means hitting the shared caches which themselves allocate memory, which means that addressing this in aristo_compute alone won't fly: what this PR is really parallelising is not so much the computation but rather the loading from disk.

Writing is in some ways the easy part because then you already know everything, it's reading that messes things up. 

In essence, we'd need to implement our own memory manager in the lowest cache layer and work with raw pointers instead of using `ref` - this is a worthy rewrite, but a bit more involved, as it involves creating a copy-on-write page manager which allows multiple versions of the in-memory to coexist in memory and synchronizes access to those versions. This is not too far off, ie a lot of the infrastructure is already there - I started laying the groundwork in a branch: https://github.com/status-im/nimbus-eth1/compare/vertex-ref-object - that branch gets rid of `ref` in VertexRef/BranchRef etc which is the first step, though it was premature at the time - the codebase should be in a better shape now to continue that work.

bhartnett: Ok good to know. Yeah, I noticed that the shared caches are not thread safe but the part I missed is that reading will allocate into the shared caches in the thread local heap. I thought that when passing in a pointer to a ref object then the allocations would be in the main thread heap but I guess not.

bhartnett:
Note that the reason we can't achieve a 16x speedup in practice here is because the problem is not perfectly parallel. When we compute each hashkey we read and then write back into each txFrame. We use a read-write lock on each txFrame so reads can happen in parallel but writes will block reads and each task which runs in a separate thread will very likely be writing to the same txFrame as other threads so we get threads blocking each other.

PR #3999, opened 2026-02-20
Feat:clean up eth_estimategas

### LINKED ISSUE #2773: the rpc method implementation is ok. but the actual implementation of rpcEstimateGas is incorrect
the rpc method implementation is ok. but the actual implementation of rpcEstimateGas itself probably need more test. will merge for now

_Originally posted by @jangko in https://github.com/status-im/nimbus-eth1/issues/2701#issuecomment-2395267317_

### LINKED ISSUE #3946: Improve eth_estimateGas binary search by doing first some short circuits
Before binary search add:

- plain value transfer check
- test execute with highest gas limit (as currently even on contract call failures, we will still iterate over the whole bin search to just fail)
- Next, start binary search with first usedGas + gasRefund

RazorClient:
fixes  #2773

- [x] fixes to pass the new [test](https://github.com/status-im/nimbus-eth1/issues/2773#issuecomment-3816992692)
- [x] change normalisation to either use maxPriorityFeePerGasNorm and maxFeePerGasNorm or eip1559TxNormalization
- [x] add 7702 eoa gas accounting into eth_estimategas

and apply the techniques stated in https://github.com/status-im/nimbus-eth1/issues/3946#issue-3875856632, thus fixes #3946

### DIFF
--- a/nimbus_verified_proxy/engine/accounts.nim
+++ b/nimbus_verified_proxy/engine/accounts.nim
@@ -48,7 +48,9 @@ proc getAccountFromProof*(
   of ValidProof:
     return ok(acc)
   of InvalidProof:
-    return err((VerificationError, proofResult.errorMsg))
+    # we leave the error untagged so the backend responsible for
+    # the error can be tagged. See tagBackend impl in types.nim
+    return err((VerificationError, proofResult.errorMsg, UNTAGGED))
 
 proc getStorageFromProof(
     account: Account, storageProof: StorageProof
@@ -66,7 +68,9 @@ proc getStorageFromProof(
   of ValidProof:
     return ok(storageProof.value)
   of InvalidProof:
-    return err((VerificationError, proofResult.errorMsg))
+    # we leave the error untagged so the backend responsible for
+    # the error can be tagged. See tagBackend impl in types.nim
+    return err((VerificationError, proofResult.errorMsg, UNTAGGED))
 
 proc getStorageFromProof*(
     stateRoot: Hash32,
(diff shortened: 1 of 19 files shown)


tests/test_rpc.nim
```diff
@@ -509,6 +509,34 @@ proc rpcMain*() =
...
+        `from`: Opt.some(signer),
+        to: Opt.some(extraAddress),
+        value: Opt.some(spendValue),
+        maxFeePerGas: Opt.some(w3Qty(hugeMaxFee.uint64)),
+        maxPriorityFeePerGas: Opt.some(w3Qty(priorityFee.uint64)),
+      )
+
+      let res = await client.eth_estimateGas(eip1559Args)
+      check res == w3Qty(gasBudget)
+    # Todo: Add https://github.com/ethereum/execution-apis/blob/main/tests/eth_estimateGas/estimate-with-eip7702.io and https://github.com/ethereum/execution-apis/blob/main/tests/eth_estimateGas/estimate-with-eip4844.io
```
kdeme: No need to add these here necessarily. These tests are used in Hive tests and thus nimbus already gets tested on this via hive.

You can run hive with this branch to verify it works. I can link you the commands for this if you haven't run hive locally yet.

RazorClient: Hey 
I have been running hive locally however for the estimate gas test cases it only checks the schema response and not the actual value returned.
Checking for the value returned makes nimbus fail all of the estimate_gas cases expect for 1-2.
Will share the details

RazorClient: however not adding any specific testcases would be a wiser move nonetheless, as estimate_gas does differ client to client

nimbus_verified_proxy/tests/test_state.nim
```diff
@@ -88,7 +88,7 @@ suite "test state verification":
       verifiedCall.get() ==
         "000000000000000000000000000000000000000000000000288a82d13c3d1600".hexToSeqByte()
       verifiedAccessList.get() == accessList
-      verifiedEstimate.get() == Quantity(22080)
+      verifiedEstimate.get() <= Quantity(30000)
```
advaita-saha, line 91: what is the basis of the test vector change ?
cc: @chirag-parmar

RazorClient: estimate gas changes are generally not static, so it would have made sense to just cap it and have a lesser than check
Reiterating from our Discord conversation

chirag-parmar: > estimate gas changes are generally not static, so it would have made sense to just cap it and have a lesser than check

+1

PR #4001, opened 2026-02-20
proxy: update docs

chirag-parmar:


nimbus_verified_proxy/README.md
```diff
@@ -73,18 +75,19 @@ A block root may be obtained from a trusted beacon node or from a trusted provid
 
 * Trusted provider, e.g. Beaconcha.in:
 
-    On the [beaconcha.in](https://beaconcha.in) website ([Goerli](https://prater.beaconcha.in)), navigate to the `Epochs` section and select a recent `Finalized` epoch. Then, scroll down to the bottom of the page. If the bottom-most slot has a `Proposed` status, copy its `Root Hash`. Otherwise, for example if the bottom-most slot was `Missed`, go back and pick a different epoch.
+    On the [beaconcha.in](https://beaconcha.in) website, navigate to the `Epochs` section and select a recent `Finalized` epoch. Then, scroll down to the bottom of the page. If the bottom-most slot has a `Proposed` status, copy its `Root Hash`. Otherwise, for example if the bottom-most slot was `Missed`, go back and pick a different epoch.
```
tersec: It's URL hyperlink, probably don't have to specify that it's a website.

nimbus_verified_proxy/docs/metamask_configuration.md
```diff
@@ -5,21 +5,22 @@ to make use of the proxy.
...
 ```bash
 # From the nimbus-eth1 repository
 TRUSTED_BLOCK_ROOT=0x1234567890123456789012345678901234567890123456789012345678901234 # Replace this
 API_KEY=abcd # Replace this
 ./build/nimbus_verified_proxy \
-    --network=goerli \
+    --network=mainnet \
     --trusted-block-root=${TRUSTED_BLOCK_ROOT} \
-    --web3-url="wss://eth-goerli.g.alchemy.com/v2/${API_KEY}"
+    --backend-url="wss://eth-mainnet.g.alchemy.com/v2/${API_KEY}" \
```
kdeme: Good for updating to the current cli options and thus my following comment is not really relevant to this PR.

But I think the current naming of the cli options is not great in terms UX.

`backend-url`: This name doesn't make it clear at all which backend. I understand it is there to differentiate clearly between backend and frontent URL but I'm not sure it achieves that goal.

`external-beacon-api-url`: I know this comes from the consensus light client (as suggested in other PR). But the external is rather implicit here because it is an cli parameter in the first place.


Helios seems to use `--execution-rpc` and `--consensus-rpc` which makes that pretty clear. We could opt for `--execution-api-url` and `--beacon-api-url` considering execution-apis and `beacon-apis are the specs repo names.

`--frontend-url` could then use a different name too if above would be changed.

chirag-parmar: makes sense

chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4003

nimbus_verified_proxy/README.md
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4001/commits/9b155e2601be292605571284331f17f391117091

PR #4009, opened 2026-02-24
ci: rebuild hive everytime

siddarthkay:
we frequently need to update hive

### DIFF
--- a/ci/hive.groovy
+++ b/ci/hive.groovy
@@ -34,7 +34,7 @@ pipeline {
     )
     string(
       name: 'TIMEOUT_MINUTES',
-      defaultValue: '20',
+      defaultValue: '40',
       description: 'Timeout for each stage in minutes'
     )
     booleanParam(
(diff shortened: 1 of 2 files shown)


markoburcul (approved):

ci/hive.groovy
```diff
@@ -17,6 +17,7 @@ pipeline {
     dockerfile {
       label 'linuxcontainer'
       filename 'docker/Dockerfile.hive'
+      additionalBuildArgs "--build-arg HIVE_CACHE_BUST=${new Date().getTime()}"
```
jakubgs: Wouldn't it be simpler to just make building hive a build stage rather than add weird hacks?

siddarthkay: I thought this way was ~~clever~~ cute, but I don't mind moving to build stage in Jenkins.

jakubgs: Much better and more clear on what is happening.

PR #4014, opened 2026-02-24
remove `blk` from BlockRef

### LINKED ISSUE #3778: Remove `blk` field from BlockRef
https://github.com/status-im/nimbus-eth1/blob/fe162f0052e4ea0fa61a4c69ad9c7ebd300f5a70/execution_chain/core/chain/forked_chain/chain_branch.nim#L20 stores a full copy of each block in memory - this block is used very rarely, ie only during replays. 

A second copy of the same data is stored in `txFrame` as the block is serialized and persisted there after verification.

When there is a longer period of non-finality, this leads to high memory consumption. Instead:

* remove `blk` from BlockRef and load it on demand from `txFrame` - this halves memory usage
* eventually, flatten `txFrame` and keep it up to date using the `heads` structure, akin to how `FcState` does it - this way, almost _no_ block payload data will be kept in memory.

https://github.com/status-im/nimbus-eth1/pull/3628 implements this idea the other way around by keeping `blk` and removing it from `txFrame` - that approach however has almost no advantages and cannot be flattened.

advaita-saha:
Fixes #3778

### DIFF
--- a/execution_chain/rpc/params.nim
+++ b/execution_chain/rpc/params.nim
@@ -1,5 +1,5 @@
 # Nimbus
-# Copyright (c) 2024 Status Research & Development GmbH
+# Copyright (c) 2026 Status Research & Development GmbH
 # Licensed under either of
 #  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
 #  * MIT license ([LICENSE-MIT](LICENSE-MIT))
@@ -13,6 +13,7 @@ import
   chronicles,
   eth/common/[addresses, headers],
   web3/eth_api_types,
+  ../transaction,
   ../transaction/call_common,
   ../evm/types,
   ../evm/evm_errors,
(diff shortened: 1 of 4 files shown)


execution_chain/core/chain/forked_chain.nim
```diff
@@ -198,17 +196,22 @@ func calculateNewBase(
...
+  for txHash in toRemove:
+    c.txRecords.del(txHash)
+
   b.txFrame.dispose()
 
   # Mark it as removed, don't remove it twice
   b.txFrame = nil
   # Clear parent and let GC claim the memory earlier
   b.parent = nil
+  b.header = default(Header)   # frees extraData seq immediately
```
arnetheduck:
```suggestion
  b.header.reset()   # frees extraData seq immediately
```

execution_chain/sync/beacon/worker/blocks.nim
```diff
@@ -105,7 +105,7 @@ template blocksCollect*(
           discard ctx.blocksUnprocFetch(ctx.subState.topNum-bottom).expect("iv")
 
         # Fetch blocks and verify result
-        let blocks = buddy.blocksFetch(nFetchBodiesRequest, info).valueOr:
+        var blocks = buddy.blocksFetch(nFetchBodiesRequest, info).valueOr:
```
tersec, line 108: `let` to `var` with no other changes? seems unlikely to be required.

advaita-saha: There is reset of `blocks[n]`, if import succeeds

execution_chain/core/chain/forked_chain/chain_serialize.nim
```diff
@@ -252,7 +259,9 @@ proc deserialize*(fc: ForkedChainRef): Result[void, string] =
     return err("Cannot find previous FC state in database")
 
   let prevBase = fc.base
-  var blocks = newSeq[BlockRef](state.numBlocks)
+  var
+    blocks = newSeq[BlockRef](state.numBlocks)
+    bodies = initTable[Hash32, Block]()
```
tersec: https://nim-lang.org/docs/tables.html#initTable
> Starting from Nim v0.20, tables are initialized by default and it is not necessary to call this function explicitly.

i.e. the Nim manual suggests instead
```nim
    bodies: Table[Hash32, Block]
```

advaita-saha: https://github.com/status-im/nimbus-eth1/pull/4014/commits/d649d41d5749e73bec1ab9b957e23aac197eacc4

PR #4016, opened 2026-02-25
Try updating to orc

bhartnett:
So far the following works with orc on linux:
- Building the standalone execution client: `make nimbus_execution_client`
- Apart from test_txpool test case, all tests pass when running: `make test`
- The era file block import appears to work fine.
- All eest blockchain tests pass when running: `make eest_blockchain_test `
- All eest engine tests pass when running: `make eest_engine_test `

### DIFF
--- a/tests/all_tests.nim
+++ b/tests/all_tests.nim
@@ -38,7 +38,7 @@ import
     test_rpc,
     test_snap,
     test_transaction_json,
-    # test_txpool,
+    test_txpool,
     test_networking,
     test_pooled_tx,
     test_stateless_witness_types,
--- a/tests/all_tests.nim
+++ b/tests/all_tests.nim
@@ -38,7 +38,7 @@ import
     test_rpc,
     test_snap,
     test_transaction_json,
-    test_txpool,
+    # test_txpool,
     test_networking,
     test_pooled_tx,
     test_stateless_witness_types,
--- a/tests/config.nims
+++ b/tests/config.nims
@@ -13,8 +13,8 @@ if not defined(windows):
   # available on some GCC versions but not all - run with `-d:limitStackUsage`
   # and look for .su files in "./build/", "./nimcache/" or $TMPDIR that list the
   # stack size of each function.
-  switch("passC", "-fstack-usage -Werror=stack-usage=2048576")
-  switch("passL", "-fstack-usage -Werror=stack-usage=2048576")
+  switch("passC", "-fstack-usage -Werror=stack-usage=1048576")
+  switch("passL", "-fstack-usage -Werror=stack-usage=1048576")
 
 switch("define", "chronicles_line_numbers")
 switch("define", "chronicles_sinks=textlines")
--- a/tests/config.nims
+++ b/tests/config.nims
@@ -13,8 +13,8 @@ if not defined(windows):
   # available on some GCC versions but not all - run with `-d:limitStackUsage`
   # and look for .su files in "./build/", "./nimcache/" or $TMPDIR that list the
   # stack size of each function.
-  switch("passC", "-fstack-usage -Werror=stack-usage=1048576")
-  switch("passL", "-fstack-usage -Werror=stack-usage=1048576")
+  switch("passC", "-fstack-usage -Werror=stack-usage=2048576")
+  switch("passL", "-fstack-usage -Werror=stack-usage=2048576")
 
 switch("define", "chronicles_line_numbers")
 switch("define", "chronicles_sinks=textlines")
--- a/config.nims
+++ b/config.nims
@@ -118,7 +118,7 @@ switch("passL", "-fno-omit-frame-pointer")
 
 --threads:on
 --opt:speed
---mm:refc
+--mm:orc
 --excessiveStackTrace:on
 # enable metric collection
 --define:metrics
--- a/config.nims
+++ b/config.nims
@@ -127,10 +127,6 @@ switch("passL", "-fno-omit-frame-pointer")
 --styleCheck:usages
 --styleCheck:error
 
-# Workaround for v2.2.8 regression; remove with v2.2.10
-# https://github.com/nim-lang/Nim/pull/25343
---mangle:cpp
-
 switch("define", "nim_compiler_path=" & currentDir & "env.sh nim")
 switch("define", "withoutPCRE")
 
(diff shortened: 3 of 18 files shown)


tests/all_tests.nim
```diff
@@ -38,7 +38,7 @@ import
     test_rpc,
     test_snap,
     test_transaction_json,
-    test_txpool,
```
bhartnett, line 41: This test is crashing with a seg fault on linux.

tests/config.nims
```diff
@@ -13,8 +13,8 @@ if not defined(windows):
   # available on some GCC versions but not all - run with `-d:limitStackUsage`
   # and look for .su files in "./build/", "./nimcache/" or $TMPDIR that list the
   # stack size of each function.
-  switch("passC", "-fstack-usage -Werror=stack-usage=1048576")
-  switch("passL", "-fstack-usage -Werror=stack-usage=1048576")
```
bhartnett, line 17: It appears that orc increases stack usage. 

```
/home/user/development/status-im/nimbus-eth1/vendor/nimbus-build-system/vendor/Nim/lib/system.nim: In function ‘_ZN7eip759434validateBlobTransactionWrapper7594EN10pooled_txs17PooledTransactionE’:
/home/user/development/status-im/nimbus-eth1/execution_chain/core/eip7594.nim:19:15: error: stack usage might be 1205200 bytes [-Werror=stack-usage=]
   19 | proc validateBlobTransactionWrapper7594*(tx: PooledTransaction):
      |               ^
lto1: some warnings being treated as errors
make[1]: *** [/tmp/ccZV498U.mk:227: /tmp/ccLQlohJ.ltrans75.ltrans.o] Error 1
make[1]: *** Waiting for unfinished jobs....
lto-wrapper: fatal error: make returned 2 exit status
compilation terminated.
/usr/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
Error: execution of an external program failed: 'g++  @all_tests_linkerArgs.txt'
```

tests/all_tests.nim
tersec: It's a stack-based segfault due to `ulimit -s` running as part of `make test`

tersec: notably, running `./env.sh nim c -r tests/all_tests` without `ulimit` does not reproduce it.

tests/config.nims
tersec: Yes, this seems to be behind the segfault

bhartnett:
> * [Nim v2.2.x ORC adds extra `nimZeroMem`/`nimCopyMem`/stack usage over refc nim-lang/Nim#25552](https://github.com/nim-lang/Nim/issues/25552)

Is this a blocker for moving to orc or do you think we could work around by just increasing stack size?

bhartnett:
@tersec So it turns out that moving to orc will be problematic for the multithreaded use case because it doesn't support atomic reference counting. Any ref types used in multiple threads concurrently will cause crashes due to ref counts being corrupted. The current parallel stateroot computation doesn't work in orc for this reason because it reads from the database from multiple threads in parallel in order to read the hashes and vertexes. The database, txFrame, and vertex types all through the code are ref types. I've found that it does work when using --mm:atomicArc so that confirms the issue is related to the reference counts.

Getting this to work in orc is possible, I could pass around ptr to object types but that would require updating much of the codebase and it leads to messy unmaintainable code in my opinion. The reason for passing these ref types into each thread is because I'm going for the shared memory model when multiple threads read and write to shared state which is generally faster than copying data between threads. When we implement full parallel execution and batch IO we need to be able to read state from the in memory layers and then the database in parallel. In order to do this, each thread needs to access the shared database and txFrame ref types.

tersec:
Is it harder than refc or just neutral? https://nim-lang.org/docs/mm.html#other-mm-modes doesn't list either `refc` or ORC as atomic.

There is `atomicArc`, but I've never tried it. ARC in general isn't designed to collect cycles, which might be too big a constraint for Nimbus.

bhartnett:
> Is it harder than refc or just neutral? https://nim-lang.org/docs/mm.html#other-mm-modes doesn't list either `refc` or ORC as atomic.
> 
> There is `atomicArc`, but I've never tried it. ARC in general isn't designed to collect cycles, which might be too big a constraint for Nimbus.

refc is better than orc for my use case because I can use/share ref types between threads. I just need to make sure the ref type in the main thread outlives the tasks where it is used by the worker threads. 

When I compile with arc or atomicArc there are some warnings about cycles so I guess we can't use arc based memory management. 

That link says 'The reference counting operations (= "RC ops") do not use atomic instructions' under the arc/orc section. The fact that atomicArc exists suggests that arc is not atomic and this matches my conclusions based on my testing where I'm seeing crashes when using orc/arc.

Actually I'm not 100% sure if refc does in fact use atomic ref counts, it might actually be working because of the thread local heaps where the ref counts are stored separately on each heap and therefore the worker threads are unable to touch the ref counts of any other threads. Either way, refc works for me as does atomicArc (but atomicArc will likely be leaking memory).

arnetheduck:
>      refc is better than orc for my use case because I can use/share ref types between threads. 

this sounds off - we can share (unmanaged) pointers to memory managed by refc but we must always ensure no refcounting takes place across threads regardless of `mm` - ie both gc and orc perform non-atomic reference counting, they differ mainly in how they deal with cycles / root tracking mechanism.

In both cases, we must manually make sure that the "owner" thread keeps the referenced memory alive until the thread is done with it - there's no _excellent_ way to pass ownership between threads outside of the `isolate` infrastructure.

bhartnett:
> this sounds off - we can share (unmanaged) pointers to memory managed by refc but we must always ensure no refcounting takes place across threads regardless of `mm` - ie both gc and orc perform non-atomic reference counting, they differ mainly in how they deal with cycles / root tracking mechanism.

Just curious, where are the ref counts stored for each ref instance? Are they stored in each thread local heap? I've noticed that with orc ref counts get corrupted and cause crashes when accessing a ref type from another thread but with refc this appears to work fine.

> In both cases, we must manually make sure that the "owner" thread keeps the referenced memory alive until the thread is done with it - there's no _excellent_ way to pass ownership between threads outside of the `isolate` infrastructure.

In my case the main thread is the originator of the ref types which are then passed via pointer to the tasks and always outlive the tasks itself so that part is fine.

PR #4018, opened 2026-02-26
ci: hive test for commit and nuke erigon

siddarthkay:
## Summary

- remove `erigon` sync tests as per Advaita's request.
- tag the built docker image in CI with `statusim/nimbus-eth1:master` 
- remove `--docker.pull` flag from hive calls so that hive relies on locally built image rather than pulling from docker hub. This change is required so that `nimbus-eth1` changes from the PR are built and used rather than `master`.

note: This is kind of misleading because we tag with `master`, and `hive` config references `master` for nimbus.


ci/hive.groovy
```diff
@@ -114,33 +113,6 @@ pipeline {
                     --sim.parallelism=${params.PARALLELISM} \
                     --sim.loglevel 4 \
                     --docker.nocache hive/clients/nimbus-el \
-                    --docker.pull true \
```
jakubgs, line 117: Why are we removing this? Neither title not descriptions explains why.

siddarthkay: my bad, will update, this flag needs to be removed to prfioritize locally built image rather than pull from docker hub

jakubgs (approved):
Much better. Thanks.

PR #4022, opened 2026-02-26
Add delete and delete range functions to kvt utils

bhartnett:
This PR adds a delete range function to the kvt backend to support bulk deletes. 

The idea here is we bypass the kvt txFrame layers and go directly to the backend since the purpose of the method is really to delete historical data on disk and a bulk delete going through each of the in memory txframe layers would be slow.

The history expiry pruner can use this by importing the kvt_utils module and calling `delRangeBe`. 

The `compactRange` parameter doesn't compact the key range immediately but rather suggests to the RocksDb background job that the key ranges should be compacted.

cc @advaita-saha

### DIFF
--- a/execution_chain/db/kvt/kvt_desc.nim
+++ b/execution_chain/db/kvt/kvt_desc.nim
@@ -48,6 +48,10 @@ type
 
   # -------------
 
+  DelKvpFn* =
+    proc(key: openArray[byte]): Result[void, KvtError] {.gcsafe, raises: [].}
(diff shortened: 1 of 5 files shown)


advaita-saha:
Ahh that's great, but if deleteRange is available bypassing the txFrame
Can you also add a delete endpoint similar to this ( not range, but single )
Or should I use the range one only, by passing only 1 ? ( not sure which one will be better )

Issue #3942, opened 2026-01-30
Execution-api JSON-RPC master tracker issue


kdeme:
Tracker issue for all [execution-api](https://github.com/ethereum/execution-apis) JSON RPC issues.

Most of these are visible / discovered via hive tests, but some are additional ones discovered (and more are likely to be discovered as these get worked on).

Some are also just spec / test vector / hive PITAs

For latest hive results on `rpc-compat`: https://hive.ethpandaops.io/#/test/generic/1769753159-b9b9d40703a67f0737e0eb880873ff13

### block tag not parsing hash

Some calls are named "...ByNumber" or "...ByHash"

These will parse either:
- latest/safe/finalized and number
- block hash

Other calls without this specific naming however are often supposed parse latest/safe/finalized + number + hash.
Several failures are due to nimbus not implementing the hash part.

#### `null` vs omitting the fields:
Our current JSON flavor is set to NOT omit null fields (= `Opt.none(T)`). This is because this behavior is actually expected/specified in `engine_api` (see https://github.com/status-im/nim-json-rpc/pull/238).
I think ideally Hive would just not see this as an error, but just as an alternative, correct response.
But we could also adjust our JSON-RPC flavor for `eth_` and `debug_` method namespace. For this we need https://github.com/status-im/nimbus-eth1/pull/3914 so that we can define different flavors per api.

#### `null` vs error:

We often return an error when a block (header, etc) is not found. The tests (and spec) however expect a `null` for not found and error for actual errors.

#### `null` vs empty array

In some cases there is also an empty array returned instead of null and vice versa.

### debug_getRawBlock
- [ ] wrong error code invalid param:     "code": -- -32000 ++ -32602

### debug_getRawHeader
- [ ] wrong error code invalid param:     "code": -- -32000 ++ -32602
### debug_getRawReceipts
- [ ] wrong error code invalid param:     "code": -- -32000 ++ -32602

### debug_getRawReceipts

- [ ] returns empty receipts on test case

### debug_getRawTransaction

- [ ] wrong error code invalid param:     "code": -- -32000 ++ -32602,

### eth_blobBaseFee

- [ ] wrong blobBaseFee calculation: response differs from expected (-- client, ++ test):

### eth_call

- [ ] invalid result on test case
- [ ] reverts not happening on abi error and panic test cases

### eth_createAccessList

- [ ] all 3 hive errors are errors because of `null` fields -> all 3 fixed with null omit

[...]

PR #4069, opened 2026-03-25
Feat:Fix Debug_* RPC endpoints

RazorClient:
This pr addresses part of https://github.com/status-im/nimbus-eth1/issues/3942

This pr proposes to fix the following testcases that are currently failing in Hive

- [X] [debug_getRawBlock/get-invalid-number](https://hive.ethpandaops.io/#/logs/generic/1774247262-8eaedf52dec0767f8097d4c8df931274/details%2F1774247259-ee87d092653a6654fcf9e1bec7b433b4c73f4248b2ad69534c00c9c6f7f65f49-0.log?begin=4176&end=4611)
- [X] [debug_getRawHeader/get-invalid-number](https://hive.ethpandaops.io/#/logs/generic/1774247262-8eaedf52dec0767f8097d4c8df931274/details%2F1774247259-ee87d092653a6654fcf9e1bec7b433b4c73f4248b2ad69534c00c9c6f7f65f49-0.log?begin=7060&end=7497)
- [x] [debug_getRawReceipts/get-block-n ](https://hive.ethpandaops.io/#/logs/generic/1774247262-8eaedf52dec0767f8097d4c8df931274/details%2F1774247259-ee87d092653a6654fcf9e1bec7b433b4c73f4248b2ad69534c00c9c6f7f65f49-0.log?begin=7555&end=11422)
- [X] [debug_getRawReceipts/get-invalid-number](https://hive.ethpandaops.io/#/logs/generic/1774247262-8eaedf52dec0767f8097d4c8df931274/details%2F1774247259-ee87d092653a6654fcf9e1bec7b433b4c73f4248b2ad69534c00c9c6f7f65f49-0.log?begin=11664&end=12105)
- [X] [debug_getRawTransaction/get-invalid-hash](https://hive.ethpandaops.io/#/logs/generic/1774247262-8eaedf52dec0767f8097d4c8df931274/details%2F1774247259-ee87d092653a6654fcf9e1bec7b433b4c73f4248b2ad69534c00c9c6f7f65f49-0.log?begin=12171&end=12677)


kdeme (commented):
I think this can be better achieved by raising the right application error from the `blockFromTag` call.

But that kind of collides with the fact that it uses Result at the moment.

So I would change that or pass the numbers along in the Result error and then raise it on failure.

kdeme:
> I think this can be better achieved by raising the right application error from the `blockFromTag` call.
> 
> But that kind of collides with the fact that it uses Result at the moment.
> 
> So I would change that or pass the numbers along in the Result error and then raise it on failure.

To further clarify what I mean with this: Simply raise in the application the `ApplicationError` with the error code defined in execution api specs. And provide a decent user focused error message (afaik hive doesn't fail on those).

The use of the `32602` error code by the execution API specs is rather a misuse of it, but we should still be following those specs and thus implement it. However, we should not be using `ApplicationError` in the core "system" logic of nim-json-rpc to fix this. That's a hack and setting the wrong example for that library usage. The "hack" should live on nimbus-eth1 side.

We already do this in several locations:
- https://github.com/status-im/nimbus-eth1/blob/master/execution_chain/beacon/api_handler/api_utils.nim#L141-L145
- https://github.com/status-im/nimbus-eth1/blob/master/execution_chain/rpc/server_api.nim#L369
- https://github.com/status-im/nimbus-eth1/blob/master/portal/rpc/rpc_portal_history_api.nim#L148

kdeme:
> Simply raise in the application the `ApplicationError` with the error code defined in execution api specs. And provide a decent user focused error message (afaik hive doesn't fail on those).

So in practice, I think that here this mostly means adapting the error result from `headerFromTag` or `blockFromTag` to the correction raising of ApplicationError with right error code.
However be careful as "block not found" type of errors should still return `null` afaik and thus not raise.

So perhaps those calls need to change function signature.
e.g. I think this can be achieved with an `Opt` inside the `Result`, or perhaps just let those `blockFromTag` type of calls raise immediately. In general we prefer usage of `Result`, but if the end of the API is supposed to use the raising of exceptions, then I think it is warranted here to use exceptions also in those calls if that makes it cleaner.

RazorClient:
All debug_* hive tests are now Fixxed!!!!!

```
Apr 16 23:34:49.699 INF API: test started suite=0 test=2 name="debug_getRawBlock/get-block-n (nimbus-el)"
Apr 16 23:34:49.702 INF API: test ended suite=0 test=2 pass=true
Apr 16 23:34:49.704 INF API: test started suite=0 test=3 name="debug_getRawBlock/get-genesis (nimbus-el)"
Apr 16 23:34:49.710 INF API: test ended suite=0 test=3 pass=true
Apr 16 23:34:49.711 INF API: test started suite=0 test=4 name="debug_getRawBlock/get-invalid-number (nimbus-el)"
Apr 16 23:34:49.715 INF API: test ended suite=0 test=4 pass=true
Apr 16 23:34:49.717 INF API: test started suite=0 test=5 name="debug_getRawHeader/get-block-n (nimbus-el)"
Apr 16 23:34:49.720 INF API: test ended suite=0 test=5 pass=true
Apr 16 23:34:49.722 INF API: test started suite=0 test=6 name="debug_getRawHeader/get-genesis (nimbus-el)"
Apr 16 23:34:49.723 INF API: test ended suite=0 test=6 pass=true
Apr 16 23:34:49.725 INF API: test started suite=0 test=7 name="debug_getRawHeader/get-invalid-number (nimbus-el)"
Apr 16 23:34:49.726 INF API: test ended suite=0 test=7 pass=true
Apr 16 23:34:49.727 INF API: test started suite=0 test=8 name="debug_getRawReceipts/get-block-n (nimbus-el)"
Apr 16 23:34:49.729 INF API: test ended suite=0 test=8 pass=true
Apr 16 23:34:49.730 INF API: test started suite=0 test=9 name="debug_getRawReceipts/get-genesis (nimbus-el)"
Apr 16 23:34:49.732 INF API: test ended suite=0 test=9 pass=true
Apr 16 23:34:49.733 INF API: test started suite=0 test=10 name="debug_getRawReceipts/get-invalid-number (nimbus-el)"
Apr 16 23:34:49.735 INF API: test ended suite=0 test=10 pass=true
Apr 16 23:34:49.736 INF API: test started suite=0 test=11 name="debug_getRawTransaction/get-invalid-hash (nimbus-el)"
Apr 16 23:34:49.738 INF API: test ended suite=0 test=11 pass=true
Apr 16 23:34:49.739 INF API: test started suite=0 test=12 name="debug_getRawTransaction/get-tx (nimbus-el)"
Apr 16 23:34:49.741 INF API: test ended suite=0 test=12 pass=true
```

The only problem was with debug_getRawTransaction/get-invalid-hash as it needs some changes to the json conversion/error handling layer to convert the decode error raised with https://github.com/status-im/nim-web3/pull/243 to have the right idiomatic fix. However, fixing the RPC first and then fixing the other problems down the line might be better.

execution_chain/rpc/debug.nim
```diff
@@ -191,31 +191,45 @@ proc setupDebugRpc*(com: CommonRef, txPool: TxPoolRef, server: RpcServer) =
...
-    for receipt in chain.baseTxFrame.getReceipts(header.receiptsRoot):
-      res.add rlp.encode(receipt)
+    for receipt in txFrame.getReceipts(header.receiptsRoot):
+      res.add rlp.encode(receipt.to(Receipt))
 
     res
 
   # https://ethereum.github.io/execution-apis/api/methods/debug_getRawTransaction
-  server.rpc("debug_getRawTransaction") do(txHash: Hash32) -> seq[byte]:
+  # we take a string input instead of a hex as in that manner we can preverse the raw json input
```
kdeme, line 224: I think it is better to just fix it in web3 immediately. You can do it in separate PR or here as you prefer.

But this is nothing urgent of a fix and probably doesn't block anybody and is just an issue for Hive.

When fixing in web3 for `Hash32` type I hope there is no other call that relies on not having the `0x`. But I don't think so. I checked some `eth_` api calls that use it but I think they all assume `0x` prefixed.

execution_chain/rpc/debug.nim
```diff
@@ -191,31 +191,45 @@ proc setupDebugRpc*(com: CommonRef, txPool: TxPoolRef, server: RpcServer) =
   server.rpc("debug_getRawBlock") do(blockTag: BlockTag) -> seq[byte]:
     ## Returns an RLP-encoded block.
     let blockFromTag = chain.blockFromTag(blockTag).valueOr:
-      raise newException(ValueError, error)
+      raise invalidParams(error)
```
kdeme, line 200: This fixes Hive errors, but behavior is opposite of the non debug calls.

As I've mentioned in the past. I think we still need to clean this up by having `null` consistently for "block not found" and errors for the actual errors.

Now we run `null` on the `eth_` api calls for everything and here error for everything. Which works for hive as hive tests just happen to be testing just one or the other.

execution_chain/rpc/debug.nim
RazorClient: the problem becomes that 
even if i make the hash-decoding for nim-web3 stricker using https://github.com/status-im/nim-web3/pull/243 but then we will still fail in the nim-json-rpc param-unpack layer.
We would only change the  decode failure message.
the only complete solution to this would be to change the  RequestDecodeError in nim-json-rpc as 32602,which again is something that is not prefered.

execution_chain/rpc/debug.nim
RazorClient: very weird 
seems to be some github issue 
https://github.com/RazorClient/nimbus-eth1/blob/d5a4f5d3690bf8e25698cd204c9653de4886ce58/execution_chain/rpc/debug.nim#L194(permalink to base branch for the pr)
does show that it is invalidParams(error) however it doesnt show up in this diff ??!!

RazorClient: without this change the hive tests wouldn't also have passed when i ran them

Issue #3944, opened 2026-01-30
Implement Execution api method eth_feeHistory


kdeme:
Implement eth_feeHistory method as described in https://ethereum.github.io/execution-apis/api/methods/eth_feeHistory

jangko:
fixed by #4112

Issue #3505, opened 2025-07-23
implement `eth_maxPriorityFeePerGas`


tersec:
https://github.com/ethereum/execution-apis/blob/bbf817f7f6404fc0d270f66a3c96f8737e6d7176/src/eth/fee_market.yaml#L29-L42

```yaml
- name: eth_maxPriorityFeePerGas
  summary: Returns the current maxPriorityFeePerGas per gas in wei.
  params: []
  result:
    name: Max priority fee per gas
    schema:
      title: Max priority fee per gas
      $ref: '#/components/schemas/uint'
  examples:
    - name: eth_maxPriorityFeePerGas example
      params: []
      result:
        name: Max priority fee per gas
        value: '0x773c23ba'
```

etan-status:
Encountered this one while playing with spamoor erctx scenario.

kdeme:
Implemented in https://github.com/status-im/nimbus-eth1/pull/4112

PR #4112, opened 2026-04-09
Feat:implement eth_feehistory and eth_maxPriorityFeePerGas 

RazorClient:
This pr introduces the eth_feehistory and eth_maxPriorityFeePerGas rpc methods into Nimbus EL and closes the following issues
- https://github.com/status-im/nimbus-eth1/issues/3505
- https://github.com/status-im/nimbus-eth1/issues/3944

hive test results
```
Apr  9 21:31:17.020 INF API: test started suite=0 test=32 name="eth_feeHistory/fee-history (nimbus-el)"
Apr  9 21:31:17.022 INF API: test ended suite=0 test=32 pass=true
```
```
>>  {"jsonrpc":"2.0","id":1,"method":"eth_feeHistory","params":["0x1","0x1b",[95,99]]}
<<  {"jsonrpc":"2.0","result":{"oldestBlock":"0x1b","baseFeePerGas":["0x000000000000000000000000000000000000000000000000000000003b9aca00","0x00000000000000000000000000000000000000000000000000000000342dc057"],"baseFeePerBlobGas":["0x0000000000000000000000000000000000000000000000000000000000000000","0x0000000000000000000000000000000000000000000000000000000000000000"],"gasUsedRatio":[0.0016543629259729885],"blobGasUsedRatio":[0.0],"reward":[["0x0000000000000000000000000000000000000000000000000000000000000001","0x0000000000000000000000000000000000000000000000000000000000000001"]]},"id":1}
```
addresses parts of https://github.com/status-im/nimbus-eth1/issues/3942

### DIFF
--- a/execution_chain/rpc/oracle.nim
+++ b/execution_chain/rpc/oracle.nim
@@ -106,8 +106,8 @@ proc processBlock(oracle: Oracle, bc: BlockContent, percentiles: openArray[float
     nextBlobBaseFee: getBlobBaseFee(calcExcessBlobGas(com, bc.header, fork), com, fork),
     gasUsedRatio: float64(bc.header.gasUsed) / float64(bc.header.gasLimit),
     blobGasUsedRatio:
-      # for pre-Cancun blocks the division becomes 0/0,which leads the evalution to result in a Nan
-      # guard the divisor to divisor to make sure we dont hit that case
+      # for pre-Cancun blocks the division becomes 0/0, which leads the evaluation to result in NaN
+      # guard the divisor to make sure this case is avoided
       if maxBlobGasPerBlock == 0'u64:
         0.0
       else:
--- a/execution_chain/rpc/oracle.nim
+++ b/execution_chain/rpc/oracle.nim
@@ -105,7 +105,13 @@ proc processBlock(oracle: Oracle, bc: BlockContent, percentiles: openArray[float
     nextBaseFee: calcBaseFee(com, bc),
     nextBlobBaseFee: getBlobBaseFee(calcExcessBlobGas(com, bc.header, fork), com, fork),
     gasUsedRatio: float64(bc.header.gasUsed) / float64(bc.header.gasLimit),
-    blobGasUsedRatio: float64(bc.header.blobGasUsed.get(0'u64)) / float64(maxBlobGasPerBlock)
+    blobGasUsedRatio:
+      # for pre-Cancun blocks the division becomes 0/0,which leads the evalution to result in a Nan
+      # guard the divisor to divisor to make sure we dont hit that case
+      if maxBlobGasPerBlock == 0'u64:
+        0.0
+      else:
+        float64(bc.header.blobGasUsed.get(0'u64)) / float64(maxBlobGasPerBlock)
   )
 
   if percentiles.len == 0:
--- a/execution_chain/rpc/oracle.nim
+++ b/execution_chain/rpc/oracle.nim
@@ -8,7 +8,7 @@
 # those terms.
 
 import
-  std/[hashes, algorithm, strutils],
+  std/[hashes, algorithm],
   eth/eip1559,
   stew/keyed_queue,
   stew/endians2,
@@ -56,7 +56,6 @@ type
     reward : UInt256
 
   BlockRange = object
-    pendingBlock: Opt[uint64]
     lastBlock: uint64
     blocks: uint64
 
@@ -76,11 +75,8 @@ func new*(_: type Oracle, chain: ForkedChainRef): Oracle =
     historyCache: KeyedQueue[CacheKey, ProcessedFees].init(),
   )
 
-func hash*(x: CacheKey): Hash =
-  var h: Hash = 0
-  h = h !& hash(x.number)
-  h = h !& hash(x.percentiles)
-  result = !$h
+template com(oracle: Oracle): CommonRef =
+  oracle.chain.com
 
 func toBytes(list: openArray[float64]): seq[byte] =
   for x in list:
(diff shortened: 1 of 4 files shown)


execution_chain/rpc/oracle.nim
```diff
@@ -100,15 +96,22 @@ func calcBaseFee(com: CommonRef, bc: BlockContent): UInt256 =
...
     blobBaseFee: getBlobBaseFee(bc.header.excessBlobGas.get(0'u64), com, fork),
-    nextBaseFee: calcBaseFee(oracle.com, bc),
-    nextBlobBaseFee: getBlobBaseFee(calcExcessBlobGas(bc.header, com, fork), com, fork),
+    nextBaseFee: calcBaseFee(com, bc),
+    nextBlobBaseFee: getBlobBaseFee(calcExcessBlobGas(com, bc.header, fork), com, fork),
     gasUsedRatio: float64(bc.header.gasUsed) / float64(bc.header.gasLimit),
-    blobGasUsedRatio: float64(bc.header.blobGasUsed.get(0'u64)) / float64(maxBlobGasPerBlock)
+    blobGasUsedRatio:
+      # for pre-Cancun blocks the division becomes 0/0,which leads the evalution to result in a Nan
+      # guard the divisor to divisor to make sure we dont hit that case
```
kdeme:
```suggestion
      # for pre-Cancun blocks the division becomes 0/0, which leads the evaluation to result in NaN
      # guard the divisor to make sure this case is avoided
```

execution_chain/rpc/oracle.nim
```diff
@@ -170,31 +172,15 @@ proc resolveBlockRange(oracle: Oracle, blockId: BlockTag, numBlocks: uint64): Re
...
-        # Pending block not supported by backend, process only until latest block.
-        resolved = headBlock
-        # Update total blocks to return to account for this.
-        dec blocks
-    else:
-      try:
-        resolved = headerFromTag(oracle.com.db, blockId)
-      except CatchableError as exc:
-        return err(exc.msg)
+    let resolved = oracle.chain.headerFromTag(blockId).valueOr:
```
kdeme, line 182: I guess eventually we have to add "pending" functionality to all calls that require it...?

But fine to progress for now like this.

execution_chain/rpc/rpc_utils.nim
```diff
@@ -417,6 +426,8 @@ proc headerFromTag*(chain: ForkedChainRef, blockTag: BlockTag): Result[Header, s
       ok(chain.finalizedHeader)
     of "safe":
       ok(chain.safeHeader)
+    of "earliest":
+      chain.headerByNumber(base.BlockNumber(0))
```
kdeme, line 430: Same remark I had here: https://github.com/status-im/nimbus-eth1/pull/4092#discussion_r3050410850

execution_chain/rpc/server_api.nim
```diff
@@ -743,3 +748,12 @@ proc setupServerAPI*(api: ServerAPIRef, server: RpcServer, am: ref AccountsManag
         return Opt.none(BlockAccessList)
 
     Opt.some(bal)
+
+  server.rpc("eth_feeHistory") do(
+    blockCount: Quantity, newestBlock: BlockTag, rewardPercentiles: Opt[seq[float64]]
+  ) -> FeeHistoryResult:
+    api.oracle.feeHistory(blockCount.uint64, newestBlock, rewardPercentiles.get(@[])).valueOr:
+      raise newException(ValueError, error)
```
kdeme, line 756: We do raise `ValueError` all over the place (mostly for errors that don't actually have a defined code in the execution-apis specifications), so it is understandable for this PR to follow that example.

But I do want to make the remark that this results in the `-32000`  code error for all those errors. 
I think we should move to raising specifically `ApplicationError` and pick a consistent default number for those that are not defined by the specs, e.g. `38000`.

Reason: JSON-RPC specs: 

`The error codes from and including -32768 to -32000 are reserved for pre-defined errors. Any code within this range, but not defined explicitly below is reserved for future use. `

`The remainder of the space is available for application defined errors.`

kdeme (commented):
lgtm

It did make me realize that we are missing the `pending` option in block tag, but that's not in scope of this PR so 👍

execution_chain/rpc/rpc_utils.nim
kdeme: Actually, since the pruner code got merged recently we can now use https://github.com/status-im/nimbus-eth1/pull/3953/changes#diff-75f9cc6813e733de906d9a8209ac17d76094e226f1616108803dc9496ec7464eR659 or similar.

RazorClient: I was thinking of raising a different pr that adds the tail to the logic and also addressing the follow-up reviews to the other previous pr.
I worked on this b4 the pruner was merged and was also thinking about the other active pr i have, so I can add the tail semantics in this pr or add in another one,which ever seems the best thing to do

execution_chain/rpc/server_api.nim
RazorClient: Now that you mention it, we should raise a different number. The question is to do this with this pr or maybe follow up with a separate issue and fix it in all the places where we are abusing 32000 and apply the fix

RazorClient:
I did have a talk with @advaita-saha about the pending case regarding this pr. The pending block tag logic is not there in the backend, and adding it would be quite a hassle.
But yes, the pending blocktag semantics would anyway not be a part of this pr

execution_chain/rpc/rpc_utils.nim
kdeme: Yes, it's fine to tackle both in another PR

execution_chain/rpc/server_api.nim
kdeme: Yes, a PR that corrects this for all instances sounds good

PR #4126, opened 2026-04-14
fix: txpool should order by per-gas tip, not total gas expenditure

advaita-saha:
The transaction pool's `calculatePrice` function in `tx_item.nim` computes a "price" used to order transactions for block building:
```nim
item.price = item.tx.gasLimit * item.tx.tip(baseFee) + item.tx.getTotalBlobGas
```
This orders by **total gas expenditure** instead of per-gas tip. A transaction with high `gasLimit` but low per-gas tip gets prioritized over one with lower `gasLimit` but much higher per-gas tip.

**Example with old logic**:
 - Tx A: gasLimit=500,000, tip=2 gwei → price = 500,000 × 2 gwei = 1×10^15
 - Tx B: gasLimit=75,000, tip=10 gwei → price = 75,000 × 10 gwei = 7.5×10^14
 - A is ordered first despite 5× lower per-gas tip

The standard approach (geth and other clients) orders by **effective gas tip per gas unit**.
The + getTotalBlobGas term is also maybe a dimensional error (adds gas units to wei), subsumed by this fix.

This PR also adds a debug rpc endpoint to check txpool behaviour

### DIFF
--- a/execution_chain/core/tx_pool/tx_tabs.nim
+++ b/execution_chain/core/tx_pool/tx_tabs.nim
@@ -1,5 +1,5 @@
 # Nimbus
-# Copyright (c) 2018-2025 Status Research & Development GmbH
+# Copyright (c) 2018-2026 Status Research & Development GmbH
 # Licensed under either of
 #  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or
 #    http://www.apache.org/licenses/LICENSE-2.0)
--- a/execution_chain/core/tx_pool/tx_tabs.nim
+++ b/execution_chain/core/tx_pool/tx_tabs.nim
@@ -11,7 +11,7 @@
 {.push raises: [].}
 
 import
-  std/[tables, heapqueue],
+  std/[tables, heapqueue, times],
   eth/common/base,
   eth/common/addresses,
   eth/common/hashes,
@@ -60,7 +60,11 @@ func removeLookup*(blobTab: var BlobLookupTab, item: TxItemRef) =
 # HeapQueue needs `<` to be overloaded for custom object
 # and in this case, we want to pop highest price first.
 # That's why we use '>' instead of '<' in the implementation.
-func `<`(a, b: TxItemRef): bool = a.price > b.price
+# Tiebreaker: earliest arrival first (FIFO), matching geth/reth.
+func `<`(a, b: TxItemRef): bool =
+  if a.price != b.price:
+    return a.price > b.price
+  a.time < b.time
 
 proc validBlobItem(item: TxItemRef;
                    fork: EVMFork;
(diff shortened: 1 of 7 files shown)


execution_chain/core/tx_pool/tx_tabs.nim
```diff
@@ -57,6 +57,11 @@ func removeLookup*(blobTab: var BlobLookupTab, item: TxItemRef) =
   for v in item.tx.versionedHashes:
     blobTab.del(v)
 
+# HeapQueue needs `<` to be overloaded for custom object
+# and in this case, we want to pop highest price first.
+# That's why we use '>' instead of '<' in the implementation.
+func `<`(a, b: TxItemRef): bool = a.price > b.price
+
```
advaita-saha, line 64: Not sure if this is the correct way to handle this scenario

advaita-saha:
The `{.used.}` pragma kinda gave me way — it was added specifically because the compiler warned the function was "declared but not used". That warning was correct: the HeapQueue generic never actually picked it up

Instead, the heap was falling back to a default ref pointer comparison, producing essentially arbitrary ordering
Moving it to module scope makes it properly visible to the generic HeapQueue code. The test confirmed this — with the local definition, the 2 gwei tx was ordered before the 10 gwei tx. After moving to module scope, ordering works correctly.

execution_chain/core/tx_pool/tx_tabs.nim
tersec: Need to make sure it's in the correct scope.

tersec: i.e. for `HeapQueue` to pick it up.

advaita-saha: It picks up now, in the earlier code it wasn't getting picked up. I have added tests to check this

jangko: I think we also need to fix the `<`. i.e. the first priority of the sorter is gas tip, but then among txs with same tip, they should be sorted by profit as second priority and there the blob gas also have a role.

advaita-saha: Blob gas fees are burned by protocol and don't go to the block builder. The builder reward is purely `gasUsed × effectiveGasTip` (execution gas only). Given this, what role should blob gas play in the tiebreaker?

jangko: right, I think I was confused by the transaction gas cost.

jangko (approved):

execution_chain/core/tx_pool/tx_tabs.nim
advaita-saha: But the tiebreaker is IMP point
Will add FIFO logic

Issue #3701, opened 2025-09-19
tx pool holding onto future txnonce txs, and proposing blocks with 0 txs


advaita-saha:
These are block proposals from out fleet
```bash
[...]

PR #4143, opened 2026-04-17
fix: txpool bugs

### LINKED ISSUE #3525: TxPool unable to fetch pooled txs
This starts happening after the client is running for few minutes 
And we are filled with 
```bash
Request pooled transactions get nothing    topics="tx-broadcast"
```

This happens in both devnet-3 and our nimbus hoodi fleet

advaita-saha:
Main issue details: https://github.com/status-im/nimbus-eth1/issues/3525#issuecomment-4266672615

FIxes #3525, #3701
Partially addresses #4155 

### Details
- **Queue-full guard (both handlers)**
  - `handleTransactionsBroadcast` and `handleTxHashesBroadcast`: check `wire.actionQueue.full` and drop the announcement instead of awaiting `addLast`
  - Previously: addLast on a full queue blocked the rlpx per-peer dispatch loop, preventing PooledTransactions responses from being read → 10s timeouts → lost peers.
- **Empty-batch short-circuit in `handleTxHashesBroadcast`**
  - Inside the inner loop that builds `PooledTransactionsRequest` batches, if `msg.txHashes.len == 0` (all hashes were already in pool), continue — skip the empty `getPooledTransactions` round-trip.
- **A slow peer no longer head-of-line-blocks the broadcast action queue; queue bump absorbs bigger bursts**
  - `actionHeartbeat` changed from a single `Future` to `seq[Future[void].Raising([CancelledError])]` to support multiple worker loops.
  - `MAX_ACTION_HANDLER`: 128 → 512
  - Added `NUM_ACTION_WORKERS` = 4
  - Constructor spawns `NUM_ACTION_WORKERS` `actionLoops` into `actionHeartbeat` instead of one
 - **Eliminated duplicate KZG blob verification**
   - the handler was validating blob proofs before `addTx`, and `addTx` was validating them again. Now the handler only keeps the cheap `isNil` protocol-breach check; `addTx` is the single source of truth. When `addTx` returns `txErrorInvalidBlob`, the handler disconnects the peer with `BreachOfProtocol`. Cuts per-blob-tx CPU in half (~4 ms → ~2 ms).
 - **Added cooperative yields**
   - `await sleepAsync(ZeroDuration)` after each `addTx` call in both `handleTxHashesBroadcast` and `handleTransactionsBroadcast`. The chronos event loop now gets a scheduling slot between every tx, so RPC/peer-dispatch latency during heavy tx ingestion is bounded by a single tx's processing time (~2 ms) instead of a whole batch (~20+ ms).
  - **Filter already announced hashes before enqueueing** 
    - Drops hashes already in the pool — no fetch needed 
    - Drops hashes already in wire.seenTransactions — another peer's action is (or recently was) handling the fetch. Record the peer in `seen.peers` and update `seen.lastSeen` so we keep the dedupe window accurate. 
    - Keeps novel hashes, marks them in seenTransactions, and enqueues a packet of only the novel hashes.


advaita-saha:
This branch might have unwanted RPC endpoints, to test block building

advaita-saha:
# Consistent good block builds on hoodi
Tested with 10+ peers 24+hrs run
```bash
❯ curl -s http://127.0.0.1:8545 \                                                                ─╯
     -H 'content-type: application/json' \
     -d '{"jsonrpc":"2.0","id":1,"method":"debug_buildTestBlock","params":[]}' | jq
{
  "jsonrpc": "2.0",
  "result": {
    "txCount": 28,
    "blobCount": 19
  },
  "id": 1
}

 ~ ─────────────────────────────────────────────────────────────────────────────── 02:32:24 PM ─╮
❯ curl -s http://127.0.0.1:8545 \                                                                ─╯
     -H 'content-type: application/json' \
     -d '{"jsonrpc":"2.0","id":1,"method":"debug_buildTestBlock","params":[]}' | jq
{
  "jsonrpc": "2.0",
  "result": {
    "txCount": 31,
    "blobCount": 21
  },
  "id": 1
}

 ~ ─────────────────────────────────────────────────────────────────────────────── 02:41:26 PM ─╮
❯ curl -s http://127.0.0.1:8545 \                                                                ─╯
     -H 'content-type: application/json' \
     -d '{"jsonrpc":"2.0","id":1,"method":"debug_buildTestBlock","params":[]}' | jq
{
  "jsonrpc": "2.0",
  "result": {
    "txCount": 32,
    "blobCount": 21
  },
  "id": 1
}
```

PR #4147, opened 2026-04-18
manual GC cleanup

advaita-saha:
Attempt to fix the sawtooth nature of the memory graph


advaita-saha:
### Comparing `master` with `memory-opt-v2` ( `5d597f6` )
Screenshot_from_2026-04-18_23-05-09

execution_chain/core/chain/forked_chain/chain_private.nim
```diff
@@ -57,7 +57,7 @@ proc processBlock*(
     c: ForkedChainRef,
     parentBlk: BlockRef,
     txFrame: CoreDbTxRef,
-    blk: Block,
+    blk: sink Block,
```
bhartnett, line 60: Why use sink here? Block is a stack variable so shouldn't effect the heap memory usage.

execution_chain/core/chain/forked_chain.nim
```diff
@@ -31,6 +31,12 @@ from std/sequtils import mapIt
 from std/heapqueue import len
 from web3/engine_api_types import ExecutionPayloadBodyV1, ExecutionPayloadBodyV2
 
+when defined(linux):
+  # Return freed heap chunks to the OS. glibc's malloc holds freed memory in
+  # per-arena free-lists by default; without this, RSS doesn't shrink after
+  # disposed frames are reclaimed.
+  proc mallocTrim(pad: csize_t): cint {.importc: "malloc_trim", header: "<malloc.h>".}
+
```
bhartnett, line 39: Since this is not available on MacOS or windows, does this fix work on those platforms? Or does it depend on using malloc_trim?

arnetheduck: the rss memory does shrink if there's memory pressure - there's no need for this kind of trimming _generally_, ie RSS is not that relevant

execution_chain/core/chain/forked_chain.nim
```diff
@@ -391,6 +397,20 @@ proc processUpdateBase(c: ForkedChainRef): Future[Result[void, string]] {.async:
...
+    # Under --mm:refc, cycles in the disposed Aristo/KVT frame graph wait for
+    # the next mark-and-sweep pass, so RSS shows a large sawtooth. Force
+    # collection here — the queue just drained, so any cycles involving
+    # just-disposed frames are guaranteed unreachable and this is the natural
+    # idle point to pay the pause cost.
+    var memBefore = 0
+    var memAfter = 0
+    if c.persistedCount > 0:
+      memBefore = getOccupiedMem()
+      GC_fullCollect()
```
arnetheduck, line 409: we want to avoid gc_fullcollect as much as possible as it is _very_ expensive.. 

The way to get rid of sawtooth is to find reference cycles and explicitly break these with `= nil` - what gc_fullcollect does is essentially to collect cycles - the sawtooth happens because the cycle collector runs rarely (and it runs rarely because it's very very slow)

advaita-saha:
Superseded by #4199

PR #4154, opened 2026-04-20
add blockBuilding test API

advaita-saha:
adds RPC endpoint `debug_buildTestBlock`

### DIFF
--- a/execution_chain/rpc/debug.nim
+++ b/execution_chain/rpc/debug.nim
@@ -31,7 +31,13 @@ type
     hash: Hash32
     rlp: seq[byte]
 
+  TestBlockSummary = object
+    txCount: int
+    blobCount: int
+
 BadBlock.useDefaultSerializationIn JrpcConv
+
+TestBlockSummary.useDefaultSerializationIn JrpcConv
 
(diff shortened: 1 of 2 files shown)


RazorClient:
could be modified to become 
https://github.com/ethereum/execution-apis/blob/main/src/testing/testing_buildBlockV1.yaml
would increase the hive RPC-compat scores a bit more then : )

advaita-saha:
> could be modified to become https://github.com/ethereum/execution-apis/blob/main/src/testing/testing_buildBlockV1.yaml would increase the hive RPC-compat scores a bit more then : )



> could be modified to become https://github.com/ethereum/execution-apis/blob/main/src/testing/testing_buildBlockV1.yaml would increase the hive RPC-compat scores a bit more then : )

Yes but I will consider this out of scope of this txpool bugs issue for now
But sure, add it in the rpc issue tracker of create a new issue

kdeme:
> But sure, add it in the rpc issue tracker of create a new issue

Yeah, it's already in there in the master tracker but I hadn't created separate issue as I didn't think it was anything urgent/necessary.

PR #4166, opened 2026-04-22
fixes for engineAPI high latency

advaita-saha:
An attempt to improve the engineAPI performance. The following changes have been done

### Changes
1. **skip redundant RLP decode in `validateVersionedHashed`**
2. **conditional yield in `processQueue`**
3. **reuse tx hashes in `txRecords` instead of re-RLP-hashing**
    After a block is processed, forked_chain  loops over all transactions and calls computeRlpHash(tx) for every one — i.e. re-RLP-encodes and hashes each tx. These hashes were already computed during RLP decode in ethBlock(payload, .... For a 150-tx block that's ~150 × ~20µs ≈ 3ms of wasted work per payload, on the critical path.
4. **non-blocking engine_api enqueue**
    When the FC action queue is at MaxQueueSize=128, queueImportBlock and queueForkChoice call await c.queue.addLast(item) which suspends the caller until space frees up. Under burst load the CL sees unbounded tail latency rather than a clean SYNCING response, and the CL will eventually time out.


advaita-saha:
Unsure about the 4th change, not sure if it's the best idea. Or we should just let the engineAPI timeout instead for this case.
For example the CL approaches the tip of the chain, it sends 30-40blocks almost together, and this is the time when this happens

execution_chain/core/chain/forked_chain.nim
```diff
@@ -471,7 +476,8 @@ proc validateBlock(
     parent: BlockRef,
     blk: Block,
     blockAccessList: Opt[BlockAccessListRef],
-    finalized: bool
+    finalized: bool,
+    txHashes: Opt[seq[Hash32]] = Opt.none(seq[Hash32])
```
tersec, line 480: Don't need to list the type here explicitly since it can infer it from the default value.

execution_chain/core/chain/forked_chain.nim
```diff
@@ -787,16 +809,23 @@ template queueImportBlock*(
...
+    c.importBlock(blk, blockAccessList, finalized, txHashes)
 
   let item = QueueItem(
     responseFut: Future[Result[void, string]].Raising([CancelledError]).init(),
     handler: asyncHandler
   )
-  await c.queue.addLast(item)
+  if c.queue.full:
+    debug "Engine action queue saturated, rejecting importBlock",
+      queueLen = c.queue.len, maxSize = MaxQueueSize
```
tersec, line 824: Why log a constant? Is there an intention to change this to be variable in the future?

execution_chain/core/chain/forked_chain.nim
```diff
@@ -810,7 +839,13 @@ template queueForkChoice*(c: ForkedChainRef,
     responseFut: Future[Result[void, string]].Raising([CancelledError]).init(),
     handler: asyncHandler
   )
-  await c.queue.addLast(item)
+  if c.queue.full:
+    debug "Engine action queue saturated, rejecting forkChoice",
+      queueLen = c.queue.len, maxSize = MaxQueueSize
```
tersec, line 844: Likewise logging a constant.

execution_chain/core/chain/forked_chain.nim
advaita-saha: No, these are only for testing for now
Mostly for me to detect scenarios where this condition is occurring

Would be removed in the final version

tersec:
> Unsure about the 4th change, not sure if it's the best idea. Or we should just let the engineAPI timeout instead for this case. For example the CL approaches the tip of the chain, it sends 30-40blocks almost together, and this is the time when this happens

`SYNCING` is basically always better than a timeout.

advaita-saha:
Closing since the performance improvement is not significant enough

PR #4171, opened 2026-04-23
Rpc error handling

RazorClient:
This pr introduces the Result[Opt[T], string] semantics to the rpc layer that allows us to handle the case where we need to differentiate between
- err(...)
- ok(none(...)) 
- ok(some(...)) 
also adds the tail semantics from the pruner to "earliest" tag resolution

address reviews
- https://github.com/status-im/nimbus-eth1/pull/4092#discussion_r3050400998
- https://github.com/status-im/nimbus-eth1/pull/4092#discussion_r3050400998


RazorClient:
This PR cleans up the RPC error handling semantics by introducing `Result[Opt[T], string]`-style handling into the RPC layer, allowing us to properly differentiate between:

* `err(...)` → invalid params / actual RPC errors
* `ok(none(T))` → valid “not found” case returning `null`
* `ok(some(T))` → successful response

The debug level failures are now removed from this pr as it doesn't fit in with the new ethjson 
To be introduced with the follow-up ethjson pr

PR #4172, opened 2026-04-24
devp2p eth/71 fixes and tests

bhartnett:
Changes in this PR:
- Implement the change to return 0x80 (empty rlp string) when the BAL doesn't exist. See here: https://github.com/ethereum/EIPs/pull/11553
- Added unit tests to cover eth/71 message flows.
- Fixed issues with receipts message request and handling for the new eth/70 receipt message type.

### DIFF
--- a/execution_chain/sync/wire_protocol/eth/eth_handler.nim
+++ b/execution_chain/sync/wire_protocol/eth/eth_handler.nim
@@ -249,8 +249,6 @@ proc getBlockHeaders*(ctx: EthWireRef,
 
 proc getBlockAccessLists*(
     ctx: EthWireRef, req: BlockAccessListsRequest): BlockAccessListsPacket =
-  const emptyBal = default(BlockAccessList)
-
   var blockHashes = req.blockHashes
   blockHashes.setLen(min(req.blockHashes.len(), MAX_BALS_SERVE))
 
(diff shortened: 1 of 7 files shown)


execution_chain/sync/wire_protocol/eth/eth_handler.nim
```diff
@@ -20,14 +20,14 @@ logScope:
...
+  MAX_BODIES_SERVE*    = 256
   # https://github.com/ethereum/devp2p/blob/master/caps/eth.md#getpooledtransactions-0x09
-  MAX_TXS_SERVE       = 256
-  MAX_BALS_SERVE      = 256
-  MAX_ACTION_HANDLER  = 512
-  NUM_ACTION_WORKERS  = 4
+  MAX_TXS_SERVE*       = 256
+  MAX_BALS_SERVE*      = 256
+  MAX_ACTION_HANDLER*  = 512
+  NUM_ACTION_WORKERS*  = 4
```
bhartnett: Exported these so they can be used in the tests.

execution_chain/sync/wire_protocol/eth/eth_responder.nim
```diff
@@ -476,10 +477,6 @@ template registerCommonThunk(protocol: ProtocolInfo, PROTO: type) =
               pooledTransactionsThunk[PROTO], PooledTransactionsPacket)
   registerMsg(protocol, GetPooledTransactionsMsg, "getPooledTransactions",
               getPooledTransactionsThunk[PROTO], PooledTransactionsRequest)
-  registerMsg(protocol, ReceiptsMsg, "receipts",
-              receiptsThunk[PROTO], ReceiptsPacket)
-  registerMsg(protocol, GetReceiptsMsg, "getReceipts",
-              getReceiptsThunk[PROTO], ReceiptsRequest)
```
bhartnett, line 482: Removed from the common registration because eth/70 and eth/71 no longer use this message format.

execution_chain/sync/wire_protocol/eth/eth_handler.nim
jangko: Or we can use `import module_name {.all.}` in the test module if we only ever to use the exported symbol in test while in production code we keep to scope pollution to a minimum.

jangko (approved):

execution_chain/sync/wire_protocol/eth/eth_handler.nim
bhartnett: > Or we can use `import module_name {.all.}` in the test module if we only ever to use the exported symbol in test while in production code we keep to scope pollution to a minimum.

Ok, I'll do that instead.

PR #4173, opened 2026-04-24
fix: bal bug in unify tx validation

advaita-saha:


jangko (approved):

jangko:
@bhartnett does it make any sense to add `doAssert tracker.hasPendingCallFrame()` into `getBlockAccessList` to prevent something like this happen again?

bhartnett:
> @bhartnett does it make any sense to add `doAssert tracker.hasPendingCallFrame()` into `getBlockAccessList` to prevent something like this happen again?

Yes that's a good idea. I guess you meant 'doAssert not tracker.hasPendingCallFrame()'. I'll put that in my next change.

PR #4174, opened 2026-04-24
async `updateBase`

advaita-saha:
updateBase is now async, yielding once after db.persist() and once every 16 ancestors during the cleanup loop — so a full 256-block batch gets

  ~17 yield points instead of 0.
  - Reordered so c.base / c.baseTxFrame pointer updates happen before the cleanup burst, with oldFrontier = base.parent captured first. If a shutdown cancels us mid-cleanup, ForkedChain invariants stay coherent.
  - Split wall-clock into persistMs + cleanupMs aggregated over the batch and surfaced in the existing "Finalized blocks persisted" log — the cheapest possible instrumentation to tell you at a glance whether future stalls are disk or cleanup.
  - processUpdateBase now awaits the async updateBase.


bhartnett:
@advaita-saha I find all the comments in the PR to be overly verbose. I would suggest removing most of them, just keeping the most necessary.

execution_chain/core/chain/forked_chain.nim
```diff
@@ -338,20 +339,45 @@ with --debug-eager-state-root."""
...
+    count = 0'u
+    sinceYield = 0
+
+  for it in ancestors(oldFrontier):
     c.removeBlockFromCache(it)
     inc count
+    inc sinceYield
+    if sinceYield >= cleanupYieldChunk:
+      sinceYield = 0
+      await sleepAsync(0.milliseconds)
```
bhartnett: This could be implemented in a simpler way. Why not just add an i index to the iterator and then if i mod 16 == 0 then run the sleep.

advaita-saha:
> I find all the comments in the PR to be overly verbose

Yes the comments are there mostly for me so intentionally verbose, as I am working on multiple branches with different things.
Will remove when the PR is finalized

execution_chain/core/chain/forked_chain.nim
advaita-saha: Yes agreed

bhartnett: Or maybe the `removeBlockFromCache` function could become async so you can simply avoid the sleep altogether.

execution_chain/core/chain/forked_chain.nim
```diff
@@ -367,9 +393,7 @@ with --debug-eager-state-root."""
       batchSizeLowerBound = 4
       batchSizeUpperBound = 256
 
-    let
-      finishTime = Moment.now()
-      runTime = (finishTime - startTime).milliseconds
+    let runTime = (finishTime - startTime).milliseconds
```
bhartnett: I would remove cleanupMs since that part is no longer blocking the event loop after this change.

Then runTime used here should just equal persistMs

advaita-saha: https://github.com/status-im/nimbus-eth1/pull/4174/commits/3d5b9bc80358f2459500f1dc217589b641a23b5c

advaita-saha:
Deploying branch to my home validator nodes ( hoodi ) for analysing improvement

advaita-saha:
Noticed minimal to no improvement in attestation perfomance

PR #4183, opened 2026-04-27
add metrics to engineAPI

advaita-saha:


execution_chain/rpc/engine_api.nim
```diff
@@ -19,6 +21,18 @@ from ../beacon/web3_eth_conv import Hash32
 
 {.push raises: [].}
 
+declareGauge nec_engine_api_request_duration_ms,
```
arnetheduck: time metrics should always be counters that accumulate total time, else a fast call will overwrite a slow call

arnetheduck: also, we typically use `toFloatSeconds` for this kind of metric

advaita-saha: well the purpose of adding the metrics to engineAPI for now is to detect the exact scenarios which causes the fCU & getPayload call to take longer than usual, which inturn causes sometimes a missed proposal or orphaned block

So for this detection, getting the average time won't really help. So I went with `declareGauge` for now

advaita-saha: Example of spikes which I want to identify and debug
Screenshot from 2026-04-27 18-23-31

arnetheduck: > getting the average time

a counter gives _cumulative_ time - if one api request takes 10s and the other takes 2s, the counter will show 12 while the gauge shows 2. You've lost the 10 and the graph will be missing information.

In the case that you call the metrics api often enough, the two are same but if you call it less frequently, a counter loses less information.

tersec (approved):

PR #4188, opened 2026-04-29
Remove ref types from account and storage leaf caches

bhartnett:
The goal here is to remove the ref types from the account and storage leaf caches so that we can implement batch IO / pre-fetching of accounts and storage slots in background thread/s. The background threads doing the pre-fetching will need to write to the shared lru caches but refc uses thread local heaps which means refs allocated and written to the lru caches from background threads will cause memory corruption issues. Using value types in the caches solves this problem.

Tthe current VertexBuf cache in the rocksdb backend may sometimes duplicate the leaf data that is already stored in the leaf cache so it would perhaps be possible to remove the leaf caches but my preference is to keep them because the pre-fetching will perform much better when the leaf data is converted and stored in the required format. Fetching from the VertexBuf cache would require an allocation to convert into the vertex ref type for every call which is not ideal.

In practice, leaf vertex refs are not shared between the txFrame layers and the LRU leaf caches so converting the caches to use value types shouldn't increase memory usage due to duplication. In fact memory usage might go down since we no longer hold the ref pointer in the caches and just store the object directly.

### DIFF
--- a/execution_chain/db/aristo/aristo_desc.nim
+++ b/execution_chain/db/aristo/aristo_desc.nim
@@ -129,7 +129,7 @@ type
 
     txRef*: AristoTxRef              ## Bottom-most in-memory frame
 
-    accLeaves*: LruCache[Hash32, AccLeafRef]
+    accLeaves*: LruCache[Hash32, Opt[(AristoAccount, StorageID, NibblesBuf)]]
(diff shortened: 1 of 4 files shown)


bhartnett:
This change doesn't perform well, most likely due to the copying of value objects.

PR #4192, opened 2026-04-29
Remove ref types from account and storage leaf caches

bhartnett:
The goal here is to remove the ref types from the account and storage leaf caches so that we can implement batch IO / pre-fetching of accounts and storage slots in background thread/s. The background threads doing the pre-fetching will need to write to the shared lru caches but refc uses thread local heaps which means refs allocated and written to the lru caches from background threads will cause memory corruption issues. Using value types in the caches solves this problem.

The current VertexBuf cache in the rocksdb backend may sometimes duplicate the leaf data that is already stored in the leaf cache so it would perhaps be possible to remove the leaf caches but my preference is to keep them because the pre-fetching will perform much better when the leaf data is converted and stored in the required format. Fetching from the VertexBuf cache would require an allocation and deserialization to convert into the vertex ref type for every call which is not ideal.

In practice, leaf vertex refs are not shared between the txFrame layers and the LRU leaf caches so converting the caches to use value types shouldn't increase memory usage due to duplication. In fact memory usage might go down since we no longer hold the ref pointer in the caches and just store the object directly.

### DIFF
--- a/execution_chain/db/aristo/aristo_desc/desc_structural.nim
+++ b/execution_chain/db/aristo/aristo_desc/desc_structural.nim
@@ -66,6 +66,23 @@ type
   StoLeafRef* = ref object of LeafRef
     stoData*: UInt256
 
+  CachedAccLeaf* = object
+    case empty*: bool
+    of false:
+      pfx*: NibblesBuf
+      account*: AristoAccount
+      stoID*: StorageID
+    of true:
+      discard
+
+  CachedStoLeaf* = object
+    case empty*: bool
+    of false:
+      pfx*: NibblesBuf
+      stoData*: UInt256
+    of true:
+      discard
+
   NodeRef* = ref object of RootRef
     ## Combined record for a *traditional* ``Merkle Patricia Tree` node merged
     ## with a structural `VertexRef` type object.
--- a/execution_chain/db/aristo/aristo_desc/desc_structural.nim
+++ b/execution_chain/db/aristo/aristo_desc/desc_structural.nim
@@ -68,20 +68,20 @@ type
 
   CachedAccLeaf* = object
     case empty*: bool
+    of true:
+      discard
     of false:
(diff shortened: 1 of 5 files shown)


execution_chain/db/aristo/aristo_desc/desc_structural.nim
```diff
@@ -66,6 +66,23 @@ type
...
+      account*: AristoAccount
+      stoID*: StorageID
+
+  CachedStoLeaf* = object
+    case empty*: bool
+    of true:
+      discard
+    of false:
+      pfx*: NibblesBuf
+      stoData*: UInt256
```
bhartnett, line 87: Using these case objects which support indicating when the value doesn't exist to avoid the Opt[Opt[T]] returned from the LruCache get calls.

bhartnett:
Block import starting from around block 8 million. Imported 100000 blocks:

**Before (master branch)**
elapsed=17m22s168ms

**After (leafcaches branch)**
elapsed=17m18s697ms

bhartnett:
Ran another import, the first 3 million blocks covering the DOS attack block range:

**Before (master branch)**
elapsed=27m45s66ms

**After (leafcaches branch)**
elapsed=27m29s616ms

bhartnett:
Ran another benchmark at block 15 million imported 50000 blocks:

**Before (master branch)**
elapsed=35m18s840ms

**After (leafcaches branch)**
elapsed=34m46s33ms

execution_chain/db/aristo/aristo_desc/desc_structural.nim
arnetheduck: you can instead use zero values here and thus save a few bytes - ie `stoData = 0` is equivalent to "doesn't exist in the database" (this will make the cachedstoleaf type smaller) ... I believe the same holds true for an account full of zeroes thought that remains to be verified.

PR #4194, opened 2026-04-30
pre-flight fcu for speed bump

advaita-saha:
makes fcu call super-fast with minimal downside 
Upon testing shows significant performance improvement, and improvement in attestations from CL side ( fewer attestation misses )


execution_chain/core/chain/forked_chain.nim
```diff
@@ -813,6 +816,46 @@ template queueForkChoice*(c: ForkedChainRef,
...
+  ## mirroring the failure paths inside `forkChoice()`. Run this from the RPC
+  ## handler before enqueuing so we don't ack a request that the queue worker
+  ## would later reject (since the ack has already been sent by then).
+  ##
+  ## Invariant: between this check and the worker actually running our
+  ## enqueued item, the only mutators of `c.heads`/`hashToBlock` are
+  ## previously-queued `validateBlock` (adds heads, never removes the
+  ## validated one) and `updateFinalized` (only prunes branches that don't
+  ## contain the finalized block we already verified is on head's chain).
+  ## So a successful preflight remains valid by the time the worker runs.
```
bhartnett, line 843: Is the logic in this function duplicated from another location? If so, should be refactored into a common helper.

execution_chain/beacon/api_handler/api_forkchoice.nim
```diff
@@ -199,13 +199,26 @@ proc forkchoiceUpdated*(ben: BeaconEngineRef,
...
+  # Pre-validate that the forkchoice update will succeed in the queue worker.
+  # We do this synchronously so we can surface a per-call error here; once the
+  # worker has been kicked off (below), errors can no longer be reported back
+  # to this fCU response.
+  chain.preflightForkChoice(headHash, finalizedBlockHash, safeBlockHash).isOkOr:
     return invalidFCU(error, chain, header)
 
+  # Enqueue the actual forkchoice apply. For ack-only fCUs we don't await
+  # the result. This avoids the response getting blocked behind the shared queue worker
+  let fcuFut = chain.queueForkChoice(headHash, finalizedBlockHash, safeBlockHash)
```
bhartnett, line 211: Is it possible for the queueForkChoice background job to fail in the background? Or after running preflightForkChoice and returning a success to the caller , should the job always succeed?

jangko:
LGTM

execution_chain/beacon/api_handler/api_forkchoice.nim
advaita-saha: Theoretically it is not possible to fail.
The only possibility is if the db is corrupt, or somehow the db connection is broken ( or other db level failures )

execution_chain/core/chain/forked_chain.nim
advaita-saha: 2 lines maybe, not fully
IMO for that 2 lines not worth it to create a common helper

bhartnett (approved):

execution_chain/core/chain/forked_chain.nim
```diff
@@ -813,6 +816,46 @@ template queueForkChoice*(c: ForkedChainRef,
...
+  let head = c.hashToBlock.getOrDefault(headHash)
+  if head.isNil:
+    return err("Cannot find head block: " & headHash.short)
+
+  if safeHash != zeroHash32 and c.hashToBlock.getOrDefault(safeHash).isNil:
+    return err("Cannot find safe block: " & safeHash.short)
```
bhartnett, line 849: This check is more strict now than the implementation in forkChoice which would silently ignore if the safeHash doesn't exist in the hashToBlock map. Not sure if this change is desired or not.

execution_chain/beacon/api_handler/api_forkchoice.nim
bhartnett, line 211: Looks like there is nothing logging the result of the forkChoice call. Even if we don't expect any failures we should probably log something. Perhaps the queueForkChoice asyncHandler could log the result of forkChoice.

bhartnett: Actually it might be possible for it to fail when processing multiple heads. The `removeBlockFromCache` function can remove blocks from the `hashToBlock` table which can cause `forkChoice` to fail. `removeBlockFromCache` gets called from two places `updateBase` and `updateFinalized`. If those functions get called asynchronously before the forkChoice task runs then you get a silent failure.

advaita-saha: Surely yeah, will add some logs. Will help identify issues

advaita-saha: Well the `queueForkChoice` just adds the fCU call to the asyncQueue. So nothing else calls it async. And the queue is processed one by one

advaita-saha: > `removeBlockFromCache` gets called from two places `updateBase` and `updateFinalized`. If those functions get called asynchronously before the forkChoice task runs then you get a silent failure.

I think not possible in the current design with `asyncQueue` implementation

execution_chain/beacon/api_handler/api_forkchoice.nim
```diff
@@ -199,13 +199,26 @@ proc forkchoiceUpdated*(ben: BeaconEngineRef,
...
-  (await chain.queueForkChoice(headHash, finalizedBlockHash, safeBlockHash)).isOkOr:
+  # Pre-validate that the forkchoice update will succeed in the queue worker.
+  # We do this synchronously so we can surface a per-call error here; once the
+  # worker has been kicked off (below), errors can no longer be reported back
+  # to this fCU response.
+  chain.preflightForkChoice(headHash, finalizedBlockHash, safeBlockHash).isOkOr:
     return invalidFCU(error, chain, header)
 
+  # Enqueue the actual forkchoice apply. For ack-only fCUs we don't await
+  # the result. This avoids the response getting blocked behind the shared queue worker
```
bhartnett, line 210: Note that because you don't await on the queue, it is possible to fill up memory with pending async tasks/futures if more api calls are made then what can be processed. You could just await on the queue rather than the future/result.

execution_chain/beacon/api_handler/api_forkchoice.nim
```diff
@@ -199,13 +199,26 @@ proc forkchoiceUpdated*(ben: BeaconEngineRef,
...
+  let fcuFut = chain.queueForkChoice(headHash, finalizedBlockHash, safeBlockHash)
+
   # If payload generation was requested, create a new block to be potentially
   # sealed by the beacon client. The payload will be requested later, and we
   # might replace it arbitrarilly many times in between.
   if attrsOpt.isSome:
+    # Payload generation reads `chain.latestHeader`, so we must wait for the
+    # apply to complete before assembling the block.
+    (await fcuFut).isOkOr:
+      return invalidFCU(error, chain, header)
```
bhartnett, line 220: Since we end up awaiting here for the block building it might be better to not use the async queue flow at all if attrsOpt.isSome. Block building would be faster in that case.

execution_chain/beacon/api_handler/api_forkchoice.nim
bhartnett: Hmm, I'm not sure. Maybe its ok because the single queue enforces the ordering as long as there are no await calls which yield to other tasks before the forkChoice task is added to the queue.

execution_chain/beacon/api_handler/api_forkchoice.nim
advaita-saha: Yes that PR is also ready on top of this. Just testing and ironing out few issues on that
( little difficult to test as I don't have much validators, so very less proposals to test )

execution_chain/beacon/api_handler/api_forkchoice.nim
advaita-saha: @jangko any thoughts on this ?

execution_chain/beacon/api_handler/api_forkchoice.nim
advaita-saha: the queue is already awaited
https://github.com/status-im/nimbus-eth1/blob/8cc94ca821facbd9e73c316285b4a17e63284e31/execution_chain/core/chain/forked_chain.nim#L802-L814

execution_chain/beacon/api_handler/api_forkchoice.nim
bhartnett: > Hmm, I'm not sure. Maybe its ok because the single queue enforces the ordering as long as there are no await calls which yield to other tasks before the forkChoice task is added to the queue.

Since the queueForkChoice is a template and contains an await call (which is good for rate limiting purposes), then the await will yield the event loop meaning something else can run concurrently perhaps before the forkChoice task is added to the queue. Maybe this is fine as long as no other code calls updateBase or updateFinalized outside of the task queue flow.

advaita-saha: Yes that is true.
For now by design `updateBase` & `updateFinalized` is not called outside the queue 

Will add some comments on this

execution_chain/core/chain/forked_chain.nim
```diff
@@ -813,6 +825,46 @@ template queueForkChoice*(c: ForkedChainRef,
   await c.queue.addLast(item)
   item.responseFut
 
+func preflightForkChoice*(c: ForkedChainRef,
+                          headHash: Hash32,
+                          finalizedHash: Hash32,
+                          safeHash: Hash32 = zeroHash32):
+                            Result[void, string] =
+  ## Pre-validate a forkchoice update against the current in-memory chain,
+  ## mirroring the failure paths inside `forkChoice()`. Run this from the RPC
```
tersec, line 834: While these are likely to be relatively static, as an ongoing maintenance point, right now the synchronization between the two has been manually checked (and I've been checking it as well, as did you and @bhartnett ). But if the engine API rules change at some point, e.g., Gloas enabling backwards head movement in engine API as had been prohibited but might be allowed, then this being out of sync is a risk.

execution_chain/core/chain/forked_chain.nim
```diff
@@ -813,6 +825,46 @@ template queueForkChoice*(c: ForkedChainRef,
...
+  ## handler before enqueuing so we don't ack a request that the queue worker
+  ## would later reject (since the ack has already been sent by then).
+  ##
+  ## Invariant: between this check and the worker actually running our
+  ## enqueued item, the only mutators of `c.heads`/`hashToBlock` are
+  ## previously-queued `validateBlock` (adds heads, never removes the
+  ## validated one) and `updateFinalized` (only prunes branches that don't
```
tersec, line 841: After:
- https://github.com/ethereum/execution-apis/pull/786

This is is still somewhat true, but e.g., the checks at:
https://github.com/status-im/nimbus-eth1/blob/2749417c07d7292552d8f0d838cd293ecdf29dc1/execution_chain/beacon/api_handler/api_forkchoice.nim#L170-L182
 https://github.com/status-im/nimbus-eth1/blob/2749417c07d7292552d8f0d838cd293ecdf29dc1/execution_chain/core/chain/forked_chain.nim#L1227-L1241

aren't valid anymore in Amsterdam. So it's not really that the head becomes not-valid; it was as valid as it ever was, it just might not be the head anymore.

This pre-flight check still looks ok, I don't see how it can result in an ultimately incorrect outcome as a result of this change, but the reasoning for it differs.

execution_chain/core/chain/forked_chain.nim
```diff
@@ -813,6 +825,46 @@ template queueForkChoice*(c: ForkedChainRef,
...
+    let fin = c.hashToBlock.getOrDefault(finalizedHash)
+    if fin.isNil:
+      return err("Cannot find finalized block: " & finalizedHash.short)
+    if fin.number > head.number:
+      return err("Invalid finalizedHash: block is newer than head block")
+    if c.heads.len > 1:
+      var found = false
+      for it in ancestors(head):
+        if it == fin:
+          found = true
```
tersec, line 861: Why set this indicator variable rather than directly early-return?

execution_chain/core/chain/forked_chain.nim
```diff
@@ -813,6 +825,46 @@ template queueForkChoice*(c: ForkedChainRef,
...
+          break
+      if not found:
+        return err("Invalid finalizedHash: block not in argument head ancestor lineage")
+
+  ok()
+
 func resolvedFinHash*(c: ForkedChainRef): Hash32 =
   c.latestFinalized.hash
```
tersec, line 870: There might be a mismatch here between the preflight and the full fork choice handling. When `finalizedHash == zeroHash32`, i.e. the change has never finalized, and there's e.g., a reorg so that the condition in
https://github.com/status-im/nimbus-eth1/blob/2749417c07d7292552d8f0d838cd293ecdf29dc1/execution_chain/core/chain/forked_chain.nim#L760-L771
doesn't fire, then it will try to run `findFinalizedPos` which per https://github.com/status-im/nimbus-eth1/blob/2749417c07d7292552d8f0d838cd293ecdf29dc1/execution_chain/core/chain/forked_chain.nim#L98-L132 won't find the 0x0 finalized block hash and become an error.

At that point, the preflight and backend handling might have a mismatch. It actually seems like preflight here might have the correct handling and the gap might be in the backend handler.

PR #4199, opened 2026-05-01
clear blk from async closure env

advaita-saha:
Screenshot from 2026-05-02 08-51-58


advaita-saha:
## 7hrs run
Screenshot from 2026-05-02 21-19-46

Complete removal of `seq[Transaction]` after fixing the block downloader

execution_chain/sync/beacon/worker/blocks/blocks_import.nim
```diff
@@ -39,7 +44,9 @@ proc importBlock*(
     try:
       # TODO: The block access list needs to be passed in when available over devp2p
       # and when the block falls within the BAL retention period.
-      (await ctx.chain.queueImportBlock(blk, Opt.none(BlockAccessListRef))).isOkOr:
+      let importRes = await ctx.chain.queueImportBlock(blk, Opt.none(BlockAccessListRef))
+      clearBlock(blk)
```
arnetheduck, line 48: this is an interesting find, but we do not really want to rely on these compiler internals .. instead, the entire `importBlock` function can be removed - it serves no purpose since `queueImportBlock` already performs the necessary sleeping - this means that the entire `nextAsyncNanoSleep` mechanism should be removed.

arnetheduck: once that is done, the queue in forkchoice can also be modernised using the same technique as was applied in https://github.com/status-im/nimbus-eth2/pull/7546, ie use an `AsyncLock` instead of an `AsyncQueue` in `forked_chain`.

Taken together, these should simplify reasoning about lifetimes - in particular, `blk` itself does not contain any reference cycles - while clearing it helps keep memory usage lower, the underlying issue of circular references is still is present.

advaita-saha: Yes agreed
But the changes would be quite larger refactor within the syncer, for error handling, and safe peer daemon shutdown

Would prefer doing this in a separate PR

arnetheduck: ok, so let's remove clearBlock entirely from this PR, then we can continue with the rest

advaita-saha: Also remove `clearBlock` from the forkedChain inside `validateBlock`?
I feel that's still important to stay even if we do the syncer changes

arnetheduck: also from there .. it should not be needed once the queue itself is refactored to use an asynclock (ie the lifetimes will be more straightforward to reason about and that's the best thing to clear / manage - if you clear the thing that "owns" the blk, the blk itself will be cleared too)

advaita-saha: Ahh you mean also the `asyncLock` changes together
Yeah that makes sense, it can be done. I need to check how it works; I haven't used it before. 

Since testing these takes quite a lot of time, and this fix brings quite a big memory improvement I was thinking let the `clearBlock` stay in forkedChain for now, and remove it from the syncer

Then in next PR I will remove the `nextAsyncNanoSleep` and syncer's `blockImport`. Which fixes the memory issue from the syncer. ( This is quite easy can have a PR up by today, and with testing can be merged in a 2-3days )

Then I can start working on the `asyncLock` in forkedChain. This refactors a lot how FC works, and all the stuff that uses FC. This might take quite a bit of time to write and test. Also few concerns about it, IIRC @jangko did a similar implementation before, and it was weirdly buggy ( not with AsyncLock, maybe semaphores ) because of some deadlock issue. AFAIK `validateBlock` calls `queueUpdateBase` — if both end up taking the same lock, deadlock. The current queue design serializes via the worker so it sidesteps this; with a lock we either need a single outer lock that wraps both, or to factor out the inner operation as a private "lock-not-held" variant.
So on the asyncLock I am not sure how it will span out, or how long it might take

arnetheduck: In general, the technique is quite straightforward - an AsyncLock is approximately equivalent to an AsyncQueue with 1 item - meaning that if you had deadlocks before, they are potentially still possible when the queue is full. 

That said, the to use AsyncLock:
* replace `await queue.add` with `await acquire; defer: release`
* remove the queue
* await the task inside the lock

for quarantine items, `asyncSpawn` instead of await - this is fine because the quarantine is bounded so there's a limit to the amount of tasks we'll have in flight.

PR #4200, opened 2026-05-02
proxy: Go binding

chirag-parmar:
### Description

The PR implements a Go binding for the verified proxy. The async event loop of chronos runs in a goroutine that is locked to a OS thread. Pointers to channels are passed as userData into the C-API so they can be used to send responses later.

### Notes:

1. Genesis SSZ is converted to a much lighter version of directly defining the validators root and the genesis time (since those two are the only things that are used from genesis) - genesis_params.nim

2. The release pipeline depends on the static libraries published as artefacts. The go tool functionality is used to download the libraries for application developers alongside the module.


jangko:
@chirag-parmar this PR breaks nightly build.
https://github.com/status-im/nimbus-eth1/actions/runs/25622006872/job/75210285837
windows-amd64: looks like `LIBEXT="lib"` not needed for some reason/misconfiguration in root makefile.
macos-arm64: need to alias `llvm-ar` to osxcross's `aarch64-apple-darwin${DARWIN_VER}-ar`.


nightly build ci can be manually triggered to run on a selected branch.

PR #4203, opened 2026-05-03
remove `nextAsyncNanoSleep` from syncer as FC already yields

advaita-saha:


advaita-saha:
Screenshot from 2026-05-04 02-27-09

</br>

Seems that this change single-handedly tackle the `seq[Transaction]` bloat @arnetheduck

arnetheduck:
> single-handedly tackle

Hm, yeah, I see the syncer is written as a giant set of nested templates which _tend to_ be harder to analyse from a memory lifetime perspective - they also _tend to_ generate longer-living memory footprints because there is a single giant closure environment that holds on to various references longer than they're needed (a template adds variables to where it's instantiated while the environment of a proc is released when the proc ends)

PR #4210, opened 2026-05-05
Add tracking of EIP-7702 delegation target in BAL on aborted calls

kdeme:
EEST zkevm tests have shown a blockAccessListHash mismatch in two test vectors which do a EIP-7702 call to delegated EOA that gets aborted. Currently in this scenario the delegation address does not get recorded in the BAL currently.

As per EIP-7928 spec, the delegation target must be recorded in the BAL once the access_cost check passes inside gas calculation, regardless of whether the call actually executes.

---

At https://eips.ethereum.org/EIPS/eip-7928#eip-7702-delegation:

> When a call target has an [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702) delegation, the target is accessed to resolve the delegation. If a delegation exists, the delegated address requires its own access_cost check before being accessed. If this check fails, the delegated address MUST NOT appear in the BAL, though the original call target is included (having been accessed to resolve the delegation).

> Note: Delegated accounts cannot be empty, so GAS_NEW_ACCOUNT never applies when resolving delegations."

So anything after the access_cost check should be added, also if the call does not execute.

### DIFF
--- a/execution_chain/block_access_list/block_access_list_tracker.nim
+++ b/execution_chain/block_access_list/block_access_list_tracker.nim
@@ -14,6 +14,7 @@ import
   eth/common/addresses,
   stint,
   ../db/ledger,
+  ../core/eip7702,
   ./block_access_list_builder
 
 export addresses, block_access_list_builder, ledger, stint
@@ -388,6 +389,16 @@ proc trackCodeChange*(
   tracker.trackAddressAccess(address)
   tracker.capturePreCode(address)
   tracker.pendingCallFrame.codeChanges[address] = newCode
+
+proc trackDelegationTarget*(tracker: BlockAccessListTrackerRef, codeAddress: Address) =
+  ## Track the delegation target address for a call to a delegated EOA.
+  ## Called after gas has been successfully charged (i.e. after the access_cost check)
+  ## but before the sub-call executes in order to cover the case where the sub-call
+  ## is aborted (i.e. depth exceeded and insufficient balance) and getCallCode never
+  ## runs to track the target itself.
+  let delegateTo = parseDelegationAddress(tracker.ledger.getCode(codeAddress))
+  if delegateTo.isSome():
+    tracker.trackAddressAccess(delegateTo.get())
 
 proc trackSelfDestruct*(tracker: BlockAccessListTrackerRef, address: Address) =
   tracker.trackBalanceChange(address, 0.u256)
--- a/execution_chain/evm/interpreter/op_handlers/oph_call.nim
+++ b/execution_chain/evm/interpreter/op_handlers/oph_call.nim
@@ -264,6 +264,8 @@ proc callOp(cpt: VmCpt): EvmResultVoid =
         gasCallDelegate: cpt.gasCallDelegate(p.codeAddress),
         contractGas:     p.gas))
 
+  if cpt.balTrackerEnabled:
+    cpt.vmState.balTracker.trackDelegationTarget(p.codeAddress)
   ? cpt.opcodeGasCost(Call, gasCost, reason = $Call)
   cpt.gasMeter.escrowSubcallRegularGas(childGasLimit)
 
@@ -335,6 +337,8 @@ proc callCodeOp(cpt: VmCpt): EvmResultVoid =
         gasCallDelegate: cpt.gasCallDelegate(p.codeAddress),
         contractGas:     p.gas))
 
+  if cpt.balTrackerEnabled:
+    cpt.vmState.balTracker.trackDelegationTarget(p.codeAddress)
   ? cpt.opcodeGasCost(CallCode, gasCost, reason = $CallCode)
   cpt.gasMeter.escrowSubcallRegularGas(childGasLimit)
 
@@ -406,6 +410,8 @@ proc delegateCallOp(cpt: VmCpt): EvmResultVoid =
         gasCallDelegate: cpt.gasCallDelegate(p.codeAddress),
         contractGas:     p.gas))
 
+  if cpt.balTrackerEnabled:
+    cpt.vmState.balTracker.trackDelegationTarget(p.codeAddress)
   ? cpt.opcodeGasCost(DelegateCall, gasCost, reason = $DelegateCall)
   cpt.gasMeter.escrowSubcallRegularGas(childGasLimit)
 
@@ -470,6 +476,8 @@ proc staticCallOp(cpt: VmCpt): EvmResultVoid =
         gasCallDelegate: cpt.gasCallDelegate(p.codeAddress),
         contractGas:     p.gas))
 
+  if cpt.balTrackerEnabled:
+    cpt.vmState.balTracker.trackDelegationTarget(p.codeAddress)
   ? cpt.opcodeGasCost(StaticCall, gasCost, reason = $StaticCall)
   cpt.gasMeter.escrowSubcallRegularGas(childGasLimit)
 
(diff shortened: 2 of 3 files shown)


kdeme:
The `access_cost`  check mentioned in the spec should be this one: https://github.com/status-im/nimbus-eth1/blob/master/execution_chain/evm/interpreter/gas_costs.nim#L404

So ideally the tracking happens at that point in code, but it is not that great that we would have to pass in a proc for just that.

This check https://github.com/status-im/nimbus-eth1/blob/master/execution_chain/evm/interpreter/gas_costs.nim#L420 is NOP for amsterdam+

I think this check https://github.com/status-im/nimbus-eth1/blob/master/execution_chain/evm/interpreter/gas_costs.nim#L441 cannot really occur and is more of a safeguard, but to be cleaner I could add a check to verify if the returned error is `OutOfGas` or not.

Or perhaps we need to move some code around.

kdeme:
Note that this happens on two zkevm specific tests:

`./tests/fixtures/eest_zkevm/blockchain_tests/for_amsterdam/amsterdam/eip8025_optional_proofs/witness_7702/witness_codes_delegated_eoa_insufficient_balance.json`

`./tests/fixtures/eest_zkevm/blockchain_tests/for_amsterdam/amsterdam/eip8025_optional_proofs/witness_validation_codes/validation_codes_missing_delegated_code_on_insufficient_balance_call.json`

Both have delegated code + insufficient balance.

These should probably also get added to the general BAL eest tests.

bhartnett:
So just to summarize our discussion about this in discord:

The test here appears to be generating an incorrect BAL. We already correctly handle tracking the delegateTo address when the call happens here: https://github.com/status-im/nimbus-eth1/blob/155442ed5b785584d2dee241d80f5a98babdeb82/execution_chain/evm/message.nim#L48

If the sender balance check or other gas checks fail then the delegateTo address code doesn't need to be loaded. There is no reason to load the delegateTo address before these checks since it would be doing unneeded work and therefore the BAL shouldn't include these addresses in the access list. 

So in my opinion the tests should be updated to fix this, rather then fixing on our side.

kdeme:
Some further info on these test cases:

E.g.
`./tests/fixtures/eest_zkevm/blockchain_tests/for_amsterdam/amsterdam/eip8025_optional_proofs/witness_7702/witness_codes_delegated_eoa_insufficient_balance.json`

BAL in test vector:
```json
"blockAccessList": [
    {
        "address": "0x00000961ef480eb55e80d19ad83579a64c007002",
        "nonceChanges": [],
        "balanceChanges": [],
        "codeChanges": [],
        "storageChanges": [],
        "storageReads": [
            "0x00",
            "0x01",
            "0x02",
            "0x03"
        ]
    },
    {
        "address": "0x0000bbddc7ce488642fb579f8b00f3a590007251",
        "nonceChanges": [],
        "balanceChanges": [],
        "codeChanges": [],
        "storageChanges": [],
        "storageReads": [
            "0x00",
            "0x01",
            "0x02",
            "0x03"
        ]
    },
    {
        "address": "0x0000f90827f1c53a10cb7a02335b175320002935",
        "nonceChanges": [],
        "balanceChanges": [],
        "codeChanges": [],
        "storageChanges": [
            {
                "slot": "0x00",
                "slotChanges": [
                    {
                        "blockAccessIndex": "0x00",
                        "postValue": "0x4e9d359b986902614246a7f0e92088ce88f2bc409749eeedaf0ae65c4e42133d"
                    }
                ]
            }
        ],
        "storageReads": []
    },
    {
        "address": "0x000f3df6d732807ef1319fb7b8bb8522d0beac02",
        "nonceChanges": [],
        "balanceChanges": [],
        "codeChanges": [],
        "storageChanges": [
            {
                "slot": "0x0c",
                "slotChanges": [
                    {
                        "blockAccessIndex": "0x00",
                        "postValue": "0x0c"
                    }
                ]
            }
        ],
        "storageReads": [
            "0x200b"
        ]
    },
    {
        "address": "0x1ad9bc24818784172ff393bb6f89f094d4d2ca29",
        "nonceChanges": [
            {
                "blockAccessIndex": "0x01",
                "postNonce": "0x01"
            }
        ],
        "balanceChanges": [
            {
                "blockAccessIndex": "0x01",
                "postBalance": "0x3635c9adc5de9af222"
            }
        ],
        "codeChanges": [],
        "storageChanges": [],
        "storageReads": []
    },
    {
        "address": "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
        "nonceChanges": [],
        "balanceChanges": [
            {
                "blockAccessIndex": "0x01",
                "postBalance": "0x018429"
            }
        ],
        "codeChanges": [],
        "storageChanges": [],
        "storageReads": []
    },
    {
        "address": "0x8d52ba50da7800224b597a539cfbd67808b00da9",
        "nonceChanges": [],
        "balanceChanges": [],
        "codeChanges": [],
        "storageChanges": [
            {
                "slot": "0x00",
                "slotChanges": [
                    {
                        "blockAccessIndex": "0x01",
                        "postValue": "0x00"
                    }
                ]
            }
        ],
        "storageReads": []
    },
    {
        "address": "0x9edafae1466ee32e367d557061f62c142dbe720e",
        "nonceChanges": [],
        "balanceChanges": [],
        "codeChanges": [],
        "storageChanges": [],
        "storageReads": []
    },
    {
        "address": "0xfcdf2078adc64abb094ef21a7e091b0c6bc0d145",
        "nonceChanges": [],
        "balanceChanges": [],
        "codeChanges": [],
        "storageChanges": [],
        "storageReads": []
    }
],
```

`0xfcdf2078adc64abb094ef21a7e091b0c6bc0d145` here is the delegated EOA.

Looks like:
```json
    "0xfcdf2078adc64abb094ef21a7e091b0c6bc0d145": {
        "nonce": "0x01",
        "balance": "0x00",
        "code": "0xef01009edafae1466ee32e367d557061f62c142dbe720e",
        "storage": {}
    },
```

code field holds: `0xef0100 9edafae1466ee32e367d557061f62c142dbe720e` -> `0x9edafae1466ee32e367d557061f62c142dbe720e`  is the actual target with contract code to be run. However, that never happens due to insufficient balance. 

```json
   "0x9edafae1466ee32e367d557061f62c142dbe720e": {
          "nonce": "0x01",
          "balance": "0x00",
          "code": "0x60425000",
          "storage": {}
      },
```

So while `0xfcdf2078adc64abb094ef21a7e091b0c6bc0d145` does get accessed, `0x9edafae1466ee32e367d557061f62c142dbe720e` does not.

Our implementation just adds `0xfcdf2078adc64abb094ef21a7e091b0c6bc0d145` to the BAL (read only) but not `0x9edafae1466ee32e367d557061f62c142dbe720e`.

The reference implementation seems to (deducted from the test vectors at least, I didn't check the code yet) also add `0x9edafae1466ee32e367d557061f62c142dbe720e` to the BAL.

From a spec PoV this is somewhat ambiguously written down, it states:

> When a call target has an [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702) delegation, the target is accessed to resolve the delegation. If a delegation exists, the delegated address requires its own access_cost check before being accessed. If this check fails, the delegated address MUST NOT appear in the BAL, though the original call target is included (having been accessed to resolve the delegation).


So it does not explicitly state what to do in the scenario we encounter, and thus up for interpretation.
However, when reasoning about what the BAL is to be used for, it seems not useful to add it as it doesn't actually get accessed?

PR #4238, opened 2026-05-14
Optimistic state prefetch

bhartnett:
**Changes in this PR:**
- All lru caches in the database now use the thread safe ConcurrentLruCache. This is needed to support concurrent state lookups on the database.
- The ConcurrentLruCache supports a single threaded mode which avoids using the shards for improved performance when none of the parallel features are enabled.
- Manual cleanup of caches using dispose.
- Added optimistic state prefetching to speed up block execution by using background threads to execute the transactions out of order and in parallel. This has the effect of filling the caches before the main thread runs the block execution. One task is created for each transaction which does both the sender recovery and the prefetching. An atomic flag is used to signal to the main thread when the sender is available so that execution in the main thread can progress without waiting for the prefetching to complete. 
- Added a optimisticStatePrefetch debug parameter which is disabled by default.
- Updated the tests to enable the taskpool and set optimisticStatePrefetch to use the default setting.


bhartnett:
Imported 50000 blocks starting at block 24 million with stateroot checks disabled using 16 threads:
master - elapsed=1h29m44s
optimistic-state-prefetch (with prefetch enabled) - elapsed=1h4m50s

That's a ~28% improvement.

execution_chain/nimbus_execution_client.nim
```diff
@@ -385,6 +386,11 @@ proc main*(config = makeConfig(), nimbus = NimbusNode(nil)) {.noinline.} =
     kzg.loadTrustedSetup(config.trustedSetupFile.get(), 0).isOkOr:
       fatal "Cannot load KZG trusted setup from file", msg = error
       quit(QuitFailure)
+  else:
+    # Load eagerly to avoid race conditions - lazy kzg loading is not thread safe
+    loadTrustedSetupFromString(kzg.trustedSetup, 0).expect(
```
tersec: It would be better to use some precompute. See:
- https://github.com/status-im/nimbus-eth2/pull/8405

`nimbus-eth2` went with 7, due to memory reasons. Given `nimbus-eth1`'s memory consumption, 7 (48MiB RAM) is reasonable, but 8 (96MiB) might be too.

tests/test_generalstate_json.nim
```diff
@@ -20,10 +20,31 @@ import
...
   unittest2,
   stew/byteutils,
   results
 
+# Load eagerly to avoid race conditions - lazy kzg loading is not thread safe
+# and would race with parallel optimistic state prefetch workers. Loading on a
+# dedicated thread because parsing the trusted setup uses ~400 KB of stack
+# which exceeds the 1 MB ulimit set for `make test`.
+proc loadKzgSetup() {.thread.} =
+  discard loadTrustedSetupFromString(kzg.trustedSetup, 0)
```
tersec: See other comment about:
- https://github.com/status-im/nimbus-eth2/pull/8405
- https://github.com/status-im/nimbus-eth1/pull/4238#discussion_r3278853345

tests/eest/eest_helpers.nim
```diff
@@ -34,6 +36,12 @@ import
   ../../execution_chain/stateless/stateless_types,
   ../../hive_integration/engine_client
 
+# Load eagerly to avoid race conditions - lazy kzg loading is not thread safe
+# and would race with parallel optimistic state prefetch workers.
+loadTrustedSetupFromString(kzg.trustedSetup, 0).expect(
```
tersec: More KZG setup, worth considering the right amount of precomputation.

bhartnett:
Imported 500000 blocks starting at block 15 million with stateroot checks disabled using 16 threads:
master - elapsed=4h21m20s
optimistic-state-prefetch (with prefetch enabled) - elapsed=4h10m2s

In a longer run there is not as much improvement. I guess this is expected because once the caches are warmed up the prefetching doesn't have as much impact.

With that in mind, having better performance when the caches are empty will still be beneficial for various scenarios such as the benchmarking tests.

execution_chain/nimbus_execution_client.nim
bhartnett: Sure, I'll try out 8 and see how it goes in a separate PR. 96MiB is not much for the EL.

arnetheduck:
> With that in mind, having better performance when the caches are empty will still be beneficial for various scenarios such as the benchmarking tests.

16 threads vs 1 thread and only a tiny improvement means that either the threads are not doing anything or the _total_ amount of work went up too much - did you look into which scenario it is?

Regarding pre-fetching in general, I suspect that a design with a few (2? 3?) worker threads could work just as well since otherwise, there's a good chance the I/O pipeline gets swamped and you don't get more benefit.

arnetheduck:
> All lru caches in the database now use the thread safe ConcurrentLruCache. This is needed to support concurrent state lookups on the database.

what's the cost of this change?

bhartnett:
> 16 threads vs 1 thread and only a tiny improvement means that either the threads are not doing anything or the _total_ amount of work went up too much - did you look into which scenario it is?

There is not much improvement in the longer run because once the caches are filled, the prefetching doesn't help as much because the caches already contain the state (with a high probability since our caches are rather large). The scenario that improves the most is when the state is very large and the caches are cold. I guess once the caches are warm we could turn off the prefetching in order to reduce wasted work but for larger machines that have the available cores probably no point doing that. I also think that this sharded lru cache design is not optimal (there are better designs) but it still performs ok for now. An improved concurrent LRU that is mostly lock free would improve this further.

Also note the background for this is that I need the database layers to be thread safe so I can implement parallel execution using BALs, that was the original driver for this change.

> 
> Regarding pre-fetching in general, I suspect that a design with a few (2? 3?) worker threads could work just as well since otherwise, there's a good chance the I/O pipeline gets swamped and you don't get more benefit.

Ok that's worth a try. Note that this is the same state prefetching algorithm used by Geth where a task is created per transaction and all tasks are submitted at once. The number of cores used is 4/5 of the number available on the machine. It was also a natural fit to extend the existing sender recovery tasks to do the prefetching because I found that creating separate prefetching tasks after the sending recover was much slower and that lead to creating a task per transaction.

bhartnett:
> > All lru caches in the database now use the thread safe ConcurrentLruCache. This is needed to support concurrent state lookups on the database.
> 
> what's the cost of this change?

I'm assuming you mean how much slower is the ConcurrentLruCache compared to the minilru cache for the single threaded use case?

Well I've implemented a single threaded mode in the ConcurrentLruCache which uses very similar code to the original minilru. Basically its a single shard without the locking overhead. This means we can still support single threaded mode for Nimbus EL and the performance should be very close to the existing minilru.

This benchmark compares the performance of each cache using a single thread:
```
LruCache vs ConcurrentLruCache single-threaded comparison 
  capacity=100000, ops=100000000
  benchmark                            elapsed(s)          ops/s      ns/op
  LruCache put                             3.8721      25825495.       38.7
  ConcurrentLruCache put (no lock)         4.1849      23895672.       41.8
  ConcurrentLruCache put                  10.1778       9825332.      101.8
  LruCache get                             3.8302      26108437.       38.3
  ConcurrentLruCache get (no lock)         4.6089      21696922.       46.1
  ConcurrentLruCache get                   6.1184      16344222.       61.2
  LruCache peek                            3.2374      30888718.       32.4
  ConcurrentLruCache peek (no lock)        3.8833      25751192.       38.8
  ConcurrentLruCache peek                  6.0639      16491168.       60.6
. (46.0s)
```

PR #4240, opened 2026-05-14
EIP-7928 BAL: Fix for EIP-7702 account call

jangko:
- Low balance, both target and delegated target in BAL.
- Message call delegated, no more multiple parseDelegation, only one call.
- OOG Case, only `codeAddress` in BAL.

Tests for these cases are in tests-bal@v7.1.1.
But this commit can be safely extracted from bal-devnet-7 branch because it does not depends on EIP-8037.

### DIFF
--- a/execution_chain/db/ledger.nim
+++ b/execution_chain/db/ledger.nim
@@ -18,7 +18,6 @@ import
   eth/common/[addresses, hashes],
   ../utils/[mergeutils, utils],
   ../evm/code_bytes,
-  ../core/eip7702,
   ../constants,
   ./[access_list as ac_access_list, core_db, storage_types],
   ./aristo/aristo_blobify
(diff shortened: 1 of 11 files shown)


kdeme (commented):
This fixes the issue discovered in this zkevm tests and trying to resolve in https://github.com/status-im/nimbus-eth1/pull/4210, correct?

jangko:
right, it is the ugliest part of EIP 7928, at least for nimbus-eth1 implementation.
The related eels implementation is here:
https://github.com/ethereum/execution-specs/blob/c6b7f1704db7c2e224ecaefabbef2973f0c75acb/src/ethereum/forks/amsterdam/vm/instructions/system.py#L422-L423

The state is accessed using get_code before the sender balance check. While our implementation access the state after balance check.

kdeme:
Yes, so we would benefit of having https://github.com/ethereum/execution-specs/pull/2473 merged, as that would clean things up. But it also just seems like the more correct way to handle this.

PR #4243, opened 2026-05-14
fix network argument to be a string and not a list

jakubgs:
Tried using TOML config format and encountered this unexpected error:
```
Failed to load secondary sources: 'nimbus-execution-client.toml(17, 11) expected '"["''
```

### DIFF
--- a/execution_chain/conf.nim
+++ b/execution_chain/conf.nim
@@ -146,12 +146,11 @@ type
         "- mainnet/1       : Ethereum main network\n" &
         "- sepolia/11155111: Testnet for smart contract testing\n" &
         "- hoodi/560048    : Testnet for staking and hard forks\n" &
-        "- custom/path     : /path/to/genesis-or-network-configuration.json\n" &
-        "Both --network: name/path --network:id can be set at the same time to override network id number"
-      defaultValue: @[] # the default value is set in makeConfig
+        "- custom/path     : /path/to/genesis-or-network-configuration.json"
+      defaultValue: none(string) # the default value is set in makeConfig
       defaultValueDesc: "mainnet(1)"
       abbr: "i"
-      name: "network" .}: seq[string]
+      name: "network" .}: Option[string]
 
     customNetwork {.
       ignore
@@ -692,7 +691,7 @@ proc parseNetworkParams(network: string): (NetworkParams, bool) =
     (params, true)
 
 proc processNetworkParamsAndNetworkId(config: var ExecutionClientConf) =
-  if config.network.len == 0 and config.customNetwork.isNone:
+  if config.network.isNone and config.customNetwork.isNone:
     # Default value if none is set
     config.networkId = MainNet
     config.networkParams = networkParams(MainNet)
@@ -703,16 +702,11 @@ proc processNetworkParamsAndNetworkId(config: var ExecutionClientConf) =
     id: Opt[NetworkId]
     simulatedCustomNetwork = false
 
-  for network in config.network:
+  if config.network.isSome:
+    let network = config.network.get
     if decOrHex(network):
-      if id.isSome:
-        warn "Network ID already set, ignore new value", id=network
-        continue
       id = Opt.some parseNetworkId(network)
     else:
-      if params.isSome:
-        warn "Network configuration already set, ignore new value", network
-        continue
       let (parsedParams, custom) = parseNetworkParams(network)
       params = Opt.some parsedParams
       # Simulate --custom-network while it is still not disabled.
--- a/execution_chain/conf.nim
+++ b/execution_chain/conf.nim
@@ -860,6 +860,15 @@ func jwtSecretOpt*(config: ExecutionClientConf): Opt[InputFile] =
   else:
     Opt.none InputFile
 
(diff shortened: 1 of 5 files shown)


vendor/nim-zlib
tersec, line 1: ?

jakubgs: Always forget to run submodule update.

jangko:
The reason it is a list is nimbus can accept:
1. `--network:hoodi`
2. `--network:1`
3. `--network:hoodi --network:123`
4. `--network:"/path/to/genesis" --network:123`

The 3rd & 4th version allow nimbus to override the numerical network id.
What need to be fixed is the TOML parser, but right now nim-toml-serialization is less flexible and missing a crucial feature to parse dynamic field type like nim-json-serialization.

jakubgs:
If overriding network ID is required why not have separate `--network-id` flag to make functionality explicit and flag parsing simpler?

Also, why would overriding the network ID be desirable in the first place?
And why providing genesis file shouldn't be a separate flag as well?

jangko:
In the past nimbus have `--network-id` and `--custom-network`, they do not make parsing simpler. That's why `--network-id` is removed, and `--custom-network` been deprecated.

And why overriding network ID is needed? Looks like geth genesis file format is very popular among client implementer. But it does not contains network ID. It force chain ID to be used as network ID which is not always true. override network ID is a clean mechanism when client user want a different network ID from chain ID. (So it's an external/historical complexity)

And why providing genesis file shouldn't be a separate flag as well? both genesis file and built in named network are the same thing. You can choose a built in named network such as `hoodi` or a `genesis.json of hoodi` and it will be the same configuration for your client.

For now, isn't using bracket for your `network` field in your TOML file is simpler than try to fix the parser?

jakubgs:
>In the past nimbus have --network-id and --custom-network, they do not make parsing simpler. 

That's not an argument, that's simply an assertion. Overloading flags usually leads to obscure, unexpected, and more complex behavior and is genrally bad for end user.

>And why overriding network ID is needed? Looks like geth genesis file format is very popular among client implementer.

Irrelevant.

>It force chain ID to be used as network ID which is not always true. override network ID is a clean mechanism when client user want a different network ID from chain ID. (So it's an external/historical complexity)

It's not clean, nor is it common. Here's most commonly used EL clients:

* Geth uses [`--${chain}`](https://geth.ethereum.org/docs/fundamentals/command-line-options), [`--networkid`](https://geth.ethereum.org/docs/fundamentals/private-network#choosing-network-id), and [`init` keyword](https://geth.ethereum.org/docs/fundamentals/private-network#initializing-geth-database) as stated in docs.
* Erigon uses [`--chain`](https://docs.erigon.tech/fundamentals/supported-networks), and [`--networkid`](https://docs.erigon.tech/fundamentals/configuring-erigon#general-options) as stated in docs.
* Besu uses [`--network`](https://besu.hyperledger.org/public-networks/reference/cli/options#network), [`--network-id`](https://besu.hyperledger.org/public-networks/reference/cli/options#network-id), and [`--genesis-file`](https://besu.hyperledger.org/public-networks/reference/cli/options#genesis-file) as stated in docs.
* Reth [uses `--chain` and separate `--network-id` as stated in dpcs](https://github.com/paradigmxyz/reth/blob/v2.2.0/docs/vocs/docs/pages/cli/reth/node.mdx)

The only outlier is Nethermind which Uses `--config` which takes names of config files for chains it provides or path:
https://github.com/NethermindEth/nethermind/tree/master/src/Nethermind/Nethermind.Runner/configs

As far as I can tell you are arguing here against the convention established by other clients. Not sure why.

It also goes against what the beacon chain node does, where the argument is just a string:
https://github.com/status-im/nimbus-eth2/blob/6a475234ba0f107f734d61d405c4293434ae56a1/beacon_chain/conf.nim#L1147-L1150

And genesis files are provided by separate flags:
https://github.com/status-im/nimbus-eth2/blob/6a475234ba0f107f734d61d405c4293434ae56a1/beacon_chain/conf.nim#L394-L400

I'm happy to refactor this PR but your arguments make no sense to me.

jangko:
You only want `--network` can accept string as argument right? And the CLI `--network` accept string as argument right? So, it works, but maybe not to your liking. You don't need `--network-id` nor `--genesis-file` right?
What need to be fixed is the TOML parser. If that is too complex, and you don't like to do it, fine. You can open an issue here or in nim-toml-serialization repo. Maybe someone will try to fix it.

> As far as I can tell you are arguing here against the convention established by other clients. Not sure why.

It's a convention, not a law. What I see is `--network:mainnet`, `--network:1`, `--network:/path/mainnet.json`, they all end up as the same thing inside the client, why have to use separate flags. I don't know what kind of bad for end user. I don't need to remember many separate flags when configuring the same thing. What I see is the convention does not make sense while we can make it simpler.

jakubgs:
That error was how I discovered this. It's not the only issue. Overloading flags is bad in general, I don't get why you insist on overloading flag meaning.

>they all end up as the same thing inside the client

Nobody using the software cares about what happens inside, they care about the interface and the outcome.

Having flags with overloaded meaning usually leads to confusion and mistakes. But I don't care enough about your software to keep arguing about it with you .

PR #4245, opened 2026-05-15
Remove BPO forks from EVMFork

jangko:
BPO forks are not used internally by EVM interpreter. They are pretty much similar to glacier forks.
If they stay in the EVMFork, it will bloat the VM interpreter dispatch generator with no benefits.

This PR removes those BPO forks from EVMFork but still allow both current correct evmFork and hardFork accessible by those who need them.


tersec:
Merge conflict with `transaction/call_common.nim`.

PR #4247, opened 2026-05-15
Enable parallel state root computation by default

bhartnett:
I've been testing the parallel state root computation for a while now and it has been running well so far. Perhaps now is a good time to enable it by default.

It can cause an increase in memory usage in some scenarios and in the worst case goes from around 16G when the caches are full to around 24G.

The speed up is somewhere between 2-4x when using 16 cores/threads. For the large stateroot computation on startup the speed up is around 2x because writing to the database is the bottleneck in that case but when most keys are in memory we can get 3-4x speedup.

### DIFF
--- a/execution_chain/conf.nim
+++ b/execution_chain/conf.nim
@@ -354,7 +354,7 @@ type
 
     parallelStateRootComputation* {.
       hidden
-      defaultValue: false
+      defaultValue: true
       desc: "Compute state root in parallel using multiple threads"
       name: "debug-parallel-state-root".}: bool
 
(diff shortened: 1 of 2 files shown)


execution_chain/conf.nim
```diff
@@ -354,7 +354,7 @@ type
 
     parallelStateRootComputation* {.
       hidden
-      defaultValue: false
+      defaultValue: true
```
bhartnett, line 357: @advaita-saha How has your node been going with this enabled so far? Do you think we should make this enabled by default?

PR #4249, opened 2026-05-18
Remove resetPayload from ConcurrentLruCache

bhartnett:
The resetPayload function is removed from the LruCache type because only supportsCopyMem types are allowed for the generic parameters and so this function is a no-op.

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -123,6 +123,10 @@ switch("passL", "-fno-omit-frame-pointer")
 --styleCheck:usages
 --styleCheck:error
 
+# Disable ABI warning: 
+# 'the ABI for passing parameters with 64-byte alignment has changed in GCC 4.6'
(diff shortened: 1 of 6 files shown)


arnetheduck:
`=destroy` does not have a deterministic release point with `--mm:refc`, in fact it might linger indefinitely - it has also been known to cause crashes but these were never investigated - instead,  we simply avoided =destroy. If we switch to orc, it becomes more predictable where these run.

bhartnett:
> `=destroy` does not have a deterministic release point with `--mm:refc`, in fact it might linger indefinitely - it has also been known to cause crashes but these were never investigated - instead, we simply avoided =destroy. If we switch to orc, it becomes more predictable where these run.

I see, ok will remove. Can just go with the manual cleanup for now.

PR #4271, opened 2026-05-27
fix hashIndexKey high byte always zero due to shl/shr mistake

kdeme:
hashIndexKey() used `shl 8` instead of `shr 8` to extract the high byte of the index, making it always 0. This reduced the key space to 256 causing silent data corruption for any block with 256 or more transactions or receipts. The high index entries overwrote the low index ones on write and the iterator never terminated early.

On import from era1/era this could only be noticed when using the debug-store-bodies and/or receipts flags. But when following the chain normally it will also occur of course.

The fix changes the storage key layout, but since existing blocks for with >= 256 txs are already corrupt anyhow, the change does not make it worse for that data. And it will actually work for new data.

Nodes will need to resync to have actual valid data for affected blocks.

### DIFF
--- a/execution_chain/db/storage_types.nim
+++ b/execution_chain/db/storage_types.nim
@@ -109,7 +109,7 @@ func fcuKey*(u: uint64): DbKey {.inline.} =
 func hashIndexKey*(hash: Hash32, index: uint16): HashIndexKey =
   result[0..31] = hash.data
   result[32] = byte(index and 0xFF)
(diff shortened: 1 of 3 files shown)


kdeme:
Note that the storage layout change should not be causing any issues (Defects) aside from returning already invalid data.

FYI: I've come to hit this issue when verifying generated `ere` files from our EL database.

PR #4279, opened 2026-06-01
BAL state prefetching (batch IO)

bhartnett:
This implements state prefetching using the read (and storage write) locations in the block access list. The prefetching runs on background threads, similar to the optimistic state prefetching. 

The bal prefetching is disabled by default for now and can be enabled using the  `--debug-bal-state-prefetch=true` parameter. 

The prefetching reads each account when no storage reads or writes exist in the BAL since storage reads would implicitly read the account anyway. The storage writes are sorted by the first block access index and then read in that order so the state of earlier accessed slots will be available first. 

The number of workers (level of parallelism) is configurable using the `--debug-bal-state-prefetch-workers=<number>` parameter. The default value for this parameter will be tuned later once we implement parallel execution using BALs (coming soon).

### DIFF
--- a/execution_chain/block_access_list/block_access_list_validation.nim
+++ b/execution_chain/block_access_list/block_access_list_validation.nim
@@ -81,6 +81,8 @@ func validate*(
     var changedSlots: HashSet[StorageKey]
     for slotChanges in accountChanges.storageChanges:
       changedSlots.incl(slotChanges.slot)
+      if slotChanges.changes.len == 0:
+        return err("Each slot in storage changes must have at least one change")
       if not slotChanges.changes.isSorted(balIndexCmpTreatEqualAsGreater):
         return err("Slot changes should be unique and sorted by blockAccessIndex")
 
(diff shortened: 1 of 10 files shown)


bhartnett:
In this https://github.com/status-im/nimbus-eth1/tree/bal-performance-testing branch I've added some code to support performance testing with BALs in the block import even for pre-amsterdam blocks. 

Beforehand I generated a file containing the BALs for a range of blocks by running the block import using the `--debug-bal-sidecar-write=filename` parameter. 
Then after that ran the performance tests with `--debug-bal-sidecar-read=filename` parameter. All tests read in the BAL from the file even if not used based on the settings so that the comparison is accurate.

Ran the test starting at block 15 million, imported 10K blocks with stateroot disabled, running with 16 taskpool threads.

With all prefetching (both optimistic and bal) disabled:
elapsed=7m32s890ms

With bal prefetching enabled with default workers = 16 and optimistic prefetch disabled:
elapsed=4m41s899ms

With bal prefetching enabled with workers = 8 and optimistic prefetch disabled:
elapsed=4m23s613ms

With bal prefetching disabled and optimistic prefetch enabled:
elapsed=5m11s394ms

Once parallel execution is implemented the number of bal prefetch workers can be tuned to a better default.

PR #4280, opened 2026-06-01
Add historical_hashes_accumulator export and era1 export

kdeme:
Add historical_hashes_accumulator export and era1 export to nimbus_history_exporter tool. This is functionality ported from the Portal eth_data_exporter tool to this tool, where instead of JSON-RPC the local nimbus database is used as data source.


kdeme:
Will eventually remove the old version of this in eth_data_exporter.

The accumulator export is required for being able to export this in Sepolia, and then include it in binaries like is donef or mainnet.

era1 export will be rather legacy once `ere` gets adopted but for the time being it's fine to keep.

PR #4290, opened 2026-06-02
Add EreDB implementation

kdeme:
EreDB implementation to be used as drop-in replacement of Era1DB but with post merge usage.

Maybe a little silly to already add this without usage. But have upcoming draft PR to see/show changes compared to era1/era. 

And once files are generated can also use this for Portal bridge.


kdeme:
Will go ahead and merge this already as https://github.com/status-im/nimbus-eth1/pull/4291 has shown this to be working fine as replacement for era1DB

PR #4291, opened 2026-06-02
Replace era1/era based import with ere based import

kdeme:
to be tested...


yorickdowne:
For your tests, note that while ` exec "build/nimbus_execution_client import --data-dir:" & tmp & " --ere-dir:tests/replay --max-blocks:1"` should work, the unified binary doesn't have an "import" command right now: That also should be tested. `nimbus executionClient import <etc>`

kdeme:
> For your tests, note that while ` exec "build/nimbus_execution_client import --data-dir:" & tmp & " --ere-dir:tests/replay --max-blocks:1"` should work, the unified binary doesn't have an "import" command right now: That also should be tested. `nimbus executionClient import <etc>`

Thanks for pointing this out. I hadn't looked yet into the `nimbus` options for this.

I've now checked and the unified binary takes over the subcommands directly. 

So the unified binary version of this is:
`exec "./build/nimbus import --data-dir:/data/dir/nimbus --ere-dir:/data/dir/nimbus/ere --max-blocks:1`

I assume it is intended to be like this as opposed to the `executionClient import` , @arnetheduck ?

I will add this to the import test case in CI here.

kdeme:
Note that this PR is still in draft because:
1. `ere` files have not been generated yet for the full history (mainnet / sepolia) (issue: https://github.com/status-im/infra-nimbus/issues/283)
2. We need to decide still how to deal with `era1`+`era` versus `ere`:
  * Swap support in EL from `era1` + `era` to `ere` 
  * Support both temporary, but eventually remove `era1` + `era` (era of course remains supported on CL side, and we even need `era` to support proofs in `ere` export)
  * Support both forever...

yorickdowne:
Note that pandaops ends their files in `.erae`. Not sure whether you check file ending, but if you do, please support both `.ere` and `.erae` ending. See https://ethpandaops.io/data/history/?history-erae-network=mainnet

I am not sure whether what pandaops serves is complete. It looks fairly complete, at least.

kdeme:
> Note that pandaops ends their files in `.erae`. Not sure whether you check file ending, but if you do, please support both `.ere` and `.erae` ending. See https://ethpandaops.io/data/history/?history-erae-network=mainnet
> 
> I am not sure whether what pandaops serves is complete. It looks fairly complete, at least.

This is because those files were generated before the `ere`format specifications where complete and got merged.

The proper extension name now is `ere`. Also, profiles were added in the name, and those files are without proofs and should thus have the `noproofs` profile which they currently don't have (see also ere format specs: https://github.com/eth-clients/e2store-format-specs/blob/ca2523a6420d64336000f5607c0b59df1a08c83b/formats/ere.md#file-name).

So if you want to already test the nimbus import with those files you have to all rename them like `mainnet-00000-a6860fef.erae` -> `mainnet-00000-a6860fef-noproofs.ere`.

yorickdowne:
Oh that’s good to know. I’ll rip out the EraE stuff from the Geth PR

arnetheduck:
> Support both temporary, but eventually remove

+1 this - we also need to specify `.erc` which is the consensus data only (that pairs well with `.ere` (and `.erb` for blobs) - this is specially important for gloas since execution data travels separately - once we have `.erc` for some overlap time, we can drop era/era1 in both cl and el.

kdeme:
Closing this for https://github.com/status-im/nimbus-eth1/pull/4301. Same implementation but keeping era1/era around for now.

PR #4294, opened 2026-06-03
nimbus_history_exporter: Add hoodi support

kdeme:
Some trickiness due to hoodi genesis not being baked in and it being a PoS only network (with GENESIS_DELAY > 0)

### DIFF
--- a/tools/nimbus_history_exporter/beacon_proof_builder.nim
+++ b/tools/nimbus_history_exporter/beacon_proof_builder.nim
@@ -17,10 +17,11 @@
 
 import
   std/times,
+  chronos,
   results,
   ssz_serialization,
   beacon_chain/era_db,
-  beacon_chain/networking/network_metadata,
+  beacon_chain/networking/[network_metadata, network_metadata_downloads],
   beacon_chain/spec/[forks, eth2_ssz_serialization],
   beacon_chain/beacon_clock,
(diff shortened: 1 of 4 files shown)


tools/nimbus_history_exporter/beacon_proof_builder.nim
```diff
@@ -96,20 +97,31 @@ proc init*(
...
+              BeaconStateHeader,
+            )
+          except SerializationError as exc:
+            return err("Cannot decode genesis state header: " & exc.msg)
+        (genesisHeader.genesis_validators_root, genesisHeader.genesis_time)
+      else:
+        # For networks where genesis is not baked in (e.g. hoodi), download it.
+        let genesisState =
+          try:
+            waitFor fetchGenesisState(metadata)
```
kdeme, line 117: Note that one could also get this from the `era` file itself, perhaps worth the change considering you have to make sure those files are valid/trusted.

kdeme: There is actually a `fetchGenesisState` version that prioritizes grabbing state from era before downloading: https://github.com/status-im/nimbus-eth2/blob/6fb05f36804d53c2e8e014cfeeea8ad7996a5efe/beacon_chain/nimbus_beacon_node.nim#L111

Need to extract and/or make that public, so that could be an improvement here.

Will already merge this version though as it works fine.

PR #4299, opened 2026-06-04
Don't load code for delegation when the address is a precompile

bhartnett:
Changes in this PR:
- Skip loading of code for delegation when the address is a pre-compile.
- Optimise performance of getPrecompile.
- Use isPrecompile template at call sites.
- Don't load precompiles in the bal state prefetch workers.

### DIFF
--- a/execution_chain/core/executor/process_block.nim
+++ b/execution_chain/core/executor/process_block.nim
@@ -20,6 +20,7 @@ import
   ../../transaction,
   ../../evm/state,
   ../../evm/types,
+  ../../evm/precompiles,
   ../../evm/interpreter/gas_costs,
   ../../block_access_list/block_access_list_validation,
   ../../concurrency/utils,
@@ -56,6 +57,7 @@ when compileOption("threads"):
       nextIndex: Atomic[int]
       balPtr: ptr BlockAccessList
       txFrame: CoreDbTxRef
+      fork: EVMFork
 
   proc recoverAndPrefetchTask(
       e: ptr Entry, ctx: ptr PrefetchCtx, tx: ptr Transaction): bool {.nimcall.} =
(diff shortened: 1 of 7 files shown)


jangko:
https://github.com/status-im/nimbus-eth1/blob/1d0df8c5b9e3d8f95919585d2c7df2e0e017891d/execution_chain/evm/interpreter/op_handlers/oph_call.nim#L190

https://github.com/status-im/nimbus-eth1/blob/1d0df8c5b9e3d8f95919585d2c7df2e0e017891d/execution_chain/evm/message.nim#L32

https://github.com/status-im/nimbus-eth1/blob/1d0df8c5b9e3d8f95919585d2c7df2e0e017891d/execution_chain/evm/interpreter_dispatch.nim#L224

Considering the call to `getPrecompile` involving a loop inside the function, I think it's time to reduce the call to `getPrecompile` and use flag maybe?

bhartnett:
> Considering the call to `getPrecompile` involving a loop inside the function, I think it's time to reduce the call to `getPrecompile` and use flag maybe?

I'm currently working on a change to precompute a per fork table of all the precompiles so that we don't loop here. What do you mean by flag?

jangko:
> What do you mean by flag?

`getPrecompile` in : `oph_call.nim.gasCallDelegate`, `oph_call.nim.getCallCode`, and `interpreter_dispatch.nim.executeOpcodes` all evaluating the same address when a child Computation is executed. Why not evaluate once and use `MsgFlags` enum to flag it as a precompile.

The same goes for top level message call.

bhartnett:
> `getPrecompile` in : `oph_call.nim.gasCallDelegate`, `oph_call.nim.getCallCode`, and `interpreter_dispatch.nim.executeOpcodes` all evaluating the same address when a child Computation is executed. Why not evaluate once and use `MsgFlags` enum to flag it as a precompile.
> 
> The same goes for top level message call.

Updated as suggested.

bhartnett:
Performance testing results:

Imported 10K blocks starting from around block 24 million.

master - 
**elapsed=18m51s135ms**

precompiles branch - 
**elapsed=15m37s970ms**

PR #4301, opened 2026-06-04
Add ere import support while keeping era1/era as fallback option

kdeme:
ere takes priority: if .ere files are found in --ere-dir, they are used
for import. If not import falls back to the era1/era path.

This is alternative for https://github.com/status-im/nimbus-eth1/pull/4291 which removed era1/era support completely.

For now support both until `ere` has been generated for full history + better tested + nimbus.guide updated + ...


kdeme:
> ere takes priority: if .ere files are found in --ere-dir, they are used
> for import. If not import falls back to the era1/era path.

Note that this does allow for a sort of potential misuse in this dual support. That is if the user has incomplete set of `ere` files it will import those and stop. Not looking at potentially a more complete era1/era set of files. That should hopefully be clear enough from the log output though, what is being used for the import.
So I think this is fine and warrants for not adding further complexity, especially considering the era1/era solution will eventually disappear. 

I could perhaps improve it slightly by letting explicitly set era1/era cli flags overrule the ere option.
For example something like this:

```patch
diff --git a/execution_chain/nimbus_import.nim b/execution_chain/nimbus_import.nim
index a915c4c8e..8ba61bfa2 100644
--- a/execution_chain/nimbus_import.nim
+++ b/execution_chain/nimbus_import.nim
@@ -197,10 +197,17 @@ proc importBlocks*(config: ExecutionClientConf, com: CommonRef) =
 
     time1 = time2
 
-  # ere takes priority: use it if files are present in the ere directory
-  # (either the explicit --ere-dir or the default <data-dir>/ere).
+  # ere takes priority unless the user explicitly set --era1-dir or --era-dir
+  # without also setting --ere-dir, in which case era1/era is used directly.
+  let tryEre =
+    config.ereDirFlag.isSome or
+    (config.era1DirFlag.isNone and config.eraDirFlag.isNone)
   let networkName = config.networkId.name
-  let ereResult = EreDB.init(config.ereDir, networkName, mergeBlockNumber(config.networkId))
+  let ereResult =
+    if tryEre:
+      EreDB.init(config.ereDir, networkName, mergeBlockNumber(config.networkId))
+    else:
+      Result[EreDB, string].err("era1/era explicitly requested")
   if ereResult.isOk():
     let db = ereResult.get()
     defer:
@@ -222,9 +229,9 @@ proc importBlocks*(config: ExecutionClientConf, com: CommonRef) =
       persistBlock()
       checkpoint()
   else:
-    # Fall back to era1/era import, legacy way to be removed eventually
-    notice "Could not open ere database, falling back to era1/era import",
-      ereDir = config.ereDir, networkName, error = ereResult.error
+    if tryEre:
+      notice "Could not open ere database, falling back to era1/era import",
+        ereDir = config.ereDir, networkName, error = ereResult.error
 
     let (cfg, genesis_validators_root, lastEra1Block, firstSlotAfterMerge) =
       getMetadataLegacy(config.networkId)
```

Not sure if it's worth the ugliness.

PR #4315, opened 2026-06-09
sync to block head with unknown header

advaita-saha:


execution_chain/beacon/api_handler/api_forkchoice.nim
```diff
@@ -107,13 +107,18 @@ proc forkchoiceUpdated*(ben: BeaconEngineRef,
...
     let header = chain.quarantine.getHeader(headHash).valueOr:
-      warn "Forkchoice requested unknown head",
-        hash = headHash.short
+      info "Forkchoice requested sync to unknown head",
+        hash = headHash.short,
+        finHash = update.finalizedBlockHash.short,
+        safe = update.safeBlockHash.short,
+        base = chain.baseNumber,
+        pendingFCU = chain.pendingFCU.short
+      com.headerTargetRequest(headHash, update.finalizedBlockHash)
```
jangko, line 121: What happen if the same `headHash` requested again to the syncer when there is already a previous request waiting for completion, can the syncer handle it?

advaita-saha: yes, the syncer handles repeated requests safely

execution_chain/sync/beacon.nim
mjfh, line 1: Setting call back functions here is probably not the right place and breaks consistency.

All  call back functions are installed on initialisation in `beacon/worker/start_stop.setupServices()`.  This includes other trigger like `chain.com.beaconSyncerProgress()`.

mjfh:
And the destructor for `chain.com.headerTargetRequest` goes into `beacon/worker/start_stop.destroyServices()`

PR #4322, opened 2026-06-12
proxy: fix closure cycles in evm

chirag-parmar:


execution_chain/evm/types.nim
```diff
@@ -80,8 +80,9 @@ type
     savePoint*:             LedgerSpRef
     instr*:                 Op
     opIndex*:               int
-    parent*, child*:        Computation
-    continuation*:          proc(): EvmResultVoid {.gcsafe, raises: [].}
+    parent* {.cursor.}:     Computation  # non-owning back pointer
+    child*:                 Computation  # owning front pointer
+    continuation*:          proc(c: Computation): EvmResultVoid {.gcsafe, raises: [].}
```
jangko, line 85: might as well add `nimcall` pragma to the pragma list to ensure there is no hidden closure.

see https://github.com/status-im/nimbus-eth1/pull/2282

chirag-parmar: two other parameters are captured `memPos` and `memLen` in `execSubCall`. Can try to refactor and try to fit them in without requiring closures, but its better if we pair this refactor with the overall refactor of evm.

config.nims
```diff
@@ -51,7 +51,7 @@ if defined(release) and not defined(disableLTO) and not defined(windows):
   extend "gcc.options.linker", " -flto=auto -Wno-stringop-overflow -Wno-stringop-overread -finline-limit=100000"  # https://github.com/nim-lang/Nim/issues/21595
   extend "gcc.cpp.options.linker", " -flto=auto -Wno-stringop-overflow -Wno-stringop-overread -finline-limit=100000"
 
-  if defined(macosx):
+  if defined(macosx) and not defined(emscripten):
```
tersec, line 54: Do non-macos LTO settings work with enscripten? It's a little surprising to juxtapose these two conditions, since emscripten works across other platforms too.

chirag-parmar: Apologies for the late reply. This link was attached to the line https://clang.llvm.org/docs/CommandGuide/clang.html#cmdoption-flto and the `object_path_lto` is only required for debug symbols. The `lto` options for every platform work and are very much needed for wasm compilation, without it WASM compiles for too long (mostly because of `presets.nim`, not sure why)

chirag-parmar: that option doesn't work with `wasm-ld` btw so the guard is needed

PR #4348, opened 2026-06-18
Activate stateless testing default in eest tests

kdeme:
+ remove eest stateless specific test.

Now that all tests pass (except for a stateless specific witness optimisation test), it makes more sense to just run them default with statelessEnabled on.

This way any new test / test suite / fork gets immediatly tested also for stateless execution and witness generation.

And it should also improve CI time as we have to only compile and
 run these tests once (albeit a little slower).


kdeme:
Closing this.

After discussing it with @bhartnett, it's better to keep separate runs of this as in the future we might not be supporting the witness generation with parallel tx execution using BALs.

PR #4350, opened 2026-06-19
Ledger block access list overlay

bhartnett:
For parallel execution using BALs, when executing each transaction in a separate context/thread we need to read the prestate from both the database and the writes in the BAL. The BAL overlay adds code to binary search the writes in the BAL based on the current BAL index.


jangko:
Parallel execution only available for transactions with disjoint set of state access. transactions with overlap state access cannot be executed together, they must be separated into different batch/group.

Because of that, before execution, the BAL will be analyzed and accounts will be grouped too. Therefore I think there is no need for BAL overlay at all, we can inject the the accounts into the ledger one disjoint group after another.

bhartnett:
> Parallel execution only available for transactions with disjoint set of state access. transactions with overlap state access cannot be executed together, they must be separated into different batch/group.

I'm not fully following what you mean here. Having the BAL enables full parallel executions of transactions without having to care about the write and read dependencies between each transaction so when parallel execution is enabled we will execute every transaction in a separate context.

> Because of that, before execution, the BAL will be analyzed and accounts will be grouped too. Therefore I think there is no need for BAL overlay at all, we can inject the the accounts into the ledger one disjoint group after another.

I don't think we need to go down the path of analyzing the BAL in order to try and batch sets of transactions together. Even so I don't think you can do this because the BAL doesn't give you the reads for each bal index, only the block level reads. You would also likely find that the overhead of the additional analyzing work slows down the critical path of starting the parallel execution.

jangko:
>  the BAL enables full parallel executions of transactions without having to care about the write and read dependencies between each transaction.

That is even better, so why need a BAL overlay?, isn't injecting prestate into ledger will be much simpler ?

bhartnett:
> That is even better, so why need a BAL overlay?, isn't injecting prestate into ledger will be much simpler ?

Yes maybe simpler to implement but worse for performance. It is possible to write the prestate of each transaction into the ledger before execution but doing so may be wasted effort because those state locations written may not even be read by the transaction. It's better to do the reads on demand (lazy read) so that we only do the work when required. We benefit from the ledger caches here as well so we only need to read each account/slot from the BAL at most once.

jangko:
Isn't balAccessIndex tell us which transaction use which accounts? If we are going to execute each transaction in their own context/thread, wouldn't it sufficient only inject prestate needed by that transaction, no wasted accounts.
They are not sharing the same ledger right?

For example, we have a list of excution context, using balAccessIndex, we populate the ledger of each context, no need for binary search.

bhartnett:
> Isn't balAccessIndex tell us which transaction use which accounts? 

Only partially. The writes are indexed by BAL index but not the reads. This means for any transaction we can only know which state locations the tx writes to. The reads in the BAL are at the block level and shared between all transactions so we don't know which transactions caused a particular read to be added to the read set.

> If we are going to execute each transaction in their own context/thread, wouldn't it sufficient only inject prestate needed by that transaction, no wasted accounts. 

This would be ideal but its not possible to compute the exact read set for any given transaction just using the BAL because the reads are not indexed by BAL index.

> They are not sharing the same ledger right?

Yes we will need to use separate ledger for each transaction.

> For example, we have a list of excution context, using balAccessIndex, we populate the ledger of each context, no need for binary search.

Since we don't know the reads for a specific transaction we would have to apply all the writes in the BAL for this to work and as I suggested before this would be too slow.

Having said that, there is one additional optimization we could do: Since every time a storage slot is written, it is also read so we know that the tx will read those locations so we can write the pre-state of those locations into the ledger. Even with this we would still need the BAL overlay for any other reads not covered by the write reads.

jangko:
> Since we don't know the reads for a specific transaction we would have to apply all the writes in the BAL for this to work

I see, too bad they don't make it easier for us. save bandwith but not computational power.

bhartnett:
> I see, too bad they don't make it easier for us. save bandwith but not computational power.

The idea to store reads for each transaction was considered but rejected because the BALs would become too large. Yeah so bandwidth vs computation is the trade off here.

PR #4356, opened 2026-06-22
syncer aware of FC state

advaita-saha:


execution_chain/sync/beacon/worker/blocks/blocks_blocks.nim
```diff
@@ -211,10 +211,13 @@ template blocksImport*(
...
-        # Skip blocks at or below the current base — `FC` would otherwise
-        # quarantine them as orphans and abort the batch.
-        if nthBn <= ctx.chain.baseNumber:
-          trace "Ignoring block less eq. base", peer, blk=nthBn,
+        # Skip blocks the live `FC` head already covers. These were imported
+        # either by us earlier or by a concurrent importer such as `el_sync`.
+        # Re-importing is redundant, and for blocks pruned below `base` the `FC`
+        # would quarantine them as orphans and abort the batch. Re-reading
+        # `latestNumber` here keeps the decision fresh as the head moves.
+        if nthBn <= ctx.chain.latestNumber:
```
arnetheduck: this doesn't look correct - what if the blocks are from a fork? this is legitimate. 

What should be happening here is that the syncer gives the block to the fc - the fc then responds something like "duplicate" or "ophaned" depending on where it's before or after `finalized`, ie logic like this should live in the fc.

Here's the flow:
* syncer asks fc "what should I sync"?
* syncer downloads a chunk of blocks and gives them to the FC - time passes
* for each block, the fc has the logic to decide if the blocks ar:
  * invalid
  * already observed (ie the block content arrived from some other source)
  * already orphaned (ie a fork choice caused the blocks to become orphaned because finalization cut off the branch to which the blocks belonged)
  * valid

See https://github.com/status-im/nimbus-eth2/blob/a183cf7d688d30ee40736153a0730698a2ed63f1/beacon_chain/consensus_object_pools/block_clearance.nim#L199

in the case of an "already orphaned" branch, it means that all other "forward" blocks from that same branch can be dropped - in the case of an "already observed", the syncer simply continues working on the next block.

Crucially, once it has completed a chunk of blocks, it needs to re-establish "what should I sync"

arnetheduck: broadly, this idea means that the return value of `validateBlock` should be updated with this kind of information

advaita-saha: Yeah was doing something similar
But having `validateBlock` return the status is a good idea

PR #4360, opened 2026-06-24
proxy: op-stack support

chirag-parmar:
### Description:

This PR adds support for op-stack L2 chains into verified proxy. Currently the integration supports running the L1 engine and the L2 engine side-by-side so the API for both can be used at the same time. 

### Notes:

1. Organization: All op-stack related code is placed under the subdirectory `op`
2. Semantics: The L2 engine uses the same `RpcVerificationEngine` type since it is an EVM compatible type. Few minor configuration overrides are done when loading/starting the secondary L2 engine. These overrides include passing the L1 network id into the EVM to avoid creating unnecessary code paths
3. Verification: the L2 engine follows a similar pattern of *syncing* anchor points (like light client updates in L1) using the L1 engine. 
     a. Every block on the L2 is added to a batch of blocks which are compressed and posted on L1. A rollup node then derives every block from the "available" data and proposes an output root (hash of state_root and block_hash and other values). This output root is **proposed** via a dispute game through the `DisputeGameFactory` against a block number. Such a proposal itself carries significant monetary capital that can be slashed. Hence, we use this proposal to mark blocks `safe` (safe according op-stack specs means the block can derived not that an output root was proposed, we use it differently in verified proxy for stricter guarantees). 
     b. Note: The output root is posted at an irregular cadence. So `safe` blocks don't come at regular intervals but they do imply through canonical-ness that everything that came before is also "safe". So we read the `DisputeGameFactory` for the latest game, verify the claim and push the "safe" block into the header store of the verified proxy.
     c. After the dispute game resolves and the L1 block at which it resolved is finalized, the L2 block against which the output root was proposed is considered finalized. We use the same semantics for the verified proxy as well. 
     d. Every time a game finalizes, either proposer or any other actor pushes the finalized output root and the block number to the `AnchorStateRegistry`. This registry contract performs a trustless verification of the output root and only then adds it to the registry. We use this registry to read the finalized output root and block number and then store the block in the header store of the proxy.
     e. To achhieve this verified anchor point update a.k.a syncing. The verified proxy first calls the `SystemConfig` contract to get the current addresses of the `OptimismPortal` contract, `DisputeGameFactory` contract and `AnchorStateRegistry` contract (fetched from `OptimismPortal` contract). 
     f. NOTE:  the header store only stores safe blocks as a canonical chain and marks the finalized blocks in the store. If a query is made against the latest block, the proxy first downloads the latest and verifies it comes from the canonical chain by walking through the history to a safe block. This is the only way a latest block can be downloaded with minimized trust.

### DIFF
--- a/nimbus_verified_proxy/library/c_frontend.nim
+++ b/nimbus_verified_proxy/library/c_frontend.nim
@@ -46,6 +46,16 @@ template callbackToC(
   else:
     fut.addCallback sendResult
 
+template requireOpFrontend(ctx: ptr Context, cb: CallBackProc, userData: pointer) =
+  if ctx.opFrontend.eth_blockNumber.isNil:
+    cb(
+      ctx,
+      RET_ERROR,
+      alloc("op_ namespace unavailable (no OP network configured)"),
+      userData,
+    )
+    return
+
 proc eth_blockNumber(
     ctx: ptr Context, cb: CallBackProc, userData: pointer
 ) {.exported.} =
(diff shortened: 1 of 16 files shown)


nimbus_verified_proxy/op/op_frontend.nim
```diff
@@ -0,0 +1,563 @@
...
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [], gcsafe.}
+
+import
+  std/strutils,
+  results,
+  stew/byteutils,
+  nimcrypto/sysrand,
```
tersec, line 14: https://nim-lang.org/docs/sysrand.html is another option here

nimbus_verified_proxy/library/verifproxy.nim
```diff
@@ -108,6 +111,17 @@ proc load(T: type VerifiedProxyConf, configJson: string): T {.raises: [ProxyErro
...
           ProxyError, "Couldn't parse `privateTxUrls` from JSON config: " & e.msg
         )
+    opExecutionApiUrls =
+      try:
+        let rawUrls = jsonNode.getOrDefault("opExecutionApiUrls").getStr("")
+        if rawUrls.len == 0:
+          UrlList(@[])
+        else:
+          parseCmdArg(UrlList, rawUrls)
+      except CatchableError as e:
```
tersec: Which exceptions actually need to be caught here?

nimbus_verified_proxy/op/op_frontend.nim
chirag-parmar: went with nimcrypto because we already use it elsewhere and because we created a shim for it to be compatible with emscripten. Haven't explored if `urandom` for `std/sysrand` provides the compatibility

nimbus_verified_proxy/library/verifproxy.nim
chirag-parmar: `ValueError` but catching all catchable is better no?

tersec: https://status-im.github.io/nim-style-guide/errors.exceptions.html#catchableerror
> Catching `CatchableError` implies that all errors are funnelled through the same exception handler. When called code starts raising new exceptions, it becomes difficult to find affected code - catching more specific errors avoids this maintenance problem.

In general, this is what `raises` annotations are for, to avoid this need for defensive programming around exception handling.

chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4360/commits/87d0ebdd94a1333ae3f7b8a98c5b81fc4e41371f

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -238,18 +264,66 @@ proc run(
   let frontend = engine.getExecutionApiFrontend()
   let frontendServers = startFrontends(frontend, config.frontendUrls)
 
+  # nil unless an OP network is configured (RpcVerificationEngine is a ref, so nil is the
+  # natural "no L2 engine" state no Option wrapping needed)
+  var
+    l2Engine: RpcVerificationEngine
+    opExecBackendClients: seq[JsonRpcClient] = @[]
+    opFrontendServers: seq[JsonRpcServer] = @[]
```
tersec: No need for either ` = @[]`.

https://nim-lang.org/docs/manual.html#statements-and-expressions-var-statement
> Variables are always initialized with a default value if there is no initializing expression. The default value depends on the type and is always a zero in binary.

And for a sequence it's documented to be `@[]` already.

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -183,10 +186,33 @@ proc run(
...
+  let networkName = config.eth2Network.get("mainnet")
+
+  # If an op-stack network is selected we run a secondary engine alongside the primary engine
+  let opParams =
+    if isOpNetwork(networkName):
+      let p = opChainParamsForNetwork(networkName).valueOr:
+        raise newException(ProxyError, "Unknown OP network: " & error)
+      some(p)
+    else:
+      none(OpChainParams)
```
tersec: Per https://status-im.github.io/nim-style-guide/libraries.results.html prefer `Opt.some` / `Opt.none`

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -238,18 +264,66 @@ proc run(
...
+      )
+
+    # the L2 EVM follows the L1 fork schedule, so it uses the L1 network id
+    let l2NetworkId = chainIdToNetworkId(l1ChainId).valueOr:
+      raise newException(
+        ProxyError, "Couldn't derive the L2 network id from the L1 chain id"
+      )
+
+    l2Engine = RpcVerificationEngine.initCore(
+      chainId = opParams.get().l2ChainId,
```
tersec: When `opParams` is an `Opt`, can replace `if opParams.isSome():` with `opParams.isErrOr`:` and then use `value` instead of `opParams.get`.

This avoids having raw `get`s and manually checking presence without efficiency loss.

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -238,18 +264,66 @@ proc run(
...
   try:
     while true:
       await sleepAsync(engine.timeParams.SLOT_DURATION)
+
       let syncRes = await engine.syncOnce()
       if syncRes.isErr():
-        debug "LC sync failed", error = syncRes.error.errMsg
+        error "LC sync failed", error = syncRes.error.errMsg
+
+      if opParams.isSome():
```
tersec: `opParams.isErrOr: ...`

nimbus_verified_proxy/op/op_anchor_utils.nim
```diff
@@ -0,0 +1,171 @@
...
+  outputRoot*: Hash32
+  l2BlockNumber*: base.BlockNumber
+
+func hashFnSig(sig: string): seq[byte] =
+  @(keccak256(sig).data[0 .. 3])
+
+proc abiEncode[T](value: T): EngineResult[seq[byte]] =
+  try:
+    ok(Abi.encode(value))
+  except CatchableError as e:
```
tersec: https://status-im.github.io/nim-style-guide/errors.exceptions.html#catchableerror
> Catching CatchableError implies that all errors are funnelled through the same exception handler. When called code starts raising new exceptions, it becomes difficult to find affected code - catching more specific errors avoids this maintenance problem.

nimbus_verified_proxy/op/op_anchor_utils.nim
```diff
@@ -0,0 +1,171 @@
...
+    err((InvalidDataError, "calldata ABI encode failed: " & e.msg, UNTAGGED))
+
+proc abiDecode(output: seq[byte], T: typedesc): EngineResult[T] =
+  try:
+    ok(Abi.decode(output, T))
+  except CatchableError as e:
```
tersec: https://status-im.github.io/nim-style-guide/errors.exceptions.html#catchableerror

nimbus_verified_proxy/op/op_frontend.nim
```diff
@@ -0,0 +1,563 @@
...
+
+    ok(blobBaseFee)
+
+  frontend.eth_gasPrice = proc(): Future[EngineResult[Quantity]] {.
+      async: (raises: [CancelledError])
+  .} =
+    opEngine.opSync(l1Engine)
+
+    let suggestedPrice = opEngine.penaltyOr(await opEngine.suggestGasPrice())
+    ok(Quantity(suggestedPrice.uint64))
```
tersec: Don't need to first convert `suggestedPrice` to `uint64` then `Quantity`, can just convert to `Quantity` (i.e. `distinct uint64`).

nimbus_verified_proxy/op/op_frontend.nim
```diff
@@ -0,0 +1,563 @@
...
+
+  frontend.eth_maxPriorityFeePerGas = proc(): Future[EngineResult[Quantity]] {.
+      async: (raises: [CancelledError])
+  .} =
+    opEngine.opSync(l1Engine)
+
+    let suggestedPrice = opEngine.penaltyOr(await opEngine.suggestMaxPriorityGasPrice())
+    ok(Quantity(suggestedPrice.uint64))
```
tersec: Don't need to first convert `suggestedPrice` to `uint64` then `Quantity`, can just convert to `Quantity` (i.e. `distinct uint64`).

nimbus_verified_proxy/op/op_anchor_utils.nim
```diff
@@ -0,0 +1,171 @@
...
+# this gives us the latest proposed output root
+proc readLatestGame*(
+    l1Engine: RpcVerificationEngine, dgfContract: Address, l1Header: Header
+): Future[EngineResult[L1OutputProposal]] {.async: (raises: [CancelledError]).} =
+  let
+    countOut =
+      ?(await l1Engine.makeCall(dgfContract, hashFnSig("gameCount()"), l1Header))
+    count = ?abiDecode(countOut, UInt256)
+
+  if count == 0.u256:
```
tersec: More efficient to use [built-in `nim-stint` support of `isZero`](https://github.com/status-im/nim-stint/blob/c3e76a01580ae2ac0e62fbd04040722f8f22b84d/stint/intops.nim#L122-L123):
```nim
func isZero*(a: StInt): bool =
  a.impl.isZero
```

nimbus_verified_proxy/library/c_frontend.nim
```diff
@@ -415,6 +425,406 @@ proc eth_sendRawTransaction(
...
+    index: culonglong,
+    cb: CallBackProc,
+    userData: pointer,
+) {.exported.} =
+  let
+    blockHashTyped = unpackArg($blockHash, Hash32).valueOr:
+      cb(ctx, RET_DESER_ERROR, alloc(error), userData)
+      return
+
+    indexTyped = Quantity(uint64(index))
```
tersec: Can convert directly: `Quantity(index)`

nimbus_verified_proxy/library/c_frontend.nim
```diff
@@ -415,6 +425,406 @@ proc eth_sendRawTransaction(
...
+    index: culonglong,
+    cb: CallBackProc,
+    userData: pointer,
+) {.exported.} =
+  let
+    blockTagTyped = unpackArg($blockTag, BlockTag).valueOr:
+      cb(ctx, RET_DESER_ERROR, alloc(error), userData)
+      return
+
+    indexTyped = Quantity(uint64(index))
```
tersec: Can convert directly: `Quantity(index)`

chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4360/commits/5429cf0d70f9d8b8462acaf50f17b24815e1b84f

nimbus_verified_proxy/library/c_frontend.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4360/commits/5429cf0d70f9d8b8462acaf50f17b24815e1b84f

nimbus_verified_proxy/op/op_anchor_utils.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4360/commits/5429cf0d70f9d8b8462acaf50f17b24815e1b84f

nimbus_verified_proxy/op/op_frontend.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4360/commits/5429cf0d70f9d8b8462acaf50f17b24815e1b84f

nimbus_verified_proxy/op/op_frontend.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4360/commits/5429cf0d70f9d8b8462acaf50f17b24815e1b84f

nimbus_verified_proxy/op/op_anchor_utils.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4360/commits/5429cf0d70f9d8b8462acaf50f17b24815e1b84f

nimbus_verified_proxy/op/op_anchor_utils.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4360/commits/5429cf0d70f9d8b8462acaf50f17b24815e1b84f

nimbus_verified_proxy/nimbus_verified_proxy.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4360/commits/5429cf0d70f9d8b8462acaf50f17b24815e1b84f

nimbus_verified_proxy/nimbus_verified_proxy.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4360/commits/5429cf0d70f9d8b8462acaf50f17b24815e1b84f

nimbus_verified_proxy/nimbus_verified_proxy.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4360/commits/5429cf0d70f9d8b8462acaf50f17b24815e1b84f

nimbus_verified_proxy/nimbus_verified_proxy.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4360/commits/5429cf0d70f9d8b8462acaf50f17b24815e1b84f

nimbus_verified_proxy/op/op_anchor_utils.nim
```diff
@@ -0,0 +1,171 @@
...
+
+const OUTPUT_ROOT_VERSION_V0* = default(array[32, byte])
+
+func computeOutputRoot*(
+    stateRoot: Hash32, messagePasserStorageRoot: Hash32, blockHash: Hash32
+): Hash32 =
+  let preimage =
+    @OUTPUT_ROOT_VERSION_V0 & @(stateRoot.data) & @(messagePasserStorageRoot.data) &
+    @(blockHash.data)
+  keccak256(preimage)
```
tersec: More efficient to use `withKeccak256`, e.g., https://github.com/status-im/nimbus-eth1/blob/b83526d719fa4ae3cf0403076feb1373fdc1eab3/execution_chain/utils/utils.nim#L103-L114 to avoid actual concatenation (as well as all the `array` to `seq` conversions, since `update` takes `openArray`).

nimbus_verified_proxy/nimbus_verified_proxy.nim
```diff
@@ -183,10 +186,33 @@ proc run(
     except Exception:
       notice "commandLineParams() exception"
 
+  let networkName = config.eth2Network.get("mainnet")
+
+  # If an op-stack network is selected we run a secondary engine alongside the primary engine
+  let opParams =
+    if isOpNetwork(networkName):
+      let p = opChainParamsForNetwork(networkName).valueOr:
```
tersec:
```nim
func isOpNetwork*(name: string): bool =
  opChainParamsForNetwork(name).isOk()
```
`isOpNetwork` is already directly defined in terms of `opChainParamsForNetwork`, so should be able to directly query the latter once, rather than indirectly, twice.

chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4360/commits/c2740a4c30d391f8b69d626bbb7706edacf78551

nimbus_verified_proxy/op/op_anchor_utils.nim
chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4360/commits/c2740a4c30d391f8b69d626bbb7706edacf78551

tersec (approved):

PR #4361, opened 2026-06-25
Only allow skipping pre-execution BAL check for finalized blocks

bhartnett:
Reverts the temporary change for bal devnets which didn't require having eth/71 implemented so non finalized blocks could be received without a valid BAL. 

eth/71 is now required so this workaround is no longer required.


bhartnett:
Actually it turns out that the engine api already enforces that a BAL is present for amsterdam blocks and for the eth/71 sync we need to be able to tolerate missing BALs for scenarios when we fail to download them from peers for various reasons. So this change isn't needed.

PR #4362, opened 2026-06-25
Fetch BALs when syncing

bhartnett:
This updates the sync code to fetch BALs for Amsterdam blocks via devp2p eth/71. It makes sense to support this since we now support state prefetching using BALs.

Only BALs within the BAL retention period (which is defined in EIP-7928 as 3533 epochs from head) are fetched.

### DIFF
--- a/execution_chain/sync/beacon/worker/blocks/blocks_bal.nim
+++ b/execution_chain/sync/beacon/worker/blocks/blocks_bal.nim
@@ -1,0 +1,76 @@
+# Nimbus
+# Copyright (c) 2026 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at
+#     https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at
+#     https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+{.push raises:[].}
+
+import
+  std/typetraits,
+  pkg/[chronos, results],
+  pkg/eth/common,
+  pkg/eth/common/[block_access_lists, block_access_lists_rlp],
+  ../../../wire_protocol,
+  ../worker_desc
+
+export block_access_lists
+
+# ------------------------------------------------------------------------------
+# Public functions
+# ------------------------------------------------------------------------------
+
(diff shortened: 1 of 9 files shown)


execution_chain/sync/beacon/worker/blocks/blocks_bal.nim
mjfh, line 1: This would do the job but there is no error handling yet.

I will need that stuff soon, too and can check/test and implement timeout handling. The main annoyance is when peers stop talking and the syncer still manages these peers.

execution_chain/sync/beacon/worker/blocks/blocks_blocks.nim
```diff
@@ -122,8 +124,36 @@ template blocksFetchCheckImpl(
...
+      let raws = (await buddy.fetchRawBlockAccessLists(balRequest)).valueOr:
+        default(seq[RawBlockAccessList])
+      var nBals = 0
+      for j in 0 ..< min(raws.len, balPos.len):
+        bals[balPos[j]] = decodeBlockAccessList(raws[j], blocks[balPos[j]].header)
+        if bals[balPos[j]].isSome:
+          inc nBals
+      trace info & ": fetched block access lists", peer, iv=($iv),
+        nReq=balRequest.blockHashes.len, nResp=raws.len, nBals
+
```
mjfh, line 156: I will need to extract this block 127..153 to a different source when I need BALs for snap sync. So that it will be separately available w/o downloading blocks. No need to do it now.

mjfh (approved):

mjfh (approved):

execution_chain/sync/beacon/worker/blocks/blocks_bal.nim
bhartnett: > This would do the job but there is no error handling yet.

Yeah I didn't add it here, just returning Opt.none because BALs are optional and can be generated during block processing. Maybe I'll return similar errors as the get block bodies code.

> I will need that stuff soon, too and can check/test and implement timeout handling. The main annoyance is when peers stop talking and the syncer still manages these peers.

Yeah that needs to be handled across the board for all message types I guess.

mjfh: It is done for any other message except BAL :)

PR #4381, opened 2026-07-01
t8n: support EIP-7843 currentSlotNumber in env

qu0b:
## Problem

t8n's `EnvStruct` has no field for the EIP-7843 slot number, and its custom env `readValue` discards unknown keys, so `blockCtx.slotNumber` is always `0`. The `SLOTNUM` opcode therefore always reads `0`, and t8n cannot faithfully replay any block whose transactions use `SLOTNUM` — storing the slot becomes a no-op.

## Fix

- `tools/t8n/types.nim` — add `currentSlotNumber*: Opt[uint64]` to `EnvStruct`
- `tools/t8n/helpers.nim` — parse `"currentSlotNumber"` in the custom env `readValue`
- `tools/t8n/transition.nim` — thread it through `envToHeader` into the header (→ `blockCtx.slotNumber`)

## Verification

On a glamsterdam-devnet block that uses `SLOTNUM`: with `currentSlotNumber` set in the env, `SLOTNUM` returns the slot and a contract storing it persists the correct value (`0x…03c0`), matching the canonical chain. Previously the store was a no-op (value `0`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tools/t8n/helpers.nim
+++ b/tools/t8n/helpers.nim
@@ -169,6 +169,7 @@ proc readValue*(r: var JsonReader[T8Conv], val: var EnvStruct)
     of "currentTimestamp":
       r.readValue(val.currentTimestamp)
       currentTimestampParsed = true
+    of "currentSlotNumber": r.readValue(val.currentSlotNumber)
     of "currentDifficulty": r.readValue(val.currentDifficulty)
     of "currentRandom": r.readValue(val.currentRandom)
     of "parentDifficulty": r.readValue(val.parentDifficulty)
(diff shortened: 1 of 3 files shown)


tools/t8n/helpers.nim
```diff
@@ -169,6 +169,7 @@ proc readValue*(r: var JsonReader[T8Conv], val: var EnvStruct)
     of "currentTimestamp":
       r.readValue(val.currentTimestamp)
       currentTimestampParsed = true
+    of "currentSlotNumber": r.readValue(val.currentSlotNumber)
```
jangko, line 172: Which one is correct, "currentSlotNumber" or "slotNumber"?

geth is using "slotNumber"
https://github.com/ethereum/go-ethereum/blob/7e625dd54822499d7fece76de9850377d65957c4/cmd/evm/internal/t8ntool/gen_stenv.go#L40

qu0b: ah its slotNumber

jangko:
Close in favor of #4392, the new PR fix a lot of t8n bugs and test it against EEST fixtures.

PR #4383, opened 2026-07-01
stateless: add guest program and integrate into EEST blockchain test

kdeme:
Add `verify_stateless_new_payload` and `run_stateless_guest` as the stateless guest entry points. This adds deserializing of the StatelessInput, executing the block purely from that input, and returning the StatelessValidationResult.

eest_blockchain.nim got refactored to:
- use the stateless input/output and run the full `run_stateless_guest` pipeline
- also validate/test fixtures with intended exec failures

The newly skipped vectors are due to fails from the latter

### DIFF
--- a/execution_chain/stateless/stateless_execution.nim
+++ b/execution_chain/stateless/stateless_execution.nim
@@ -12,14 +12,23 @@
 import
   std/tables,
   results,
-  eth/common/[headers, blocks],
-  eth/rlp,
+  stew/endians2,
+  eth/common/[headers, blocks, hashes],
+  eth/trie/ordered_trie,
+  beacon_chain/spec/eth2_merkleization,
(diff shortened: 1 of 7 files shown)


kdeme:
arm-64 runner for some reason takes so much longer... and hangs.

jangko:
looks like OOM, it happen with #4392 too. After I separate the `make eest_full_test` from `make test t8n_test` the arm64 CI now finished.

kdeme:
> looks like OOM, it happen with #4392 too. After I separate the `make eest_full_test` from `make test t8n_test` the arm64 CI now finished.

ok good. Thanks!  I'll rebase then once that gets merged.

Or this run might actually pass then however, considering I removed the develop set of tests for stateless in this run (I think I can leave that as such to have it run slightly faster).

PR #4399, opened 2026-07-03
Fix payload id hit

advaita-saha:


jangko:
The comment in line 212 already correct
https://github.com/status-im/nimbus-eth1/blob/d0bacc63a1109b9d4632b0574fca43165ef75956/execution_chain/beacon/api_handler/api_forkchoice.nim#L210-L212

CL can replace the payload many times. 

And the root cause have been found #4405, but I think we can modify the test a bit to add engine api test. when #4405 only add txpool test.

PR #4404, opened 2026-07-03
evm: error on missing in-range BLOCKHASH instead of returning zero

kdeme:
getBlockHash now returns Result[Hash32, string] and propagates the error up through getAncestorHash -> computation.getBlockHash -> blockhashOp. A block hash that is missing while within the accessible 256-block range should be considered an error for stateless (an incomplete witness in stateless execution), so the BLOCKHASH opcode now reverts the transaction rather than silently pushing a zero hash. The block then fails at the state-root check.
For stateful execution this should never occur as those block headers and/or hashes should be available. So perhaps an assert would have been better here.

Out-of-range requests (older than 256 blocks or not a past block) are unchanged and still return the default zero hash. This matches the execution-specs, where an in-range miss raises rather than returning 0.

The error causes an transaction-level revert and not full block exec stop, which is not ideal. But it is not evident to propogate errors all the way up outside of the opcodes handling.

### DIFF
--- a/execution_chain/core/executor/process_transaction.nim
+++ b/execution_chain/core/executor/process_transaction.nim
@@ -169,6 +169,11 @@ proc processTransaction*(
   var callResult = tx.txCallEvm(sender, vmState, intrinsic)
   vmState.captureTxEnd(tx.gasLimit - callResult.gasUsed)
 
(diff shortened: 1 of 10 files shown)


kdeme:
closing for https://github.com/status-im/nimbus-eth1/pull/4407

PR #4407, opened 2026-07-03
stateless: fail block on missing in-range BLOCKHASH

kdeme:
An in-range BLOCKHASH whose hash is absent used to silently return the zero hash. During stateless execution that hides an incomplete witness. getBlockHash now records a fatalError on such a miss, which processTransaction turns into a block validation error under stateless execution and an assert (corrupt database) on a full node. The zero hash is still returned so the async EVM path is unaffected (i.e. the verified proxy).

### DIFF
--- a/execution_chain/core/executor/process_transaction.nim
+++ b/execution_chain/core/executor/process_transaction.nim
@@ -155,7 +155,7 @@ proc processTransaction*(
     rollbackReads: bool = false;
       ): Result[LogResult, string] =
   ## Modelled after `https://eips.ethereum.org/EIPS/eip-1559#specification`_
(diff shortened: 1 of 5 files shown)


kdeme:
This is a bit less convoluted version of https://github.com/status-im/nimbus-eth1/pull/4404
It also adds stateless flag, which if done in that version would be more cumbersome as we would also have to divide between stateless and async EVM (verified proxy's way). And those would mix up with the witness generation version which should assert in theory.

kdeme:
Still not fully convinced of doing this by means of a flag, but seems better than the alternatives.

Compared with Python execution spec, what is done there for these cases is just let an Exception (index error in this case) raise up all the way and just catch it in the guest program.

PR #4408, opened 2026-07-03
Parallel execution using block access lists

bhartnett:
This implements parallel transaction execution using block access lists. The approach taken is as follows:
- Every transaction runs in a separate task which executes on the taskpool threadpool in a background thread.
- For each transaction we setup a new vmState, ledger and tracker. 
- The ledger is setup with the BAL overlay which enables reading the correct prestate for each transaction.
- The tracker is setup using a thread safe BAL builder which is shared between all the tasks and also the main thread. This shared builder is what enables the main thread to build the BAL and verify it matches the BAL hash in the header.
- After executing each transaction, the ledger is not persisted into the database, instead the final post state value for each account/slot in the BAL is used to update the state before the stateroot computation. The purpose of the execution is to collect the BAL reads/writes and to build the logs for the transaction receipts.
- Pre-execution and post-execution system calls still execute in the main thread (for now)
- Block level data such as gas used is collected and processed in the main thread after executing each task.

Parallel execution is disabled by default and can be enabled using `--debug-bal-parallel-execution=true`


bhartnett:
Ran the test starting at block 15 million, imported 10K blocks with stateroot disabled, running with 16 taskpool threads.

With bal parallel execution disabled:
elapsed=7m32s890ms

With bal parallel execution enabled:
elapsed=4m15s196ms

The improvement is similar to the improvement from just running the BAL state prefetching. 

The parallel execution speedup is not that much, likely due to:
- Write contention on the shared BAL builder.
- Parallel execution context starts with empty ledger with empty caches, so code is always fetched from the disk (no shared code cache)
- More allocations (vmState and ledger created in each tx task)
- More copies (code and logs get copied from worker threads to main thread)

To improve this further in future PRs we can try:
- Using a lock free hash table in the shared BAL builder.
- Moving the code caches out of the ledger into the db layer so that they get shared between all ledgers.
- Perhaps the ledger could become a value object to reduce allocations.

tersec:
- https://github.com/nim-lang/Nim/issues/25992

is the root-cause Nim bug noted by https://github.com/status-im/nimbus-eth1/tree/bal-parallel-execution-arm64-bug / https://github.com/status-im/nimbus-eth1/commit/977a2f61dfb60fbf35ad96f23edd5a17ffe0436b demonstrated by [`linux-arm64`](https://github.com/status-im/nimbus-eth1/actions/runs/28851903745/job/85569061618):
```
Snap Data Validator ..... (0.01s)
Snap Data Edge Cases .... (0.00s)
Transactions tests .................................................................................................................................................................................................................... (0.47s)
transactions / receipts storage (>256) .. (0.00s)
TxPool test suite ..............F............. (4.76s)
===============================
  /home/runner/work/nimbus-eth1/nimbus-eth1/build/all_tests 'TxPool test suite::Test TxPool with blobhash block'
-------------------------------
    /home/runner/work/nimbus-eth1/nimbus-eth1/tests/test_txpool.nim(547, 21): Check failed: blockValue == bundle.blockValue
    blockValue was 29628290609190
    bundle.blockValue was 0

  [FAILED ] (  0.24s) Test TxPool with blobhash block

(4.76s)   TxPool test suite

TxPool BAL post-Amsterdam .. (0.00s)
TxFrame blobify round-trip ................. (0.00s)
```

PR #4409, opened 2026-07-04
fix(engine): preserve withdrawals for V4 (Amsterdam) payload attributes

lodekeeper:
## Problem

`setWithdrawals` (`execution_chain/beacon/beacon_engine.nim`) only copies withdrawals from `PayloadAttributes` for versions `V2`/`V3`; every other version falls into the `else` branch and sets `xp.withdrawals = @[]`.

From Amsterdam, the CL drives block building via `engine_forkchoiceUpdatedV4` with `PayloadAttributesV4` (`version == Version.V4`). That version isn't enumerated, so **all withdrawals are silently discarded** during block assembly. The built payload then has `withdrawalsRoot = calcWithdrawalsRoot(@[])` (empty root), which consensus clients reject as a withdrawals-root mismatch.

## Impact — observed on `glamsterdam-devnet-6`

Every block built by a `nimbus-el` proposer that carried a withdrawal was rejected network-wide, stalling finality:

- **187 distinct slots over ~40h** had their envelopes rejected — **100% `nimbus-el` proposers**, across 4 different consensus clients (teku, lodestar, grandine, prysm). No other EL is affected.
- On the Nimbus CL self-build path the mismatch is explicit: `withdrawals_from_cl_len=1 → withdrawals_from_el_len=0` (66/66 build attempts in a 13h window, `el_len` always `0`).
- The withdrawals involved are builder-payment withdrawals (`get_builder_withdrawals`; `validator_index` carries `BUILDER_INDEX_FLAG = 2**40`), only because those are the withdrawals occurring on that devnet. Regular validator withdrawals traverse the same path and would be dropped identically.

Present on `master` (which already has the V4/Amsterdam FCU handling) and live on the `glamsterdam-devnet-6` branch (`2a30edf`).

## Fix

Invert the guard so only `V1` (pre-Shanghai, no withdrawals) empties the list, and every later version copies the withdrawals through. This also prevents the same regression when future payload-attributes versions are added.

```nim
case attrs.version
of Version.V1:
  xp.withdrawals = @[]
else:
  xp.withdrawals = ethWithdrawals attrs.withdrawals.get
```

`setWithdrawals` is the only withdrawal setter and is called once (`beacon_engine.nim:170`), so nothing else covers V4.

Happy to add a regression test (V4 payload attributes preserve withdrawals) — point me at the preferred harness and I'll follow up.

🤖 Generated with AI assistance

### DIFF
--- a/execution_chain/beacon/beacon_engine.nim
+++ b/execution_chain/beacon/beacon_engine.nim
@@ -87,7 +87,7 @@ func setWithdrawals(xp: TxPoolRef, attrs: PayloadAttributes) =
   case attrs.version
   of Version.V1:
     xp.withdrawals = @[]
-  of Version.V2 .. Version.V6:
+  of Version.V2 .. Version.V4:
     xp.withdrawals = ethWithdrawals attrs.withdrawals.get
 
 template wrapException(body: untyped): auto =
--- a/execution_chain/beacon/beacon_engine.nim
+++ b/execution_chain/beacon/beacon_engine.nim
@@ -89,6 +89,8 @@ func setWithdrawals(xp: TxPoolRef, attrs: PayloadAttributes) =
     xp.withdrawals = @[]
   of Version.V2 .. Version.V4:
     xp.withdrawals = ethWithdrawals attrs.withdrawals.get
+  of Version.V5, Version.V6:
+    raiseAssert "unreachable: PayloadAttributes has no V5/V6 variant"
 
 template wrapException(body: untyped): auto =
   try:
--- a/execution_chain/beacon/beacon_engine.nim
+++ b/execution_chain/beacon/beacon_engine.nim
@@ -85,10 +85,10 @@ const
 
 func setWithdrawals(xp: TxPoolRef, attrs: PayloadAttributes) =
   case attrs.version
-  of Version.V2, Version.V3:
-    xp.withdrawals = ethWithdrawals attrs.withdrawals.get
-  else:
+  of Version.V1:
     xp.withdrawals = @[]
+  of Version.V2 .. Version.V6:
+    xp.withdrawals = ethWithdrawals attrs.withdrawals.get
 
 template wrapException(body: untyped): auto =
   try:
--- a/tests/test_engine_api.nim
+++ b/tests/test_engine_api.nim
@@ -311,6 +311,48 @@ proc newPayloadV4InvalidRequestType(env: TestEnv): Result[void, string] =
 
   ok()
 
+proc payloadAttrV4PreserveWithdrawalsTest(env: TestEnv): Result[void, string] =
+  # Regression: setWithdrawals used to drop withdrawals for V4 (Amsterdam) payload
+  # attributes. A PayloadAttributes with a targetGasLimit resolves to Version.V4,
+  # which fell into the `else` branch and had its withdrawals replaced with an empty
+  # seq, so the assembled payload carried no withdrawals even when the attributes
+  # requested some. Verify they are preserved.
+  let
+    ben = BeaconEngineRef.new(env.txPool)
+    time = getTime().toUnix
+    withdrawals = @[
+      WithdrawalV1(
+        index:          w3Qty(1'u64),
+        validatorIndex: w3Qty(100'u64),
+        address:        default(Address),
+        amount:         w3Qty(42'u64)
+      )
+    ]
+    # targetGasLimit set -> PayloadAttributes.version == Version.V4
+    attr = PayloadAttributes(
+      timestamp:             w3Qty(time + 1),
+      prevRandao:            default(Bytes32),
+      suggestedFeeRecipient: default(Address),
+      withdrawals:           Opt.some(withdrawals),
+      parentBeaconBlockRoot: Opt.some(default(Hash32)),
+      targetGasLimit:        Opt.some(w3Qty(30_000_000'u64))
+    )
+
+  if attr.version != Version.V4:
+    return err("expected PayloadAttributes version V4, got " & $attr.version)
+
+  let bundle = ? ben.generateExecutionBundle(attr)
+
+  if bundle.payload.withdrawals.isNone:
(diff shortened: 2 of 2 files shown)


execution_chain/beacon/beacon_engine.nim
```diff
@@ -84,11 +84,14 @@ const
...
+  # Withdrawals are present from Shanghai (V2) onward; only V1 has none.
+  # Enumerating explicit versions here dropped all withdrawals for V4
+  # (Amsterdam) payload attributes, yielding an empty withdrawals root.
   case attrs.version
-  of Version.V2, Version.V3:
-    xp.withdrawals = ethWithdrawals attrs.withdrawals.get
-  else:
+  of Version.V1:
     xp.withdrawals = @[]
+  else:
```
tersec: Generally, to maintain the compiler-checking of case-exhaustiveness without boilerplate, something like
`  Version.V2 .. Version.V4:` is preferred here. This formulation extends sans additional ceremony to any future `V5`, `V6`, etc. The issue before was that `else:` defeated that exhaustiveness analysis, which allowed this to happen silently.

execution_chain/beacon/beacon_engine.nim
```diff
@@ -84,11 +84,14 @@ const
 # ------------------------------------------------------------------------------
 
 func setWithdrawals(xp: TxPoolRef, attrs: PayloadAttributes) =
+  # Withdrawals are present from Shanghai (V2) onward; only V1 has none.
+  # Enumerating explicit versions here dropped all withdrawals for V4
```
tersec: Please drop this comment. It's inaccurate and the problem was precisely _not_ enumerating versions. Part of using Nim's case exhaustiveness check is that the intended actions for each version are specified.

execution_chain/beacon/beacon_engine.nim
lodekeeper: Good call — done in the latest push. Dropped the `else` so the case is exhaustive and the compiler will now flag any unhandled `Version`.

One wrinkle: the shared `Version` enum runs `V1..V6` (V5/V6 exist for `GetPayloadResponse`/`ExecutionPayload`), so `V1` + `V2 .. V4` alone isn't exhaustive. `version(PayloadAttributes)` only ever yields `V1..V4`, so V5/V6 are unreachable here, but I folded them into the withdrawals arm (`Version.V2 .. Version.V6`) as the safe default rather than adding a separate unreachable branch. Happy to switch to an explicit `V5/V6` arm (e.g. `raiseAssert`) if you'd prefer to make the impossibility explicit.

execution_chain/beacon/beacon_engine.nim
lodekeeper: Done — dropped the comment. Agreed on both counts: with the exhaustive `case` the per-version intent is specified in the code itself, and it was the missing V4 arm (the `else` absorbing it), not the enumeration, that caused the silent drop.

execution_chain/beacon/beacon_engine.nim
lodekeeper: Done — changed to `Version.V2 .. Version.V4` so the exhaustiveness check catches any future version additions explicitly.

execution_chain/beacon/beacon_engine.nim
lodekeeper: Dropped — the comment was already removed in the previous force-push.

execution_chain/beacon/beacon_engine.nim
```diff
@@ -85,10 +85,10 @@ const
 
 func setWithdrawals(xp: TxPoolRef, attrs: PayloadAttributes) =
   case attrs.version
-  of Version.V2, Version.V3:
-    xp.withdrawals = ethWithdrawals attrs.withdrawals.get
-  else:
+  of Version.V1:
     xp.withdrawals = @[]
+  of Version.V2 .. Version.V4:
```
tersec, line 90:
```
/home/runner/work/nimbus-eth1/nimbus-eth1/execution_chain/beacon/beacon_engine.nim(87, 3) Error: not all cases are covered; missing: {V5, V6}
```

lodekeeper: Fixed — added an explicit `of Version.V5, Version.V6: raiseAssert` arm so the exhaustiveness check is fully satisfied while making the impossibility explicit to future readers.

jangko:
here to put the test https://github.com/status-im/nimbus-eth1/blob/master/tests/test_engine_api.nim

lodekeeper:
Added in `4abba1b` (`test(engine): regression test for V4 payload attributes preserving withdrawals`), registered in `testList` as "PayloadAttributesV4 preserve withdrawals".

It builds a `PayloadAttributes` with `targetGasLimit` set (so `attr.version == Version.V4`) and a non-empty withdrawals list, runs it through `generateExecutionBundle`, and asserts the assembled payload still carries the withdrawals. That fails on the old `setWithdrawals` (V4 fell into the `else` branch → empty seq) and passes with the fix.

I used the `generateExecutionBundle` path rather than a full `forkchoiceUpdatedV4` → `getPayloadV6` RPC round-trip because the latter needs an Amsterdam env (plus blobs-bundle / execution-requests plumbing) that the current `setupEnv` doesn't wire up, and this isolates the `setWithdrawals` behaviour directly. Happy to rewrite it as a full RPC/Amsterdam integration test instead if you'd prefer that style — just say the word.

One heads-up: I don't have a Nim toolchain on this machine, so I couldn't compile/run it locally — I mirrored the existing tests' patterns and types, but please give CI a spin (or point out anything off) and I'll fix it promptly.

jangko:
It's ok, I can add the rpc test later, there is something I still need to fix too there.

lodekeeper:
Sounds good, thanks! The `generateExecutionBundle` test covers the `setWithdrawals` V4 regression in the meantime, and you can swap in the RPC version whenever it fits with your other fixes. Let me know if there's anything else I can help with on the PR.

PR #4413, opened 2026-07-05
fix: txpool crash

### LINKED ISSUE #4412: crash in `txPool` with AssertionDefect: the length of the table changed while iterating over it
Issue discovered by @tersec 

The execution client crashes with an unhandled AssertionDefect originating from the tx-broadcast subsystem. The crash is a mutate-while-iterating race on the shared seenTransactions table: the periodic cleanup task iterates the table while awaiting inside the loop, and a concurrently-dispatched peer message inserts into the same table during that await, changing its length mid-iteration.

## Cause
Because the cleanup action and the peer message handler both run on the same event loop but are not mutually serialized (only actions in actionQueue are serialized against each other; peer dispatch is independent), any await while holding a live iterator over seenTransactions is unsafe
This makes it an intermittent, traffic-dependent crash rather than a deterministic one.

### Test code for reproduction in `master`
```bash
  test "periodic cleanup survives concurrent seenTransactions mutation":
    ## Regression for the production crash:
    ##   `len(t) == L` the length of the table changed while iterating over it
    ##   [AssertionDefect]
    ##
    ## Black-box: drives the real "Periodical cleanup" action that tickerLoop
    ## enqueues (the only cleanup entry point that exists in BOTH the buggy and
    ## fixed trees), then runs it while another task mutates seenTransactions —
    ## exactly as a concurrently-dispatched handleTxHashesBroadcast would.
    ##
    ## FAILS on the buggy code: the cleanup awaits (awaitQuota) *inside* its
    ## `for key, seen in wire.seenTransactions` scan, so the concurrent insert
    ## changes the table length mid-iteration and trips the pairs-iterator
    ## assertion. PASSES after the fix: the scan is fully synchronous.
    proc runTest() {.async.} =
      let env = newBroadcastTestEnv()

      # Take over scheduling: stop the auto-started ticker/action loops so we
      # can enqueue exactly one cleanup action and run it ourselves.
      await env.wire.tickerHeartbeat.cancelAndWait()
      for fut in env.wire.actionHeartbeat:
        await fut.cancelAndWait()

      # Force awaitQuota to actually suspend: a capacity-1 bucket that
      # replenishes slowly means every throttled consume yields to the event
      # loop. On the buggy code that yield happens *inside* the table scan.
      env.wire.quota = TokenBucket.new(1, chronos.milliseconds(5))

      proc mkHash(i: int): Hash32 =
        var a: array[32, byte]
        a[0] = byte(i and 0xff)
        a[1] = byte((i shr 8) and 0xff)
        a.to(Hash32)

      # Populate with expired entries so the cleanup scan/deletion does work.
      let expiredAt = getTime() - initDuration(minutes = 25)
      for i in 0 ..< 32:
        env.wire.seenTransactions[mkHash(i)] =
          SeenObject(lastSeen: expiredAt, peers: initHashSet[NodeId]())

      # Make tickerLoop take its cleanup branch promptly: a short (but not-yet-
      # finished) cleanupTimer is kept and fires quickly; a long brUpdateTimer
      # never wins the `one()` race. tickerLoop runs synchronously up to its
      # first await, so these assignments are seen before the timers are read.
      env.wire.cleanupTimer = sleepAsync(chronos.milliseconds(50))
      env.wire.brUpdateTimer = sleepAsync(chronos.hours(1))

      let tl = tickerLoop(env.wire)

      # Wait for the cleanup action to be enqueued, then stop the ticker.
      var waited = 0
      while env.wire.actionQueue.len == 0 and waited < 200:
        await sleepAsync(chronos.milliseconds(10))
        inc waited
      await tl.cancelAndWait()
      check env.wire.actionQueue.len == 1

      # Concurrently insert fresh entries while the cleanup action runs,
      # mimicking a handleTxHashesBroadcast dispatch landing during a yield.
      proc mutator() {.async: (raises: [CancelledError]).} =
        for i in 0 ..< 40:
          await sleepAsync(chronos.milliseconds(1))
          env.wire.seenTransactions[mkHash(1000 + i)] =
            SeenObject(lastSeen: getTime(), peers: initHashSet[NodeId]())

      let mutatorFut = mutator()
      let action = await env.wire.actionQueue.popFirst()

      # On the buggy code this raises the AssertionDefect and the test fails.
      # On the fixed code it completes cleanly.
      let completed = await withTimeout(action(), chronos.seconds(5))
      check completed
      await mutatorFut.cancelAndWait()

      env.close()

    waitFor runTest()
```

advaita-saha:
- _**Fix**_: Never await while iterating the live table. Collect the expired keys in a single synchronous pass (no suspension → no concurrent mutation possible), then throttle/delete afterwards while iterating a plain seq (safe, since no table iterator is live across those awaits)
- Added regression test

Fixes #4412


tersec (approved):

advaita-saha:
Branch looks healthy on my nodes

PR #4425, opened 2026-07-06
stateless: fail block on incomplete witness during persist

kdeme:
A stateless witness is untrusted and may be partial/incomplete. A write (merge/delete) into a path whose node is absent from the witness fails in Aristo, and the ledger persist path escalates that to raiseAssert. This is fine for stateful execution, as it is a scenario that should not occur under proper working circumstances. However, due to the nature of stateless it can occur there and would be crashing the guest on untrusted input instead of rejecting the block.

We handle this type of error for stateless with the combination of stateless/fataError flags, like is done already for the blockhash miss. We could continue to make the persist call also result based but the issue (aside from handling all call sites of this) is that the the system calls where persist is done also do not propagate these errors. Two of them are void and need to (spec demands to) error silently. So if we were to start using result system there too, this would require a bigger re-architecture.

However, this fix does perhaps look/feel hacky due to the nature that an error is not immediatly handled but later checked upon.

We could also investigate just throwing an exception up, similar to how now the assert is thrown.

### DIFF
--- a/execution_chain/core/executor/process_block.nim
+++ b/execution_chain/core/executor/process_block.nim
@@ -741,6 +741,10 @@ proc procBlkEpilogue(
   if vmState.balTrackerEnabled:
     vmState.balTracker.commitCallFrame()
 
+  # Catch a fatal condition from the reward persist or the post-execution system
+  # calls before getStateRoot below, which would otherwise assert on it.
+  vmState.ledger.abortOnFatalError()
+
   if not skipValidation:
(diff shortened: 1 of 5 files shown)


kdeme:
See the exception based impact here: https://github.com/status-im/nimbus-eth1/pull/4434

Plus, we still need to flag in that case for the BLOCKHASH opcode fails. Also moving there to exceptions would have even bigger impact.

PR #4427, opened 2026-07-07
builds MUST start on forkchoiceState.headBlockHash

### LINKED ISSUE #4382: `generateExecutionBundle` builds on last imported block, not fcU-specified head
https://github.com/status-im/nimbus-eth1/blob/b6fdbd5b0d1e3e5e8cb0be9d6bc39c1d78f9b209/execution_chain/beacon/beacon_engine.nim#L149-L157
i.e.
https://github.com/status-im/nimbus-eth1/blob/b6fdbd5b0d1e3e5e8cb0be9d6bc39c1d78f9b209/execution_chain/core/chain/forked_chain.nim#L885-L886
but
https://github.com/status-im/nimbus-eth1/blob/b6fdbd5b0d1e3e5e8cb0be9d6bc39c1d78f9b209/execution_chain/core/chain/forked_chain.nim#L91-L99
called by
https://github.com/status-im/nimbus-eth1/blob/b6fdbd5b0d1e3e5e8cb0be9d6bc39c1d78f9b209/execution_chain/core/chain/forked_chain.nim#L228-L234 called by

sets `c.fcuHead`, not `c.latest`; rather, it does it while importing blocks:
https://github.com/status-im/nimbus-eth1/blob/b6fdbd5b0d1e3e5e8cb0be9d6bc39c1d78f9b209/execution_chain/core/chain/forked_chain.nim#L64-L79
and pruning:
https://github.com/status-im/nimbus-eth1/blob/b6fdbd5b0d1e3e5e8cb0be9d6bc39c1d78f9b209/execution_chain/core/chain/forked_chain.nim#L292-L308

But fcU-induced block assembly should use the fcU-specified head.

advaita-saha:
Fixes #4382 
- `api_forkchoice` passes headHash to `generateExecutionBundle`
- `generateExecutionBundle` resolves that header (erroring if unknown) and uses it for the parent-timestamp guard
- `assembleBlock` gained an mandatory parentHash and `updateVmState` an overload taking an explicit parent, so the packer VM runs on `chain.txFrame(headHash)`. All other callers have been updated to pass the parentHash

### DIFF
--- a/execution_chain/core/tx_pool.nim
+++ b/execution_chain/core/tx_pool.nim
@@ -166,9 +166,9 @@ func getWrapperVersion(com: CommonRef, timestamp: EthTime): WrapperVersion =
 
 proc assembleBlock*(
     xp: TxPoolRef,
+    parentHash: Hash32,
     someBaseFee: bool = false,
-    gasLimit: Opt[GasInt] = Opt.none(GasInt),
-    parentHash: Opt[Hash32] = Opt.none(Hash32)
+    gasLimit: Opt[GasInt] = Opt.none(GasInt)
 ): Result[AssembledBlock, string] =
   # Packing mutates the ledger (tx nonces/balances are persisted) and the
   # per-block accumulators (blobGasUsed, gas counters, receipts), and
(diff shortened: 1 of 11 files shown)


execution_chain/core/tx_pool.nim
```diff
@@ -167,7 +167,8 @@ func getWrapperVersion(com: CommonRef, timestamp: EthTime): WrapperVersion =
 proc assembleBlock*(
     xp: TxPoolRef,
     someBaseFee: bool = false,
-    gasLimit: Opt[GasInt] = Opt.none(GasInt)
+    gasLimit: Opt[GasInt] = Opt.none(GasInt),
+    parentHash: Opt[Hash32] = Opt.none(Hash32)
```
tersec: What is the use case for not specifying `parentHash`?

advaita-saha: Not sure
Just kept the option open for the interface, if it has some dependencies elsewhere 

If I am not very wrong, it has some old usage in the tests

advaita-saha: Not much dependency on this as I checked ( was there only one at the debug_rpc endpoint of test block building )
Makes sense to remove it

tersec: https://github.com/status-im/nimbus-eth1/pull/4427/commits/83bbf908dda7a955107eac202ea98b31b88f2352

tersec (approved):

jangko (approved):

PR #4434, opened 2026-07-07
stateless: fail block on incomplete witness during persist

kdeme:
Make persist raise a BlockAbortError on a write failure. Turn it into a validation error under stateless execution, and assert for stateful.

Note that we cannot do the same easily for the BLOCKHASH opcode error as it would mean changing the raises signature for VmOpFn.


kdeme:
closing this for https://github.com/status-im/nimbus-eth1/pull/4425

PR #4447, opened 2026-07-09
Updated client identification code to NE

rolfyone:
The code in use (NB) was for nimbus CL, and because they're position independent this was basically a collision to have Nimbus EL using NB as well.

I've also raised an execution PR to reserve `NE` for Nimbus EL.

reference PR https://github.com/ethereum/execution-apis/pull/848


tersec:
Fine merging this if the corresponding spec PR goes in.

arnetheduck:
is it a problem that it uses the same for both? ie there's only one nimbus at the end, it just happens to have two components

rolfyone:
> is it a problem that it uses the same for both? ie there's only one nimbus at the end, it just happens to have two components

Conceptually it's kinda weird because it appears in both sides of the stack... From the execution api it's defined CL's and EL's and theres a clear split except for Nimbus. I'm not sure how we'd differentiate what NB is acting as, it basically breaks the way we've specified the codes if there's no difference, it'd arguably look like 2 CL's building a full block...

rolfyone:
> Fine merging this if the corresponding spec PR goes in.

spec PR got merged...

PR #4449, opened 2026-07-09
--network now can look genesis.json into folder

### LINKED ISSUE #4387: Separate `--max-peers` in unified client
The unified client uses `--max-peers` for both CL and EL. When specified, it appears to apply to the CL.

A solution that doesn't break existing `nimbus executionClient` tooling could be to create an alias on the EL side, say `--el-max-peers`, that can be used when running unified.

### LINKED ISSUE #4388: Custom networks in unified client
I believe it is not possible to run a custom network (such as devnetX) in the unified client. I might be missing something, of course.

`--bootstrap-node` is a different format for CL and EL, but a shared parameter name across both. Ditto for `--network`, which should point to the config directory for the CL, and the genesis.json for the EL.

If this is indeed not working right now, a solution could be to create alias parameters on the EL side, so that existing `nimbus executionClient` tooling works, while unified tooling could use `--bootstrap-node` and `--el-bootstrap-node`. For `--network`, maybe just specifying it three times is the right move - id, directory and json. It might also already work that way: I haven't done testing with custom networks yet, this is all based on reading `--help` text at the moment.

jangko:
And also add alias --el-bootstrap-node and --el-bootstrap-file to --bootstrap-node and --bootstrap-file.

--network with an input path:
first it will look for a `genesis.json`, if there is no `genesis.json`, it will try to parse the existing json files, and accept the first match.

fix #4388
fix #4387
@yorickdowne


yorickdowne:
Awesome, thank you! Tried it, it works.

PR #4454, opened 2026-07-09
Adjust discovery flag as just turn on/off bool

kdeme:
Some questions here:
1. OK to adjust flag without backwards compatibility ? 
2. should we name it `--discovery` or `--discv5` to be consistent with `nimbus_beacon_node`? Keeping also in mind `nimbus` unified client
3. Should we wait for discv4 removal or not before making this cli change?

My answers:
1. I'd say yes as the nec is in alpha
2. Not sure here. I would change it to `--discv5` for consistency, but wished we had instead named it `--discovery` in `nimbus_beacon_node`. But I think even calling it `--discv5` is would be ok, even if for the time being it still means also running discv4.
3. I would change it the sooner the better, so that anybody that might want to only set 1 of the two currently doesn't get their setup broken in the future.

### DIFF
--- a/execution_chain/conf.nim
+++ b/execution_chain/conf.nim
@@ -263,10 +263,10 @@ type
       defaultValueDesc: "any"
       name: "nat" .}: NatConfig
 
-    discovery* {.
+    discv5* {.
       desc: "Enable peer discovery. When disabled, peers must be added manually"
       defaultValue: true
-      name: "discovery" .}: bool
+      name: "discv5" .}: bool
 
     netKey* {.
       desc: "P2P ethereum node (secp256k1) private key (random, path, hex)"
--- a/execution_chain/conf.nim
+++ b/execution_chain/conf.nim
@@ -75,10 +75,6 @@ type
     Debug                         ## enable debug_ set of RPC API
     Admin                         ## enable admin_ set of RPC API
 
-  DiscoveryType* {.pure.} = enum
-    V4
-    V5
-
   ExecutionClientConf* = object
     ## Main configuration for the execution client - when updating, coordinate
     ## options shared with other executables (logging, metrics etc)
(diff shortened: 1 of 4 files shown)


tersec:
> Some questions here:
> 
>     1. OK to adjust flag without backwards compatibility ?
> 
>     2. should we name it `--discovery` or `--discv5` to be consistent with `nimbus_beacon_node`? Keeping also in mind `nimbus` unified client
> 
>     3. Should we wait for discv4 removal or not before making this cli change?
> 
> 
> My answers:
> 
>     1. I'd say yes as the nec is in alpha
> 
>     2. Not sure here. I would change it to `--discv5` for consistency, but wished we had instead named it `--discovery` in `nimbus_beacon_node`. But I think even calling it `--discv5` is would be ok, even if for the time being it still means also running discv4.
> 
>     3. I would change it the sooner the better, so that anybody that might want to only set 1 of the two currently doesn't get their setup broken in the future.

Broadly agree with all of this.

kdeme:
> Broadly agree with all of this.

OK, then I'll change the flag to `--discv5` like it is in `nimbus_beacon_node`. I'll check if there are any scripts here or in hive that use this. And then we can already merge this.

tersec:
> I'll check if there are any scripts here or in hive that use this. And then we can already merge this.

Did any remaining blockers emerge?

kdeme:
> 
> Did any remaining blockers emerge?

No, I've scanned this repo and hive repo and discovery flag is nowhere explicitly set. 

I did notice it gets disabled in this script in nimbus-eth2: https://github.com/status-im/nimbus-eth2/blob/4110bc7828a45518d22d60e2f60438ae81ff17e9/scripts/start_nimbus_el_nodes.sh#L47

I'll create a PR for that and its not a blocker for this to get merged.

PR #4456, opened 2026-07-10
remove nested `waitFor`

arnetheduck:
These are not well defined in chronos


execution_chain/sync/sync_sched.nim
mjfh, line 1: OK, might be some legacy code. Will test that anyway.

PR #4476, opened 2026-07-13
networking: remove DiscoveryV4 support, use DiscoveryV5 only

kdeme:
Removes discv4 protocol, kademlia, shared port hack and related tests.


kdeme:
In draft until it's actually OK to remove (Which would mean a nethermind release with default discv5 enabled out for certain amount of time).

This is the basic removal of discv4.. There is however still quite some boiler plate (the extra discovery file)  that can be removed if some things get restructured.
And the same for the Enode conversion which can be totally skipped now (and the `Node` instances).
Might tackle this here in follow-up commits.

PR #4479, opened 2026-07-13
Update to zkevm v0.6.1 and spec related changes

kdeme:
This is an update of zkevm fixtures and code changes to stateless chain config and schema id as per spec for zkevm v0.6.1

These are however changes on top of tests-glamsterdam-devnet@v7.2.0 so we will first need a new devnet 7 branch.


kdeme:
This has been cherry-picked in https://github.com/status-im/nimbus-eth1/pull/4498

PR #4477, opened 2026-07-13
stateless: Fix witness tracking for EIP-7997 related access

kdeme:

PR #4480, opened 2026-07-13
stateless: Fix witness tracking for EIP-7997 related access

kdeme:


kdeme:
I think I prefer this version over https://github.com/status-im/nimbus-eth1/pull/4477

kdeme:
Moved to draft as it is not clear if this access should be tracked or not.

Currently the execution-specs do not have this implemented, hence the difference in witness. In practice it would also be unneeded to. It's not a real tx based access.

But both go-ethereum and nimbus do it like this at the moment.

kdeme:
Closing this as it is no longer needed with https://github.com/status-im/nimbus-eth1/pull/4580

PR #4493, opened 2026-07-15
add debug flag to enable txpool by default for hive

advaita-saha:


tersec:
The premise of this seems questionable. It's creating a special Hive-only mode, which seems to defeat the point of Hive as a test of Nimbus as actually deployed.

advaita-saha:
Other clients use sync check ( is the syncer is running )
we used a time based check ( I actually like this better, as it saves bandwidth )

So I don't want to switch to the behaviour like other clients, thus came up with this flag
I am not sure about this either

PR #4498, opened 2026-07-15
glamsterdam devnet 7

jangko:


kdeme:
I've cherry picked the commit from https://github.com/status-im/nimbus-eth1/pull/4479 and re-enable the stateless tests, with some added fixtures in skipFiles.

kdeme:
OOM on windows perhaps because latest zkvm fixtures...have more test cases? Not much else changed. Unless the runners themselves just have less memory sometimes.

And then another PR that can get merge into this branch if approved: https://github.com/status-im/nimbus-eth1/pull/4509

jangko:
> OOM on windows perhaps because latest zkvm fixtures...have more test cases? Not much else changed. 

both `eest_blockchain_test` and `eest_engine_test` peak memory consumption is around 8GB+(not good either).
but `eest_stateless_execution_test` consume 12GB+ of memory even on linux, not only on windows. 
The CI provides 16GB memory for each platform. 

`eest_t8n_test` and `eest_evmstate_test` consume less memory at around 2.7GB, and when the GC kicks in, the consumption lowered to below 1GB. But they also have their own parser/framework that different from other eest test.

Looks like there is something wrong with our eest test framework/parser.

PR #4503, opened 2026-07-16
skip malformed enode entry

advaita-saha:


tests/networking/test_bootnodes.nim
```diff
@@ -34,3 +34,29 @@ suite "Bootnodes":
...
+    ]
+    var boot: BootstrapNodes
+    let res = parseBootstrapNodes(bn, boot)
+    # The bad entry is still reported...
+    check res.isErr
+    check "incorrect TCP port" in res.error
+    # ...but the four valid nodes are accepted instead of being dropped.
+    check boot.enodes.len == 4
+
+  test "a fully valid list still succeeds":
```
tersec: As a phrasing note, perhaps drop `still` here; it reads fine as part of a narrative progression within this test module, but less so in Nimbus test output

advaita-saha: https://github.com/status-im/nimbus-eth1/pull/4503/commits/87eb9bc8f560dd38d3b2080d2b2c013216f7b729

execution_chain/networking/bootnodes.nim
```diff
@@ -153,16 +153,29 @@ proc loadBootstrapNodes*(fileName: string, boot: var BootstrapNodes): Result[voi
   let text = io2.readAllChars(fileName).valueOr:
     return err($error)
 
+  var errors: seq[string]
```
tersec: It seems like this part of `loadBootstrapNodes` is basically `parseBootstrapNodes`? It's suboptimal to have to repeat the error handling in both places, and means any updates to it have to be done in both, or else the error-handling policies diverge. Probably preferable to refactor so `loadBootstrapNodes` does the load and the `splitLines` then calls a common/shared helper.

advaita-saha: https://github.com/status-im/nimbus-eth1/pull/4503/commits/8081d64d23eb2bc83aea3f7555a95b72a6aba64d

tersec (approved):

kdeme (approved):

PR #4512, opened 2026-07-17
async block builder implementation

advaita-saha:


tersec:
Most ELs seem to have a continuous-payload-improvement loop as well -- send a `PayloadAttributes` fcU, then keep sending `getPayload` and it improves gradually. This PR is a good initial improvement, so the concept is worthwhile on its own, and not waiting for such an ongoing loop, but worth considering if it's worthwhile. If nothing else, transactions keep coming into the pool as time passes, so even the same algorithm can improve over time.

(Also, merge conflict in `engine_api.nim`.)

advaita-saha:
> continuous-payload-improvement 

Yeah good idea, but out of scope of this PR. Can do a followup PR
One caveat would be, engineAPI based tests currently can assume fcU → one deterministic block. With a rebuild loop, the block you get depends on when you asked. Tests would need a way to pin/freeze the loop.

advaita-saha:
Noticed an increase of proposal misses in hoodi with this branch

tersec:
Any idea why?

PR #4521, opened 2026-07-19
Use BoringSSL P-256 for the P256VERIFY precompile

bhartnett:
Performance testing results using benchmarkoor running the p256verify tests:

**Before**
Screenshot_2026-07-20_10-45-59

**After**
Screenshot_2026-07-20_10-59-27


tersec:
Benchmarks?

bhartnett:
> Benchmarks?

See above. Looks like a 4-5x speedup.

PR #4522, opened 2026-07-20
Use BoringSSL for the modexp precompile

bhartnett:
Performance testing results using benchmarkoor running the modexp tests:

**Before (libtommath 64bit)**
623584940-53c1c728-cdfe-4242-9e32-82a7326d3cc1

**After**
Screenshot_2026-07-20_20-08-07

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -201,11 +201,6 @@
 	url = https://github.com/vacp2p/nim-intops.git
 	ignore = untracked
 	branch = master
-[submodule "vendor/libtommath"]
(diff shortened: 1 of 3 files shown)


advaita-saha:
Faster than #4519 ?

bhartnett:
> Faster than #4519 ?

Yes looks like around twice as fast as the previous version even after the last improvement (at least on linux).

jangko:
please remove libtommath submodule

jangko (approved):

PR #4530, opened 2026-07-21
proxy: EIP-2935 anchor jump

chirag-parmar:


nimbus_verified_proxy/engine/types.nim
```diff
@@ -277,6 +278,10 @@ type
     filter*: FilterOptions
     blockMarker*: Opt[Quantity]
 
+  EngineState* = object
+    archive*: bool
+    privateTx*: bool
```
tersec: Seems to be only written to but not read from

nimbus_verified_proxy/library/verifproxy.nim
```diff
@@ -136,6 +147,7 @@ proc load(T: type VerifiedProxyConf, configJson: string): T {.raises: [ProxyErro
       of "Auto": StdoutLogKind.Auto
       else: StdoutLogKind.None
     maxBlockWalk = jsonNode.getOrDefault("maxBlockWalk").getBiggestInt(1000)
+    maxWindowJumps = jsonNode.getOrDefault("maxWindowJumps").getBiggestInt(128)
```
tersec: CLI uses 500 and this uses 128 -- deliberately different or?

chirag-parmar: not deliberate. missed during copy paste.

chirag-parmar: https://github.com/status-im/nimbus-eth1/pull/4530/commits/f063f10f4c7facaa3ecd40425b564d272a1c897d

nimbus_verified_proxy/engine/types.nim
chirag-parmar: meant for consistency and future upgrades. But I agree it is redundant for the current functioning of the verified proxy. Should I remove it? (kept it in there because it is the only other thing the engine is blind about in the running context)

tersec: Yeah, preferably. Easier to add these kinds of things later than remove them, in API terms

tersec: https://github.com/status-im/nimbus-eth1/pull/4530/commits/28ceb7de8bd339f86f3d851c3f45a1e498b8a72e

PR #4531, opened 2026-07-21
Fix aristo_compute compilation with --threads:off

kdeme:
Gate references to the threads-only `lock`/`taskpool` fields behind threads compile flag so the module compiles when threads are disabled.


bhartnett:
@kdeme I've pushed an update to this so that more of the code is disabled when threads are off.

bhartnett (approved):

PR #4533, opened 2026-07-21
Support ConcurrentLruCache with --threads:off

kdeme:
The sharded ConcurrentLruCache needs ReadWriteLock / Atomic / countProcessors, unavailable without --threads:on. Gate it behind `when compileOption("threads")` and add a non-threaded fallback wrapping the plain LruCache, also with withGet/withPeek reading in place via getPtr/peekPtr.

Benchmarking (bench_lru.nim): minilru vs the local LruCache shows no measurable difference for int values, but with larger values the zero-copy get/peek is clearly faster. The threadSafe=false path also performs the same as the --threads:off fallback.

### DIFF
--- a/execution_chain/concurrency/lru.nim
+++ b/execution_chain/concurrency/lru.nim
@@ -16,7 +16,10 @@
 
 {.push raises: [], gcsafe.}
 
-import std/[atomics, cpuinfo, hashes, math, typetraits], results, ./readwritelock
+import std/[hashes, math, typetraits], results
+
+when compileOption("threads"):
+  import std/[atomics, cpuinfo], ./readwritelock
 
 export hashes, results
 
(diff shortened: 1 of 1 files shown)


kdeme:
Draft because I want to investigate just not having the cache in the first place for stateless.

(noticed it gets disabled anyhow at runtime for stateless)

bhartnett:
> Draft because I want to investigate just not having the cache in the first place for stateless.
> 
> (noticed it gets disabled anyhow at runtime for stateless)

Yeah caches should be completely disabled for stateless since it is a different scenario where all the state is provided already in the witness and all already stored in memory after we load the subtree into the in memory db. The ledger still has single threaded minilru caches which can be kept for performance to skip walking the txFrames but doesn't make sense to use the caches in the db layer.

kdeme:
> > Draft because I want to investigate just not having the cache in the first place for stateless.
> > (noticed it gets disabled anyhow at runtime for stateless)
> 
> Yeah caches should be completely disabled for stateless since it is a different scenario where all the state is provided already in the witness and all already stored in memory after we load the subtree into the in memory db. The ledger still has single threaded minilru caches which can be kept for performance to skip walking the txFrames but doesn't make sense to use the caches in the db layer.

Yeah, I've noticed that. It's however a bit of mechanical work in more different locations to not have it compiled in. But I can make a PR that works like that.

kdeme:
Closing for https://github.com/status-im/nimbus-eth1/pull/4551

PR #4536, opened 2026-07-22
optimize PREVRANDAO benchmarks

tersec:
Benchmarkoor results files:
[result_prevrandao_opt.json](https://github.com/user-attachments/files/30252918/result_prevrandao_opt.json)
[result_statusquo.json](https://github.com/user-attachments/files/30252919/result_statusquo.json)

Each is aggregate time running the PREVRANDAO tests in `master` compared with this PR. Mgas/s in `master` is flat at 4Mgas/s across across the sizes (it's mostly measuring database read times, because it's taking the slow path in https://github.com/status-im/nimbus-eth1/blob/1728ad55f0185bff4ae5dd6595ac5e88f5202d6f/execution_chain/common/common.nim#L507-L515
`  1M` gas: 0.252s in `master`; 0.005s in this branch (217Mgas/s)
`  5M` gas: 1.248s in `master`; 0.019s in this branch (279Mgas/s)
` 10M` gas: 2.56s in `master`; 0.033s in this branch (311Mgas/s)
` 30M` gas: 7.77s in `master`; 0.102s is this branch (297Mgas/s)
` 60M` gas: 15.373s in `master`; 0.196s in this branch (308Mgas/s)
`100M` gas: 25.711s in `master`; 0.315s in this branch (318Mgas/s)
`150M` gas: 39.159s in `master`; 0.474s in this branch (317Mgas/s)

There's a tradeoff here regarding always running
```nim
  self.proofOfStake = com.proofOfStake(Header(
    number: parent.number + 1,
    parentHash: blockCtx.parentHash,
    difficulty: blockCtx.difficulty,
  ), ledger.txFrame)
```
regardless of whether that any given block will actually contain it, but it's both slow enough that makes a significant difference hoisted out of a hot inner loop, yet fast enough that running it once costs on the order of milliseconds or less even from RocksDB. There are other approaches here, but they potentially create data races in parallel BALs, require pre-scanning transactions for the `PREVRANDAO` opcode,
 or have other architectural tradeoffs.

### DIFF
--- a/execution_chain/evm/state.nim
+++ b/execution_chain/evm/state.nim
@@ -8,10 +8,10 @@
 # at your option. This file may not be copied, modified, or distributed except
 # according to those terms.
 
-{.push raises: [].}
+{.push raises: [], gcsafe.}
 
 import
-  std/[options, sets, strformat],
+  std/strformat,
   stew/assign2,
   ../db/ledger,
   ../common/common,
@@ -56,6 +56,11 @@ proc init(
   self.allLogs.setLen(0)
   self.gasRefunded = 0
   self.balTracker = tracker
+  self.proofOfStake = com.proofOfStake(Header(
+    number: parent.number + 1,
+    parentHash: blockCtx.parentHash,
+    difficulty: blockCtx.difficulty,
+  ), ledger.txFrame)
 
 func blockCtx(header: Header): BlockContext =
   BlockContext(
@@ -72,7 +77,7 @@ func blockCtx(header: Header): BlockContext =
 
 # --------------
 
-proc `$`*(vmState: BaseVMState): string =
+func `$`*(vmState: BaseVMState): string =
   if vmState.isNil:
     result = "nil"
   else:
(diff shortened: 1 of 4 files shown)


execution_chain/evm/state.nim
```diff
@@ -56,6 +56,11 @@ proc init(
   self.allLogs.setLen(0)
   self.gasRefunded = 0
   self.balTracker = tracker
+  self.proofOfStake = com.proofOfStake(Header(
+    number: parent.number + 1,
+    parentHash: blockCtx.parentHash,
+    difficulty: blockCtx.difficulty,
+  ), ledger.txFrame)
```
bhartnett, line 63: For parallel execution and optimistic state prefetch we create a new vmState per transaction so this bool would need to be set in those contexts as well (in process_block_parallel.nim). Ideally only computed once per block so the bool would need to be passed into the tasks from the main thread where the value is pre-computed.

bhartnett, line 63: Not sure why we even need this proofOfStake flag considering that we have the fork and hardFork in the vmState already. We might be able to avoid this by just checking if the fork is greater than or eq to the merge fork or something similar.

bhartnett:
I can confirm that this does fix the slow randao benchmarkoor tests.

execution_chain/evm/state.nim
bhartnett: @tersec I think we can do this instead: https://github.com/status-im/nimbus-eth1/pull/4538

tersec: https://github.com/status-im/nimbus-eth1/pull/4536/commits/a389bc2eef0ecc71d3855edc8795604ac0a17186

tersec: That's what the EEST failures were -- the parallel BALs tests failing.

tersec: Wouldn't this fail precisely around the actual merge fork? i.e. it works cleanly before/after, but there's one fork during which both `DIFFICULTY`/`PREVRANDAO` are in sequence valid, precisely determined by the TTD.

tersec: Whether or not we care about this fork anymore is a different question, but the tests might still cover it.

bhartnett: This works because shanghai is always a proof of stake fork. In other words all shanghai blocks should be proof of stake because the merge transition based on TTD happens during the merge fork (the previous fork).

tersec:
- #4538

is better.

PR #4538, opened 2026-07-22
Optimize proofOfStake check for PREVRANDAO opcode

bhartnett:
When running the benchmarkoor tests (`filter: "regex:.*block_context.*"`) the PREVRANDAO opcode shows up as  slow in some tests (in red). This improves the performance by skipping the proof of stake calculation when the fork is already past the merge fork.

**Before**
Screenshot_2026-07-22_15-08-00

**After**
Screenshot_2026-07-22_15-08-40


tersec:
This is safe/correct because
- https://github.com/ethereum/consensus-specs/pull/3350

i.e. it's guaranteed that the merge has happened when `vmState.fork >= FkShanghai`.

tersec (approved):

PR #4543, opened 2026-07-22
Export hashes_rlp from aristo_compute for computeKey callers

kdeme:
computeKey is generic, so its account rlp encoding needs append(Hash32) in the caller's scope, which (some) callers didn't have (not explicitly imported). So it sometimes failed to resolve depending on ordering. Exporting hashes_rlp should fix it for any ordering.


kdeme:
Note, this was hitting an actual compile error for me on some recent master commit compiled with threads:off

PR #4547, opened 2026-07-23
Use BLST pippenger for BLS MSM precompiles

bhartnett:
BLS12-381 MSM both G1 and G2 show up as some of the slowest precompiles in the benchmarkoor tests.

This improves the performance by using the blst mult_pippenger library functions.

**Before**
Screenshot_2026-07-23_12-20-25

**After**
Screenshot_2026-07-23_15-16-00


jangko (approved):

advaita-saha:
A little surprised to see, we were not using pippenger earlier

Maybe we can also check other precompiles, and compare with optimizations added in constantine
Since integrating that have been an issue, better port those optimizations to our implementations or use something similar

bhartnett:
> A little surprised to see, we were not using pippenger earlier
> 
> Maybe we can also check other precompiles, and compare with optimizations added in constantine Since integrating that have been an issue, better port those optimizations to our implementations or use something similar

I'm going through all the precompiles that show up as slow in the benchmarks (starting with the slowest). Yeah true, we can get inspiration from the Constantine library when applicable.

PR #4550, opened 2026-07-23
Fix EVM stack memory leak from double setCode on system calls

kdeme:
System call frames set their code twice (setupComputation, then prepareDispatch), and each setCode allocated a fresh stack, orphaning the first one. This leak drove the EEST stateless suite to ~17GB VIRT and OOM on CI. setCode now allocates the buffers only when absent and reuses the still empty stack.

Also add dispose for the computation on two other early return paths that skipped it.

### DIFF
--- a/execution_chain/evm/computation.nim
+++ b/execution_chain/evm/computation.nim
@@ -143,8 +143,11 @@ func getTransientStorage*(c: Computation, slot: UInt256): UInt256 =
 func setCode*(c: Computation, code = CodeBytesRef(nil)) =
   if not code.isNil:
     c.code = CodeStream.init(code)
-    c.memory = EvmMemory.init()
-    c.stack = EvmStack.init()
+    # A frame created with code (system calls) has setCode run again from prepareDispatch.
+    # Allocate the buffers only once so the empty stack is not orphaned.
+    if c.stack.isNil:
+      c.memory = EvmMemory.init()
+      c.stack = EvmStack.init()
 
 func newComputation*(vmState: BaseVMState,
                      keepStack: bool,
--- a/execution_chain/evm/interpreter/op_handlers/oph_create.nim
+++ b/execution_chain/evm/interpreter/op_handlers/oph_create.nim
@@ -79,7 +79,9 @@ proc execSubCreate(c: Computation; childMsg: Message;
   if c.fork >= FkAmsterdam:
     newAccountCharged = not c.accountExists(child.msg.contractAddress)
     if newAccountCharged:
-      ? c.gasMeter.chargeStateGas(CREATE_ACCOUNT_STATE_GAS, "Create op new account")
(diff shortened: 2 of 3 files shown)


execution_chain/transaction/system_call.nim
```diff
@@ -96,4 +96,8 @@ proc systemCall*(params: CallParams, T: type): T =
   if c.isSuccess:
     c.execCallOrCreate(params)
     ledger.persist(clearEmptyAccount = true)
+  else:
+    # execCallOrCreate normally disposes the computation, dispose here too
+    # otherwise the EVM stack leaks.
+    c.dispose()
```
kdeme, line 102: This leak was not actually being hit in the tests, but adding it anyhow as setupComputation does allocate

execution_chain/evm/interpreter/op_handlers/oph_create.nim
```diff
@@ -79,7 +79,9 @@ proc execSubCreate(c: Computation; childMsg: Message;
   if c.fork >= FkAmsterdam:
     newAccountCharged = not c.accountExists(child.msg.contractAddress)
     if newAccountCharged:
-      ? c.gasMeter.chargeStateGas(CREATE_ACCOUNT_STATE_GAS, "Create op new account")
+      c.gasMeter.chargeStateGas(CREATE_ACCOUNT_STATE_GAS, "Create op new account").isOkOr:
+        child.dispose()
```
kdeme, line 83: This leak was being hit a few times, but it was so minor that it wasn't really noticeable in tests.

execution_chain/evm/computation.nim
```diff
@@ -143,8 +143,11 @@ func getTransientStorage*(c: Computation, slot: UInt256): UInt256 =
 func setCode*(c: Computation, code = CodeBytesRef(nil)) =
   if not code.isNil:
     c.code = CodeStream.init(code)
-    c.memory = EvmMemory.init()
-    c.stack = EvmStack.init()
+    # A frame created with code (system calls) has setCode run again from prepareDispatch.
+    # Allocate the buffers only once so the empty stack is not orphaned.
+    if c.stack.isNil:
+      c.memory = EvmMemory.init()
+      c.stack = EvmStack.init()
```
kdeme, line 150: This was the main issue.

Note that we could also decide to not `init` the `CodeStream`, because afaik there is no path that practically hits this twice where the code actually changes? But done like this to be safe.

Same reason why I didn't move it actually to `prepareDispatch` in front of `setCode` call there. Although, perhaps that would be better to make sure we don't do this check when not needed (all other `setCode` usage).

However, I kind of consider this a quick fix, and some better architectural change might be required / preferable still.

kdeme:
Note that the leak caused to fail at specific stateless eest testing enabling, because the stateless tests run the execution 3 times (Once statefull, once stateless with fixture input, once stateless with our generated input), hence aggravating the leak in those tests. The other tests were also hitting the leak however.

PR #4551, opened 2026-07-23
Disable leaves caches when compiling with threads disabled

kdeme:
The current only threads disabled use case is the stateless guest, for which these caches are not useful anyhow.


kdeme:
This is the better alternative for https://github.com/status-im/nimbus-eth1/pull/4533

See also comment https://github.com/status-im/nimbus-eth1/pull/4533#issuecomment-5037014646

bhartnett (approved):

PR #4552, opened 2026-07-23
System call should not call topFrameAuthAndDelegation

jangko:
This is a follow up after #4550.
`topFrameAuthDelegation` should only be called by message call/create and not by system call.
Check for `MsgFlags.SystemCall` not needed in `beforeExecCreate` because a system call top frame only call `beforeExecCall`.

### DIFF
--- a/execution_chain/evm/computation.nim
+++ b/execution_chain/evm/computation.nim
@@ -141,13 +141,15 @@ func getTransientStorage*(c: Computation, slot: UInt256): UInt256 =
     cpt = cpt.parent
 
 func setCode*(c: Computation, code = CodeBytesRef(nil)) =
+  # If we call setCode when c.code already set to something,
+  # it means c.memory and c.stack also have been set.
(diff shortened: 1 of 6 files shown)


kdeme:
I guess we could then also remove the additional check that I had added in setCode, as it won't be run now anymore?

jangko:
> I guess we could then also remove the additional check that I had added in setCode, as it won't be run now anymore?

Turn it into an assertion check, calling it again when stack and memory object have been created is a programming bug.

https://github.com/status-im/nimbus-eth1/pull/4552/commits/2e31b028ec4ae959ba72432c09990e667e71a6fd

And you're right about the architectural changes too. kudos!

https://github.com/status-im/nimbus-eth1/pull/4552/commits/6f6fe6845b09bdfa101345fcfc12cc6ab63ca0f7

With this PR merged, I think we can conclude glamsterdam-devnet-7 branch.

PR #4563, opened 2026-07-25
fix: syncer stalling from el_sync

### LINKED ISSUE #4537: syncer stalls due to el_sync
### The one-line cause
A concurrent importer (el_sync) overshoots the devp2p syncer's committed target T, which permanently disables the syncer's catch-up valve (blocksUnprocReconcile can't resolve a block above its own session head). The session is then stuck one step short of done, and because a "done" session is the only thing that unlocks the header cache, every subsequent `forkchoice` update from the CL is silently discarded.

### Situation at the point of stall
The wedge needs three things true at once: `syncState == blocks, topNum < headNum` (job unfinished), and the fetch queue empty (avail == 0). Here's how a healthy syncer arrives there.

### How we arrive at this position
The node runs two block importers against one ForkedChain:
  - `el_sync` — reads blocks straight from the CL's DAG and feeds them to the engine API. Faster
  - `devp2p beacon syncer `— fetches block bodies from peers. Slower

So the ordering that strands it:
  1. Syncer targets `T`, to-do list `[base+1 .. T]`
  2. `el_sync` races ahead. While `fcLatest ≤ T`, reconcile keeps `topNum` glued to `fcLatest` and trims the to-do list to `(fcLatest .. T]`. Still healthy.
  3. `el_sync`'s head crosses `T`. Reconcile goes permanently silent. `topNum` freezes at its last value, strictly below `T`, with `(topNum .. T]` still on the to-do list.
  4. Those leftover blocks were all already imported by `el_sync` — but the syncer doesn't know that, so a peer picks up the range to fetch it. This moves the range from the to-do list into borrowed.
  5. That peer disconnects/errors before finishing the import (peer churn is constant — the log shows Connect timeout ×137, handshake drops ×232). The borrowed range is never returned (only a successful blocksUnprocCommit returns it ).

**_Now: to-do list empty (avail == 0) → `blocksCollectOk` is `false` for every peer → nobody ever fetches again → `topNum` can never advance → `topNum < headNum` forever → session never finishes → `hdrCache` stays locked → every incoming FCU is dropped. Fully wedged._**

## The blind spot
Anticipated this race and added `blocksUnprocReconcile`, called at the top of every fetch/import step. Each tick it says: "el_sync moved the FC head forward — fast-forward my topNum to match and drop the blocks I no longer need to fetch."

But it can only fast-forward to a block it can look up in its own session:
```nim
  let fcLatest = ctx.chain.latestNumber
  let lineageHash = ctx.hdrCache.getHash(fcLatest).valueOr: return   # ← blind spot
```
getHash only resolves blocks inside the session range. So the valve works only while el_sync stays at or below T. The instant el_sync's imported head passes T, `getHash(fcLatest)` fails and reconcile bails at its first line — every time after that

advaita-saha:
Fixes #4537 

The problem
- Two things import blocks: el_sync (fast) and the devp2p syncer.
- el_sync races ahead to block 10. The syncer is still stuck aiming at block 8.
- The syncer waits for blocks 6,7,8 to arrive... but they're already imported. So it waits. Forever.

Fix
- Add the check - is my target (block 8) already sitting in the chain, imported, on the real canonical line? If yes, great, jump my counter to 8, mark session done, move on
- Every heartbeat, run the check. So even a node with zero peers un-sticks itself. As the el_sync is always running


execution_chain/sync/beacon/worker/update.nim
mjfh, line 1: This approach looks suspicious for two reasons.

*Firstly, running a a job (quasi-)parallel to downloading and importing peers should not go here. This is a **state-transition** function that should not do other actions. The right place is `worker.daemon()`, the function calling the state updater (I would move that to the right place when I do some syncer clean up,)

*Secondly and more important, this approach of running a background function changing the syncer work context will lead to other problems. If it does not work when placing to the peers, it will not work here in general, either. There is no control when a (quasi-)parallel job is continued. If the download state sanity depends on it, then this becomes a **race condition** asking for more trouble (I hope I am wrong.)

The right way to react to serious external changes in a particular way is via the FSA: suspending all peers (which costs some waiting time if there are many peers, though) and entering a new special-purpose syncer state.

execution_chain/sync/beacon/worker/blocks/blocks_unproc.nim
mjfh, line 1: It seems, that the approach of updating the syncer state work context (i.e. the TODO book keeping) on-the-fly does not work reliably (see other comment at `update.nim`) or is not fully manageable. Otherwise there would not be that update. Did that original solution introduce a race condition?

mjfh (changes requested):

mjfh:
> Fix
> 
>     * Add the check - is my target (block 8) already sitting in the chain, imported, on the real canonical line? If yes, great, jump my counter to 8, mark session done, move on
> 
>     * Every heartbeat, run the check. So even a node with zero peers un-sticks itself. As the el_sync is always running

The solution should be event driven. Apparently, the problem can be detected by the peers in action. So let the peers do the book keeping update for its particular job (by just pretending the job is done, maybe partially .)

advaita-saha:
Closed in favour of #4574

PR #4567, opened 2026-07-26
optimize bls12 pairing check

advaita-saha:
run one batched Miller loop instead of K separate ones multiplied together


advaita-saha:
Closing in favor of #4577 
Both does the same optimization, and have similar perf improvement

PR #4570, opened 2026-07-27
Suppress unified client complains about --rpc

### LINKED ISSUE #4461: Nimbus Unified complains about `--rpc`
Nimbus Unified on the CL side complains about the `--rpc` flag. It still works, the EL has an RPC running as it should. The error message is cosmetic.

` WRN 2026-07-12 08:45:17.465+01:00 Ignoring deprecated configuration option   topics="beacnde" opt=rpc msg="Superceded by REST API as of v1.7.0"`

jangko:
fix #4461


execution_chain/nimbus.nim
```diff
@@ -211,8 +211,17 @@ proc elSyncLoop(
       # startup
       debug "Execution client not ready", err = exc.msg
 
+proc filterCommandLineParams(): seq[string] =
+  # nimbus-eth2 has made rpc* flags obsolete and print obsolation message to the terminal.
+  # While at the same time nimbus-eth1 is using rpc* flags to configure it's rpc server.
+  # So we need to remove the confusing obsolation message when invoking unified client.
+  for x in commandLineParams():
+    if "rpc" notin x:
```
advaita-saha: Can we maintain a filter list instead of using `rpc` harcoded in the if statement?
Might be useful in the future to add more easily

tersec: Another possibility is just removing the message from `nimbus-eth2`

advaita-saha: I like this better

tersec: - https://github.com/status-im/nimbus-eth2/pull/8807

PR #4574, opened 2026-07-27
update topNum when blocks imported are behind base

### LINKED ISSUE #4537: syncer stalls due to el_sync

advaita-saha:
Fixes #4537 

Refer : https://github.com/status-im/nimbus-eth1/issues/4537#issuecomment-5042484639

### DIFF
--- a/execution_chain/sync/beacon/worker/blocks/blocks_blocks.nim
+++ b/execution_chain/sync/beacon/worker/blocks/blocks_blocks.nim
@@ -248,6 +248,8 @@ template blocksImport*(
         if nthBn <= ctx.chain.baseNumber:
           trace "Ignoring block less eq. base", peer, blk=nthBn,
             B=ctx.chain.baseNumber, L=ctx.chain.latestNumber
+          ctx.updateLastBlockImported nthBn          # block already imported
+          ctx.updateEtaBlocks()
           blocks[n].reset()
           continue
 


mjfh:
> Fixes #4537
> 
> Refer : [#4537 (comment)](https://github.com/status-im/nimbus-eth1/issues/4537#issuecomment-5042484639)

Is the `blocksUnprocReconcile()` function still needed?

execution_chain/sync/beacon/worker/blocks/blocks_blocks.nim
mjfh, line 1: :)

mjfh (approved):

advaita-saha:
for correctness, no — it's no longer needed. For efficiency, it still earns its keep

PR #4580, opened 2026-07-28
Remove EIP-7997 in-protocol factory deployment

kdeme:
EIP-7997 has been altered to be purely informative and no longer requires any in-protocol deployment of the deterministic-deployment factory.

Allows us also to remove a lot of skipped tests from zkevm eest test fixtures. These were mismatching witness as applyEip7997 would cause added accesses in the witness.


kdeme:
See also current version of EIP: https://eips.ethereum.org/EIPS/eip-7997

And the discussion at ACDT (https://github.com/ethereum/pm/issues/2170 & https://www.youtube.com/watch?v=Xh2_yW-P1Mw), which resulted in going for current spec/eip version, see also closing of:
- https://github.com/ethereum/EIPs/pull/11991
- and https://github.com/ethereum/EIPs/pull/12009

execution_chain/core/executor/process_block.nim
```diff
@@ -120,9 +119,6 @@ proc procBlkPreamble(
     if com.daoForkSupport and com.daoForkBlock.get == header.number:
       ledger.applyDAOHardFork()
 
-    if com.amsterdamTransition(vmState.parent.timestamp, header.timestamp):
```
jangko, line 123: nowhere else is using `amsterdamTransition`, remove it too from CommonRef?

PR #4582, opened 2026-07-29
devnet 8 will be discv5-only

tersec:


kdeme:
If they require it to be selectable we can also still add a `--debug-discv4` flag (not sure why). Else, this is fine.

tersec:
> If they require it to be selectable we can also still add a `--debug-discv4` flag (not sure why). Else, this is fine.

https://notes.ethereum.org/@ethpandaops/glamsterdam-devnet-8
> EL discovery runs **discv5-only** (per-client flags under Networking)
>
> ...
>
> **discv5-only verification**: each EL must publish — and refresh at the fork — the eth forkid ENR entry on its discv5 record; geth's dial-candidate filter drops nodes without it. Confirm bootnode ENR distribution, and watch for besu's silent V4 fallback masking failures.
>
> ...
>
> **EL discovery: discv5-only.** Per-client flags:

Pretty sure discv4 isn't going to be turned back on.

PR #4591, opened 2026-07-31
Add engine_newPayloadWithWitnessV4 and V5

kdeme:
Implement the newPayloadWithWitness methods, returning the execution witness (geth/ethrex ExtWitness RLP) on PayloadStatusV1. Not yet part of the Engine API spec.

Requires --stateless-provider to produce a witness. Might have to adjust this to be turned on just for the call in the future.


kdeme:
I initially had kept this PR in draft because the added calls are not actually specified yet (and potentially never will be with the upcoming of REST API).

However, this becomes quite cumbersome to maintain, so I think I'll have it merged into master (even though that means in the future removing the calls again for cleanup, but not the underlying logic, as the REST version would still use that).

PR #4601, opened 2026-08-04
stateless: Fix witness generation for precompile

kdeme:
The execution-specs still access the account which causes that to end up in the witness. We always shortcut that for precompiles as it isn't necessary for precompiles and wastes an access.

Add the access anyhow when collectWitness is set.


kdeme:
Tested locally, this currently doesn't allow strict compare of witness, as we have less than the ones generated by python spec. 

Fixes about 50 tests cases, these will be enabled on strict compare in a follow-up PR that fixes more.

kdeme:
closing for https://github.com/status-im/nimbus-eth1/pull/4620 where same fix is included in

PR #4612, opened 2026-08-06
Optimize EVM balance opcode

bhartnett:
This updates the default hash of the Address type to use rapidhash. This improves the performance of the balance opcode for the worst case scenario when addresses are iterated through rather than uniformly distributed. An account can be created in the state by sending a small value transfer to it, and then once created balance lookups are very slow.

**Before**
Screenshot_2026-08-03_23-10-52

**After**
Screenshot_2026-08-06_16-17-32


bhartnett:
CI failures are unrelated to this change.

PR #4622, opened 2026-08-07
Use genesis block hash to validate network matches data directory

bhartnett:


execution_chain/nimbus_execution_client.nim
```diff
@@ -261,29 +260,19 @@ proc init*(T: type NimbusNode, config: ExecutionClientConf, com: CommonRef, para
...
   let
-    kvt = db.baseTxFrame()
-    calculatedId = calcHash(params)
-    dataDirIdBytes = kvt.get(dataDirIdKey().toOpenArray).valueOr:
-      # an empty database
-      writeDataDirId(kvt, calculatedId)
+    storedHeader = db.baseTxFrame().getBlockHeader(0'u64).valueOr:
       return
+    storedHash = storedHeader.computeBlockHash
+    expectedHash = params.genesisBlockHash()
```
jangko: Maybe faster get the genesis hash from CommonRef rather than recompute it from NetworkParams?

jangko (commented):
Might as well remove `common/chain_config_hash.nim` and add comment to `dataDirId` in `db/storage_types.nim`

execution_chain/nimbus_execution_client.nim
bhartnett: Actually that doesn't work because CommonRef gets the genesis hash from the database and so we need to compute the genesis hash independently from the network params in order to verify it.

PR #4633, opened 2026-08-12
Block access list pruning

bhartnett:
This PR adds pruning of block access lists. 

A async background job runs every 12 seconds and prunes the BALs in batches when required. The bal pruning is enabled by default on networks having the amsterdam hardfork but can be disabled using `--debug-bal-pruning=false` . 

The bal tail is stored in the database which represents the last pruned block number. BALs are pruning between the tail and the retention cut off which is determined by binary searching the headers to find the largest header outside of the retention period.


advaita-saha:
Is there some advantage/requirement on keeping bal and block history pruning separately async?

bhartnett:
> Is there some advantage/requirement on keeping bal and block history pruning separately async?

Yeah mainly because the requirements for BAL pruning are different to the block history pruning:
- BAL pruning is only activated for amsterdam while the regular pruning runs for all forks.
- The BAL retention period is different to the history expiry periods.
- BAL pruning should be enabled by default but the current pruner is disabled by default.

Issue #3207, opened 2025-04-10
Excessive memory consumption when syncing a long way up to the `canonical head`


mjfh:
Since PR #3191 the Nimbus EL has an annoying memory problem in the `FC` module as the syncer does not update `base` anymore while importing blocks. This happens at least when the syncer has to catch up a long way.

Previously, there was a kludge related to the syncer which used the `forkChoice()` function for `base` update.

Now `base` can only be updated if the CL triggers a `forkChoiceUpdated` which has no effect if the update is out of scope for the `FC` module, which in turn happens when syncing for an old or pristine database state.

In fact, this leads to a similar situation to when `mainnet` was unable to finalise transactions globally.

For the attached screenshot, I ran the syncer overnight (with turned off CL) and had the following memory usage in the morning
* 78.9GiB virtual (from metrics screen)
* 41.4GiB physical (from metrics screen)
* 22GiB extra swap space freed after stopping the process

As it seems, a big machine can handle the situation to an extend but the execution throughput decreases.

![Image](https://github.com/user-attachments/assets/a69408fd-5aeb-4683-8917-2c7cb2f3403f)

KolbyML:
@mjfh is #3202 the PR you meant to link?

mjfh:
Ops, was the wrong one -- lol
Thanks for noticing

mjfh:
was somehow related to issue 3202 :)

jangko:
fixed by #3204

jangko:
Looks like the the problem not cured thoroughly. Need more investigation.

advaita-saha:
> Looks like the the problem not cured thoroughly. Need more investigation.

has the memory usage improved than before or exactly same as before ?

jangko:
When syncing with hoodi using an empty database, initially everything looks ok, the base can move forward.
(apply #3237).

But when the gap is wider, the base stop moving. I'm not sure why CL suddenly request sync from head < 10K, then jump to > 200K+.

The problem is syncer downloading forward from known FC base(even though the segment request is reverse), but FC expecting the syncer to download backward from head.

Of course the FC expectation not satisfied by the syncer because the finalized hash(pendingFCU) not resolved into `latestFinalizedNumber`.

IIRC from discord discussion we agree on the syncer have two phase:
- download headers from head to known base, and put it into cache. (probably also start another new session)
- then download the block or body forward and import into FC.
-
That is what assume how the syncer works. But looks like not like that.

[...]

PR #4637, opened 2026-08-13
Remove kvt TxFrame

### LINKED ISSUE #3601: remove kvt `TxFrame`
The `TxFrame` type for the kvt was introduced to ensure that database updates for data that potentially will go stale would not end up littering the database, at a time when there was no cleanup of forked chains because ForkedChain itself did not exist and therefore there was not enough information to prune unviable histories.

Since then, `ForkedChain` has been implemented and keeps track of the existing histories via the `heads` mechanism - with head tracking in place, all tracked histories are known and instead of using TxFrame to prune them, we can optimistically write them to the KVT on disk and then prune the irrelevant histories on base updates, deleting them from disk - this will save significant amounts of memory in particular during long periods of nonfinality.

This change implies removing TxFrame handling from the KVT part of the database - aristo still needs TxFrame since this is the primary mechanism for rolling back state changes on reorgs.

Instead of writing to a `TxFrame` that is bound to the block being processed, a temporary TxFrame-like "transaction" can be used to atomically collect all changes that are due to new information coming in. For example, when the finalized block is updated via FCU, the base may be updated and some histories might get pruned - this, together with the update to `heads`, would atomically get committed at the end of FCU processing such that the block database remains consistent and complete with respect to the `heads` list.

jangko:
- [ ] fix FC serialize
- [ ] fix FC deserialize
- [ ] remove dead branches from db when finalization move forward

fix #3601


advaita-saha:
Then are we planning to keep the blocks of different branches in memory ?
Might again bring the issue of `seq[Transaction]` bloat

If we are writing and fetching from the db directly, we should be wary of the performance cost and measure it

jangko:
Instead of trying to solve all issue in one PR, this PR can be viewed as the first in multi staged refactoring.
We will place as much information in db as possible, the goal is to solve #3207.

> If we are writing and fetching from the db directly, we should be wary of the performance cost and measure it

rocksdb have something called writeBatch with index, it works similarly to the txFrame, we can use it for active section of chain e.g. latest 128 blocks.

The current goal for this PR is to eliminate and simplify the kvt database logic. More PR will come.

advaita-saha:
> we can use it for active section of chain e.g. latest 128 blocks.

Sounds good

arnetheduck:
> If we are writing and fetching from the db directly,

something to keep in mind is that the blocks are a negligible amount of data from a database perspective (though not from a memory perspective) - eth2 has been following this strategy since its inception where blocks are kept in the database (where the EL block is a subset of the CL block).

The blocks are only ever "read back" when replaying history to create a new fork and the creation of a new fork is a rare event - this is where the win lies: we do not need to keep a long history of blocks in memory only to occasionally access them - instead, we write them to disk as soon as we know they're valid and (almost) never look at them again - in the meantime, we have the memory free for more important things (like processing the next block).

PR #4652, opened 2026-08-17
Parallel stateroot computation - lock free merge keys

bhartnett:
This improves the performance of the parallel stateroot computation by:
1. Removes the locking when writing keys back into the txFrame layers. Instead we write each key into a SharedSeq buffer (one per task), then after the tasks complete the keys are updated into the layers in the main thread. The merge function is also optimized to avoid the linear time `deltaAtLevel` for every key.
2. Reduces cpu usage in the main thread when it has no work to do by implementing a backoff mechanism. First we use cpuRelax and then sleep so that the main thread doesn't spin needlessly while waiting for the tasks.

This also fixes an issue where the newest snapshot was not having its keys updated which could cause rehashing of those in memory keys over time.

### DIFF
--- a/execution_chain/concurrency/shared_types.nim
+++ b/execution_chain/concurrency/shared_types.nim
@@ -93,10 +93,13 @@ proc `[]`*[E](s: var SharedSeq[E], i: int): var E =
 template len*[E](s: SharedSeq[E]): int =
   s.count
 
-
+const seqGrowThresholdBytes = 128 * 1024
 
 proc grow[E](s: var SharedSeq[E], minCap: int) =
-  s.reallocTo(nextPowerOfTwo(max(minCap, seqInitialCapacity)))
+  if max(s.cap, minCap) * sizeof(E) >= seqGrowThresholdBytes:
+    s.reallocTo(max(minCap, s.cap + (s.cap div 2)))
+  else:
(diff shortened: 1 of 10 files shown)


jangko:
I have a branch, testing `rlpimport` using EEST fixtures.
When rebased with this branch, this line below hang without cpu activity. Don't know if it's related to this PR or something else.

https://github.com/status-im/nimbus-eth1/blob/14c6f13500b11de4d400e53b2b07fe2595b41d44/tests/eest/eest_rlpimport.nim#L157-L158

bhartnett:
> I have a branch, testing `rlpimport` using EEST fixtures. When rebased with this branch, this line below hang without cpu activity. Don't know if it's related to this PR or something else.
> 
> https://github.com/status-im/nimbus-eth1/blob/14c6f13500b11de4d400e53b2b07fe2595b41d44/tests/eest/eest_rlpimport.nim#L157-L158

The changes in this branch appear to be working in my testing but its possible there is a bug somewhere. I'll have a look.

bhartnett:
> I have a branch, testing `rlpimport` using EEST fixtures. When rebased with this branch, this line below hang without cpu activity. Don't know if it's related to this PR or something else.

Looks like an chronos event loop usage issue not related to stateroot computation.

PR #4671, opened 2026-08-21
KVT put and receipt root optimization

bhartnett:
This adds a variant of the kvt put API that moves instead of copying the value. 

Also a custom receipt rlp encoding function is used to avoid copying logs while building the receipts root.

### DIFF
--- a/execution_chain/db/kvt/kvt_layers.nim
+++ b/execution_chain/db/kvt/kvt_layers.nim
@@ -54,7 +54,7 @@ func layersGet*(db: KvtTxRef; key: openArray[byte]|seq[byte]): Opt[seq[byte]] =
 # Public functions: put function
 # ------------------------------------------------------------------------------
 
-func layersPut*(db: KvtTxRef; key: openArray[byte]; data: var seq[byte]) =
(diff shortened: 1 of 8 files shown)


execution_chain/db/kvt/kvt_layers.nim
```diff
@@ -54,9 +54,16 @@ func layersGet*(db: KvtTxRef; key: openArray[byte]|seq[byte]): Opt[seq[byte]] =
 # Public functions: put function
 # ------------------------------------------------------------------------------
 
+func layersPutMove*(db: KvtTxRef; key: openArray[byte]; data: var seq[byte]) =
```
arnetheduck, line 57: why not `sink seq[byte]`?

bhartnett: Using sink didn't perform well (at least with refc) because in layersPutMove when using swap a copy of the parameter is created to swap into, a temporary seq allocation which isn't needed or used.

**Benchmarks:**
| Variant | 110 B | 512 B | 4096 B |
|---------|-------|-------|--------|
| sink + move | 1.32x | 1.01x | 1.21x |
| sink + swap | 1.54x | 1.45x | 1.59x |
| var + swap (committed) | 1.58x | 1.78x | 2.86x |

PR #4687, opened 2026-08-26
rpc: support the "pending" block tag

barnabasbusa:
## Problem

Any RPC taking a block tag rejects `"pending"`:

```
$ eth_getTransactionCount(addr, "pending")
→ -32000 "`eth_getTransactionCount` raised an exception", data: "Unsupported block tag pending"
```

go-ethereum's `ethclient.PendingNonceAt` always asks for `"pending"`, so tooling built on it (e.g. [buildoor](https://github.com/ethpandaops/buildoor), which crash-looped at startup on every `buildoor-*-nimbusel` node of glamsterdam-devnet-8) cannot fetch a nonce from nimbus at all. `ethers`/`web3.js` default some calls to `"pending"` as well.

## Changes

- `TxPoolRef.getPendingNonce`: head-state nonce advanced over the sender's gap-free run of pooled transactions — the next usable nonce, which is what `eth_getTransactionCount(addr, "pending")` returns on geth & co.
- `eth_getTransactionCount` short-circuits `"pending"` to that.
- `headerFromTag` / `blockFromTag`: `"pending"` resolves to the head like `"latest"`. No pending block is assembled outside of payload building, so this mirrors erigon rather than erroring. (Also drops the now-stale "pending not supported" note on `eth_getTransactionByBlockNumberAndIndex`.)

## Tests

`tests/test_rpc.nim`: new `eth_getTransactionCount pending` case — equals `latest` with an empty pool, and `latest + 1` after pooling the signer's next tx (the tx is removed again so later cases are unaffected).

```
[Summary] 55 tests run (1.06s): 55 OK, 0 FAILED, 0 SKIPPED
```

Companion buildoor-side fallback: ethpandaops/buildoor#179.

### DIFF
--- a/execution_chain/core/tx_pool.nim
+++ b/execution_chain/core/tx_pool.nim
@@ -115,6 +115,7 @@ export
   addTx,
   getItem,
   getNonce,
+  getPendingNonce,
   contains,
   removeTx,
   removeExpiredTxs,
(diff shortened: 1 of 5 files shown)


jangko:
fix #2094

PR #4700, opened 2026-08-28
use shallow=1 to speed up submodule checkout in Nix CI

jakubgs:
Lets see.


tersec:
Seems to work, but is it faster?

jakubgs:
No, I don't think it is.

PR #4701, opened 2026-08-28
Ledger: differentiate accountExists and accountAlive

jangko:
Based on a report prepared by Stefan. In a network where the genesis contains *dead* account, Nimbus test it as account's leaf exists instead of alive. Geth, reth, and besu agree on the block. But nimbus reject it because of block gas used mismatch.

Post-EIP-161 the shape(dead account) is not constructible at runtime.

This bug will split Nimbus from EELS on any EEST EIP-7610 vector in this class. And any future network whose genesis carries such an account inherits a permanent fork.

This PR fix the bug in three places suggested by the report:
- in call_common.nim: top level transfer
- CALL opcode
- SELFDESTRUCT beneficiary

In addition to that, this PR also fix:
- the same issue in CREATE opcode 
- cleanup other CALL* opcode which is not using the `accountAlive` check
- selfdestruct from spurious dragon to osaka in addition to amsterdam

This is the report. It will expired after 60 days of publish.
https://panda-uploads-production.devops-539.workers.dev/panda/uploads/9cc1a0/nimbus-eth1-storage-only-account-liveness-20260827.html

### DIFF
--- a/execution_chain/evm/computation.nim
+++ b/execution_chain/evm/computation.nim
@@ -87,14 +87,14 @@ proc getBlockHash*(c: Computation, number: BlockNumber): Hash32 =
     return default(Hash32)
   c.vmState.getAncestorHash(number)
 
-template accountExists*(c: Computation, address: Address): bool =
+template accountAlive*(c: Computation, address: Address): bool =
   if c.balTrackerEnabled:
     c.vmState.balTracker.trackAddressAccess(address)
 
   if c.fork >= FkSpurious:
     not c.vmState.readOnlyLedger.isDeadAccount(address)
   else:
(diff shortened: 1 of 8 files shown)


execution_chain/evm/computation.nim
```diff
@@ -88,13 +88,19 @@ proc getBlockHash*(c: Computation, number: BlockNumber): Hash32 =
   c.vmState.getAncestorHash(number)
 
 template accountExists*(c: Computation, address: Address): bool =
```
bhartnett, line 90: There is now no bal tracking in accountExists. Not sure if that is intended or not.

PR #4705, opened 2026-08-28
make BoringSSL optional for the sha256 and P256VERIFY precompiles

kdeme:
Adds -d:enable_boringssl, defaulting to true so the node is unchanged. Disabling it restores the pre-BoringSSL backends: nimcrypto/sha2 for sha256 (#4524) and bearssl/secp256r1_verify for P256VERIFY (#4521), both of which are portable and freestanding.

modexp has no such backend yet and uses BoringSSL either way.

For the BearSSL the previous init and verify function calls had to be united into one call to match the API used currently.


execution_chain/evm/secp256r1verify.nim
```diff
@@ -10,32 +10,93 @@
...
+    # The secp256r1 field prime.
+    FieldPrime =
+      UInt256.fromHex("0xffffffff00000001000000000000000000000000ffffffffffffffffffffffff")
+
+  func isInfinityByte(data: openArray[byte]): bool =
+    ## True when every byte is zero, i.e. the encoded point at infinity.
+    for b in data:
+      if b != 0:
+        return false
+    true
```
tersec, line 62: This entire function can be written without efficiency loss as `data.allIt(it == 0)` or `not data.anyIt(it != 0)`; in this case the former is more straightforward.

Separately, this is not a very efficient way of checking against zero due to the mechanics of Nim's loops, but that's a more involved benchmarking exercise.

kdeme: Yes, you're right.

I took the pre-BoringSSL implementations (see linked PRs in original description) and put them behind this compile flag. I didn't really pay attention to performance improvements.

They could most likely be improved a lot. And I'm also still missing modexp implementation (the old stint one was incomplete, but usable to some extent), making this PR in it's current state not that useful.

I'm not fully sure if I should put more effort in a non BoringSSL solution / alternative. I was adding this so that I could more easily test building the stateless guest for zkvm platforms (zisk, sp1, ...), and BoringSSL would be a pain (if not impossible, at least without pulling in way more than needed) to add in there.

However, these precompiles will/are also covered with zkvm accelerators (zkvm native implementations for performance), so these implementations behind the compile flag wouldn't even be needed on the long run. So I could also just keep these changes locally for my own testing until I have it all working with the zkvm accelerators.

Unless of course it is also useful for other binaries (perhaps verified proxy) to have? Else it's probably just more work and maintenance?

tersec: Another consideration is just the testing surface, that is, it creates a scenario which can bitrot quickly unless tested. And the more of these we have, combinatorially the more can bitrot and turn out not to be working.

So I'd generally say that if it's important enough to (re)add some backends for this, it's worth at least testing that they compile.

> However, these precompiles will/are also covered with zkvm accelerators (zkvm native implementations for performance), so these implementations behind the compile flag wouldn't even be needed on the long run.

A question, then, is timeline -- the EF seems to want to move fairly quickly on these things, but if the only use so far is near-guaranteed to be ephemeral, then no, it probably doesn't make sense to keep in the main tree.

PR #4707, opened 2026-08-29
accountExists and accountAlive take 2

jangko:
This PR only fix the bug in 1 place instead of 3 places like #4701. The actual bug is only  in prepareDispatch of call_common.nim. While the rest only change `not isDeadAccount` to `isAccountAlive` to avoid future confusion.

`Alive` flag in `LedgerRef` also renamed to `LeafExists` to match the spec.
3 test added to t8n test suite:
- account exists and empty
- account not exists
- account exists and not empty


bhartnett:
Does this line need to be updated? https://github.com/status-im/nimbus-eth1/blob/36e75fb4f19bcb40761b5b8a11767967603e7a50/execution_chain/transaction/eoa_delegation.nim#L105

jangko:
No, this is the EELS counterpart line, it is using `account_exists` and not `is_account_alive`

https://github.com/ethereum/execution-specs/blob/c4deda5b3cfc5c1c8429dcd9159a6fb5636d8486/src/ethereum/forks/amsterdam/vm/eoa_delegation.py#L295

bhartnett (approved):

Issue #4717, opened 2026-09-01
new base images required


advaita-saha:
After the rocksdb update with the PR https://github.com/status-im/nimbus-eth1/pull/4699
New base images needs to built and updated

advaita-saha:
nightly build fails with 
```bash
PLATFORM=linux_amd64
g++ (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

ROCKSDBVER=10.10.1
Building: Nim compiler
Building custom Nimble commit: 42ef70c2102a942c46f13eb76872326edd525cec
Building Nimble...
Replacing nimble binary...

Tip of the day: this will probably build faster if you use "make -j4 ...".

Building: build/generate_makefile
Build completed successfully: build/generate_makefile
Using cached RocksDB
Building: build/nimbus
/usr/bin/ld: cannot find /home/user/nimbus-eth1/vendor/nim-rocksdb/build/liburing.a: No such file or directory
collect2: error: ld returned 1 exit status
make[1]: *** [nimcache/release/nimbus/nimbus.makefile:4425: build] Error 1
make: *** [Makefile:204: nimbus] Error 2
make: *** [Makefile:538: dist-linux-amd64] Error 2
```

advaita-saha:
nim-rocksdb 11.x links liburing on Linux. build_static_deps.sh builds and ships liburing.a — but only when it actually runs. In the dist container it never does: `USE_CACHED_ROCKSDB=1` sets `USE_SYSTEM_ROCKSDB=1`, so the Makefile rocksdb target just prints "Using cached RocksDB", and `copy_rocksdb()` blindly copies `/usr/rocksdb/*` from an image built before liburing was a dependency.

  That's also exactly why only amd64 and arm64 fail — the passl is Linux-guarded, so macOS and Windows don't reference it.

advaita-saha:
Needs another base update due to #4724

advaita-saha:
Created

PR #4718, opened 2026-09-01
update base image

advaita-saha:
For #4717


jangko:
If we rebuild the base image, I think we only need to update the hash for downstream docker file. We don't need the previous base image anyway.
There is no objection to change the base image version although there is almost no significant changes to the base image implementation except the cached rocksdb.

PR #4727, opened 2026-09-02
Update base images

bhartnett:
Since rocksdb was updated to v11.8.1 we need to update the base images again.

### DIFF
--- a/docker/dist/Dockerfile.linux-amd64
+++ b/docker/dist/Dockerfile.linux-amd64
@@ -9,7 +9,7 @@
 # according to those terms.
 
 # The build is reproducible only if this base image stays the same.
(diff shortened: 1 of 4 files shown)


docker/dist/Dockerfile.linux-amd64
```diff
@@ -9,7 +9,7 @@
 # according to those terms.
 
 # The build is reproducible only if this base image stays the same.
-FROM statusim/nimbus-eth1:dist_base_eth1_v2_linux_amd64@sha256:a48f20ca970f136b69c062d847356a1ba95f863436361a35745a7ff5dc2659f8
+FROM statusim/nimbus-eth1:dist_base_eth1_v2_linux_amd64@sha256:468cae93bf70700da8a1814d473fa10e984a70ee7073c7b7227f1d717a5902fc
```
advaita-saha: Cross check this hash once, seems different to me 

Image

bhartnett: Thanks, updated

PR #4748, opened 2026-09-06
fix: peer overload in mainnet

advaita-saha:


jangko (approved):

execution_chain/networking/peer_pool.nim
```diff
@@ -209,12 +227,14 @@ proc run(p: PeerPoolRef) {.async: (raises: [CancelledError]).} =
 func newPeerPool*[Network](
     network: Network,
     discovery: Eth1Discovery,
-    minPeers = 10,
+    maxPeers = 25,
```
tersec, line 230: This 25 constant is used twice in a way which seems like it's meant to be consistent -- in that case, better to create a `const` somewhere and use that so future maintenance can't get them out of sync.

advaita-saha: it is set from config only, this is just a default fallback number
Can create a constant

execution_chain/networking/rlpx.nim
```diff
@@ -386,6 +403,38 @@ proc initPeerState(
...
+      try:
+        await peer.ping().wait(peerIdleTimeout)
+      except AsyncTimeoutError:
+        debug "Ping could not be delivered, closing connection",
+          remote = peer.remote, clientId = peer.clientId
+        peer.transport.close()
+        return
+      except EthP2PError as exc:
+        trace "Failed to send ping", remote = peer.remote, err = exc.msg
+        return
```
kdeme, line 436: Should this error close the transport or not? what is the difference with a timeout error in terms of required behavior?

Either way, the combination of returning the `keepAliveLoop` without closing the transport seems incorrect and a leak.

execution_chain/networking/rlpx.nim
```diff
@@ -386,6 +403,38 @@ proc initPeerState(
...
+      # which runs the regular disconnect path - handlers, pool removal, metrics
+      peer.transport.close()
+      return
+
+    if idle >= peerPingInterval:
+      # A live peer answers with a pong, which resets `lastReceived`. The write
+      # is bounded because a peer that stopped reading fills our send buffer
+      # and would otherwise park this loop past the idle timeout.
+      try:
+        await peer.ping().wait(peerIdleTimeout)
```
kdeme, line 428: awaiting here kind of makes your `peerLivenessInterval` no longer an interval.

execution_chain/networking/rlpx.nim
```diff
@@ -386,6 +403,38 @@ proc initPeerState(
   peer.lastReqId = Opt.some(0u64)
   peer.initPeerStates peer.dispatcher.activeProtocols
 
+proc keepAliveLoop(peer: Peer) {.async: (raises: [CancelledError]).} =
+  ## Reap connections whose remote end has gone away without closing the socket.
+  ## Nothing else does this: `dispatchMessages` blocks in `recvMsg` forever and
+  ## peers that the syncer has no slot for are never written to either, so a
+  ## dead connection is never noticed.
```
kdeme, line 410: For correctness: I think on a TCP timeout they would still get cleaned up, but yeah, that's a very slow mechanism and the ping loop is much quicker in cleanup.

execution_chain/networking/peer_pool.nim
kdeme: Could make a constant that is used in the code as defaults and then the config can also import that one and and set as config default.

execution_chain/networking/rlpx.nim
advaita-saha: https://github.com/status-im/nimbus-eth1/pull/4748/commits/805f9a5c3ab2aad0eb05f4e8a10e66b0d862b6e1

execution_chain/networking/rlpx.nim
advaita-saha: https://github.com/status-im/nimbus-eth1/pull/4748/commits/8d135c93d7052a51bc026468da21e5802e95c4d2

execution_chain/networking/peer_pool.nim
advaita-saha: https://github.com/status-im/nimbus-eth1/pull/4748/commits/e6b1c6107406275796c584f3eedc960d6fe81de5

PR #4758, opened 2026-09-07
Refactor CallParams and TxContext

jangko:
Upcoming EIP-8141 Frame Transaction requires almost all Transaction fields are inspectable by EVM. With current approach it means fields are copied twice. Once to CallParams, and once to TxContext. With this PR, only the pointer of the transaction is copied around without the need to copy the fields. This new approach also eliminating future changes to CallParams and TxContext if the Transaction gets additional fields.

Tradeoff made here is less copies and allocations but more indirections.

### DIFF
--- a/execution_chain/rpc/params.nim
+++ b/execution_chain/rpc/params.nim
@@ -25,13 +25,20 @@ const
 func sender*(args: TransactionArgs): Address =
   args.source.get(ZeroAddr)
 
-func destination*(args: TransactionArgs): Address =
-  args.to.get(ZeroAddr)
+func txType(n: TransactionArgs): TxType =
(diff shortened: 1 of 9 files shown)


execution_chain/rpc/params.nim
```diff
@@ -52,50 +60,27 @@ proc toCallParams*(vmState: BaseVMState,
       cap = globalGasCap,
       gasLimit = globalGasCap
 
-  var gasPrice = GasInt args.gasPrice.get(0.Quantity)
```
jangko, line 55: This gasPrice calculation counterpart is `effectiveGasPrice` in `callParams` function of call_evm.nim

PR #4760, opened 2026-09-08
stateless: build the guest into a ZisK ELF, with zkVM accelerators

kdeme:
Cross-compiles and links against newlib, adding the ZisK linker script and syscall glue. Route sha256, modexp and P256VERIFY to the zkVM's accelerators behind -d:enable_zkvm_accelerators.

And move guest builds to their own CI job.


kdeme:
This is a first step,just enough to run, test and benchmark fixtures on a real zkVM, which gives us a base to measure against.

I have some more work in the pipeline such as adding more of the accelerators, trying the zkVM own allocators, and comparing clang against gcc, etc. But that all makes more sense once this base is in place.

I do still in this PR want to separate/structure better the build targets from the main Makefile, because as we will have to add several more zkVM targets, and potentially some compiler options (gcc/clang), etc. it might grow a bit out of hand.

PR #4761, opened 2026-09-09
update to Nim 2.2.12

tersec:


tersec:
```nim
proc d() =
  for r in [0]: (proc() = discard r)()
d()
```
is what https://github.com/status-im/nimbus-eth1/pull/4761/commits/9cd9b0d7fa18345f9fc8c09fcc5da627dd8c20f9 addresses. The capture rules around `lent` aren't new, but 2.2.12 adds a couple of cases of collection iteration which become `lent`. This is good for efficiency, but creates occasional incompatibilities to fix when the iteration variable genuinely is captured, as it the case via `nim-unittest2`'s `runtimeTest`.

PR #4763, opened 2026-09-09
treat InvalidCmpOp and UnreachableElse warnings as errors

tersec:
`InvalidCmpOp`:
```
/home/runner/work/nimbus-eth1/nimbus-eth1/execution_chain/sync/snap/worker/extra_types/types_scalar.nim(34, 1) Warning: define `==` instead of `!=` to implement user defined comparison operator. it allows you to use `!=` automatically. [InvalidCmpOp]
```

https://nim-lang.org/docs/system.html#%21%3D.t%2Cuntyped%2Cuntyped defines this operator regardless, in terms of `==`.

`ProveField` is misleading, since it only shows up by default at `--verbosity=2` and above anyway, so without more targeted testing, it's more of a trap than useful.


arnetheduck:
I'm starting to think that all these warning switches should sit in the Makefile instead, ie be explicitly passed as CLI flags to compilation rather than sitting in the top-level nims (where they start applying to libraries and during development, ie when executing a single test, where they are _mostly_ annoying - fixing an extra import isn't a primary concern while trying things out in new code - it's a post-processing cleanup step.

For example, it's not possible to compile an eth2 test nested in eth1 since it gets a "duplicate import" warning due to libbacktrace getting imported twice (from each .nims).

tersec:
As a general point, for every one of these, in at least one of `nimbus-eth1`, `nimbus-eth2`, and `nimbos`  I have personally spent time fixing these when inadvertently introduced. I don't particularly want to have to always have the vigilance to be checking CI logs for Nim warnings. It seems like a bad use of time. Furthermore, for all but a couple (`DuplicateModuleImport` in that scenario is interesting, and `UnusedImport` routinely can occur during local development and/or Nim bug hunting, so I often remove that one locally for that purpose), even in intermediate stages of code there's no real utility to triggering these warnings.

Most of them don't have any particular issue with being composed with each other either (`ConvFromXtoItselfNotNeeded` doesn't randomly appear because one has attempted to use a particular `nimbus-build-system` setup in terms of `nimbus-eth2` development within `nimbus-eth1`), so the argument is weaker there.

Also, broadly, the goal is to align the kind of errors during local development with the errors the CI triggers. The CI should as little as possible be triggering errors that building locally does not.

I do find working with `Makefile`s to be unergonomic and constraining, because it feels like using workarounds for a DSL which has been stretched far beyond its initial design goals makes it more of a game of how to encode the desired behavior into a base language which has no straightforward capacity for it, operates in purely string-substitution based way, etc and is full of random footguns.

Finally, if the goal is to migrate to Nimble, well, I'd tend to want to decrease rather than increase reliance on makefiles.

But can look into options.

arnetheduck:
> CI logs for Nim warnings

I think you misunderstand. Keep them as errors but move their usage to the makefile where CI and `make nimbus` will enforce them with equal fervor (but local compilation using `nim c tests/mytest.nim` or `nim c vendor/nimbus-eth2/tests/othertest` will not).

arnetheduck:
this btw is the same strategy we use for other lint stuff like `nph` in projects that enable it - ie allow local deviation during development, enforce that the common repo is clean with CI - this is pragmatically very convenient.

if you want local enforcement ahead of ci, there's also `~/.config/nim/nim.cfg`.

arnetheduck:
> Finally, if the goal is to migrate to Nimble, well, I'd tend to want to decrease rather than increase reliance on makefiles.

ditto for nimble, ie `nimble build` and `nimble test` can enforce these without a top-level config.nims.

> align the kind of errors during local development with the errors the CI triggers

it's useful to have this option, sure - but the more of these warnings-as-errors there are, it also gets in the way of iteration - ie getting errors about unused locals and imports when commenting out a bit of code locally is very disruptive.

In fact, if they stay as warnings during local development, you see them and can attend to them before committing (where they get checked as errors instead)

tersec:
> ie getting errors about unused locals

I deliberately have never done `XDeclaredButNotUsed` and won't. It, yes, has too many tradeoffs.

tersec:
But I would generally note it as an example of what happens when these are not enforced: they just kind of accumulate, occasionally people decide to notice/clean up/care, but on average they just increase over time.

arnetheduck:
>  are not enforced: 

Again, I'm not suggesting to remove the enforcement from the shared repo - just move it to CI and leave it as a warning during development so that it's visible but not enforced.

PR #4776, opened 2026-09-10
Add block proof for Gloas and later

### LINKED ISSUE #4109: eraE / Portal: Add / implement new block proof for >= gloas
Issue for implementing a new block proof type for gloas fork.

With gloas we will need to adapt the current used block proofs based on `historical_summaries`: https://github.com/status-im/nimbus-eth1/blob/master/portal/eth_history/block_proofs/block_proof_historical_summaries.nim

### Currently (pre-gloas) a chain of proofs via:

`BeaconBlock → BeaconBlockBody → ExecutionPayload → block_hash`

### With gloas (ePBS):

`execution_payload` is removed from `BeaconBlockBody`.
The execution payload now lives in a separate `ExecutionPayloadEnvelope` signed by builder.

Instead `BeaconBlockBody` has `signed_execution_payload_bid: SignedExecutionPayloadBid `.  And `ExecutionPayloadBid` contains `block_hash`.

So I think the new proof chain that becomes possible is:

`BeaconBlock → BeaconBlockBody → SignedExecutionPayloadBid → ExecutionPayloadBid → block_hash`

This does only proof what a proposer has commited to as `block_hash`. But I think that is sufficient for what this proof is used for, that is it would come together with the actual EL block data and thus can be verified by comparing `block_hash` with the actual block hash of the EL block.

kdeme:
The execution payload is no longer part of the BeaconBlockBody, so prove the parent_block_hash of the payload bid instead, which is built from the beacon block that follows the one committing the payload.

fixes https://github.com/status-im/nimbus-eth1/issues/4109

### DIFF
--- a/execution_chain/history/block_proofs/block_proof_common.nim
+++ b/execution_chain/history/block_proofs/block_proof_common.nim
@@ -13,24 +13,54 @@ from beacon_chain/spec/datatypes/capella import ExecutionPayload
 from beacon_chain/spec/datatypes/deneb import ExecutionPayload
 
 const
-  # TODO:
-  # This proof only works up until Fulu fork.
-  # For gloas+ the execution payload is no longer part of the BeaconBlockBody and thus
-  # an additional proof type is required.
+  # Bellatrix up to and including Fulu hold the execution payload in the
+  # BeaconBlockBody, so the block hash is proven straight from the payload.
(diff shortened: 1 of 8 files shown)


kdeme:
Created also spec accompanying PRs for this:
- https://github.com/ethereum/portal-network-specs/pull/416
- https://github.com/eth-clients/e2store-format-specs/pull/22

PR #4777, opened 2026-09-10
Verify the genesis block of PoS only networks against the genesis data

kdeme:
Their genesis block predates the beacon chain, so its execution payload is never part of a beacon block, which makes it impossible to create a proof of the same type as for the other blocks.

Thus we skip regular proof verification and instead verify it against the genesis block itself.

Other option would be to create another proof type specifically for block 0 on PoS only networks. But that seems overkill as the block can already be verified/created with genesis data.


kdeme:
spec PR on this: https://github.com/eth-clients/e2store-format-specs/pull/23

PR #4779, opened 2026-09-11
`create` opcode use code cache

### LINKED ISSUE #2929: `create` opcode should use code cache
When creating a new contract using the `create`/`create2` opcodes, we should check whether the code already exists in the cache and reuse the cached version - otherwise, we need to recompute the code jump destination filter each time which is slow.

While executing the code the first time, the cache should not be updated (`peek`) - only when the transaction is fully successful should the cache be updated to include the new code, to avoid contracts stuffing the cache with junk code.

advaita-saha:
Fix #2929


bhartnett (approved):
Looks good to me. The comments are a bit verbose in some places so I'd suggest keeping them to a minimum where possible.

PR #4797, opened 2026-09-16
Storage trie static vids

bhartnett:
The idea here is to write storage tries using the same static vid vertex id allocation as is used for the account trie. This enables jumping (directly or near) to the leaf trie node being searched instead of having to walk the full storage trie from root to leaf. In order to implement this the vertex ids are no longer globally unique and so the storage root vertex id is needed and always used together with the trie node vertex id.

This is implemented by storing a static vid hint in each account record which is used to estimate the trie level where most leafs are located. The hint is then used in the static vid lookup to jump to the leaf trie node. The hint is added to the account trie when updating/write storage leafs and stores the level of the last write.

**Performance testing results**

At block 5100000 imported 100000 blocks with all caches disabled:
master - elapsed=14m18s396ms
branch - elapsed=11m40s600ms

At block 10 million imported 50000 blocks with all caches disabled:
master - elapsed=17m23s860ms
branch - elapsed=13m55s283ms

At block 10 million imported 1 million blocks with all caches enabled:
master - elapsed=4h36m34s
branch - elapsed=4h13m15s

The read performance when looking up slots improves around 10% percent for small storage tries (32 slots) and improves 50% for large storage tries with 100000 slots. Larger storage tries would likely see larger improvements.

Average backend vertex fetches per read:
| workload | master | branch, walking | branch, static vid probing | saved | ratio |
|----------|--------|-----------------|-----------------|-------|-------|
| small storage trie slot hit | 4.40 | 4.40 | 3.16 | 1.24 | 1.39x |
| small storage trie slot miss | 3.41 | 3.41 | 2.99 | 0.42 | 1.14x |
| large storage trie slot hit | 3.51 | 3.63 | 1.69 | 1.82 | 2.08x |
| large storage trie slot miss | 2.46 | 2.58 | 1.59 | 0.87 | 1.55x |

**Disk usage**

It turns out that this change actually saves disk space in practice because when using static vids for each storage trie each vid uses less space then the dynamic vids. This is because static vids are smaller and therefore use less significant bytes (the zeros are removed when encoding). The only increase introduced with this change is the additional stoHint byte added to accounts having storage but this is small compared to the vid serialization saving.

State database size at block 10 million (without computed hashes):
master - 28.3 GB
branch - 24.2 GB

**Migration**

The changes are implemented in a fully backwards compatible way where existing databases continue to work but existing storage tries don't use static vids and so won't benefit from them but new storage tries created after updating will start using static vids and get the performance improvement. A full resync would provide the best performance because all the storage tries would then get the static vids.

The database is modified after running this new version and so rolling back to a previous commit/version will fail on startup showing a defect.


execution_chain/db/aristo/aristo_init/rocks_db/rdb_init.nim
```diff
@@ -88,14 +88,15 @@ proc init*(rdb: var RdbInst, opts: DbOptions, baseDb: RocksDbInstanceRef) =
...
 
   rdb.rdVtxSize =
     opts.rdbVtxCacheSize div
-    (sizeof(VertexID) + sizeof(VertexBuf) + lruOverhead)
+    (sizeof(RootedVertexID) + sizeof(VertexBuf) + lruOverhead)
 
   rdb.rdBranchSize =
-    opts.rdbBranchCacheSize div (sizeof(typeof(rdb.rdBranchLru).V) + lruOverhead)
+    opts.rdbBranchCacheSize div
+    (sizeof(RootedVertexID) + sizeof(typeof(rdb.rdBranchLru).V) + lruOverhead)
```
bhartnett, line 99: This fixes a bug in the size calculation where the RootedVertexID key was not previously included.

execution_chain/db/aristo/aristo_blobify.nim
```diff
@@ -18,7 +18,7 @@ import
 
 export aristo_desc, results
 
-const MAX_VERTEX_BLOB_SIZE = 117
+const MAX_VERTEX_BLOB_SIZE = 118
```
bhartnett, line 21: Increased to accommodate the account leaf storage hint.

execution_chain/db/aristo/aristo_blobify.nim
```diff
@@ -154,6 +154,10 @@ proc blobifyTo*(pyl: AccLeafRef, data: var VertexBuf) =
     lens += uint16(tmp.len - 1) shl 8 # 3 bits
     data &= tmp.data()
 
+  if 0 < pyl.stoHint:
+    mask = mask or 0x10
+    data &= [pyl.stoHint]
```
bhartnett, line 159: The 0x10 bit indicates that the stoHint byte is present in the serialized blob.

execution_chain/db/opts.nim
```diff
@@ -41,7 +41,7 @@ const
     ## Cache of branches and leaves in the state MPTs (world and account)
   defaultRdbKeyCacheSize* = 1280 * 1024 * 1024
     ## Hashes of the above
-  defaultRdbBranchCacheSize* = 1024 * 1024 * 1024
+  defaultRdbBranchCacheSize* = 1252 * 1024 * 1024
```
bhartnett, line 44: Increased to match the currently configured memory usage. Updated here because the cache size calculation was fixed.

PR #4816, opened 2026-09-18
Add parallel flag to enable/disable parallel features

bhartnett:
Adds a new `--debug-parallel` cli parameter which can be used to enable or disable all parallel features.

When the flag is not supplied the default parallel features are enabled (currently just parallel sender recovery). When the flag is set to true, all parallel features are turned on. When the flag is set to false, all parallel features are disabled.


bhartnett:
cc @advaita-saha

PR #4817, opened 2026-09-18
Use nim-ssz-serialization digest in sha256 precompile

bhartnett:
Use nim-ssz-serialization digest in sha256 precompile instead of boringssl.

| input   | iters   | BoringSSL (med) | ssz/blst (med) | delta (med) | delta (min) |
|---------|---------|-----------------|----------------|-------------|-------------|
| 0 B     | 12.5M   | 46.76 ns        | 42.24 ns       | −9.7%       | −9.5%       |
| 32 B    | 8.3M    | 47.12 ns        | 45.29 ns       | −3.9%       | −3.8%       |
| 64 B    | 6.25M   | 65.58 ns        | 44.07 ns       | −32.8%      | −33.2%      |
| 128 B   | 4.2M    | 96.14 ns        | 94.04 ns       | −2.2%       | −1.0%       |
| 256 B   | 2.5M    | 154.66 ns       | 152.18 ns      | −1.6%       | −1.8%       |
| 1024 B  | 735k    | 505.97 ns       | 503.66 ns      | −0.5%       | −0.5%       |
| 4096 B  | 400k    | 1892.32 ns      | 1902.88 ns     | +0.6%       | −0.2%       |

The biggest improvement is for 64 B inputs likely due to the hashtree library used by the digest function which is optimized for 64 byte inputs.


bhartnett:
cc @kdeme

PR #4838, opened 2026-09-23
eth2: bump

arnetheduck:
along with check_vendor_eth2.sh --update bumps


tersec:
This is going to run into
- https://github.com/status-im/nimbus-eth2/pull/9085

arnetheduck:
indeed it did .. they could just put that json file in a release or something :/

tersec:
- #4840

